PR #4, opened 2021-04-27
Add caching of node_modules and eslintcache to Github Actions

mattrunyon:
Cache all node_modules folders. If any package-lock changes the cache will miss and trigger `npm ci`. Most of the time this should let us skip `npm ci` which also calls `lerna bootstrap` saving ~2min per run.

Also cache the `.eslintcache` files generated in each package. Currently those will rarely miss, if ever, but cache for each branch will be preferred. This should greatly speed up the eslint Jest test we have in each package. Locally `npm run test:ci` took ~54s w/o eslintcache and ~30s w/ eslintcache on a rerun without changing files.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -64,8 +64,8 @@ jobs:
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
 
-      - name: Test
-        run: npm run test:ci
-
       - name: Build
         run: npm run build
+
+      - name: Test
+        run: npm run test:ci
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -52,8 +52,8 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            '**/.eslintcache'
-            '**/.stylelintcache'
+            **/.eslintcache
+            **/.stylelintcache
           key: ${{ runner.os }}-lintcache-${{ github.head_ref }}
           restore-keys: |
             ${{ runner.os }}-lintcache-
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -54,7 +54,7 @@ jobs:
           key: ${{ runner.os }}-eslintcache-${{ github.head_ref }}
           restore-keys: |
             ${{ runner.os }}-eslintcache-
-            ${{ runner.os}}-
+            ${{ runner.os }}-
 
       # Only need to install and bootstrap deps if package-locks changed
       - name: Install dependencies
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -25,14 +25,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Cache node_modules
-        id: cache-node-modules
-        uses: actions/cache@v2
-        with:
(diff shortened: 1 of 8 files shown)


.github/workflows/node.js.yml
```diff
@@ -36,9 +46,23 @@ jobs:
...
+          path: '**/.eslintcache'
+          key: ${{ runner.os }}-eslintcache-${{ github.head_ref }}
+          restore-keys: |
+            ${{ runner.os }}-eslintcache-
+            ${{ runner.os }}-
+
+      # Only need to install and bootstrap deps if package-locks changed
       - name: Install dependencies
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
```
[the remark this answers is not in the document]
mofojed: @devinrsmith why is using the cache hit to skip a step hacky? That's exactly what this action is for: https://github.com/actions/cache#skipping-steps-based-on-cache-hit

@mattrunyon  There's no more sym linked license files so you should be good from there to try yarn again. I think `npm ci` should be clearing all the node_modules first, to make sure it has a false pass when a package is removed or something.

.github/workflows/node.js.yml
```diff
@@ -19,14 +19,25 @@ jobs:
...
-      - name: Cache node modules
+
+      - name: Cache node_modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        with:
+          # If no package-locks or the main package.json have changed, it should be safe to restore all node_modules as they were
+          # Changing the main package.json may change the postinstall scripts, so it should be unchanged to cache install results
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/package-lock.json', 'package.json') }}
```
mofojed, line 35: Use the Node - lerna example: https://github.com/actions/cache/blob/main/examples.md#node---lerna

mofojed (changes requested):

.github/workflows/node.js.yml
mattrunyon: Effectively doing that, except since I'm skipping `npm ci` on cache hit I'm also hashing the main package.json file.

If we were to change the lerna flags on install/bootstrap and we weren't hashing that file, then it would be a cache hit and skip the install with different flags.

.github/workflows/node.js.yml
```diff
@@ -31,8 +31,10 @@ jobs:
         with:
           # If no package-locks or the main package.json have changed, it should be safe to restore all node_modules as they were
           # Changing the main package.json may change the postinstall scripts, so it should be unchanged to cache install results
-          path: '**/node_modules'
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/package-lock.json', 'package.json') }}
+          path: |
+            node_modules
+            packages/*/node_modules
+          key: ${{ runner.os }}-modules-${{ hashFiles('package.json', 'package-lock.json', 'packages/*/package-lock.json') }}
```
mofojed: Should include `packages/*/package.json` as well. In the case where someone accidentally updates a package.json but doesn't update the associated package-lock.json, we don't want to pull from the cache

> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

mofojed (changes requested):

mofojed (approved):

PR #10, opened 2021-05-06
convert util ColorUtils to TS

spasovski:
Still new to this so expect gremlins.

### DIFF
--- a/packages/utils/src/ColorUtils.ts
+++ b/packages/utils/src/ColorUtils.ts
@@ -8,22 +8,22 @@ class ColorUtils {
     const d = document.createElement('div');
     d.style.display = 'none';
     d.style.color = background;
-    const colorTokens = getComputedStyle(
-      document.body.appendChild(d)
-    ).color.match(/\d+/g);
+
+    const computedColor = getComputedStyle(document.body.appendChild(d)).color;
+    const colorTokens = computedColor.match(/\d+/g);
     let color: number[] = [];
+
     if (colorTokens) {
       color = colorTokens.map(a => parseInt(a, 10));
     } else {
-      throw new Error(
-        'Invalid color received. Expected something like ["123", "123", "123"].'
-      );
+      throw new Error(`Invalid color received. Got ${computedColor}`);
     }
     document.body.removeChild(d);
     const brightness = ColorUtils.getBrightness(color);
     return brightness < 125;
   }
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/ColorUtils.ts
```diff
@@ -4,25 +4,30 @@ class ColorUtils {
...
+    const colorTokens = getComputedStyle(
+      document.body.appendChild(d)
+    ).color.match(/\d+/g);
+    let color: number[] = [];
+    if (colorTokens) {
+      color = colorTokens.map(a => parseInt(a, 10));
+    } else {
+      throw new Error(
+        'Invalid color received. Expected something like ["123", "123", "123"].'
+      );
```
mattrunyon, line 20: I would add the color received to the message. And I think the expected string is just any valid CSS color string? (hex, keyword, rgba, etc.)

packages/utils/src/ColorUtils.ts
```diff
@@ -4,25 +4,30 @@ class ColorUtils {
...
+    }
     document.body.removeChild(d);
     const brightness = ColorUtils.getBrightness(color);
     return brightness < 125;
   }
 
-  static getBrightness(color) {
+  static getBrightness(color: number[]): number {
```
mattrunyon, line 27: You can change the type to `[number, number, number]` to require it be an array of 3 numbers passed in.

However, this would mean you need to do some assertion before calling this method typically which would break it if we used anywhere else (like the call in `isDark`).

Could just add an array length check and error if it's not 3. Or leave it as is. We don't seem to call this anywhere in OSS except for in `isDark` above

mattrunyon (changes requested):

packages/utils/src/ColorUtils.ts
dsmmcken: Just noting that the function input is any css string, be it a color or a css variable, but the output from getComputedStyle and what is passed to the next step will always be an array of 3-4 numbers, even if the passed in string is complete gibberish, the dom will compute it to something valid. colorTokens can never be null, even if background param was.

Also, this technically if you pass something with alpha #FF00000F the alpha is a decimal and getting parsed into two segments by the match, doesn't matter in this case cause the get brightness only looks at the first 3, but still...

mattrunyon (approved):

PR #11, opened 2021-05-06
convert util DbNameValidator to TS

spasovski:
Tests pass locally.

### DIFF
--- a/packages/utils/src/DbNameValidator.ts
+++ b/packages/utils/src/DbNameValidator.ts
@@ -73,7 +73,13 @@ const STERILE_COLUMN_AND_QUERY_REGEX = /[^A-Za-z0-9_$]/g;
  * Table and Column names.
  */
 class DbNameValidator {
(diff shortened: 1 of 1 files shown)


packages/utils/src/DbNameValidator.ts
```diff
@@ -95,16 +101,14 @@ class DbNameValidator {
     }
 
     // If name starts with a number, append prefix to the front
-    // Intentionally using isNaN rather than Number.isNaN
-    // eslint-disable-next-line no-restricted-globals
-    if (!isNaN(legalName.charAt(0))) {
+    if (!Number.isNaN(Number(legalName.charAt(0)))) {
```
mattrunyon, line 104: This should be safe since it's only 1 character. Important to note though that it wouldn't be safe if you were doing an entire string. `Number(false)` would be coerced to 0 and then pass the number check.

But since it's 1 char it's fine like this (and is pretty much what global `isNaN` does)

mattrunyon (approved):

PR #16, opened 2021-05-07
convert util TextUtils to TS

spasovski:


packages/utils/src/TextUtils.ts
```diff
@@ -7,7 +7,7 @@ class TextUtils {
    * One, Two, and Three
    * @param {String[]} items The items to join in a list
    */
-  static join(items) {
+  static join(items: string[]): string {
```
mattrunyon, line 10: Looks like this can be `string[] | null`. Declaring it that way instead of making it optional will make the compiler warn you if you just omit the param vs. passing a possibly null param. There should be no reason for us to explicitly call the function w/o params

packages/utils/src/TextUtils.ts
```diff
@@ -30,7 +30,7 @@ class TextUtils {
    * @param {String} text The text to convert to lower case
    * @param {Boolean} isNullAllowed True if a null string should return an empty string from this function. If false an error is thrown if null is passed in.
    */
-  static toLower(text, isNullAllowed = true) {
+  static toLower(text: string, isNullAllowed = true): string {
```
mattrunyon, line 33: Should be `string | null`

mattrunyon (changes requested):

mattrunyon (approved):

PR #19, opened 2021-05-10
convert util TimeUtils to TS

spasovski:
Was wondering if we could do a custom type for time as `XX:XX:XX` but otherwise pretty straightforward.

### DIFF
--- a/packages/utils/src/TimeUtils.ts
+++ b/packages/utils/src/TimeUtils.ts
@@ -29,7 +29,7 @@ class TimeUtils {
    * Minutes aren't paded, as thats a slower change
    * @param {integer} time in seconds
    */
-  static formatElapsedTime(time) {
+  static formatElapsedTime(time: number): string {
     if (!Number.isInteger(time)) {
       throw new Error(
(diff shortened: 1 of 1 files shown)


packages/utils/src/TimeUtils.ts
```diff
@@ -73,7 +76,7 @@ class TimeUtils {
    * Parse time in seconds from the provided time string
    * @param {string} timeString Time string in hh:mm:ss format
    */
-  static parseTime(timeString) {
+  static parseTime(timeString: string): number {
```
mofojed, line 79: As you suggested in your comment, we can specify a better type using Type Literals: https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

In this case in particular, we can define a `TimeString` type, something like:
```
type TimeString = `${number}:${number}:${number}`;
```
Of course, that would match something like `0:4325423985:3928932`, which is obviously not a valid time string. I don't know if there's an even better way to write it, but even defining a type alias with a comment of the format will be better.

mofojed (changes requested):

PR #25, opened 2021-05-11
Ts components chunk 1

mattrunyon:
AutoResizeTextArea
BasicModal
Checkbox
Collapse

Also adds a linter override for ts/tsx to allow static class members in TS. Otherwise, you get the error "X class does not have property propTypes" when you specify `X.propTypes = ...` on a class component


packages/components/src/Checkbox.tsx
```diff
@@ -0,0 +1,129 @@
...
+          {children}
+        </label>
+      </div>
+    );
+  }
+);
+
+// Forward ref causes a false positive for display-name in eslint:
+// https://github.com/yannickcr/eslint-plugin-react/issues/2269
+Checkbox.displayName = 'Checkbox';
```
spasovski, line 85: Didn't know this. Quite odd.

packages/components/src/Collapse.tsx
```diff
@@ -4,52 +4,74 @@ import classNames from 'classnames';
...
+
+  static defaultProps = {
+    className: '',
+    autoFocusOnShow: false,
+  };
+
+  static handleEnter(elemParam: HTMLElement): void {
     const elem = elemParam;
-    elem.style.height = 0;
+    elem.style.height = '0';
```
spasovski, line 29: Ah TS, technically correct...the best kind of correct.

packages/components/src/Collapse.tsx
```diff
@@ -4,52 +4,74 @@ import classNames from 'classnames';
...
 
-  static handleEntering(elemParam) {
+  static handleEntering(elemParam: HTMLElement): void {
     const elem = elemParam;
     elem.style.height = `${Collapse.getHeight(elem)}px`;
   }
 
-  static handleExiting(elemParam) {
+  static handleExiting(elemParam: HTMLElement): void {
     const elem = elemParam;
```
spasovski, line 38: I question the need for these consts - I realize this is a conversion so feel free to ignore.

spasovski (approved):
This looks quite solid though I think your TS is a few notches above mine. I like that it already caught some stuff that was being coalesced behind the scenes.

packages/components/src/Collapse.tsx
mattrunyon: The const is an eslint thing I think. Pretty sure it would complain about changing param properties (but this is ok apparently)

Issue #23, opened 2021-03-17
Separate dashboard into it's own package


mofojed:
Web UI Dashboard should be it's own package that can be imported by the OSS or Enterprise app, which allows plugins to be registered to handle events, modify dashboard data, send commands, and register their own components. Then OSS can register console, table, command history, and grid plugins, and Enterprise can register the same plus PQ Explorer, Query Monitor, etc plugins. Allow reusability/extensibility of the dashboard.

PR #27, opened 2021-05-17
Separate out @deephaven/redux package

mofojed:
- Separate all the app selectors/actions to a package
- This package will be used by the main app and by other Deephaven packages


mofojed:
Precursor to #23

vbabich (approved):

PR #29, opened 2021-05-18
Ts components chunk 2

mattrunyon:
Popper
AutoCompleteInput (relies on Popper)
SearchInput
ComboBox (relies on SearchInput)

Typescript caught a bug in ComboBox which caused the first item in the dropdown to always show as active even if a different item were selected. The actual value was still correct though


packages/components/package.json
```diff
@@ -78,10 +78,11 @@
     "@types/enzyme": "^3.10.7",
     "@types/enzyme-adapter-react-16": "^1.0.6",
     "@types/jest": "^26.0.15",
+    "@types/memoizee": "^0.4.5",
     "@types/react": "^16.14.0",
     "@types/react-beautiful-dnd": "^13.0.0",
     "@types/react-dom": "^16.9.8",
-    "@types/react-transition-group": "^4.4.0",
+    "@types/react-transition-group": "^2.x",
```
mattrunyon, line 85: We use react-transition-group 2.x so the types version should match

spasovski (approved):
Looks good!

mofojed (approved):

mofojed (approved):

PR #31, opened 2021-05-19
Fix issue with layout initialization race condition

mofojed:


packages/code-studio/src/dashboard/DashboardContainer.jsx
```diff
@@ -276,6 +276,17 @@ export class DashboardContainer extends Component {
...
 
+    await new Promise(resolve => {
+      if (layout.isInitialised) {
+        resolve();
+        return;
+      }
+      layout.on('initialised', () => {
+        log.debug('layout initialized');
+        resolve();
+      });
```
vbabich, line 287: Should probably remove the handler, just in case - 
```suggestion
      const onInit = () => {
        log.debug('layout initialized');
        layout.off('initialised', onInit);
        resolve();
      };
      layout.on('initialised', onInit);
```

vbabich (approved):

PR #32, opened 2021-05-19
convert util PromiseUtils to TS

spasovski:
This one was a bit more difficult.


packages/utils/src/PromiseUtils.ts
```diff
@@ -1,16 +1,20 @@
 import CanceledPromiseError from './CanceledPromiseError';
 import TimeoutError from './TimeoutError';
 
+interface CancelablePromise<T> extends Promise<T> {
```
mattrunyon, line 4: I would export this interface. The Pending promise change should import the interface from this file instead of redefining it and I think it makes most sense to export it from utils

packages/utils/src/PromiseUtils.ts
```diff
@@ -1,16 +1,20 @@
...
+  cancel: () => void;
+}
+
 class PromiseUtils {
-  static isCancelablePromise(promise) {
+  static isCancelablePromise(promise: CancelablePromise<unknown>): boolean {
```
mattrunyon, line 9: Make this a type guard that returns a type predicate. The idea is that this function is called with some unknown value and we want to know if it's a cancelable promise. By using a user-defined type guard that makes TS understand it's saying something is a specific type

```suggestion
  static isCancelablePromise(promise: unknown): promise is CancelablePromise<unknown> {
```

```
const a: any = someUnknownThing;
if (isCancelablePromise(a)) {
  // TS will understand a is now type CancelablePromise<unknown> in this block
}
```

packages/utils/src/PromiseUtils.ts
```diff
@@ -1,16 +1,20 @@
...
     return (
       this.isPromise(promise) &&
       promise.cancel != null &&
       typeof promise.cancel === 'function'
     );
   }
 
-  static isPromise(promise) {
+  static isPromise(promise: Promise<unknown>): boolean {
```
mattrunyon, line 17: Same thing about using a type predicate.

You might have to cast the values for these such as `(promise as Promise<unknown>).then != null`

```suggestion
  static isPromise(promise: unknown): promise is Promise<unknown> {
```

mattrunyon (changes requested):

packages/utils/src/PromiseUtils.ts
```diff
@@ -1,32 +1,42 @@
...
   }
 
-  static isPromise(promise) {
+  static isPromise(promise: unknown): promise is Promise<unknown> {
     return (
       promise != null &&
-      promise.then != null &&
-      typeof promise.then === 'function'
+      (promise as Promise<unknown>).then != null &&
+      typeof (promise as Promise<unknown>).then === 'function'
```
mattrunyon, line 24: Talked w/ Bender about these and we came to the conclusion we can just remove these 2 type guards. They're not actually used anywhere except unit tests. So remove these 2 methods + corresponding unit tests

mattrunyon (changes requested):

packages/utils/src/PromiseUtils.test.js
```diff
@@ -1,40 +0,0 @@
...
-    // Need to add a catch block or the tests crash.
-    // https://github.com/facebook/jest/issues/5311
-    const cancelable = PromiseUtils.makeCancelable(item);
-    cancelable.catch(() => {});
-    return cancelable;
-  };
-
-  expect(PromiseUtils.isPromise(makeCancelable())).toEqual(true);
-  expect(PromiseUtils.isCancelablePromise(makeCancelable())).toEqual(true);
-});
```
mattrunyon, line 40: Keep this test and just make it

```
const cancelable = makeCancelable();
expect(typeof cancelable.cancel).toEqual('function')
```

mattrunyon (changes requested):

mattrunyon (approved):

PR #33, opened 2021-05-19
convert util Pending to TS

spasovski:
Lots of `unknown`s but it *is* a util.

### DIFF
--- a/packages/utils/src/Pending.ts
+++ b/packages/utils/src/Pending.ts
@@ -1,8 +1,4 @@
-import PromiseUtils from './PromiseUtils';
-
-interface CancelablePromise<T> extends Promise<T> {
-  cancel: () => void;
-}
+import PromiseUtils, { CancelablePromise } from './PromiseUtils';
 
 /**
  * Helper class for managing pending promises.
@@ -20,14 +16,11 @@ class Pending {
    * @param cleanup The cleanup function to use when the promise is cancelled
    * @returns CancelablePromise Returns a cancelable promise.
    */
-  add(
-    item: unknown,
-    cleanup: (val: unknown) => void | null
-  ): CancelablePromise<unknown> {
-    const promise = PromiseUtils.makeCancelable(
-      item,
-      cleanup
-    ) as CancelablePromise<unknown>;
+  add<T>(
+    item: Promise<T> | T,
+    cleanup: (val: T) => void | null
+  ): CancelablePromise<T> {
+    const promise = PromiseUtils.makeCancelable(item, cleanup);
     this.pending.push(promise);
     promise.then(
       () => {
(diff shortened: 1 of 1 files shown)


packages/utils/src/Pending.ts
```diff
@@ -1,23 +1,33 @@
 import PromiseUtils from './PromiseUtils';
 
+interface CancelablePromise<T> extends Promise<T> {
+  cancel: () => void;
+}
+
```
mattrunyon, line 6: Import this from PromiseUtils since they're referencing the same interface really

packages/utils/src/Pending.ts
```diff
@@ -1,23 +1,33 @@
...
+   * @param item Item to track.
+   * @param cleanup The cleanup function to use when the promise is cancelled
    * @returns CancelablePromise Returns a cancelable promise.
    */
-  add(item, cleanup = null) {
-    const promise = PromiseUtils.makeCancelable(item, cleanup);
+  add(
+    item: unknown,
+    cleanup: (val: unknown) => void | null
+  ): CancelablePromise<unknown> {
```
mattrunyon, line 26:
```suggestion
  add<T>(
    item: Promise<T> | T,
    cleanup: (val: T) => void | null
  ): CancelablePromise<T> {
```

While unknown works, templating it will enforce the cleanup function uses the same param type as the item

packages/utils/src/Pending.ts
```diff
@@ -1,23 +1,33 @@
...
+    const promise = PromiseUtils.makeCancelable(
+      item,
+      cleanup
+    ) as CancelablePromise<unknown>;
```
mattrunyon, line 30: You shouldn't need to cast this. Probably required it since PromiseUtils is in JS on this branch still

packages/utils/src/Pending.ts
```diff
@@ -34,7 +44,7 @@ class Pending {
    * Remove a promise from tracking.
    * @param {Promise} promise Promise to stop tracking
    */
-  remove(promise) {
+  remove(promise: Promise<unknown>): void {
```
mattrunyon, line 47: Should this be `CancelablePromise<unknown>`? I suppose Promise works just fine, but if we are converting to CancelablePromises always when adding (and the tracking array is only cancelable), then this should be Cancelable I think.

Since it's checking the promise is the same as the one to remove, passing it the non-cancelable version of the promise would mean it's not actually removed I think

packages/utils/src/Pending.ts
```diff
@@ -43,15 +53,15 @@ class Pending {
     }
   }
 
-  resolve(promise) {
+  resolve(promise: CancelablePromise<unknown>): void {
     this.remove(promise);
     this.resolved.push(promise);
```
mattrunyon, line 58: Not part of this change, but I don't think this is actually resolving the promise. Just removes it from tracking array and then says it's resolved even if it isn't

mattrunyon (changes requested):

mattrunyon (approved):

PR #36, opened 2021-05-20
Separate out @deephaven/console component

mofojed:
- Includes everything in the Console and CommandHistory components
- Also includes StorageTable - that should probably be factored out into a separate package as well (@deephaven/storage?)


packages/console/src/common/ObjectIcon.tsx
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { dhTable, vsGraph, vsPreview } from '@deephaven/icons';
 import dh from '@deephaven/jsapi-shim';
-import { ReactComponent as PandasIcon } from '../../assets/svg/pandas_mark_white_tiny.svg';
```
vbabich, line 5: If we don't need the Pandas icon, maybe delete the svg too, to not keep the unused asset in the codebase.

vbabich (dismissed):

packages/console/src/common/ObjectIcon.tsx
dsmmcken: Why are we dropping the pandas icon?

mofojed: It should be put into @deephaven/icons instead #37 created

dsmmcken: Can't, it's multi-colored.

vbabich (approved):

packages/console/src/common/ObjectIcon.tsx
dsmmcken: This merged with no resolution, and no follow up tickets.

mofojed: It'll be captured in #37 . If we can't add it to @deephaven/icons, then need to add direct to @deephaven/console, and have `svgr` somewhere in the pipeline.

PR #40, opened 2021-05-25
Fix ComboBox UX issues

### LINKED ISSUE #30: Fix ComboBox selection issues with non-matching value
There are at least 3 problems with selection currently in ComboBox.

1. Currently if you type a value in the ComboBox component (Input with Select) it will default to showing the 1st item in the list as selected if you open the dropdown. There should be nothing selected if this is the case. The list should still be entered if using the arrow keys.

image

2. If you use arrow keys and reach below the bottom visible item, the box does not scroll into view.
3. If you use arrow keys, the navigation does not wrap from bottom to top or top to bottom (going upward)

- [ ] Default to no highlighted item if non-matching value is entered
- [ ] Scroll items into view so the currently selected item is always visible
- [ ] Let arrow key navigation wrap

mattrunyon:
Fixes #30 

This started as a single problem and then turned into fixing several small UX things in the component

- [ ] Default to no highlighted item if non-matching value is entered
- [ ] Scroll items into view so the currently selected item is always visible
- [ ] Let arrow key navigation wrap
- [ ] No highlighted item arrow down/up should go to first/last item respectively
- [ ] Filtering should highlight the previously highlighted item, or nothing if the previous item does not match the new filter
- [ ] Pressing escape clears filter, but should not close component. Pressing escape w/ empty filter closes component


dsmmcken (dismissed):

dsmmcken:
My approval is for behaviour only, get someone else to review the code.

dsmmcken (approved):
Approved for behaviour.

packages/components/src/ComboBox.tsx
```diff
@@ -187,7 +194,9 @@ class ComboBox extends Component<ComboBoxProps, ComboBoxState> {
 
     switch (event.key) {
       case 'Enter':
-        this.updateInputValue(menuOptions[keyboardOptionIndex].value);
+        if (menuOptions[keyboardOptionIndex]?.value) {
```
mofojed, line 197: Should check for != null here, the value could be '' (empty string, I think?)

packages/components/src/ComboBox.tsx
```diff
@@ -223,21 +239,19 @@ class ComboBox extends Component<ComboBoxProps, ComboBoxState> {
     switch (direction) {
       case ComboBox.MENU_NAVIGATION_DIRECTION.UP:
         if (keyboardOptionIndex > 0) {
```
mofojed, line 241: There's a clever way to write this kind of thing, so you can add and mod correctly without having an else case for the wrap. Just add the menuOptionsLength after subtracting, something like:
```
let delta = 0;
switch ComboBox.Up:
  delta = -1;
  break;
switch ComboBox.Down:
  delta = 1;
  break;

if (delta != 0) {
  this.setState({ keyboardOptionIndex = (keyboardOptionIndex + delta + menuOptionsLength) % menuOptionsLength })
}
```

mofojed (changes requested):

packages/components/src/ComboBox.tsx
mattrunyon: Yep. Good catch

packages/components/src/ComboBox.tsx
mattrunyon: I switched it to this way

mofojed (approved):

PR #50, opened 2021-06-01
Context Menu Typescript

mattrunyon:
Updates all the context menu parts to typescript.

Also a small tweak to PromiseUtils

Git is being really obtuse about ContextMenuItem being renamed and changed. I split it into 2 commits (1 for rename, 1 for changes), but it still says the result is a deleted file and a new file. You can browse the 2nd commit to see it as a side-by-side of just the changes


packages/components/src/context-actions/ContextActions.test.jsx
```diff
@@ -129,12 +129,7 @@ it('renders a promise returning menu items properly', async () => {
 
 it('renders an empty menu for a rejected promise', () => {
   const mock = DEFAULT_MOCK();
-  const promise = {
-    then: () => {},
-    catch: callback => {
-      callback();
-    },
-  };
+  const promise = new Promise((resolve, reject) => reject());
```
mofojed, line 132: You can just write this as `Promise.reject()`
```suggestion
  const promise = Promise.reject();
```

packages/components/src/context-actions/ContextActions.test.jsx
```diff
@@ -155,12 +151,7 @@ it('renders an empty menu for a rejected promise', () => {
 
 it('renders a menu from a promise returned from a function', () => {
   const mock = DEFAULT_MOCK();
-  const promise = {
-    then: () => {},
-    catch: callback => {
-      callback();
-    },
-  };
+  const promise = new Promise(resolve => resolve());
```
mofojed, line 154:
```suggestion
  const promise = Promise.resolve();
```

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -37,35 +52,64 @@ const log = Log.module('ContextActions');
...
+  ContextActionsProps,
+  ContextActionsState
+> {
+  /**
+   * Group you can assign to context menu actions to group them together.
+   * Lower group IDs appear at the top of the list.
+   * Groups are separated by a separator item.
+   * Items within groups are ordered by their order property, then by their title.
+   */
+  static groups = {
```
mofojed, line 65: Doesn't necessarily need to be changed now, but we should be using `SCREAMING_SNAKE_CASE` for constants

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -37,35 +52,64 @@ const log = Log.module('ContextActions');
...
       return;
     }
 
-    const mouseEvent = new MouseEvent('contextmenu', {
-      clientX,
-      clientY,
-      bubbles: true,
-      cancelable: true,
-    });
+    const mouseEvent: Partial<ContextActionEvent> = new MouseEvent(
```
mofojed, line 85: Nice, `Partial` is where it's at

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -87,32 +131,41 @@ class ContextActions extends Component {
...
         'keydown',
         this.handleKeyDown
       );
     }
   }
 
-  handleContextMenu(e) {
-    if (!e.contextActions) {
-      e.contextActions = [];
+  container: React.RefObject<HTMLDivElement>;
```
mofojed, line 160: I think this should be put before the constructor

packages/components/src/context-actions/ContextMenuRoot.tsx
```diff
@@ -22,44 +33,56 @@ class ContextMenuRoot extends Component {
...
         this.handleContextMenu
       );
     }
   }
 
-  handleContextMenu(e) {
+  container: React.RefObject<HTMLDivElement>;
+
+  openMenu: React.RefObject<ContextMenu>;
+
```
mofojed, line 57: Same here, I think these should be before the constructor

packages/components/src/context-actions/ContextMenuRoot.tsx
```diff
@@ -1,19 +1,30 @@
...
-  constructor(props) {
+class ContextMenuRoot extends Component<
+  ContextMenuRootProps,
+  ContextMenuRootState
+> {
+  constructor(props: ContextMenuRootProps) {
     super(props);
 
-    this.container = null;
+    this.container = React.createRef();
```
mofojed, line 24: Should be after the binds

packages/components/src/context-actions/GlobalContextAction.tsx
```diff
@@ -1,12 +1,22 @@
...
 
-class GlobalContextAction extends Component {
-  constructor(props) {
+interface GlobalContextActionProps {
+  action: ContextAction;
+}
+
+class GlobalContextAction extends Component<
+  GlobalContextActionProps,
+  KeyState
```
mofojed, line 17: I don't reaaaaally like using `KeyState` directly here, should define a `type GlobalContextActionState = KeyState` above and use that.

packages/components/src/context-actions/GlobalContextAction.tsx
```diff
@@ -40,8 +50,17 @@ class GlobalContextAction extends Component {
...
 
-  handleContextMenu(e) {
+  handleContextMenu(e: MouseEvent): void {
+    if (!ContextActionUtils.isContextActionEvent(e)) {
+      (e as ContextActionEvent).contextActions = [];
+    }
+
+    if (!ContextActionUtils.isContextActionEvent(e)) {
+      return;
+    }
```
mofojed, line 60: What's going on here? The second if statement doesn't make sense after the first one, it'll always be a ContextActionEvent at that point? And then below there's already a spot where it sets the `contextActions` array ... why not just leave that place below and get rid of this?

packages/utils/src/index.js
```diff
@@ -4,6 +4,7 @@ export { default as DbNameValidator } from './DbNameValidator';
 export { default as DOMUtils } from './DOMUtils';
 export { default as Pending } from './Pending';
 export { default as PromiseUtils } from './PromiseUtils';
```
mofojed, line 6: Could also just do an `export class PromiseUtils` from PromiseUtils then don't need this line since next line will export it all

mofojed (changes requested):

packages/components/src/context-actions/ContextActions.tsx
mattrunyon: Leaving for now. It's used in the current casing in ~25 places in OSS and probably more in enterprise

packages/components/src/context-actions/ContextActionUtils.ts
```diff
@@ -1,13 +1,78 @@
...
       ctrlKey: false,
     };
 
+    if (shortcutParam === '' || shortcutParam === undefined) {
+      return keyState;
+    }
+
     const shortcut = shortcutParam.replace('^', ContextActionUtils.CTRL_KEY);
 
     if (shortcut) {
```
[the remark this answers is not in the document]
mattrunyon: Good catch

packages/components/src/context-actions/ContextActions.tsx
mattrunyon: I do too. This was where the current eslint config told me to put it. I've adjusted the config to place instance variables before the constructor, but after static members.

mattrunyon: Adjusting the config throws other things off w/ linting. We use `getCachedThing = memoize(...)` in places which is an instance variable. Might have to leave the eslint config and make these an exception to it?

packages/components/src/context-actions/GlobalContextAction.tsx
mattrunyon: This was a typescript thing

The event param must be a `MouseEvent` to use `addEventListener`. In order to check that `contextActions` existed on the event there's the util function which by returning a type predicate tells TS what the type is. This was just a way to set a default empty array for `contextActions` and then tell TS that e is a ContextActionEvent.

I switched it to something better I think. Uses type assertion (technically there is no "casting" since TS doesn't do any runtime checks), but doesn't need the double call to that method.

packages/utils/src/index.js
mattrunyon: Good call

packages/components/src/context-actions/ContextActions.tsx
mofojed: Yea I'm alright with leaving it... only other thing I can think of is moving the memoizes as well (need to think about that), or defining an interface first that it implements, eg.:

```
interface ContextActionsInterface {
  container: React.RefObject<HTMLDivElement>;
}

class ContextActions extends Component<...> implements ContextActionsInterface { ... }
```

Not sure that will actually work

packages/components/src/context-actions/ContextMenuRoot.tsx
mattrunyon: Moved it

packages/components/src/context-actions/ContextActions.tsx
mattrunyon: I'll make another ticket for adjusting the linter config so we can move these w/o needing rule overrides

mofojed (approved):

Issue #51, opened 2021-06-01
Opening Code Studio in Safari 12 shows a blank screen, error in logs


mofojed:
**Steps to Reproduce**
1. Open Code Studio with Safari Version 12.1.2 (14607.3.9)

**Expected Results**
1. Code studio opens, can do stuff

**Actual Results**
1. Blank page is shown, error in the log:
```
TypeError: function is not a constructor (evaluating 'new EventTarget')
e — LogProxy.ts:115
(anonymous function) — LogInit.ts:12
i — ide:1:1308
r — ide:1:1174
t — ide:1:1037
Global Code — main.18e248e9.chunk.js:1:102
```

mofojed:
In LogProxy.ts, we create a new EventTarget: https://github.com/deephaven/web-client-ui/blob/53749323846b09a4b0e6f5778306a35c80d3d3f1/packages/log/src/LogProxy.ts#L45

EventTarget constructor was only added to Safari in version 14: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/EventTarget

We're [not targetting Safari](https://github.com/deephaven/deephaven-core/issues/286), and it can be fixed by updating Safari, the recommendation is simply to update Safari.

However, we _can_ fix this in Safari <14 by using `event-target-shim` (as we already do in GridModel). It's easy to do and will increase compatibility.

PR #52, opened 2021-06-01
Add event-target-shim to LogProxy

mofojed:
Needed for Safari <14

### DIFF
--- a/packages/code-studio/package-lock.json
+++ b/packages/code-studio/package-lock.json
@@ -4,16 +4,6 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"3d-view": {
-			"version": "2.0.0",
-			"resolved": "https://registry.npmjs.org/3d-view/-/3d-view-2.0.0.tgz",
-			"integrity": "sha1-gxrpQtdQjFCAHj4G+v4ejFdOF74=",
-			"requires": {
-				"matrix-camera-controller": "^2.1.1",
-				"orbit-camera-controller": "^4.0.0",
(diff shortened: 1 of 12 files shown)


mofojed:
I ran `./node_modules/.bin/lerna add event-target-shim --scope=@deephaven/log ` to add the event-target-shim lib, so I think the updates to the package-lock.jsons is also correct (removing the common packages from each package-lock.json to have it only in the highest level package-lock.json)

packages/code-studio/package-lock.json
```diff
@@ -1282,124 +1264,6 @@
...
-		"@deephaven/chart": {
-			"version": "0.1.1",
-			"resolved": "https://registry.npmjs.org/@deephaven/chart/-/chart-0.1.1.tgz",
-			"integrity": "sha512-QH7Zgp4K5+NeRZCTdjTXK3gen91NfUB/Fw0oCuvcCuvWciegJXHW/RIeW9isCPUt6F7iB7K/fMTLv3Zxyz0odA==",
-			"requires": {
-				"plotly.js": "^1.54.7",
-				"react-plotly.js": "^2.4.0"
-			}
-		},
-		"@deephaven/components": {
```
mattrunyon, line 1294: Weird that all our packages were removed from this package-lock

mattrunyon (approved):
Make sure our packages should have been removed from code-studio package lock

packages/code-studio/package-lock.json
mofojed: I believe it is correct - since it's pulling from within the same repo

mofojed:
Fixes #51

PR #57, opened 2021-06-06
TS Components Chunk 5

mattrunyon:
Menu
DropdownMenu
MaskedInput
TimeInput
CustomTimeSelect
TimeSlider
DragUtils
HierarchicalCheckboxMenu


packages/components/src/menu-actions/Menu.tsx
```diff
@@ -22,12 +58,12 @@ class Menu extends PureComponent {
...
     this.oldFocus = document.activeElement;
-    this.rAF = null;
+    this.rAF = 0;
 
     const { options } = props;
-    let keyboardIndex = options.initialKeyboardIndex;
+    let keyboardIndex = options.initialKeyboardIndex || -1;
     if (!Number.isInteger(keyboardIndex)) {
       keyboardIndex = -1;
     }
```
mofojed, line 69: Use `??` instead of `||`, otherwise can't select `0`. Also, shouldn't need the next `if` condition, as it will always be a number
```suggestion
    const keyboardIndex = options.initialKeyboardIndex ?? -1;
```

packages/components/src/menu-actions/Menu.tsx
```diff
@@ -132,20 +170,28 @@ class Menu extends PureComponent {
       e.key === 'ArrowUp' ||
       (e.key === 'Tab' && e.shiftKey === true)
     ) {
-      newFocus = ContextActionUtils.getNextMenuItem(newFocus, -1, menuItems);
+      newFocus = ContextActionUtils.getNextMenuItem(
+        newFocus || 0,
```
mofojed, line 174: Doesn't really matter here, but `??` instead of `||`
```suggestion
        newFocus ?? 0,
```

packages/components/src/menu-actions/Menu.tsx
```diff
@@ -132,20 +170,28 @@ class Menu extends PureComponent {
...
+        -1,
+        menuItems
+      );
     } else if (
       e.key === 'ArrowDown' ||
       (e.key === 'Tab' && e.shiftKey === false)
     ) {
-      newFocus = ContextActionUtils.getNextMenuItem(newFocus, 1, menuItems);
+      newFocus = ContextActionUtils.getNextMenuItem(
+        newFocus || 0,
```
mofojed, line 183: `??` instead of `||`
```suggestion
        newFocus ?? 0,
```

packages/utils/src/TimeUtils.ts
```diff
@@ -75,6 +75,11 @@ class TimeUtils {
     return `${hours}:${minutes}:${seconds}`;
   }
 
+  static isTimeString(s: string): s is TimeString {
+    const TIME_PATTERN = '([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]';
```
mofojed, line 79: Declare as a static const

mofojed (changes requested):

packages/components/src/menu-actions/Menu.tsx
mattrunyon: Good catch

PR #65, opened 2021-06-09
Enterprise update

mofojed:
- Add storage classes
- Fix up iris-grid/chart testing
- DH-10134 Abstract WorkspaceStorage impl to allow different backing stores
- Show spinner during multiSeries loading
- CommandHistory - remove WorkspaceStorage dependency, fix ConsoleInput filtering
- Re-implement PouchStorageTable with new storage table abstraction


packages/console/src/command-history/CommandHistory.jsx
```diff
@@ -13,14 +13,16 @@ import {
   vsPlay,
   vsTerminal,
 } from '@deephaven/icons';
+import { StorageTableViewportUpdater } from '@deephaven/storage';
```
vbabich, line 16: This seems unused...  Looks like the eslint config is missing for the console package.

packages/storage/README.md
```diff
@@ -0,0 +1,9 @@
+# @deephaven/utils
```
vbabich, line 1: This needs to be updated

vbabich (changes requested):

vbabich (changes requested):

packages/console/src/command-history/CommandHistory.jsx
mofojed: Nice catch!

vbabich (approved):

PR #66, opened 2021-06-11
Publish v0.1.2

mofojed:


packages/console/package.json
```diff
@@ -70,15 +70,15 @@
...
-    "@deephaven/components": "^0.1.1",
-    "@deephaven/icons": "^0.1.0",
-    "@deephaven/jsapi-shim": "^0.1.1",
-    "@deephaven/log": "^0.1.1",
-    "@deephaven/mocks": "^0.1.1",
-    "@deephaven/redux": "^0.1.1",
-    "@deephaven/storage": "^0.1.1",
-    "@deephaven/tsconfig": "^0.1.0",
-    "@deephaven/utils": "^0.1.1",
+    "@deephaven/components": "^0.1.2",
```
vbabich, line 73: Do we need to update `peerDependencies` as well?

mofojed: No... and really I should be writing the version for dev/peer deps in the same format ("^0.1.2"). But anyway, this is already published so we should merge

vbabich (approved):

PR #67, opened 2021-06-11
Revive autocomplete deephaven-core#682

JamesXNelson:


packages/code-studio/src/dashboard/DashboardContainer.jsx
```diff
@@ -230,7 +230,7 @@ export class DashboardContainer extends Component {
     setDashboardLinks(id, [...links]);
   }
 
-  initLayout() {
+  async initLayout() {
```
JamesXNelson, line 233: gwhaaa..... @mofojed I think I inherited this from you... unsure what I should do w/ it.

JamesXNelson:
Do NOT merge until https://github.com/deephaven/deephaven-core/pull/682 merges, else we'll have very broken clients.

packages/code-studio/src/dashboard/DashboardContainer.jsx
mofojed: Get rid of it, if you rebase off the latest it should already have this.

mofojed (changes requested):

packages/code-studio/src/dashboard/DashboardContainer.jsx
JamesXNelson: done, tyvm.

mofojed (approved):

packages/console/src/ConsoleInput.jsx
```diff
@@ -6,8 +6,9 @@ import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import Log from '@deephaven/log';
 import { PromiseUtils } from '@deephaven/utils';
 import { getCommandHistoryStorage } from '@deephaven/redux';
-import MonacoUtils from './monaco/MonacoUtils';
-import MonacoTheme from './monaco/MonacoTheme.module.scss';
+import MonacoUtils from '../monaco/MonacoUtils';
```
mofojed, line 9: Import shouldn't be changed

```suggestion
import MonacoUtils from './monaco/MonacoUtils';
```

packages/console/src/ConsoleInput.jsx
```diff
@@ -6,8 +6,9 @@ import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
...
+import MonacoCompletionProvider from './MonacoCompletionProvider';
+import MonacoTheme from '../monaco/MonacoTheme.module.scss';
```
mofojed, line 11: Import shouldn't be changed
```suggestion
import MonacoTheme from './monaco/MonacoTheme.module.scss';
```

mofojed (changes requested):

JamesXNelson:
Fixed imports. not sure if you want me to reorder the import that I do have to make it cleaner or not?

packages/console/src/ConsoleInput.jsx
```diff
@@ -7,6 +7,7 @@ import Log from '@deephaven/log';
 import { PromiseUtils } from '@deephaven/utils';
 import { getCommandHistoryStorage } from '@deephaven/redux';
 import MonacoUtils from './monaco/MonacoUtils';
+import MonacoCompletionProvider from './MonacoCompletionProvider';
```
mofojed, line 10: It would make sense to move this into `console/src/monaco` as well, then make an index file that exports all of them, so you can just do `import { MonacoCompletionProvider, MonacoTheme, MonacoUtils } from './monaco'`, would be nice little cleanup.

mofojed (approved):

mofojed (changes requested):
You'll need to update `ScriptEditor` as well, there's a couple of commented out lines there calling MonacoUtils.openDocument and MonacoUtils.closeDocument (it's already calling getCompletionItems)

packages/console/src/ConsoleInput.jsx
JamesXNelson: Does that work for MonacoTheme, even though it is imported `from ./monaco/MonacoTheme.module.scss` ?

mofojed (approved):

PR #71, opened 2021-06-14
Implement react redux registry

mofojed:
Allows modular code to register reducers

Blog post: http://nicolasgallagher.com/redux-modules-and-code-splitting/

### DIFF
--- a/packages/redux/src/reducerRegistry.ts
+++ b/packages/redux/src/reducerRegistry.ts
@@ -3,6 +3,9 @@ import { Reducer, ReducersMapObject } from 'redux';
 export type ReducerRegistryListener = (reducers: ReducersMapObject) => void;
 
 /**
+ * A registry for registering reducers. Whenever a new reducer is added,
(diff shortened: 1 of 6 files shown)


packages/redux/src/reducerRegistry.ts
```diff
@@ -0,0 +1,28 @@
...
+ * http://nicolasgallagher.com/redux-modules-and-code-splitting/
+ */
+export class ReducerRegistry {
+  reducers = {} as ReducersMapObject;
+
+  listener: ReducerRegistryListener | null = null;
+
+  register(name: string, reducer: Reducer): void {
+    this.reducers = { ...this.reducers, [name]: reducer };
+    this.listener?.(this.reducers);
```
spasovski, line 19: Haven't seen this in a function call before - freaky!

spasovski (approved):
Looks good though I am certainly not familiar with Nicolas' approach of a reducer registry. In ether case I am glad this works.

PR #73, opened 2021-06-15
Fix sass watching 100,000+ files in dev

### LINKED ISSUE #60: sass watch is watching 100,000s of files
**Steps to Reproduce**
1) Run `npm start` from the root directory
2) Check the number of watched files (from [StackOverflow](https://stackoverflow.com/a/60705427): `curl -s https://raw.githubusercontent.com/fatso83/dotfiles/master/utils/scripts/inotify-consumers | bash`)

**Expected Results**
2) Only the scss files in our src directories are watched

**Actual Results**
2) 100,000s of files are being watched... all of node_modules maybe?
```
    COUNT     PID USER     COMMAND
--------------------------------------
  104768   168748 bender   node /home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/chart/node_modules/.bin/sass --load-path=../../node_module
  104243   168787 bender   node /home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/iris-grid/node_modules/.bin/sass --load-path=../../node_mo
  103492   168876 bender   node /home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/console/node_modules/.bin/sass --load-path=../../node_modu
   93774   168722 bender   node /home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/components/node_modules/.bin/sass --load-path=../../node_m

```

mattrunyon:
Fixes #60 

This only seems to affect Linux (not WSL or Mac). The others use a different file watching protocol through chokidar. Not sure why WSL does since it should be Linux, but I had 0 inotify watches when trying it out on WSL.

### DIFF
--- a/packages/chart/src/Chart.module.scss
+++ b/packages/chart/src/Chart.module.scss
@@ -1,4 +1,4 @@
-@import '@deephaven/components/scss/custom.scss';
+@import '../../components/scss/custom.scss';
 @import './Chart.scss';
 
 /* stylelint-disable property-no-unknown */
--- a/packages/components/scss/custom.scss
+++ b/packages/components/scss/custom.scss
@@ -1,7 +1,7 @@
 // Styling overrides for bootstrap
 
 //Make bootstrap functions available for use in overrides
-@import 'bootstrap/scss/_functions.scss';
+@import '../../../node_modules/bootstrap/scss/_functions.scss';
 
 // Override / set color variables
 $red: #f95d84;
@@ -226,8 +226,8 @@ $form-feedback-icon-invalid: str-replace(
 //_variable imports come after bootstrap default overrides,
 // makes all other variables and mixins from bootstrap available
 /// with just importing customer.scss
-@import 'bootstrap/scss/_variables.scss';
-@import 'bootstrap/scss/_mixins.scss';
+@import '../../../node_modules/bootstrap/scss/_variables.scss';
+@import '../../../node_modules/bootstrap/scss/_mixins.scss';
 
 //New variables come after imports
 //---------------------------------------
--- a/packages/chart/package.json
+++ b/packages/chart/package.json
@@ -20,7 +20,7 @@
     "build-dev": "cross-env NODE_ENV=development run-s clean build-steps",
     "build-steps": "run-p build:*",
     "babel": "babel ./src --out-dir ./dist --extensions \".ts,.tsx,.js,.jsx\" --source-maps --root-mode upward --verbose",
-    "sass": "sass --load-path=../../node_modules --load-path=node_modules ./src:./dist",
+    "sass": "sass --load-path=node_modules ./src:./dist",
     "types": "tsc",
     "build:babel": "npm run babel",
     "build:sass": "npm run sass",
--- a/packages/chart/package.json
+++ b/packages/chart/package.json
@@ -20,7 +20,7 @@
     "build-dev": "cross-env NODE_ENV=development run-s clean build-steps",
     "build-steps": "run-p build:*",
     "babel": "babel ./src --out-dir ./dist --extensions \".ts,.tsx,.js,.jsx\" --source-maps --root-mode upward --verbose",
-    "sass": "sass --load-path=node_modules ./src:./dist",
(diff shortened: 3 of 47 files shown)


packages/chart/src/Chart.module.scss
```diff
@@ -1,4 +1,4 @@
-@import '@deephaven/components/scss/custom.scss';
+@import '../../components/scss/custom.scss';
```
mofojed, line 1: I don't really like having to do a relative dir for components like this - it makes it more fragile, if we move one of these scss files or one of those folders we break the chain.

Did you take a look at the sass watch source at all to see if there's any way to restrict the types of files it actually watches? Would love to have an upstream fix (or issue) filed as well.

packages/components/scss/custom.scss
```diff
@@ -1,7 +1,7 @@
 // Styling overrides for bootstrap
 
 //Make bootstrap functions available for use in overrides
-@import 'bootstrap/scss/_functions.scss';
+@import '../../../node_modules/bootstrap/scss/_functions.scss';
```
mofojed, line 4: I'd be _okay_ keeping this one relative (as it's just in one spot), but add a comment why we need relative instead of doing the normal thing with load-path.

Would doing `load-path=node_modules` in the other projects be enough? Then just need the one relative import here. Would like a comment, and possibly a link to an upstream sass issue.

Also you say this doesn't happen on Windows or Mac? What's different with the watches?

mofojed: Even having glob paths to specify would help: https://github.com/sass/dart-sass/issues/616

I don't know, the watch seems limiting!

mofojed (changes requested):

packages/chart/src/Chart.module.scss
mofojed: I found where they have it hardcoded to ignore files that aren't sass... https://github.com/sass/dart-sass/blob/f6819060ec54358e318fafa6395238975030834c/lib/src/executable/watch.dart#L129

```
              if (extension != '.sass' &amp;&amp; extension != '.scss' &amp;&amp; extension != '.css') {
                continue;
              }
```

mofojed: I think I'd be using chokidar, so this watch? https://github.com/sass/dart-sass/blob/f6819060ec54358e318fafa6395238975030834c/lib/src/io/node.dart#L218

I wonder if it _would_ be better with globbing (specifying **/*.scss or something). They mention being judicious about dirs you're watching as it may recursively watch a lot: https://github.com/paulmillr/chokidar#how

They've also got another watch impl here but I don't think I'm hitting this one: https://github.com/sass/dart-sass/blob/f6819060ec54358e318fafa6395238975030834c/lib/src/io/vm.dart#L90

There's also a couple relevant issues that we could track/add to:
https://github.com/sass/dart-sass/issues/1292
https://github.com/sass/sass/issues/2739 (which speaks about a fix not requiring modifying load-paths)

mattrunyon: These could be kept for now. There's 3 or 4 packages which need to watch their local node_modules and add a few thousand watchers each.

Could also specifically watch `node_modules/@deephaven/components/scss` but then the imports are less clear since it would just be `import 'custom.scss'`

Pretty sure I tried watching globs and it didn't work. It's also really annoying because from the source it looks like it should detect only sass files first and then only pass those to watch, but that isn't working on Linux it seems.

packages/components/scss/custom.scss
mattrunyon: `load-path=node_modules` would work for the others, it just adds 1-2k watchers for 3 different packages. Which isn't that big of a deal compared to it watching 100,000+.

I'm not sure what WSL was doing since it should be running the Linux kernel, but my guess is the FS is a bit different from a normal Linux FS and doesn't use inotify. Mac uses FSEvents through chokidar

mofojed: Okay let me know if you want me to test that out

mofojed (approved):

PR #74, opened 2021-06-16
Add lerna-changelog to devDependencies

mofojed:
lerna-changelog will be used to update the release notes.


mofojed:
I'm second guessing myself here... I kind of want to get into using conventional commits, which requires github to enforce it.

I'm going to close this for now.

PR #79, opened 2021-06-23
Make Keyboard Shortcuts User Customizable

### LINKED ISSUE #14: Make Keyboard Shortcuts User Customizable
Allow user overrides for shortcuts. Save the overrides with workspace data.

Try to check if the shortcuts will conflict w/ Monaco shortcuts. Monaco doesn't have a good API for replacing shortcuts unfortunately, but we can try to check it at least using the undocumented API we use to remove a few shortcuts currently

mattrunyon:
Closes #14 

Still needs a few things

- [x] Prompt to clear conflicting shortcuts
- [x] Clear conflicting shortcuts if confirmed
- [x] Allow removal of shortcut with Backspace key
- [x] Restore to default
- [x] Probably prompt to confirm changing the shortcut/cancel change

The majority of things are working for review though. Shortcut class, displaying shortcuts in settings panel, shortcuts defined and used throughout the code instead of context actions from string literals.

Shortcuts will override but not save since workspace storage isn't added yet. The shortcut change will persist until a page reload. Context menus should have updated shortcut strings when changing a shortcut


packages/components/src/context-actions/GlobalContextActions.tsx
```diff
@@ -12,11 +12,11 @@ class GlobalContextActions extends Component<GlobalContextActionsProps> {
     const actionElements = [];
     for (let i = 0; i < actions.length; i += 1) {
       const action = actions[i];
-      const shortcut = ContextActionUtils.getShortcutFromAction(action);
+      const { shortcut } = action;
       if (action.title || action.menuElement || shortcut) {
         const actionElement = (
           <GlobalContextAction
-            key={`${action.title}.${shortcut}`}
+            key={`${action.title}.${shortcut?.name}`}
```
mofojed, line 19: shortcut.name? Or shortcut.id

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -0,0 +1,324 @@
...
+  Y = 'Y',
+  Z = 'Z',
+  BACKSPACE = 'Backspace',
+  ESCAPE = 'Escape',
+  ENTER = 'Enter',
+  DELETE = 'Delete',
+  SLASH = '/',
+}
+
+type ShortcutKeys = [...MODIFIER[], KEY];
```
mofojed, line 49: Neat. I sort of expected the syntax to be `[...MODIFIER, KEY]` (without the array on MODIFIER), but I guess I'm wrong

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -0,0 +1,324 @@
...
+  static isAllowedKey(key: string | null): key is KEY {
+    return Object.values(KEY).includes(key as KEY);
+  }
+
+  /**
+   * Checks if a KeyState has a valid key.
+   * @param state KeyState with any string as the key
+   * @returns True if KeyState is is using an allowed keyCode
+   */
+  static isValidKeyState(state: KeyState): state is ValidKeyState {
```
mofojed, line 95: Why are we doing a check like this? What about ESCAPE, BACKSPACE, arrow keys... why don't we let a user set a keyboard shortcut to just a letter and not require a modifier key?

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -0,0 +1,324 @@
...
+  V = 'V',
+  W = 'W',
+  X = 'X',
+  Y = 'Y',
+  Z = 'Z',
+  BACKSPACE = 'Backspace',
+  ESCAPE = 'Escape',
+  ENTER = 'Enter',
+  DELETE = 'Delete',
+  SLASH = '/',
```
mofojed, line 46: What about the number keys (0-9), backtick/tilde, -/=, comma, period, colon, semicolon, backslash, F keys (eg. `F1`, etc), PageUp/PageDown, Home/End, NumLock, ScrollLock, Pause

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    if (keyState.ctrlKey) {
+      display += 'Ctrl+';
+    }
+    if (keyState.altKey) {
+      display += 'Alt+';
+    }
+    if (keyState.shiftKey) {
+      display += 'Shift+';
+    }
```
mofojed, line 183: Add a case for meta here? Not sure if those are allowed at all on Windows, if not add a comment why not

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    if (keyState.keyValue === KEY.ESCAPE) {
+      display += 'Esc';
+    } else if (Shortcut.isAllowedKey(keyState.keyValue)) {
```
mofojed, line 187: If the key isn't allowed - should we throw in that case? Or fallback to just appending the keyvalue anyway for the display text? Not sure why this check is here

packages/components/src/shortcuts/ShortcutRegistry.ts
```diff
@@ -0,0 +1,64 @@
...
+  static readonly shortcutsByCategory = new Map<string, Shortcut[]>();
+
+  /**
+   * Creates a Shortcut and adds it to the registry
+   * @param params The constructor params for the {@link Shortcut}
+   * @param category The category this will show up under in the settings menu. If omitted, shortcut will be hidden in settings
+   * @returns The created Shortcut
+   */
+  static createAndAdd(
+    params: ConstructorParameters<typeof Shortcut>[0]
```
mofojed, line 15: `ConstructorParameters`, neat

packages/components/src/shortcuts/ShortcutRegistry.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (ShortcutRegistry.shortcutMap.has(shortcut.id)) {
+      throw new Error(
+        `Skipping attempt to add duplicate shortcut ID to registry: ${shortcut.id}`
+      );
+    }
+
+    const category = shortcut.id.split('.')[0];
+
+    ShortcutRegistry.shortcutMap.set(shortcut.id, shortcut);
+    if (ShortcutRegistry.shortcutsByCategory.has(category)) {
```
mofojed, line 36: I don't think you need to have a `shortcutsByCategory` map in the registry itself, unless we enforce something like a shortcut can't conflict with something in the same category.
Even then - we should probably let the user set a conflicting shortcut, and just inform them that there's an existing shortcut for that combo.

mofojed (changes requested):

packages/components/src/shortcuts/ShortcutRegistry.ts
mattrunyon: The thought was to only warn about conflicting shortcuts within categories. Should be easy enough to just calculate the category list each time instead of keeping a map of it.

packages/components/src/shortcuts/Shortcut.ts
mattrunyon: I guess it's not common, but I'll do "Win+"

We should maybe try to differentiate between Linux/Windows just for this key and display "Meta+" for Linux? Not sure what the expected value is for Linux users

packages/components/src/shortcuts/Shortcut.ts
mattrunyon: Since `MODIFIER` is an enum (and thus the type refers to any value in the enum), you can't spread it. Specifically TS says "a rest element must be an array"

This mostly only helps with defining the shortcuts, but in 99% of cases we will probably be defining directly in the constructor params. The other 1% (tests mostly) you can declare the type I believe. It treats it as `(MODIFIER | KEY)[]` which is not compatible directly with this type

packages/components/src/shortcuts/Shortcut.ts
dsmmcken: Yes to function keys, and character keys, no to the others.

packages/code-studio/src/settings/ShortcutsSectionContent.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { Shortcut, Tooltip, ShortcutRegistry } from '@deephaven/components';
+import type { KeyState } from '@deephaven/components';
+import './ShortcutsSectionContent.scss';
+
+export default function ShortcutSectionContent(): JSX.Element[] {
+  function formatCategoryName(name: string): string {
+    return name
```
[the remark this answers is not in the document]
mattrunyon: I thought about putting it in a utils, but felt it wasn't really robust enough. It only converts screaming snake case to title case. And it's kind of a fake title case since it doesn't check for things like "a", "the", "and" which would be lowercase in title case

packages/code-studio/src/settings/SettingsMenu.jsx
```diff
@@ -60,7 +67,19 @@ class SettingsMenu extends Component {
     }));
   }
 
+  handleShortcutSectionToggle(sectionKey) {
+    const { expandedSectionKey } = this.state;
+    if (expandedSectionKey !== sectionKey) {
```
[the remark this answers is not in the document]
mattrunyon: Ya that works better

packages/components/src/shortcuts/Shortcut.ts
mofojed: Is that what windows says if you open up keyboard shortcuts?
Linux says "Super+":
image

dsmmcken: browser meta keycode = "Win" key for windows, and shouldn't be an allowed shortcut key for windows users. It conflicts with everything the OS assigns.

mattrunyon: Should be prevented from use on Windows/Linux

packages/components/src/shortcuts/Shortcut.ts
mattrunyon: Appends the keyvalue now so it displays the full invalid shortcut

packages/code-studio/src/settings/Shortcuts.ts
```diff
@@ -0,0 +1,50 @@
...
+  }),
+  RENAME: ShortcutRegistry.createAndAdd({
+    id: 'FILE_EXPLORER.RENAME',
+    name: 'Rename',
+    shortcut: [KEY.ENTER],
+    macShortcut: [KEY.ENTER],
+  }),
+};
+
+const CHART = { name: 'Chart' };
```
mofojed, line 42: Is there a reason to have this section if there's no shortcuts for it?

packages/code-studio/src/settings/ShortcutsSectionContent.tsx
```diff
@@ -0,0 +1,275 @@
...
+}
+
+type ShortcutItemProps = {
+  shortcut: Shortcut;
+  displayText: string;
+  categoryName: string;
+  onChange(shortcut: Shortcut): void;
+};
+
+function ShortcutItem({
```
mofojed, line 106: Can this be in it's own file? There seems to be too many things in this file

packages/code-studio/src/styleguide/DropdownMenus.jsx
```diff
@@ -28,17 +34,25 @@ class DropdownMenus extends Component {
...
       {
         title: `Log Message`,
         icon: dhFilePrint,
         action: () => {
           console.log(`Logging a message!`);
         },
-        shortcut: '⌃L',
-        macShortcut: '⌘L',
+        shortcut: new Shortcut({
+          id: 'STYLEGUIDE.LOW_MESSAGE',
```
mofojed, line 51:
```suggestion
          id: 'STYLEGUIDE.LOG_MESSAGE',
```

mofojed (changes requested):

packages/code-studio/src/settings/SettingsMenu.jsx
```diff
@@ -61,6 +65,7 @@ class SettingsMenu extends Component {
   }
 
   handleClose() {
+    ContextActionUtils.enableAllActions();
```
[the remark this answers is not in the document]
mattrunyon: Good catch

packages/code-studio/src/settings/Shortcuts.ts
mattrunyon: Nope. I think this was a remnant from when I initially made a list of contexts for shortcuts

packages/code-studio/src/settings/ShortcutsSectionContent.tsx
mattrunyon: Split ShortcutItem to its own file. It started a lot smaller. Darn added functionality requiring more code

mofojed (approved):

PR #98, opened 2021-07-07
convert util TestUtils to TS

spasovski:


packages/utils/src/TestUtils.ts
```diff
@@ -1,5 +1,28 @@
+interface MockContext {
+  arc: () => void;
+  beginPath: () => void;
+  clip: () => void;
+  closePath: () => void;
+  createLinearGradient: () => void;
```
mattrunyon, line 6: I think this should be `createLinearGradient: jest.Mock<{ addColorStop: jest.Mock<void> }>`

packages/utils/src/TestUtils.ts
```diff
@@ -1,5 +1,28 @@
...
+  rect: () => void;
+  restore: () => void;
+  setTransform: () => void;
+  save: () => void;
+  stroke: () => void;
+  strokeRect: () => void;
+  translate: () => void;
+  scale: () => void;
+  createPattern: () => void;
+}
```
mattrunyon, line 22: I think these would be more accurate if they were `jest.Mock<void>` except for the 2 which return something. Those should be `jest.Mock<{ width: number }>` and `jest.Mock<{ addColorStop: jest.Mock<void> }>`

The first generic is the return type (defaults to any). You can add a 2nd generic which is the arg type, but it defaults to any. I think we should leave as default instead of explicitly stating `jest.Mock<void, never>` because it's likely many of these get called with params, but we don't care what they get called with in the tests

mattrunyon (changes requested):

mattrunyon (approved):

PR #101, opened 2021-07-08
Add validation for renaming files

### LINKED ISSUE #91: Notebook UI: New item validation
Re-enable validation in the NewItemModal

mofojed:
- Inline Renaming now validates files correctly
- Rename from the right click menu also validates correctly
- Prompt to overwrite existing file
- Fixes #91

### DIFF
--- a/packages/file-explorer/src/NewItemModal.tsx
+++ b/packages/file-explorer/src/NewItemModal.tsx
@@ -16,6 +16,7 @@ import FileStorage, { FileType } from './FileStorage';
 import FileUtils from './FileUtils';
 
 import './NewItemModal.scss';
+import FileExistsError from './FileExistsError';
 
 const log = Log.module('NewItemModal');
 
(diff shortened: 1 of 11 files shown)


packages/file-explorer/src/NewItemModal.tsx
```diff
@@ -332,10 +345,13 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
...
-          if (!(e instanceof FileExistsError) || type === 'directory') {
-            throw e;
+          if (e instanceof FileExistsError) {
+            if (type !== 'directory' && e.info.type !== 'directory') {
+              this.setState({ showOverwriteModal: true });
+              return;
+            }
           }
-          this.setState({ showOverwriteModal: true });
+          throw e;
```
vbabich, line 354: I'm getting an error `Cannot overwrite existing directory` when trying to save a file with the same name as an existing directory. Should be allowed, I think.

vbabich (changes requested):
Should this change include validation for the inline rename field? I'm getting `404 Not Found` error when using the inline rename.

packages/file-explorer/src/NewItemModal.tsx
mofojed: As discussed in meeting, it should not be allowed to rename a file to overwrite a directory.

vbabich (approved):

PR #103, opened 2021-07-08
Make btn CSS selectors more specific

### LINKED ISSUE #102: Fix import order of Bootstrap and our CSS
Update Problem
-
Find a fix for CRA bundling bootstrap with the main CSS chunk but OSS component css with the vendor chunk in Enterprise. This causes bootstrap to import after our classes instead of before.

To test locally, copy the `packages/components` folder into enterprise at `node_modules/@deephaven/components`. Symlinking won't work properly. Then do a build and use something like `serve -s build -l 3000` to test the production version of enterprise.

Original Problem/Solution
-
Currently, the CSS selectors for many button styles are the same specificity as Bootstrap's selectors (1 class name).

This can cause a problem where CSS import order determines which rule takes effect when we use multiple classes on items. For example, MenuItems have class names `btn btn-navigation-menu-item`. Both classes set display type differently. The CSS import order then takes precedence when we always want our display setting from `btn-navigation-menu-item` to take priority.

Adjust the selectors for `.btn-*` rules to be `.btn-*, .btn.btn-*`. This should prevent any rules from breaking if we didn't apply both classes, but also make our selector more specific than Bootstrap's and always take precedence regardless of CSS import order.

This is a more robust fix for #42

mattrunyon:
Fixes #102 

Made any CSS rules with a single class selector starting with `.btn-*` also include a selector for `.btn.btn-*`


dsmmcken (changes requested):
We aren't sharing bootstrap as customized via the custom override, or the index.scss styles, I think it makes sense to move that into components, and hopefully that also fixes our ordering issue.

- Remove from code studio index.jsx the `import bootstrap`  in core and enterprise, it does nothing.
- publish index.scss from as part of the components/scss/index package (which imports bootstrap)
- import that componets/index.scss file in index.jsx in both core and enterprise

mattrunyon:
I tested importing the `index.scss` file in enterprise and it results in the same error.

It does dedupe some code though so my proposal is to leave this fix and then I'll adjust the files in branch where I was testing enterprise importing the scss files from OSS and make that its own package. Just makes things a bit cleaner since the SCSS imports rely on relative paths for bootstrap imports.

dsmmcken:
I would rather just the .btn.btn version rather than both, to make underspecified components more obviously broken. 
(and ditch the stylelint disables).

We need to thoroughly review enterprise though and look for other potential breakages that are non-btn related.

mattrunyon:
For some reason force pushing this branch closed the PR?

It didn't merge into main. No clue why github did that

PR #108, opened 2021-07-12
File Explorer drag and drop to move files

### LINKED ISSUE #86: Notebook UI: Drag and Drop to move
Related to #72 

Add drag and drop behaviour to move files into File Explorer

mofojed:
- Add back the drag and drop to move files functionality
- Update selection after drop
- Emit signal for all renamed files
- Fixed Shift+Click for selection, added unit tests
Fixes #86 

I didn't end up consolidating with ItemList - I was trying to use DraggableItemList which wasn't quite right since it has a lot of styling we don't need, was going to change to ItemList but then realized I'd have to do all the drag and drop styling again, figured I may just be wasting time.


mofojed:
Replaced with #109 , which just uses ItemList and gets rid of SingleClickItemList

PR #113, opened 2021-07-14
Properly register console actions and console disconnected messages

### LINKED ISSUE #112: Allow Console to properly register additional actions and a disconnect view
For Enterprise we need to have the ability to disconnect/restart a console session, and display a message when the session has disconnected. Add ability to do that.

Related to DH-11323.

mofojed:
- ContextActions can now be registered by the parent container instead of passed in
- Overflow actions are passed in and shown
- Pass in a node for displaying when console is disconnected unexpectedly

Fixes #112

### DIFF
--- a/packages/console/src/Console.jsx
+++ b/packages/console/src/Console.jsx
@@ -729,18 +729,6 @@ export class Console extends PureComponent {
 
   getObjects = memoize(objectMap => [...objectMap.values()]);
 
-  getContextActions = memoize(actions => [
-    ...actions,
-    {
-      action: this.handleClearShortcut,
-      shortcut: SHORTCUTS.CONSOLE.CLEAR,
-    },
-    {
-      action: this.handleFocusHistory,
-      shortcut: SHORTCUTS.CONSOLE.FOCUS_HISTORY,
-    },
-  ]);
-
   addCommand(command, focus = true, execute = false) {
     if (!this.consoleInput.current) {
       return;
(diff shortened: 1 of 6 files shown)


packages/console/src/Console.jsx
```diff
@@ -825,6 +810,8 @@ export class Console extends PureComponent {
     } = this.state;
     const consoleMenuObjects = this.getObjects(objectMap);
     const inputMaxHeight = Math.round(consoleHeight * 0.7);
+    const isDisconnected = disconnectedChildren != null;
```
spasovski, line 813: Nit: This is only used once, though it does make the JSX more readable.

packages/console/src/Console.jsx
```diff
@@ -753,6 +729,18 @@ export class Console extends PureComponent {
 
   getObjects = memoize(objectMap => [...objectMap.values()]);
 
+  getContextActions = memoize(actions => [
```
spasovski, line 732: Not sure where this is used.

packages/console/src/Console.jsx
mofojed: What do you mean it's only used once? `isDisconnected` is used in a few places, I figured it was easier than putting `disconnectedChildren != null` everywhere.

mofojed: What do you mean it's only used once? `isDisconnected` is used in a few places, I figured it was easier than putting `disconnectedChildren != null` everywhere.

packages/console/src/Console.jsx
mofojed: Good catch!

packages/console/src/Console.jsx
spasovski: Ah I see. Sorry was looking at the new additions only...ignore this.

spasovski (approved):
Looks good!

mofojed:
Found a couple other issues I needed to address before merging

spasovski (approved):
Good catch on `contextActions`.

PR #116, opened 2021-07-15
Fix import order of Bootstrap and our CSS

### LINKED ISSUE #102: Fix import order of Bootstrap and our CSS

mattrunyon:
Fixes #102 

This should ensure that bootstrap's classes are always included earlier than all of our CSS in both OSS and Enterprise.

This is the OSS part of the enterprise fix. The bug occurs in enterprise when using a production build.

To test locally, copy the `packages/components` folder into enterprise at `node_modules/@deephaven/components`. Symlinking won't work properly. Then do a build and use something like `serve -s build -l 3000` (npm module) to test the production version of enterprise.

The CSS should be correct. Easy one to find is the table menu buttons are way too big if the import order is wrong.


package.json
```diff
@@ -10,7 +10,7 @@
     "clean:build": "lerna run clean --stream",
     "clean:modules": "lerna clean --yes",
     "clean": "npm run clean:build && npm run clean:modules",
-    "bootstrap": "lerna bootstrap --hoist -- --no-audit --prefer-offline",
+    "bootstrap": "lerna bootstrap --hoist --nohoist=bootstrap -- --no-audit --prefer-offline",
```
dsmmcken, line 13: what's the purpose of this?

mattrunyon: `custom.scss` imports from the Bootstrap version in node_modules. Since we're shipping that scss file to enterprise without compiling it, it still needs to be able to import Bootstrap. Sass has no real module resolution, so in OSS the import would be 3 or 4 levels up, but in enterprise it would be just 2 levels up if bootstrap were also installed to enterprise.

Nohoist emulates the package structure (with regular dependencies) when installed via npm. So importing `../node_modules/boostrap/...` will work in OSS and enterprise as both will have bootstrap in the components node_modules folder

dsmmcken (approved):

PR #117, opened 2021-07-16
Convert components tests to typescript

### LINKED ISSUE #111: Finished converting @deephaven/components test files to TypeScript
There's still a few test files that have not been converted to TypeScript in the @deephaven/components package:

image

mattrunyon:
Closes #111


packages/components/src/ItemList.test.tsx
```diff
@@ -17,8 +17,8 @@ function makeItemList({
   rowHeight = 20,
   offset = 0,
   items = makeItems(),
-  onSelect = () => {},
-  onViewportChange = () => {},
+  onSelect = () => false,
+  onViewportChange = () => false,
```
mofojed, line 21: `false`? Why that instead of `undefined`?

packages/components/src/MaskedInput.tsx
```diff
@@ -34,7 +34,7 @@ type MaskedInputProps = {
   /** One or more examples of valid values. Used when deciding if next keystroke is valid (as rest of the current value may be incomplete) */
   example: string | string[];
   /** The current selection to use for the input */
-  selection: SelectionSegment | null;
+  selection?: SelectionSegment | null;
```
mofojed, line 37: If it's optional, there shouldn't be a need for `null` as well:

```suggestion
  selection?: SelectionSegment;
```

packages/components/src/MaskedInput.tsx
```diff
@@ -472,6 +472,11 @@ MaskedInput.defaultProps = {
     // no-op
   },
   getNextSegmentValue: (range, delta, segmentValue) => segmentValue,
+  selection: {
```
mofojed, line 475: Why was a default prop needed?

Random side note, was reading this in the TypeScript cheat sheet: https://github.com/typescript-cheatsheets/react/blob/main/README.md#you-may-not-need-defaultprops

packages/components/src/TimeInput.test.tsx
```diff
@@ -160,12 +173,16 @@ describe('arrow left and right jumps segments', () => {
    * @param {number|Array<number>} movement Keyboard movement to simulate, positive for right, negative for left. Eg. 2 means 2 right arrow presses, -3 means 3 left arrow presses
```
mofojed, line 173: You can remove the types from the comments, since the types are specified in the function definition now.

mofojed (changes requested):

packages/components/src/MaskedInput.tsx
mattrunyon: The linter plugin has a false positive with forwardRef components

https://github.com/yannickcr/eslint-plugin-react/issues/2856

packages/components/src/ItemList.test.tsx
mattrunyon: No reason. Maybe I didn't feel like typing out undefined that day. I changed it to undefined now.

packages/components/src/MaskedInput.tsx
mattrunyon: Yep. I think at times when converting these I erred on adding `| null` since we use null to indicate that we explicitly set no value sometimes. But in this case it's just the starting value in the `TimeInput` component so I set it to undefined.

packages/components/src/TimeInput.tsx
```diff
@@ -31,7 +31,9 @@ const TimeInput = React.forwardRef<HTMLInputElement, TimeInputProps>(
       onBlur = () => false,
     } = props;
     const [value, setValue] = useState(TimeUtils.formatTime(propsValue));
-    const [selection, setSelection] = useState<SelectionSegment | null>(null);
+    const [selection, setSelection] = useState<SelectionSegment | undefined>(
+      undefined
+    );
```
mofojed, line 36: You don't need to specify `undefined` with `useState`, there's a convenience overload when argument is omitted that returns `SelectionSegment | undefined`:

```suggestion
    const [selection, setSelection] = useState<SelectionSegment>();
```

From `@types/react`:
```
    /**
     * Returns a stateful value, and a function to update it.
     *
     * @version 16.8.0
     * @see https://reactjs.org/docs/hooks-reference.html#usestate
     */
    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
    // convenience overload when first argument is omitted
    /**
     * Returns a stateful value, and a function to update it.
     *
     * @version 16.8.0
     * @see https://reactjs.org/docs/hooks-reference.html#usestate
     */
    function useState<S = undefined>(): [S | undefined, Dispatch<SetStateAction<S | undefined>>];
```

mofojed (changes requested):
When you fix up, remember to re-request the review from me

packages/components/src/TimeInput.tsx
mattrunyon: Nice

mofojed (approved):

mofojed (approved):

PR #120, opened 2021-07-19
Get console type from the API, don't allow switching language

### LINKED ISSUE #58: UI: Select script session language
Unfortunately we can't properly support multiple script sessions without a bit of work to ensure that compiler context / query library / and query scope are populated correctly from various callback hooks (such as table listeners).

The work around for now is to allow a single global script session. In deephaven/deephaven-core#592, I have this plumbed in via a docker environment variable. Ideally, the UI would offer you the list of support sessions, then you pick one. After the initial session is created, the new GetConsoleTypes gRPC method needs to return that global session type instead of all supported types. If the UI sees just a single console type supported; then it can auto-create the console.

mofojed:
Fixes #58

- The server does not support switching languages, remove that functionality
- Get the console types (languages) from the connection, use the first one as default
- Requires server changes https://github.com/deephaven/deephaven-core/pull/863

### DIFF
--- a/packages/code-studio/src/dashboard/panels/ConsolePanel.jsx
+++ b/packages/code-studio/src/dashboard/panels/ConsolePanel.jsx
@@ -133,6 +133,10 @@ class ConsolePanel extends PureComponent {
 
       log.info('Available types:', types);
 
+      if (types.length === 0) {
+        throw new Error('No console types available');
+      }
+
       const type = types[0];
 
       log.info('Starting session with type', type);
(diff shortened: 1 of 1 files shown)


vbabich (dismissed):
Can `getConsoleTypes` potentially resolve with an empty array?

mofojed:
> Can `getConsoleTypes` potentially resolve with an empty array?

That's a good point. I'll add a case to handle that instead of assuming it's not empty.

spasovski (approved):
Looks good.

PR #121, opened 2021-07-20
Add prompt when closing tab

### LINKED ISSUE #84: Prompt before closing console
Right now if you hit the keyboard shortcut to close the browser tab, it does not warn you and it could close and you lose your data.

mofojed:
- Add onbeforeunload listener at the app level to prompt when closing the tab
- For testing, you need to have interacted with the page (click or keyboard) or else prompt will not show
- Removed unused disconnect/restart session wiring

Fixes #84


dsmmcken (approved):
I am just assuming this doesn't impact enterprise.

Issue #21, opened 2021-03-17
Convert all js in grid to ts


mofojed:
Convert everything in the grid package to TypeScript. This will be particularly useful with the grid metrics and just helping grid usage.

mofojed:
`GridModel` and `TableColumnFormatter` should probably change to interfaces or abstract classes.

PR #123, opened 2021-07-20
Convert Linker to TypeScript

mofojed:
- Convert the Linker to TypeScript
- Precursor to fully separating dashboard out #23


packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -248,21 +321,21 @@ class Linker extends Component {
...
         return;
       }
 
-      // FILTER_SOURCE links have a limit of 1 link per target
+      // filterSource links have a limit of 1 link per target
       // New link validation passed, delete existing links before adding the new one
-      if (type === LinkType.FILTER_SOURCE) {
+      if (type === 'filterSource') {
         const { links } = this.props;
         const existingLinkPanelId = isReversed ? start.panelId : end.panelId;
```
mattrunyon, line 333: This might be safer in a switch/case statement since TS will check it exhaustively. If we were to add a new link state for any reason, in theory it should immediately catch that it's unhandled here (or caught by a default)

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -152,7 +222,7 @@ class Linker extends Component {
       link => link.start?.panelId === panelId || link.end?.panelId === panelId
```
mattrunyon, line 222: My VSCode is yelling at me about this being implicitly any. I think it's something to do with the `ConnectedProps` in `LinkerProps` because when I remove that this (and 4 or 5 other spots) have proper types instead of any

We don't have a linter rule and I'm not sure that one exists for it. I found some that are close. They require a bit of config w/ eslint since you have to tell it about the TS runtime so it can do type analysis

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -29,15 +29,85 @@ import {
...
+  timeZone: string;
+}
+
+interface OwnProps {
+  layout: GoldenLayout;
+  panelManager: PanelManager;
+  localDashboardId: string;
+}
+
+const mapState = (state: LinkerState, ownProps: OwnProps) => ({
```
mattrunyon, line 82: This needs an explicit return type currently since it's inferring from JS as any. So it has the right structure, but every value is any. I believe it's related to the `ConnectedProps` problem I mentioned further down in this file

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -29,15 +29,85 @@ import {
...
+  activeTool: getActiveTool(state),
+  isolatedLinkerPanelId: getIsolatedLinkerPanelIdForDashboard(
+    state,
+    ownProps.localDashboardId
+  ),
+  links: getLinksForDashboard(state, ownProps.localDashboardId),
+  timeZone: getTimeZone(state),
+});
+
+const connector = connect(mapState, {
```
mattrunyon, line 92: Also probably needs explicit return type (or actions moved to TS). This one may be fixed by `mapState` having an explicit return type. Or you can specify `connect<State>(...)` possibly

packages/code-studio/src/dashboard/linker/LinkerLink.tsx
```diff
@@ -15,28 +14,52 @@ const TRIANGLE_HYPOTENUSE = Math.sqrt(
...
-export class LinkerLink extends PureComponent {
-  static makeCirclePath(x, y, r) {
+export type LinkerLinkProps = {
+  x1: number;
+  y1: number;
+  x2: number;
+  y2: number;
+  id: string;
+  className: string;
+  onClick: (id: string) => void;
```
mattrunyon, line 24: Since the param names in function types are just for documentation, you might be able to make the param name more descriptive? `clickedId` or `thisId` or `linkId`? Or leave it. Doesn't really matter

packages/code-studio/src/dashboard/linker/LinkerOverlayContent.test.tsx
```diff
@@ -6,7 +6,7 @@ import LinkerOverlayContent from './LinkerOverlayContent';
 const LINKER_OVERLAY_MESSAGE = 'TEST_MESSAGE';
 
 function makeEventHub() {
-  const callbacks = {};
+  const callbacks = {} as Record<string, (arg: unknown) => void>;
```
mattrunyon, line 9: I think you can move the type before the equals. `const callbacks: Record<> = {};`. Not sure if there's any difference between the 2

packages/code-studio/src/dashboard/linker/LinkerOverlayContent.tsx
```diff
@@ -1,45 +1,78 @@
...
   }
 
-  componentDidMount() {
+  componentDidMount(): void {
     window.addEventListener('mousemove', this.handleMouseMove, true);
   }
 
   // eslint-disable-next-line react/sort-comp
-  componentDidCatch(error, info) {
+  componentDidCatch?(error: Error, info: ErrorInfo): void {
```
mattrunyon, line 66: What is the question mark for?

packages/code-studio/src/dashboard/linker/LinkerOverlayContent.tsx
```diff
@@ -121,15 +154,14 @@ export class LinkerOverlayContent extends Component {
...
-      .filter(item => item != null);
+      .filter(item => item != null) as VisibleLink[];
 
     return (
       <div className="linker-overlay">
         <svg>
           {visibleLinks.map(({ x1, y1, x2, y2, id, className }) => (
             <LinkerLink
               className={className}
-              disabled={disabled}
```
mattrunyon, line 132: Why was this deleted with no replacement?

packages/iris-grid/src/formatters/TableColumnFormatter.js
```diff
@@ -39,7 +39,7 @@ class TableColumnFormatter {
 
   /**
    * @param {Object} value The value to format
-   * @param {Object} format Optional format object with value transformation options
+   * @param {number} format Optional format object with value transformation options
```
mattrunyon, line 42: The changed type and comment don't align. Type is number, comment says object

packages/code-studio/src/dashboard/linker/LinkerUtils.tsx
```diff
@@ -3,10 +3,37 @@ import { TableUtils } from '@deephaven/iris-grid';
...
 
+export type LinkPointUntyped = {
+  panelId: string;
+  panelComponent?: string;
+  columnName: string;
+};
+
+export type LinkPoint = LinkPointUntyped & {
+  columnType: string;
+};
```
mattrunyon, line 17: Idk if changing these to `LinkPoint` and `TypedLinkPoint` would be better. Not sure about naming conventions really. Or `LinkPointWityType`.

I also only see `LinkPointUntyped` used in 1 place and I'm not sure why these couldn't just be a single type. If you wanted a partial type, you could use `Pick<LinkPoint, 'panelId' | 'columnName'>` in that spot instead if you only want to enforce it has those 2 props and they're always the same type as `LinkPoint` here.

packages/code-studio/src/dashboard/linker/LinkerUtils.tsx
```diff
@@ -98,13 +139,14 @@ class LinkerUtils {
    * @param {string} cloneId Cloned panel id
    * @returns {object[]} Cloned links array or empty array if no new links added
```
mattrunyon, line 140: Remove types from comment

packages/code-studio/src/dashboard/linker/LinkerUtils.tsx
```diff
@@ -98,13 +139,14 @@ class LinkerUtils {
...
    */
-  static cloneLinksForPanel(links, panelId, cloneId) {
-    const clonedLinks = [];
+  static cloneLinksForPanel(
+    links: Link[],
+    panelId: string,
+    cloneId: string
+  ): Link[] {
+    const clonedLinks = [] as Link[];
```
mattrunyon, line 147: Could use `const clonedLinks: Link[] = [];`. Again not aware of any difference for either syntax

mattrunyon (changes requested):

packages/code-studio/src/dashboard/linker/LinkerOverlayContent.tsx
mofojed: The prop didn't exist on LinkerLink at all. It just passes a `.disabled` class on the className now.

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -1,7 +1,7 @@
-import React, { Component } from 'react';
-import { connect } from 'react-redux';
-import PropTypes from 'prop-types';
+import React, { Component, ErrorInfo } from 'react';
+import { connect, ConnectedProps } from 'react-redux';
 import shortid from 'shortid';
+import GoldenLayout from 'golden-layout';
```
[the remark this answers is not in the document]
mofojed: Nice

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -29,15 +29,85 @@ import {
...
+  glEventHub: GoldenLayout.EventEmitter;
+};
+
+export type Panel = Component<PanelProps>;
+
+export type LinkFilterMapValue = {
+  columnType: string;
+  text: string;
+  value: unknown;
+};
```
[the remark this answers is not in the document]
mofojed: Yea good call, I'll add the template to default to T = unknown

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -29,15 +29,85 @@ import {
...
+
+export type LinkFilterMap = Map<string, LinkFilterMapValue>;
+
+export type LinkDataMapValue = {
+  type: string;
+  text: string;
+  value: string;
+};
+
+export type LinkDataMap = Record<string, LinkDataMapValue>;
```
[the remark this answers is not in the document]
mofojed: It's the way it was. Not changing any of the APIs right now

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -29,15 +29,85 @@ import {
...
+
+export type LinkablePanel = Panel & {
+  setFilterMap: (filterMap: LinkFilterMap) => void;
+  unsetFilterValue: () => void;
```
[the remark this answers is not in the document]
mofojed: Yea true. I think I prefer the arrow style, since it's the similar syntax as if you're doing any other type (can mark readonly, make it optional, etc), and I think you'd have to do it like that anyway if you were doing a union or anything like that (such as `ContextAction[] | (() => ContextAction[])`. 
I don't really know the advantages the other way - you save a couple chars, and when you hover over it, it shows up as `method` instead of `property`... there's probably something I'm missing.

packages/code-studio/src/dashboard/linker/LinkerUtils.tsx
mofojed: Yea I just got rid of LinkPointUntyped - I don't think it was actually needed.

packages/code-studio/src/dashboard/linker/LinkerUtils.tsx
mofojed: Yea I don't care too much one way or the other. I guess having the type explicitly after the variable name is clearer than inferring the type with an `as` later

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -76,16 +146,16 @@ class Linker extends Component {
     }
   }
 
-  componentDidCatch(error, info) {
+  componentDidCatch?(error: Error, info: ErrorInfo): void {
```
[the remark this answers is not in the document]
mofojed: ??? I don't know how that got there

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -184,49 +254,52 @@ class Linker extends Component {
...
    */
   columnSelected(
-    panel,
-    column,
+    panel: Panel,
+    column: LinkColumn,
     isAlwaysEndPoint = false,
-    overrideIsolatedLinkerPanelId = null
+    overrideIsolatedLinkerPanelId?: string
   ) {
```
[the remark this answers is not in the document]
mofojed: It did not. I should probably look into _why_ it's not showing up in the linter. I've added the return type.

packages/code-studio/src/dashboard/linker/Linker.tsx
mofojed: Yup adding the return type here fixes the `any` problem.

packages/code-studio/src/dashboard/linker/LinkerLink.tsx
mofojed: I'm going to leave it

packages/code-studio/src/dashboard/linker/LinkerOverlayContent.tsx
mofojed: I think I copied from the d.ts and it had it there. Fixed

packages/code-studio/src/dashboard/linker/Linker.tsx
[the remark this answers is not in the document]
mattrunyon: https://stackoverflow.com/questions/39156315/function-property-vs-method

Looks like you can apply readonly to a property, but not a method. Also there are some more strict checks if generics are used for a property vs if it is a method.

Linked in that thread is this page which has an example https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#example

So it looks like as a property might be the slightly safer type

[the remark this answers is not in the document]
mofojed: I'll stick with property for now

packages/code-studio/src/dashboard/linker/Linker.tsx
```diff
@@ -41,10 +41,10 @@ export type PanelProps = {
...
 
-export type LinkFilterMapValue = {
+export type LinkFilterMapValue<T = unknown> = {
   columnType: string;
   text: string;
-  value: unknown;
+  value: T;
 };
 
 export type LinkFilterMap = Map<string, LinkFilterMapValue>;
```
mattrunyon, line 50: I think with the generic we might want to add it here as well with a default.
```suggestion
export type LinkFilterMap<T = unknown> = Map<string, LinkFilterMapValue<T>>;
```

packages/iris-grid/src/formatters/DateTimeColumnFormatter.ts
```diff
@@ -1,17 +1,19 @@
 /* eslint class-methods-use-this: "off" */
 import dh from '@deephaven/jsapi-shim';
 import Log from '@deephaven/log';
-import TableColumnFormatter from './TableColumnFormatter';
+import TableColumnFormatter, {
+  TableColumnFormat,
```
mattrunyon, line 5: Just FYI I looked into the `import type` more and what it does is enforces that something that isn't a type can only be used for its types within the file.

If you use a regular import for a class and then only use it for its types, it will be stripped from the compiled JS import. So I guess import type probably isn't needed other than to be explicit. Pretty sure that previous comment I had about `import type` would result in the same code regardless of the import type.

Basically import type should help if there's a build failure due to the isolated modules flag and other than that it's just us being explicit. Isolated modules flag will cause an error if there's some imports that. I think this example is an example of that flag causing a failure: https://github.com/microsoft/TypeScript/issues/28481#issuecomment-441284387

Basically re-exporting types is problematic for Babel because it only looks at a single file. So it has no idea if the re-exported value is a type or a value (I think it errs on considering it a value). That can result in unnecessary/broken imports in the final code

https://github.com/microsoft/TypeScript/pull/35200#issuecomment-645565073

packages/iris-grid/src/formatters/TableColumnFormatter.ts
```diff
@@ -0,0 +1,69 @@
+/* eslint class-methods-use-this: "off" */
+/* eslint no-unused-vars: "off" */
```
mattrunyon, line 2: Curious why this rule is ignore. Can't think of any reason unused vars would be needed in a file

packages/iris-grid/src/formatters/TableColumnFormatter.ts
```diff
@@ -0,0 +1,69 @@
...
+  static TYPE_CONTEXT_CUSTOM: TableColumnFormatType = 'type-context-custom';
+
+  /**
+   * Validates format object
+   * @param format Format object
+   * @returns true for valid object
+   */
+  static isValid(format: TableColumnFormat): boolean {
+    return true;
+  }
```
mattrunyon, line 33: Can you add a comment that says why this just returns true? Is it something that is a TODO, is it something that should be added by the consumer of this class? Something we were initially checking client-side but switched to server-side and it's just for maintaining existing code?

packages/iris-grid/src/formatters/TableColumnFormatter.ts
```diff
@@ -0,0 +1,69 @@
...
+  static isSameFormat(
+    formatA?: TableColumnFormat,
+    formatB?: TableColumnFormat
+  ): boolean {
+    throw new Error('isSameFormat not implemented');
+  }
+
+  /**
+   * Returns format object
+   */
```
mattrunyon, line 50: Add param/return tags

packages/iris-grid/src/formatters/DateTimeColumnFormatter.ts
```diff
@@ -128,11 +157,11 @@ class DateTimeColumnFormatter extends TableColumnFormatter {
     );
   }
 
-  getEffectiveFormatString(baseFormatString) {
+  getEffectiveFormatString(baseFormatString: string): string {
     return this.formatStringMap.get(baseFormatString) || baseFormatString;
   }
 
-  format(value, format) {
+  format(value: unknown, format?: TableColumnFormat): string {
```
mattrunyon, line 164: Can this value be set using the generic in `TableColumnFormatter`? Or is it always unknown? If always unknown, the `TableColumnFormatter` class might not need to have a generic type

packages/iris-grid/src/formatters/TableColumnFormatter.ts
mattrunyon: Nevermind I saw the comment at the top of the file saying it's the default formatter

Should we maybe use `abstract` classes and methods in this case? At least the class should be `abstract` I think if it's supposed to be extended and not directly used. Not sure if we want defaults for the methods or mark some `abstract` so the implementation has to define it

mattrunyon (changes requested):

packages/iris-grid/src/formatters/TableColumnFormatter.ts
mofojed: Yea it _should_ be an abstract class, or possibly even just an interface. I added a comment on #21 to consider that change. I didn't want to change any of the existing functionality as I was just changing the one class so I don't get errors in Linker

packages/iris-grid/src/formatters/DateTimeColumnFormatter.ts
mofojed: I think always `unknown`. I will revisit when converting grid to TS (#21)

packages/iris-grid/src/formatters/TableColumnFormatter.ts
mofojed: I think it might've been for params when it was a js file - for some reason it complains about unused params in JS, in TS it does not. I've removed since this file is now ts.

mattrunyon (approved):

Issue #122, opened 2021-07-20
Bring jest up to root of lerna


mofojed:
We can save resources running tests by using one common jest runner in the root web-client-ui dir instead of configuring it for each project:
Multiple project configs:
https://jestjs.io/docs/configuration#projects-arraystring--projectconfig

Using Jest with lerna:
https://github.com/facebook/jest/issues/3112#issuecomment-398581705

mattrunyon:
Also might be a good idea to update the GitHub action to cache jest and only run tests with changed files since the PR base branch. Could make a new ticket for that too

mofojed:
Should also implement Code Coverage reporting as an option

PR #129, opened 2021-07-23
Import golden-layout as a package

### LINKED ISSUE #20: Import golden-layout as a package
We currently have a fork of golden-layout: https://github.com/illumon-public/golden-layout/

However, the newest version of GoldenLayout dropped React support: https://github.com/golden-layout/golden-layout

They recommend switching the FlexLayout, but there's issues in FlexLayout that @dsmmcken has already fixed (such as handling long tab names, overflow when you have more tabs in the stack that can be shown that doesn't suck, being able to drag a tab and have the edge of the stack scroll, a tab placeholder that represents the actual content size, etc).

Since our previous fork will never be able to merge back to their fork, we may as well import it into our repo. We should also publish an actual package with an updated version, rather than referencing it by SHA/old repo.

Eventually we should either switch to FlexLayout (and fix the above issues that we've already fixed), or rewrite our golden-layout without jquery.

mattrunyon:
Fixes #20

I combined the license files in a single LICENSE file in our golden-layout repo. The last commit is the updates of usages of golden-layout within the app. First 2 commits are just bringing golden-layout into the repo, but not depending on it.


mofojed:
Now with the updated build process and version in `illumon-public/golden-layout`, I think the friction we've experienced as part of it being a separate repo is much less now. My main hesitation now with bringing in golden-layout to our monorepo would be related to performance - how much does this increase our startup/runtimes? I wonder if we should wait until #122 is complete

packages/golden-layout/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
```
mofojed, line 1: Doing an `npm install` or `npm run clean` causes a bunch of tracked changes:

```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    packages/golden-layout/dist/goldenlayout.js
        deleted:    packages/golden-layout/dist/goldenlayout.min.js
        modified:   packages/golden-layout/package-lock.json
        deleted:    packages/golden-layout/src/css/goldenlayout-base.css
        deleted:    packages/golden-layout/src/css/goldenlayout-base.css.map
        deleted:    packages/golden-layout/src/css/goldenlayout-dark-theme.css
        deleted:    packages/golden-layout/src/css/goldenlayout-dark-theme.css.map
        deleted:    packages/golden-layout/src/css/goldenlayout-light-theme.css
        deleted:    packages/golden-layout/src/css/goldenlayout-light-theme.css.map
        deleted:    packages/golden-layout/src/css/goldenlayout-soda-theme.css
        deleted:    packages/golden-layout/src/css/goldenlayout-soda-theme.css.map
        deleted:    packages/golden-layout/src/css/goldenlayout-translucent-theme.css
        deleted:    packages/golden-layout/src/css/goldenlayout-translucent-theme.css.map
```

packages/golden-layout/build/ns.js
```diff
@@ -0,0 +1 @@
+var lm={"config":{},"container":{},"controls":{},"errors":{},"items":{},"utils":{}};
```
mofojed, line 1: What's up with this file and why is it in a `build` folder? Any reason it shouldn't be in the `src` folder?

Not going to block merge for this since this isn't something we wrote, but we should clean it up unless there's a reason not to.

mofojed (changes requested):
package-lock.json is also changing on `npm install`. 

Should maybe just add `dist` and `src/css` to `.gitignore`? Why does scss output to src/css actually, shouldn't that be in the dist folder?

packages/golden-layout/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "test": "test"
+  },
+  "dependencies": {
+    "jquery": "^3.5.1"
+  },
+  "scripts": {
+    "build": "cross-env NODE_ENV=production run-s clean build-steps",
+    "build-dev": "cross-env NODE_ENV=development run-s clean build-steps",
+    "build-steps": "run-p build:*",
+    "sass": "sass ./src/scss:./src/css -s compressed",
```
mofojed, line 34: Can this output to dist/css instead?

mofojed (changes requested):

mattrunyon:
Waiting for #145 first

packages/golden-layout/package.json
mattrunyon: Moved to dist/css and ignored the dist folder

Updated our imports of the css to use the dist folder

packages/golden-layout/build/ns.js
mattrunyon: It's the build starter file I guess (the build process is just concatenating several JS files)

I moved it to src and renamed it `base.js` since `ns.js` has no meaning to me

packages/golden-layout/.gitignore
mattrunyon: Test this again. I think it should be fixed

mattrunyon:
@dsmmcken Can you build this locally and make sure that everything looks right with golden-layout?

packages/code-studio/src/layout/golden-layout.js
```diff
@@ -1,7 +1,7 @@
 import './jquery';
 import './ReactDOM';
 
-import 'golden-layout';
-import 'golden-layout/src/css/goldenlayout-base.css';
-import 'golden-layout/src/css/goldenlayout-dark-theme.css';
+import '@deephaven/golden-layout';
+import '@deephaven/golden-layout/dist/css/goldenlayout-base.css';
+import '@deephaven/golden-layout/dist/css/goldenlayout-dark-theme.css';
 import './GoldenLayout.scss';
```
dsmmcken, line 7: Not your breakage, but noticed while testing:

in lm_header in the GoldenLayout.scss, within codestudio, I think I added overflow: hidden to fix this issue:
image

where extremely narrow panels icons overflow, but it entirely breaks the tab list dropdown, so I think it's best to remove the overflow: hidden for now.

image

Rest looks fine.

dsmmcken: I can't think of an easy fix for that with out re-parenting that last in the dom.

dsmmcken (approved):
Thanks. This should be much easier to manage fixes in GL now, especially with it being updated to build with current node etc.

mofojed (approved):

PR #132, opened 2021-07-28
Update workflow_dispatch action to take a git SHA

mofojed:
workflow_dispatch action allows you to select the branch you want to execute the workflow from, however it doesn't let you select a PR that's currently in review if it's created from a fork.

By allowing a SHA to be input, we should be able to input the commit from the PR and it will publish from that.


spasovski (approved):
We did some battle-testing of this yesterday. Worked great!

Issue #64, opened 2021-06-09
Error message when unable to find API is non-intuitive


mofojed:
**Steps to Reproduce**
1) Start up the web UI without having the server started up

**Expected Results**
1) An error message indicating the API couldn't be found

**Actual Results**:
1) A pretty confusing error message, see screenshot

I think could check in `index.jsx` and display an appropriate error message, or something to that effect.

![screenshot738](https://user-images.githubusercontent.com/4505624/121340550-c30ac780-c8ed-11eb-9f7e-8d2a136dcac5.png)

PR #133, opened 2021-07-28
Show an error message if the API is not found

mofojed:
Write out an error message when the API can't load.

Since we depend on the API for our initialization (eg. ChartBuilder depends on `dh.plot.SeriesPlotStyle.*` to initialize it's static types), we can't count on our app fully loading and/or executing our index.jsx entrypoint. It may be possible to style this to look better, but it may be a pain. I think this way gets the message across instead of just giving the user a blank screen.

### DIFF
--- a/packages/code-studio/public/index.html
+++ b/packages/code-studio/public/index.html
@@ -36,12 +36,13 @@
     <script
       src="%REACT_APP_CORE_API_URL%/%REACT_APP_CORE_API_NAME%"
       type="text/javascript"
+      onerror="document.write('Unable to load API. Ensure the server is running and you are able to reach %REACT_APP_CORE_API_URL%/%REACT_APP_CORE_API_NAME%, then refresh the page')"
     ></script>
     <title>Deephaven</title>
   </head>
-  <body style="background:black">
+  <body>
     <noscript> You need to enable JavaScript to run this app. </noscript>
-    <div id="root">Broken</div>
+    <div id="root"></div>
     <!--
       This HTML file is a template.
       If you open it directly in the browser, you will see an empty page.
--- a/packages/code-studio/public/index.html
+++ b/packages/code-studio/public/index.html
@@ -39,9 +39,9 @@
     ></script>
     <title>Deephaven</title>
   </head>
-  <body>
+  <body style="background:black">
     <noscript> You need to enable JavaScript to run this app. </noscript>
-    <div id="root"></div>
+    <div id="root">Broken</div>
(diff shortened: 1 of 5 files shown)


mofojed:
Nevermind this only works in dev, not in a production build :(

packages/components/scss/new_variables.scss
```diff
@@ -6,7 +6,7 @@ $spacer-3: map-get($spacers, 3);
 $spacer-4: map-get($spacers, 4);
 $spacer-5: map-get($spacers, 5);
 
-/* Marching Ants for golden layout dropzone and drag and drop */
+//Marching Ants for golden layout dropzone and drag and drop
```
mofojed, line 9: I changed this because when I was inspecting the styles outputted, this comment was in every <style> tag in <head>. I don't know why it includes it if it's in `/* */`, but I just changed it to // to be consistent with the rest of the comment and so it doesn't appear it _every_ style tag.

packages/code-studio/public/index.html
```diff
@@ -37,6 +37,12 @@
       src="%REACT_APP_CORE_API_URL%/%REACT_APP_CORE_API_NAME%"
       type="text/javascript"
     ></script>
+    <script>
```
dsmmcken, line 40: I would do this immediately before the </body>

```
    <script>
      (function () {
        // If for some reason the API doesn't load, our app will not work. Write out a message to help the user.
        if (!window.dh) {
          document.getElementById('root').insertAdjacentHTML(
            'afterbegin',
            `<div class="modal  d-block">
              <div class="modal-dialog modal-dialog-centered theme-bg-light">
                <div class="modal-content">
                  <div class="modal-body">
                    <h5 class="modal-title">Error: Unable to load API</h5>
                    <p class="text-break">Ensure the server is running and you are able to reach %REACT_APP_CORE_API_URL%/%REACT_APP_CORE_API_NAME%, then refresh the page.</p>
                  </div>
                </div>
              </div>
            </div>`
          );
        }
      })();
    </script>
```

dsmmcken (commented):
Unrelated, but why not fix the advanced filter creator error, there's an explicit proptype we have in the shim.
```
import { dh, PropTypes as DhPropTypes } from '@deephaven/jsapi-shim';
...
filters: PropTypes.arrayOf(DhPropTypes.FilterCondition).isRequired,
```

mofojed:
Fixes #64

dsmmcken (approved):

PR #135, opened 2021-07-28
Create a publish nightly action

mofojed:
Right now it's publishing to GitHub packages. Once we've verified it's working there, then I'll update to publish to npmjs


spasovski (approved):
Looks good - I like the "beta" preid.

PR #137, opened 2021-07-29
Disable aggregations if they're not enabled

### LINKED ISSUE #136: Some table features are shown even though they do not work correctly
See https://github.com/deephaven/deephaven-core/issues/37 and https://github.com/deephaven/deephaven-core/issues/697. These options are shown even though they are not enabled on the server.

Make iris-grid customizable as to which options to shown, and only show options that are available.

mofojed:
- Related to https://github.com/deephaven/deephaven-core/pull/931
- Rollup, column statistics already checked if the methods existed, just needed to add the same check for totals
- Fixes #136 

Requires this PR on core well: https://github.com/deephaven/deephaven-core/pull/931

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModel.js
+++ b/packages/iris-grid/src/IrisGridTableModel.js
@@ -313,7 +313,7 @@ class IrisGridTableModel extends IrisGridModel {
   }
 
   get isTotalsAvailable() {
-    return true;
+    return this.table.getTotalsTable != null;
   }
 
   get isEditable() {


packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -313,7 +313,7 @@ class IrisGridTableModel extends IrisGridModel {
   }
 
   get isTotalsAvailable() {
-    return true;
+    return this.table.getTotalsTable != null;
```
spasovski, line 316: Curious why you're loose checking null and not strict `undefined`.

mofojed: No reason in particular. It wouldn't work if it was null anyway.

One day I'll convert all the grid stuff to TypeScript, and it'll be easier to be strict checking.

spasovski (approved):
Looks good.

PR #141, opened 2021-07-30
Bring Jest up to the root of lerna

### LINKED ISSUE #122: Bring jest up to root of lerna

mattrunyon:
Fixes #122 

GitHub action should only run tests that changed or the source changed since the base branch

Code-studio tests still fail. Need to fix those before making this a full PR


mattrunyon:
Closing to move it to a PR from forked repo to ensure the changes will work with that workflow

PR #143, opened 2021-08-09
Fix the nightly publish action

mofojed:
- Add a `check_date` step that ensures it only runs if there's been a commit in the last 24 hours
- Add the `canary` flag so it publishes a proper version (instead of bringing up a prompt)


spasovski (approved):
👍

PR #145, opened 2021-08-10
Bring Jest up to the root of lerna

### LINKED ISSUE #122: Bring jest up to root of lerna

mattrunyon:
Fixes #122

GitHub action should only run tests that changed or the source changed since the base branch


packages/code-studio/jestBabelTransform.js
```diff
@@ -0,0 +1,32 @@
+// This was created from ejecting a blank CRA-4 project w/ TypeScript support
+// Code-studio tests do not run properly without using this transform
```
mofojed, line 2: Any more detail on why they don't run without this transform? I'm a little confused about this part.

mofojed (approved):

packages/code-studio/jestBabelTransform.js
mattrunyon: Fixed to use babel-jest transform just for code-studio instead of ts-jest. This removes the need for the transform from CRA

mofojed (approved):

PR #147, opened 2021-08-11
Application Mode - UI

### LINKED ISSUE #127: Export/import layout
For Application Mode, UI needs to be able to export and import a layout. Provide that functionality.

### LINKED ISSUE #165: Weird space behaviour from monaco
Select text in a monaco editor that has indentations. It goes wacky when selected.

https://user-images.githubusercontent.com/1576283/130279720-0c232dba-1ec8-4a68-b2cc-308b2bac4225.mp4

**Versions**
 - Deephaven: Version 0.2.1
 - OS: Win
 - Browser: Chome
 - Docker: ...

mofojed:
- Panels menu added to include the objects available
- Export/Import layout from the Panels menu
- Layouts are loaded from the `data/layouts` directory. If the user doesn't already have a layout, the first layout from the layouts directory is used.
- If the layout loaded is empty, then show an empty dashboard screen with an auto fill button and hint.
- Requires JS API to be added (https://github.com/deephaven/deephaven-core/issues/854)
- Update how fonts preload

Fixes #127 
Fixes #165 

BREAKING CHANGES: The `#preload-fonts` id has been removed from @deephaven/components


dsmmcken (dismissed):
UX is fine 👍

packages/code-studio/src/redux/selectors.js
```diff
@@ -0,0 +1,14 @@
+// Session
+/**
+ * Retrieve the current session and it's config
+ * @param {Store} store The redux store that is used
+ * @returns {{ session: dh.IdeSession, config: { type: string }}} The active session and it's options
```
vbabich, line 5: I think there's also `id` in the `config`. Maybe use the `SessionConfig` type here?
TBH, `getSession` returning an object that is not a session is a bit confusing. I can't come up with a better name, ok to leave as is unless you have any suggestions.

packages/file-explorer/src/WebdavFileStorage.ts
```diff
@@ -52,9 +52,17 @@ export class WebdavFileStorage implements FileStorage {
   }
 
   async loadFile(name: string): Promise<File> {
-    const content = await this.client.getFileContents(name, {
+    let content = await this.client.getFileContents(name, {
       format: 'text',
     });
+
+    // There appears to be a bug where data is returned as an object instead of a string
```
vbabich, line 59: Already fixed, can be dropped.

vbabich (changes requested):

packages/code-studio/src/main/WebdavLayoutStorage.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    return files.map(file => file.basename);
+  }
+
+  async getLayout(name: string): Promise<ItemConfigType[]> {
+    let content = await this.client.getFileContents(`${this.root}${name}`, {
+      format: 'text',
+    });
+
+    // For some reason, even though we specify the format as 'text', it comes back as whatever it wants
```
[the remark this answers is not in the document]
mofojed: No

packages/code-studio/src/main/WidgetList.tsx
```diff
@@ -0,0 +1,230 @@
...
+};
+
+export type SelectStartEvent = {
+  x: number;
+  y: number;
+  widget: WidgetDefinition;
+};
+
+export interface WidgetListProps {
+  onSelect: (widget: WidgetDefinition, e?: _MouseEvent) => undefined;
```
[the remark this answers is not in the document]
mofojed: NativeMouseEvent, I'll make it clearer

packages/code-studio/src/main/WidgetList.tsx
```diff
@@ -0,0 +1,230 @@
...
+      }
+      selectStartEvent.current = undefined;
+    },
+    [handleMouseMove, selectStartEvent, sendSelect]
+  );
+
+  const filteredWidgets = useMemo(
+    () =>
+      widgets.filter(widget =>
+        searchText ? widget.name.includes(searchText) : true
```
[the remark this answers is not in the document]
mofojed: Nope, good catch!

packages/code-studio/src/redux/selectors.js
mofojed: I do agree that `getSession` returning not a session is confusing... `getSessionWrapper`?

mofojed: `getSessionWrapper`

packages/code-studio/src/dashboard/DashboardContainer.jsx
```diff
@@ -103,6 +107,17 @@ export class DashboardContainer extends Component {
     this.init();
   }
 
+  componentDidUpdate(prevProps) {
+    const { layoutConfig } = this.props;
+    const { layoutConfig: currentLayoutConfig } = this.state;
```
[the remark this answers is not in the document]
mofojed: No, the `layoutConfig` in the state is what's emitted from the GoldenLayout after it has loaded, it's different than the prop that is passed in. If I don't keep a ref to it, then it gets stuck in a load cycle:
prop = layoutConfigA
GoldenLayout loads, emits layoutConfigB
prop changes to layoutConfigB, which is diff than prevProp.layoutConfigA, so reloads
GoldenLayout loads, emits layoutConfigC

etc.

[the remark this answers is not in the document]
vbabich: Got it, thanks

vbabich (approved):

packages/components/src/BaseStyleSheet.scss
```diff
@@ -37,19 +37,6 @@ body {
   }
 }
 
-#preload-fonts {
```
dsmmcken, line 40: enterprise uses this class, unless you have a fix to remove it from enterprise, but really the same preload fix should happen in it's app loader too, 

note that preloading is required not just before monaco init, but also any iris-grid init.

mofojed: When updating Enterprise to the latest published package I'll add the fix then, as it's a breaking change.

dsmmcken (approved):

vbabich (approved):

PR #148, opened 2021-08-11
Add IrisGrid.setFilters method

vbabich:
Add `setFilters` method for the Filter View Manager

### DIFF
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -965,7 +965,14 @@ export class IrisGrid extends Component {
     this.startLoading('Rebuilding filters...', true);
     this.setState({
       quickFilters: newQuickFilters,
-      advancedFilter: newAdvancedFilters,
+      advancedFilters: newAdvancedFilters,
+    });
+  }
+
+  setFilters({ quickFilters, advancedFilters }) {
+    this.setState({
+      quickFilters,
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -965,7 +965,14 @@ export class IrisGrid extends Component {
     this.startLoading('Rebuilding filters...', true);
     this.setState({
       quickFilters: newQuickFilters,
-      advancedFilter: newAdvancedFilters,
+      advancedFilters: newAdvancedFilters,
```
mofojed, line 968: Great catch!

mofojed (approved):
The fix is good, adding the `setFilters` method doesn't really change much though, are you sure it's necessary? It's just calling `irisGrid.setFilters({ quickFilters, advancedFilters })` instead of `irisGrid.setState({ quickFilters, advancedFilters })`
I guess it's more explicit. Your call

vbabich:
> The fix is good, adding the `setFilters` method doesn't really change much though, are you sure it's necessary? It's just calling `irisGrid.setFilters({ quickFilters, advancedFilters })` instead of `irisGrid.setState({ quickFilters, advancedFilters })`
> I guess it's more explicit. Your call
Yea, the fix doesn't add much, I just thought direct `setState` across different packages is a bad practice. A bit safer with a method call.

PR #150, opened 2021-08-12
Improve console error message handling

### LINKED ISSUE #149: Console Error selection is difficult
**Description**
When trying to select the text of an error message in the console history, it is very difficult to do.

**Steps to reproduce**
1. Open a console session in the web UI
2. Enter a command that will return the error
3. Click the command to expand it
4. Try and select text by clicking outside of the error and then dragging into it
5. Try and extend a selection by holding Shift and then clicking in the error

**Expected results**
4. The selection should stay as you have selected
5. The selection should extend to where you clicked

**Actual results**
4. The error collapses
5. The error collapses for this as well

mofojed:
Before, it would collapse the error message often while you were trying to select text. A couple of fixes:
- Sets `isClicking` on mouse down to start the click. Prevents cases where you drag into the error message from outside collapsing the error message
- Don't register the click if the user is holding Shift - they could be trying to update the selection

Fixes #149

### DIFF
--- a/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
+++ b/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
@@ -21,7 +21,7 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
 
     this.mouseX = null;
     this.mouseY = null;
-    this.isDragging = false;
+    this.isClicking = false;
 
     this.state = {
       isExpanded: false,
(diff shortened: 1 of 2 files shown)


packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
```diff
@@ -56,18 +56,23 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
         Math.abs(event.clientY - this.mouseY) >=
           ConsoleHistoryResultErrorMessage.mouseDragThreshold
       ) {
-        this.isDragging = true;
+        this.isClicking = false;
       }
+    } else if (this.isClicking) {
+      // Rare case - could happen if you mouse down, switch window focus, release the mouse, then come back, mouse down outside of the error, drag into the error, then release the mouse
```
spasovski, line 62: Nit: wrap this line.

spasovski (approved):
Looks good and this should work. Components like this are ripe with unpredictable edge cases unfortunately.

Issue #151, opened 2021-08-12
Unexpected File Explorer Click Handling


mattrunyon:
**Description**

The file explorer currently handles clicks in an unexpected way. There are many cases to handle. "Preview mode" files have italic names on the panel tab. "Normal mode" files have regular font names on the panel tab.

The main problem is with single click actions. Currently, the `ItemList` component simply has an `onSelect` handler. This can be set to select on double click, but then the single click actions are not used. We emit `NotebookEvent.SELECT_NOTEBOOK` which focuses the notebook based on if it is currently open or not. So really double clicks are not handled and instead it is the 2nd single click event which focuses the notebook and editor.

`ItemList` needs a double click handler (can use regular mouse click event and check `e.detail === 2`). We may also want to adjust `onSelect` to emit info about what kind of select action was used (keyboard, mouse, etc.)

GoldenLayout may be focusing the created tab (even though we don't focus the editor) on a single click event currently. Need to figure out how to maintain focus on the clicked item for single clicks.

**Single Clicks**

Any single click should always maintain focus on the file explorer item. Doesn't matter if the file is open or not. Always keep focus on the file explorer.

If the file is not open, open it in preview mode.
If the file is open in preview mode, bring the tab into view, but do not focus it.
If the file is open in normal mode, bring the tab into view, but do not focus it.

**Double Clicks**

Double clicks should always open the file in normal mode and focus the editor. The cursor should in the editor should start blinking.

**Steps to reproduce**
1. Open the file explorer
2. Single click on a file

**Expected results**

File explorer keeps focus

**Actual results**

Focus moves to the notebook tab.

**Versions**
 - Deephaven: 0.2.1

PR #152, opened 2021-08-12
Fix Some Shortcut Problems

### LINKED ISSUE #142: Various issues with keyboard shortcuts.
**Description**

Some of the expected keyboard shortcuts are not working as expected. 
These manifest differently between windows and macos. 

I found these on Deephaven Enterprise but will effect OSS as well. 

**Windows**
Ctrl + L 
Expected clear console. Actually brings up linking tool. 

Ctrl + H  when history tab is not active
Expected history tab becomes active and gets focus. Actually nothing happens. 

Delete when file explorer tab is active
Expected delete notebook dialog. Actually delete command is entered into currently active notebook. 

Enter when file explorer tab is active
Expected change name dialog. Actually enter command is entered into currently active notebook. 

Ctrl + S on Input Table
Expected commit changes. Actually brings up search bar. 

Ctrl + E
Expected clear all table filters on all tables. Actually moves focus to title bar.

**MacOs**

Control + C
Expected clearing console history. Actually nothing happens. 

Cmd + Shift + E
Expected clearing filters on currently active table. Actually all filters on all tables are cleared. 

Cmd + S on Input table
Expected to commit input table changes. Actually toggles search bar. 

Cmd + E 
Expected clear all table filters on all tables. Actually highlights Edit but no action is preformed. 

**Versions**
 - Deephaven Enterprise: 1.20210201.147beta
 - Web Browser: crhome

mattrunyon:
Fixes #142 (mostly). See #151 which will address remaining shortcut problems.

Removes the global shortcut handlers from being attached to the capture phase instead of bubble phase. This makes local shortcuts take precedence over global shortcuts. Also means linker won't trigger when inside of monaco, but monaco has its own shortcuts and Don said we should just leave monaco shortcuts alone instead of removing them so our shortcuts work globally (really it's only linker)

### DIFF
--- a/packages/code-studio/src/main/AppControlsMenu.jsx
+++ b/packages/code-studio/src/main/AppControlsMenu.jsx
@@ -169,7 +169,7 @@ const AppControlsMenu = props => {
           onClearFilter();
         },
         order: 60,
-        shortcut: GRID_SHORTCUTS.TABLE.CLEAR_FILTERS,
+        shortcut: GRID_SHORTCUTS.TABLE.CLEAR_ALL_FILTERS,
       },
     ],
     [handleControlSelect, handleToolSelect, onClearFilter]
--- a/packages/code-studio/src/main/AppControlsMenu.jsx
+++ b/packages/code-studio/src/main/AppControlsMenu.jsx
@@ -7,7 +7,7 @@ import React, {
 } from 'react';
 import classNames from 'classnames';
 import { DropdownMenu, GLOBAL_SHORTCUTS } from '@deephaven/components';
-import { SHORTCUTS as GRID_SHORTCUTS } from '@deephaven/iris-grid';
+import { SHORTCUTS as IRIS_GRID_SHORTCUTS } from '@deephaven/iris-grid';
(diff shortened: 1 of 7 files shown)


packages/code-studio/src/main/AppControlsMenu.jsx
```diff
@@ -169,7 +169,7 @@ const AppControlsMenu = props => {
           onClearFilter();
         },
         order: 60,
-        shortcut: GRID_SHORTCUTS.TABLE.CLEAR_FILTERS,
+        shortcut: GRID_SHORTCUTS.TABLE.CLEAR_ALL_FILTERS,
```
mofojed, line 172: It bugs me that you named it `GRID_SHORTCUTS` here, but `IRIS_GRID_SHORTCUTS` in AppMainContainer. Would like it to be consistent.

mofojed (approved):

mofojed (changes requested):
You'll need to fix that failing test

mofojed (approved):

PR #153, opened 2021-08-13
Hide the suggestion widget when it's showing a status message

### LINKED ISSUE #144: Sometimes an empty tooltip appears for autocomplete when editing a console input
**Description**

**Steps to reproduce**
1. Focus in the console input
2. Type something that does not autocomplete to anything, eg `qwert`, then press Tab

**Expected results**
2. No empty tooltip appears

**Actual results**
2. The tooltip is empty

Selection_048

mofojed:
We were already supposed to be doing this, but we weren't hiding the right element so it was showing an empty box.
Update it to hide the correct element.

Fixes #144

### DIFF
--- a/packages/components/src/BaseStyleSheet.scss
+++ b/packages/components/src/BaseStyleSheet.scss
@@ -651,12 +651,10 @@ input[type='number']::-webkit-inner-spin-button {
   display: none;
 }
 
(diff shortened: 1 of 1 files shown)


mattrunyon (approved):
Shows no tooltip if there are no suggestions

PR #155, opened 2021-08-13
Improve context menu selection in Command History/Item List

### LINKED ISSUE #154: Bug when copying commands from history panel
Margaret reported in slack: https://illumon.slack.com/archives/G01K6UHJ00K/p1628008266061600

Quoting:
I'm getting a bug when I copy commands my from history panel
I've copied the new table script from the history, and it's pasting that snapshot thing instead

Don says is related to this JIRA: https://deephaven.atlassian.net/browse/DH-9097

mofojed:
- Right-click now updates the selection before showing context menu
- Fixes #154

### DIFF
--- a/packages/golden-layout/package-lock.json
+++ b/packages/golden-layout/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@deephaven/golden-layout",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/packages/components/src/ItemList.test.tsx
+++ b/packages/components/src/ItemList.test.tsx
@@ -219,6 +219,42 @@ describe('mouse', () => {
     it('extends selection when shift+right-click outside current selection', () => {
       testContextMenu(3, 6, [[3, 6]], { shiftKey: true });
     });
+
+    it('maintains selection if right-clicked item is selected', () => {
+      const onSelect = jest.fn();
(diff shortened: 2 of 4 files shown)


dsmmcken (changes requested):
Logic should be if the context click occurs on a selected row that is part of an existing selection , it should be maintained. (To match standard OS behaviour)

Only create a new selection if context menu click occurs on a blank row.

ie.

Select items a, b, c,
Context click on b, and copy
selection maintained, copied contents should be abc, not just b.

Context click on unselected item z, and copy
select z, and copied contents should be z

mofojed:
@dsmmcken If I'm being very accurate, should I also style context-selected items differently? On Mac OS it looks like they get an outline when context selected, as opposed to the primary selection format, eg:

Clicking outside existing selection:
image

Clicking inside existing selection:
image

dsmmcken:
Counterpoint, win10 changes the selection, with no additional styling. So I would say we are safe to skip the extra styling work.

right click

packages/golden-layout/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@deephaven/golden-layout",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
dsmmcken, line 3: should this be here?

dsmmcken (approved):
Approved for behaviour.

vbabich (approved):

PR #156, opened 2021-08-13
Fix up styling of console find widget

### LINKED ISSUE #118: Fix console log button background
The clear and settings icons appear to have a transparent background. If there are long log messages, they appear behind these icons such as below

image

mofojed:
- Added background behind console menu buttons
- Aligned find widget with top of log view
- Fixes #118

### DIFF
--- a/packages/console/src/log/LogView.scss
+++ b/packages/console/src/log/LogView.scss
@@ -8,9 +8,11 @@ $log-pane-buttons-width: 48px;
 
 .log-pane-menu {
   position: absolute;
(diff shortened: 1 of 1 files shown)


mattrunyon:
image

If you resize the log window then the messages don't seem to reflow properly and end up under the scroll bar.

But the issue I reported for this ticket is fixed, so make a new ticket for resizing messages ending up under the scroll bar if you don't want to do it now

mattrunyon (approved):

PR #174, opened 2021-08-31
Make scrollbar styling consistent

dsmmcken:
This change updates scrollbars (monaco, grid, native) to all look _similar_. Since we can't control the monaco scrollbar styling, we instead match it. This style should look better for adding "selection-ticks to grid scrollbars" eventually anyways.

Monaco:
- Unchanged

Native:
- Changes native scroll bars in blink/webkit to be styled similar monaco scrollbars
- Adds a light scrollbar variant for areas in the app that use a light bg
- Remove explicitly setting colors in firefox, in favor of just 'auto'

Grid:
- Changes grid scroll bar styling to have just an inner casing, rather than both sides
- Makes scroll thumb square instead of rounded to match
- Make it slight wider by default
- Adjusts colors

image
(left: monaco, bottom-right: grid, top-right: native)

### DIFF
--- a/packages/components/src/BaseStyleSheet.scss
+++ b/packages/components/src/BaseStyleSheet.scss
@@ -25,7 +25,7 @@ body {
   padding: 0;
   overscroll-behavior: none;
   -ms-scroll-chaining: none;
-  scrollbar-color: auto; //applies to firefox only
+  scrollbar-color: rgba($white, 0.1) rgba($black, 0.1); //applies to firefox only
 }
 
 #root {
@@ -648,6 +648,7 @@ input[type='number']::-webkit-inner-spin-button {
 // used in inverted color sections, like light modals
 .theme-bg-light {
   --scrollbar-color: 0, 0, 0;
+  scrollbar-color: rgba($black, 0.5) rgba($black, 0.3); // firefox
 }
 
 /********** Monaco Overides *********/
--- a/packages/components/src/BaseStyleSheet.scss
+++ b/packages/components/src/BaseStyleSheet.scss
@@ -648,6 +648,7 @@ input[type='number']::-webkit-inner-spin-button {
 // used in inverted color sections, like light modals
 .theme-bg-light {
   --scrollbar-color: 0, 0, 0;
+
   scrollbar-color: rgba($black, 0.5) rgba($black, 0.3); // firefox
 }
 
--- a/packages/components/src/BaseStyleSheet.scss
+++ b/packages/components/src/BaseStyleSheet.scss
@@ -25,6 +25,7 @@ body {
   padding: 0;
   overscroll-behavior: none;
   -ms-scroll-chaining: none;
+  scrollbar-color: auto; //applies to firefox only
 }
 
 #root {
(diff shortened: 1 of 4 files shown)


packages/components/src/BaseStyleSheet.scss
```diff
@@ -25,7 +25,7 @@ body {
   padding: 0;
   overscroll-behavior: none;
   -ms-scroll-chaining: none;
-  scrollbar-color: rgba($white, 0.1) rgba($black, 0.1); //applies to firefox only
+  scrollbar-color: auto; //applies to firefox only
```
mofojed, line 28: I'm a little confused why you made this auto? When dragging in firefox now the scrollbar is red.

https://user-images.githubusercontent.com/4505624/131732375-4b4e5a48-bfc5-481c-9969-1e4da839dd4a.mp4

packages/grid/src/GridRenderer.js
```diff
@@ -2069,19 +2069,18 @@ class GridRenderer {
       context.fillRect(
         width - rowHeaderWidth - scrollBarSize + scrollBarCasingWidth,
         height - columnHeaderHeight - scrollBarSize + scrollBarCasingWidth,
-        scrollBarSize - scrollBarCasingWidth * 2,
-        scrollBarSize - scrollBarCasingWidth * 2
+        scrollBarSize - scrollBarCasingWidth,
```
mofojed, line 2072: Ah yes, the casing is only on one side now

packages/iris-grid/src/IrisGridTheme.module.scss
```diff
@@ -24,13 +24,13 @@ $selection-outline-casing-color: $gray-900;
...
-  scroll-bar-color: $gray-400;
-  scroll-bar-hover-color: $gray-300;
-  scroll-bar-active-color: $gray-200;
+  scroll-bar-bg: $gray-850;
+  scroll-bar-hover-bg: $gray-800;
+  scroll-bar-casing-color: $gray-600;
+  scroll-bar-corner-color: $gray-850;
+  scroll-bar-color: $gray-500;
+  scroll-bar-hover-color: $gray-400;
+  scroll-bar-active-color: $gray-300;
```
mofojed, line 33: This gray isn't matching the native scroll bar gray when dragging in Chrome

mofojed (changes requested):
I think you want the colours to be the same at least between native and grid? Right now they're all different, in the idle, hover, or dragging stage.

Also you _can_ customize the monaco scroll bar a bit: https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditorscrollbaroptions.html
and some colour options:
https://microsoft.github.io/monaco-editor/playground.html#customizing-the-appearence-exposed-colors

packages/components/src/BaseStyleSheet.scss
dsmmcken: Oh firefox, my special little child. I guess it goes wild on ubunutu, didn't realize that.

I'll revert that I guess.

packages/iris-grid/src/IrisGridTheme.module.scss
dsmmcken: I am going to say that is acceptable, I don't want to introduce a new color step just for the grid active state, and since they are all constructed differently, this is close enough.

mofojed (approved):

PR #177, opened 2021-09-02
Wire up minRange/maxRange to plotly

nbauernfeind:
Fixes deephaven/deephaven-core#1112


nbauernfeind:
Also, can you help me understand what exactly the github action is complaining about and how I can go about making it happy?

mofojed:
@nbauernfeind as mentioned in Slack:
> Okay so the tests are complaining because you're adding a range, which has it fall into the if block at ChartUtils:811 if (getRangeParser && range && !autorange) {, which ends up calling TableUtils.isDateType, which isn't a function in our mock in packages/chart/src/__mocks__/@deephaven/iris-grid.js . So should just add that function to the mock.
As for plotly not updating the zoom, you can listen for onSelected or onSelecting... my guess is that by explicitly setting the range, we need to explicitly update it as well. Though not seeing that in the docs, haven't dug in the source yet. So adding an event handler in Chart and updating the axes with the new ranges (I think)

spasovski:
@nbauernfeind let me know if you can still work on this or if you want us to take over or close it

spasovski:
This seems to be low-priority and we don't have the resources to attack it in the short-term. The issue will remain open but I'll close this PR for now.

PR #180, opened 2021-09-02
Using the http-proxy-middleware to proxy

mofojed:
Alternative to Don's PR #172 . Needed to add the `ws: true` bit so websockets worked for the api as well.

### DIFF
--- a/packages/code-studio/package.json
+++ b/packages/code-studio/package.json
@@ -32,6 +32,7 @@
     "cross-env": "^7.0.2",
     "deep-equal": "^2.0.4",
     "fira": "github:mozilla/fira#4.202",
+    "http-proxy-middleware": "^2.0.1",
     "jquery": "^3.5.1",
     "jszip": "3.2.2",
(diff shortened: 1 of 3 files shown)


packages/code-studio/package.json
```diff
@@ -32,6 +32,7 @@
     "cross-env": "^7.0.2",
     "deep-equal": "^2.0.4",
     "fira": "github:mozilla/fira#4.202",
+    "http-proxy-middleware": "^2.0.1",
```
dsmmcken, line 35: dev-dependencies

mofojed: The instructions do specify it as a dependency...: https://create-react-app.dev/docs/proxying-api-requests-in-development/

I'll try

mofojed: Looks like it worked, good call

dsmmcken: CI did not like that 🤷, sorry.

dsmmcken (approved):

PR #186, opened 2021-09-06
Enable Scala syntax highlighting

nbauernfeind:
Although this is not yet available via the DHC server, the change-set itself is valid and mergeable.

I did notice that the python implementation both is similar and dissimilar to the official monaco-languages python definition. We appear to be missing a handful of keywords. Also, I suspect that the python implementation needs to properly reference the original MIT license.

I wasn't sure of exactly how to make it clear the scala implementation was sourced from a piece of work under the MIT license. I took a first stab at it, but suspect it is inadequate.

I'd be happy to give a demo with or without notice. Looking forward to hearing how to proceed.

### DIFF
--- a/packages/code-studio/.env
+++ b/packages/code-studio/.env
@@ -18,7 +18,7 @@ BUILD_PATH=./build
 # Define the language key and display name that are available on the server.
 # Omit the display name to use the key as the display name.
 # Defaults to the first language, or the value of `REACT_APP_SESSION_LANGUAGE` if defined.
-REACT_APP_SESSION_LANGUAGES=python=Python,groovy=Groovy
+REACT_APP_SESSION_LANGUAGES=python=Python,groovy=Groovy,scala=Scala
 
 # We run our own eslint configuration as part of jest tests in src/test/eslint.test.js
 # We don't need to run it as part of react-scripts, we don't need it running on npm build or npm start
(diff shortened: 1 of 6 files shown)


mofojed (approved):
Licensing looks fine to me.

I don't know _why_ we're building scala support in the UI first without having it on the server first, but whatever.

The python/groovy ones are a little different than the Microsoft provided ones, James created them to parse out Deephaven Java code within strings (I think) so there's a bit of nesting there.

nbauernfeind:
> I don't know why we're building scala support in the UI first without having it on the server first, but whatever.

I am contributing a Scala solution purely from free time. That time is pretty limited (with children and all), and saving 15m each weekend is pretty beneficial to me. I have been committing changes to the core repository that are building blocks to support scala over-time, too. The JavaParser change, for example, ate at least two weekends worth of free time -- not including several weeknights. I would never finish if I couldn't commit finished pieces.

My ambitions here are huge. I appreciate being able to make progress one step at a time. Thank you for the review =).

I have a lot of use cases working from the re-imagined Scala REPL, but there are still more things to do. I am also contributing back to the REPL project that I've chosen to chase; I am still using locally built versions at this point. As I'm crossing t's and dotting i's, the syntax highlighting was becoming one of the larger pain points. Getting this into the next release means I can stop re-building, re-deploying, re-basing, etc the web-client-ui as I continue on the path to introducing something, I believe, is going to be very exciting even for people who don't have an interest in Scala.

P.S. I also have another idea for the web-client-ui that I'm building in addition to this change. Understanding the web-team's build, review, and merge process will save me time on any future contributions. (The process-learning goal is also the reason for the plotly change, which I have not yet finished.)

PR #194, opened 2021-09-10
Golden layout fallback for unregistered components

### LINKED ISSUE #184: Implement fallback for unregistered components in GoldenLayout
GoldenLayout currently throws an error and fails to render the layout with a config that contains unregistered components.
We should be able to specify a fallback component that would render in place of the unregistered components.

vbabich:
Fixes #184

### DIFF
--- a/packages/code-studio/src/dashboard/DashboardContainer.jsx
+++ b/packages/code-studio/src/dashboard/DashboardContainer.jsx
@@ -31,6 +31,7 @@ import {
   NotebookPanel,
   PandasPanel,
   PanelManager,
+  PanelPlaceholder,
 } from './panels';
 import MarkdownUtils from '../controls/markdown/MarkdownUtils';
 import {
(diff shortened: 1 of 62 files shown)


packages/golden-layout/src/js/utils/ReactComponentHandler.js
```diff
@@ -13,6 +13,7 @@ lm.utils.ReactComponentHandler = function (container, state) {
   this._container = container;
   this._initialState = state;
   this._reactClass = this._getReactClass();
+  console.log('ReactComponentHandler', this._reactClass, container, state);
```
mofojed, line 16: Remove extra console.log

packages/golden-layout/src/js/LayoutManager.js
```diff
@@ -209,15 +222,21 @@ lm.utils.copy(lm.LayoutManager.prototype, {
    * @returns {Function}
    */
   getComponent: function (name) {
-    if (this._components[name] === undefined) {
-      throw new lm.errors.ConfigurationError(
-        'Unknown component "' + name + '"'
-      );
-    }
-
+    console.log('getComponent', name);
```
mofojed, line 225: Remove console.log

packages/code-studio/src/dashboard/DashboardContainer.jsx
```diff
@@ -56,29 +57,6 @@ const log = Log.module('DashboardContainer');
 const RESIZE_THROTTLE = 100;
 
 export class DashboardContainer extends Component {
-  static dehydratePanelConfig(config) {
```
mofojed, line 59: I think I did something similar in my separating dashboards branch (moving dehydratePanelConfig to LayoutUtils)

mofojed (changes requested):
Great stuff, just remove the `console.log` statements and it should be good.

PR #202, opened 2021-09-17
Don't show custom context menu on right-click in Monaco editor

vbabich:
Fix the issue with an unusable Monaco context menu when Monaco is nested in a component having its own `ContextMenu` with a non-empty `actions` list.

### DIFF
--- a/packages/components/src/context-actions/ContextActions.tsx
+++ b/packages/components/src/context-actions/ContextActions.tsx
@@ -149,8 +149,13 @@ class ContextActions extends Component<
       let el = e.target as Element | null;
       while (el != null) {
         const { classList } = el;
-        if (ignoreClassNames.some(className => classList.contains(className))) {
-          log.debug2('Contextmenu event ignored based on the target className');
+        const ignoredClassName = ignoreClassNames.find(className =>
+          classList.contains(className)
+        );
+        if (ignoredClassName !== undefined) {
+          log.debug2(
+            `Contextmenu event ignored based on the target className "${ignoredClassName}"`
+          );
           return;
         }
         el = el.parentElement;
--- a/packages/components/src/context-actions/ContextActions.tsx
+++ b/packages/components/src/context-actions/ContextActions.tsx
@@ -150,6 +150,7 @@ class ContextActions extends Component<
       while (el != null) {
         const { classList } = el;
         if (ignoreClassNames.some(className => classList.contains(className))) {
+          log.debug2('Contextmenu event ignored based on the target className');
           return;
         }
         el = el.parentElement;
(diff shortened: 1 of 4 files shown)


packages/components/src/context-actions/ContextMenuRoot.tsx
```diff
@@ -56,6 +56,14 @@ class ContextMenuRoot extends Component<
   openMenu: React.RefObject<ContextMenu>;
 
   handleContextMenu(e: MouseEvent): void {
+    let el = e.target as Element | null;
+    while (el != null) {
+      if (el.classList.contains('monaco-editor')) {
```
mofojed, line 61: At a high-level, the `@deephaven/components` package shouldn't really have a dependency on `monaco-editor` (or anything monaco really). I see there's already some references to monaco-editor styling in `BaseStyleSheet.scss`, but it doesn't really "belong" there.

As for this... can we instead pass in an array of element classes we should ignore the context menu in? Or pass in a function for whether is should be handled or not? Leaving monaco out of this package entirely.

mofojed (changes requested):

packages/components/src/context-actions/ContextMenuRoot.tsx
vbabich: Agreed. Added `ignoredClassNames` prop, moved the class list check to the `ContextActions` component to make this change more decentralized.
I'm exposing the `INPUT_CLASS_NAME` const in `ConsoleInput` instead of just using a wrapper class, as there's an area around the input - paddings, ">" icon on the left, that should still display our custom context menu and not be ignored.

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -143,6 +144,17 @@ class ContextActions extends Component<
   container: React.RefObject<HTMLDivElement>;
 
   handleContextMenu(e: MouseEvent): void {
+    const { ignoreClassNames = [] } = this.props;
+    if (ignoreClassNames.length > 0) {
+      let el = e.target as Element | null;
+      while (el != null) {
+        const { classList } = el;
+        if (ignoreClassNames.some(className => classList.contains(className))) {
+          return;
```
mofojed, line 153: debug2 log might be helpful here

mofojed (dismissed):

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -150,6 +150,7 @@ class ContextActions extends Component<
       while (el != null) {
         const { classList } = el;
         if (ignoreClassNames.some(className => classList.contains(className))) {
+          log.debug2('Contextmenu event ignored based on the target className');
```
mofojed, line 153: Why not print the className?

vbabich: Good point, updated.

mofojed (approved):

PR #203, opened 2021-09-21
Separate dashboard into own package

### LINKED ISSUE #23: Separate dashboard into it's own package

mofojed:
- Create new package `@deephaven/dashboard` that contains the new `Dashboard.tsx`, which replaces `DashboardContainer.jsx`
- `Dashboard` uses plugins now instead of internally registering it's own list of panels and event handlers
- Dashboard plugins handle registration of panels, hydration/dehydration of those panels, the same functionality as event handlers
- All plugins currently written as hooks with Typescript, though they could be any React component
- Dashboard plugins pulled into a separate package, @deephaven/dashboard-core-plugins. Didn't feel a need to separate them further than that at this time.
- Updated redux actions to write dashboard data from @deephaven/dashboard, @deephaven/dashboard-core-plugins, pulled out of @deephaven/redux
- Fixes #23

### DIFF
--- a/packages/console/src/ConsoleShortcuts.ts
+++ b/packages/console/src/ConsoleShortcuts.ts
@@ -1,6 +1,7 @@
 import { ShortcutRegistry, KEY, MODIFIER } from '@deephaven/components';
 
 const CONSOLE = {
+  name: 'Console',
   CLEAR: ShortcutRegistry.createAndAdd({
     id: 'CONSOLE.CLEAR',
     name: 'Clear',
--- a/packages/dashboard-core-plugins/src/panels/index.js
+++ b/packages/dashboard-core-plugins/src/panels/index.js
@@ -12,3 +12,4 @@ export { default as PandasPanel } from './PandasPanel';
 export { default as Panel } from './Panel';
 export { default as WidgetPanel } from './WidgetPanel';
 export { default as WidgetPanelTooltip } from './WidgetPanelTooltip';
+export { default as SHORTCUTS } from './Shortcuts';
--- a/packages/file-explorer/src/index.ts
+++ b/packages/file-explorer/src/index.ts
@@ -4,6 +4,7 @@ export * from './FileExplorer';
 export * from './FileListContainer';
 export * from './FileList';
 export * from './FileStorage';
+export { default as SHORTCUTS } from './FileExplorerShortcuts';
 export { default as NewItemModal } from './NewItemModal';
 export { default as FileExplorerToolbar } from './FileExplorerToolbar';
 export { default as FileUtils } from './FileUtils';


packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+const log = Log.module('DashboardLayout');
+
+const EMPTY_OBJECT = Object.freeze({});
+
+const DEFAULT_LAYOUT_CONFIG: DashboardLayoutConfig = [];
+
+const DEFAULT_CALLBACK = () => undefined;
+
+type DashboardData = {
```
vbabich, line 35: It seems this should be somehow dependent on the registered plugins. I.e. contain `links` for the `Linker` plugin, etc.

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -0,0 +1,254 @@
...
+      return cleanup;
+    },
+    [hydrateMap, dehydrateMap, layout, store]
+  );
+  const hydrateComponent = useCallback(
+    (name, props) => hydrateMap.get(name)?.(props, id),
+    [hydrateMap, id]
+  );
+  const dehydrateComponent = useCallback(
+    (name, props) => dehydrateMap.get(name)?.(props, id),
```
vbabich, line 115: We need to try to dehydrate the components that are not in the map. Maybe add a fallback here.. or everywhere `dehydrateComponent` is called.

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const handleLayoutStateChanged = useCallback(() => {
+    // we don't want to emit stateChanges that happen during item drags or else
+    // we risk the last saved state being one without that panel in the layout entirely
+    if (isItemDragging) return;
+
+    const glConfig = layout.toConfig();
+    const contentConfig = glConfig.content;
+    const dehydratedLayoutConfig = LayoutUtils.dehydrateLayoutConfig(
+      contentConfig
```
vbabich, line 148: Should probably pass the `dehydrateComponent` here if we want to be able to use custom dehydrate functions

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -0,0 +1,254 @@
...
+      while (layout.root.contentItems.length > 0) {
+        layout.root.contentItems[0].remove();
+      }
+
+      // Add the new content. It is usally just one item from the root
+      for (let i = 0; i < content.length; i += 1) {
+        layout.root.addChild(content[i]);
+      }
+
+      // TODO: Wire up empty dashboard?
```
vbabich, line 218: TODO

packages/dashboard/src/PanelManager.ts
```diff
@@ -208,35 +279,42 @@ class PanelManager {
...
+  }
+
+  addClosedPanel(glContainer: GoldenLayout.Container): void {
     const config = LayoutUtils.getComponentConfigFromContainer(glContainer);
-    if (config && this.dehydrateComponentConfigMap[config.component]) {
-      const dehydratedConfig = this.dehydrateComponentConfigMap[
-        config.component
-      ](config);
+    if (config) {
+      const dehydratedConfig = this.dehydrateComponent(
```
vbabich, line 307: Make sure this doesn't throw on unregistered components

packages/golden-layout/src/js/LayoutManager.js
```diff
@@ -222,6 +232,12 @@ lm.utils.copy(lm.LayoutManager.prototype, {
    * @returns {Function}
    */
   getComponent: function (name) {
+    if (this._components[name] === undefined) {
```
vbabich, line 235: I think this would conflict with the fallback functionality

packages/dashboard/src/PanelManager.ts
```diff
@@ -195,9 +268,7 @@ class PanelManager {
     // Rehydrate the panel before adding it back
     const { component } = panelConfig;
     let { props } = panelConfig;
-    if (this.hydrateComponentPropMap[component]) {
-      props = this.hydrateComponentPropMap[component](props);
-    }
+    props = this.hydrateComponent(component, props);
```
vbabich, line 271: Make sure this doesn't break on unregistered components

vbabich (changes requested):

packages/golden-layout/src/js/LayoutManager.js
mofojed: I'm going to roll this and the other changes into my other cleanup changes (separating plugins into their own package).

packages/code-studio/src/dashboard/plugins/ConsolePlugin.tsx
```diff
@@ -0,0 +1,200 @@
...
+      ),
+    ];
+
+    return () => {
+      cleanups.forEach(cleanup => cleanup());
+    };
+  }, [registerComponent]);
+
+  useEffect(() => {
+    addMissingPanels();
```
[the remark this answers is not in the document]
mofojed: Not only does it not seem to work... but this might be functionality we should only have on Enterprise side? With application mode in community, we allow exporting of layouts, and re-importing without ConsolePanel or whatever panels you don't want. By adding them back, we're not doing what the user may expect.

packages/dashboard/src/Dashboard.tsx
```diff
@@ -0,0 +1,121 @@
...
+    fallbackComponent,
+    onGoldenLayoutChange,
+    setIsInitialized,
+    setLayout,
+  ]);
+
+  const handleResize = useMemo(
+    () =>
+      throttle(() => {
+        // TODO: This doesn't seem to be throttling
```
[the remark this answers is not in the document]
mofojed: Double checked, it does appear to be throttling. I must have had this comment when there was some other bug occurring, but it's not fixed.

packages/dashboard/src/PanelManager.ts
mofojed: This path doesn't get triggered on unregistered components, since they don't have a `Panel` implemented that emits the CLOSED event.
With that said... it'd probably make sense to have the Panel wrapper right at the dashboard level, but I'll leave that as a possible future change.

packages/golden-layout/src/js/LayoutManager.js
mofojed: Cleaned this up

vbabich: Can you push the update?

packages/code-studio/src/dashboard/plugins/ConsolePlugin.tsx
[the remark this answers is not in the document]
mofojed: I'm just going to remove this community side.

vbabich (changes requested):

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -32,16 +32,21 @@ const DEFAULT_LAYOUT_CONFIG: DashboardLayoutConfig = [];
 
 const DEFAULT_CALLBACK = () => undefined;
 
+// If a component isn't registered, just pass through the props so they are saved if a plugin is loaded later
+const FALLBACK_CALLBACK = (props: unknown) => props;
```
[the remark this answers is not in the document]
mofojed: @vbabich I'm not getting the error you described? It's showing up with the fallback component when I'm testing...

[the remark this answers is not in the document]
mofojed: I handled this in LayoutManager.toConfig

vbabich (dismissed):

PR #206, opened 2021-09-22
Fix up some issues with input tables

mofojed:
- Ctrl+S only triggers a save if the table is editable, has pending data, and there are no errors
- Ctrl+Alt+S only discards if the table is editable and there's pending data
- Save error message is now correctly displayed by PendingDataBottomBar
- Use a circle checkmark instead of the square checkbox

Related to https://deephaven.atlassian.net/browse/DH-11388


spasovski (approved):
looks good, wasn't aware of some of these model properties

PR #208, opened 2021-09-22
DH-11336 support Layout Hints on web

spasovski:
This handles `.hide()`, `.atFront()` and `.atEnd()`. A separate ticket and PR will address `.freeze()`.


packages/iris-grid/src/ColumnStatistics.scss
```diff
@@ -23,6 +23,10 @@
   white-space: nowrap;
 }
 
+.column-statistics-status {
+  margin: $spacer-1 0;
```
mofojed, line 27: In the screenshot on the ticket, the left side of the lock is aligned with the text. Adjust the margin to align it correctly.
The "Generate Statistics" text should also left align, may as well fix that up while you're in there as well.

packages/iris-grid/src/IrisGridMetricCalculator.js
```diff
@@ -8,6 +8,28 @@ import TableUtils from './TableUtils';
...
+  getVisibleColumnWidth(
+    column,
+    state,
+    firstColumn = this.getFirstColumn(state),
+    treePaddingX = this.calculateTreePaddingX(state)
+  ) {
+    const { model } = state;
+    const hiddenColumns = model.layoutHints?.hiddenColumns ?? [];
+
+    if (this.userColumnWidths.has(column)) {
```
mofojed, line 20: A confusing thing here - this method takes the _visible_ column index as the input, but if you've moved columns around, that will no longer match the _model_ column index.

You need to translate from the visible column index to the model column index. You can do that by calling `const modelColumn = this.getModelColumn(column, state);`. getModelColumn applies the moved column transforms to get the model index.

See `GridMetricCalculator.getVisibleColumnWidth` for an example.

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,8 +458,36 @@ class IrisGridTableModel extends IrisGridModel {
     return '';
   }
 
+  getColumnsWithLayoutHints = memoize((columns, hints) => {
+    if (hints) {
+      const frontColumns = new Array(hints.frontColumns.length);
+      const backColumns = new Array(hints.backColumns.length);
+      const middleColumns = [];
+
+      columns.forEach(col => {
```
mofojed, line 467: You could just allocate the array once and fill it in. eg.

```
const newColumns = new Array(columns.length);
let middleIndex = hints.frontColumns.length;
forEach...
if (frontColumn) {
  newColumns[indexOf] = col
} else if (backColumn) {
  // Get the correct back index
  newColumns[columns.length - indexOf] = col
} else {
  newColumns[middleIndex] = col;
  middleIndex += 1;
}
```
Also instead of doing `includes` and `indexOf`, which would require iterating the array twice (unless there's compiler optimization magic?), could just get the indexOf right away, and check if indexOf >= 0.

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -1044,8 +1072,17 @@ class IrisGridTableModel extends IrisGridModel {
     }
   );
 
-  isColumnMovable(x) {
-    return !this.isKeyColumn(x);
+  isColumnMovable(column) {
+    const columnName = Number.isInteger(column)
```
mofojed, line 1076: This should always be the columnIndex (a `number`) passed in, not the column name. Then with the column index, retrieve the column with just `this.columns[column].name`

This method is defined in `GridModel`, and `Grid` itself has no concept of columns with names, it just references columns by numbers for performance. Moved columns are applied as a transformation.

packages/iris-grid/src/ColumnStatistics.jsx
```diff
@@ -169,6 +169,12 @@ class ColumnStatistics extends Component {
         {description && (
           <div className="column-statistics-description">{description}</div>
         )}
+        {!model.isColumnMovable(column.name) && (
```
mofojed, line 172: As mentioned in another comment, `isColumnMovable` should take a column index instead of a name. So you'll need to make a method here that you can call, and essentially get the index by calling `model.columns.findIndex(col => col.name === column.name)`, then call `isColumnMovable` with that index.

IrisGridModel has the concept of columns with names, but GridModel does not. So need to respect the GridModel interface.

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,8 +458,36 @@ class IrisGridTableModel extends IrisGridModel {
...
+      });
+
+      return [...frontColumns, ...middleColumns, ...backColumns];
+    }
+    return columns;
+  });
+
   get columns() {
-    return this.table.columns;
+    return this.getColumnsWithLayoutHints(
```
mofojed, line 483: Before checking this in, we're going to need to update a few other problematic references to `table.columns` directly, and instead have them go through `model.columns`. Otherwise, we may get some weird things out of sync.

The only problem is that I don't know how to get an exhaustive list of those cases easily. You can look for `table.columns`, but it could also be written `const { columns } = table;`. And you'll need to replace those with `model.columns` instead, which could require passing the model through instead of just the table.

IrisGridUtils with all it's dehydrate/hydrate functions has a few places that need to be updated, for example.

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -460,20 +460,19 @@ class IrisGridTableModel extends IrisGridModel {
...
-        if (hints.frontColumns.includes(col.name)) {
-          frontColumns[hints.frontColumns.indexOf(col.name)] = col;
-        } else if (hints.backColumns.includes(col.name)) {
-          backColumns[hints.backColumns.indexOf(col.name)] = col;
-        } else {
-          middleColumns.push(col);
-        }
-      });
-
+      const columnMap = new Map();
```
mofojed, line 463: As we briefly discussed on the phone, we only want the column to appear once in the final array, so if it's added to the front AND the back, we should handle that (or throw an error). Also, this doesn't warn if there's columns that don't even exist, it's just filtering them out (I think?). To make it easy... see if we can just throw an error and if that's handled correctly? Eg.
```
if (hints) {
  if (// Check if the same name exists in front and back columns ||  // Check a name is in front/back cols that isn't a valid column) {
    throw new Error('Layout hints are invalid');
  }
  
}
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridModel.js
```diff
@@ -199,6 +199,10 @@ class IrisGridModel extends GridModel {
     throw new Error('set totalsConfig not implemented');
   }
 
+  get layoutHints() {
```
mofojed, line 202: Can you add JsDocs for this? What kind of object it returns, etc.
Note it could return `undefined` (in the case the API doesn't support `getLayoutHints` yet)

packages/iris-grid/src/IrisGridTableModel.js
spasovski: Added a check for invalid column names. Duplicates seem handled fine (silently) at the API.

packages/code-studio/src/dashboard/panels/IrisGridPanel.jsx
```diff
@@ -748,7 +746,7 @@ export class IrisGridPanel extends PureComponent {
         advancedSettings,
         pendingDataMap
       ),
-      this.getDehydratedGridState(table, movedColumns, movedRows)
+      this.getDehydratedGridState(model, movedColumns, movedRows)
```
mofojed, line 749: You changed it here, but the method signature of `getDehdyratedGridState` still says `table`, update it there as well:
```  
getDehydratedGridState = memoize((table, movedColumns, movedRows) =>
    IrisGridUtils.dehydrateGridState(table, { movedColumns, movedRows })
  );
```

mofojed (changes requested):

PR #245, opened 2021-10-20
WIP Matplotlib

mofojed:


mofojed:
This PoC worked, but we'll eventually implement it with Deephaven Extensions which will encompass much more functionality: https://github.com/deephaven/deephaven-core/discussions/1468

Issue #264, opened 2021-10-28
Split shortcut registry from components


mattrunyon:
`ShortcutRegistry` is currently in the `@deephaven/components` package and is the only item that is a singleton. Since it's a singleton, it requires `@deephaven/components` to be a peer dependency so only 1 version of it exists within a repo.

Split out the shortcut registry and change `@deephaven/components` to a regular dependency within our packages. Add `@deephaven/shortcuts` as a peer dependency. Alternatively, provide some way for plugins/modules to add shortcuts and combine the shortcut definitions into the `@deephaven/shortcuts` package instead of having each package create the shortcuts it uses (might not be the best choice though)

PR #265, opened 2021-10-28
Clean up dev/peer dependencies

### LINKED ISSUE #222: Clean up devDependencies/peerDependencies
As a developer, I want less peer dependencies so that it's easier to import @deephaven scoped packages.

Right now a lot of the @deephaven imports have a lot of dev/peerDependencies that should just be `dependencies`. Make sure the outputted package size does not increase much because of these changes.

mattrunyon:
Fixes #222 

Also removed any unused dependencies by packages and removed the gh-pages deploy scripts since we are not using gh-pages for anything.

`@deephaven/log` as a peer dependency because of LogProxy used for capturing all logs

`@deephaven/components` is also a peer dependency because of the `ShortcutRegistry`. This is a singleton which means it should be a peer dependency. Opened #264 to address this


mofojed (approved):
Nice.

mofojed (approved):

PR #272, opened 2021-11-01
DH-11632 handle LH frozen columns in the UI

spasovski:
There is some possible refactoring optimization to be had in `drawColumnHeaders` but it looked quite ugly locally since the loops contents relied on too many scoped+calculated vars. Still open to suggestions in either case.

### DIFF
--- a/packages/grid/src/GridRenderer.js
+++ b/packages/grid/src/GridRenderer.js
@@ -762,7 +762,20 @@ class GridRenderer {
 
   drawGridLinesForRows(context, state, rows) {
     const { metrics } = state;
-    const { visibleRowYs, maxX } = metrics;
+    const {
+      visibleRowYs,
+      maxX: metricsMaxX,
+      floatingLeftColumnCount,
+      visibleColumnXs,
+      visibleColumnWidths,
+    } = metrics;
+    let maxX = metricsMaxX;
+
+    if (floatingLeftColumnCount) {
+      maxX =
+        visibleColumnXs.get(floatingLeftColumnCount - 1) +
+        visibleColumnWidths.get(floatingLeftColumnCount - 1);
+    }
 
     // Draw row lines
     for (let i = 0; i < rows.length; i += 1) {
@@ -1754,7 +1767,7 @@ class GridRenderer {
       context.fillStyle = theme.selectionColor;
       context.fill();
 
-      /* 
+      /*
       draw an "inner stroke" that's clipped to just inside of the rects
       to act as a casing to the outer stroke. 3px width because 1px is outside
       the rect (but clipped), 1px is "on" the rect (technically this pixel is
--- a/packages/grid/src/GridRenderer.js
+++ b/packages/grid/src/GridRenderer.js
@@ -1147,6 +1147,8 @@ class GridRenderer {
     } = theme;
     const hiddenSeparatorHeight = columnHeaderHeight * 0.5;
     const hiddenY = columnHeaderHeight * 0.5 - hiddenSeparatorHeight * 0.5;
+    const containsFrozenColumns = floatingLeftColumnCount > 0;
+    let floatingLeftColumnsWidth = 0;
 
     context.save();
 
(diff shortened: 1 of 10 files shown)


packages/grid/src/GridUtils.js
```diff
@@ -231,11 +231,20 @@ class GridUtils {
...
 
-  // Returns the column index if the x/y coordinates provided are close enough to the separator, otherwise false
-  static getColumnSeparatorIndex(x, y, metrics, theme) {
+  /**
+   * Gets the column index if the x/y coordinated provided are close enough
+   * @param {Number} x Mouse x coordinate
+   * @param {Number} y Mouse y coordinate
+   * @param {GridMetrics} metrics The GridMetricCalculator metrics
+   * @param {GridTheme} theme The grid theme with potantial user overrides
+   * @param {Number} floatingColumnsWidth The total width of the frozen columns
```
mofojed, line 240: As mentioned in another comment, don't have `floatingColumnsWidth` as a param here - the floating columns are already in the metrics, and you should technically handle left/right floating columns.

It might be easiest to test left/right floating columns in the styleguide, modifying one of the mocks there (or making your own). http://localhost:4000/styleguide

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1024,6 +1024,18 @@ export class IrisGrid extends Component {
     }
   }
 
+  getAlwaysFetchColumns = memoize((alwaysFetchColumns, model) => {
+    const columnSet = new Set(alwaysFetchColumns);
+
+    model.columns.forEach(({ name }) => {
```
mofojed, line 1030: This works, you could also build it just by looking at `model.floatingLeftColumnCount` and `model.floatingRightColumnCount` and then pulling from `model.columns`

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1024,6 +1024,18 @@ export class IrisGrid extends Component {
     }
   }
 
+  getAlwaysFetchColumns = memoize((alwaysFetchColumns, model) => {
```
mofojed, line 1027: This isn't going to memoize correctly if the floating columns change after you've initialized (ie. you've changed them from the UI).
Really the memoized function should just take the alwaysFetchColumns and the frozenColumns and join them

mofojed (changes requested):
Some cleanup needed.

Also, this should work for `floatingRightColumnCount`, but we're not actually using that yet, I would file a follow up ticket for that (or do it right away).

packages/grid/src/GridRenderer.js
```diff
@@ -762,7 +761,20 @@ class GridRenderer {
...
+      visibleColumnXs,
+      visibleColumnWidths,
+    } = metrics;
+    let maxX = metricsMaxX;
+
+    if (floatingLeftColumnCount) {
+      maxX =
+        visibleColumnXs.get(floatingLeftColumnCount - 1) +
+        visibleColumnWidths.get(floatingLeftColumnCount - 1);
+    }
```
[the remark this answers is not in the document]
spasovski: Without it we get the thick grid lines everywhere but they are only needed in the frozen columns.

image

packages/grid/src/GridRenderer.js
```diff
@@ -1142,16 +1158,67 @@ class GridRenderer {
...
       const hiddenColumns = [];
+
+      // Draw visible column separators.
       let isPreviousColumnHidden = false;
       for (let i = 0; i < visibleColumns.length; i += 1) {
         const column = visibleColumns[i];
         const columnX = visibleColumnXs.get(column);
         const columnWidth = visibleColumnWidths.get(column);
+
+        if (!(columnX < floatingLeftColumnsWidth - columnWidth)) {
```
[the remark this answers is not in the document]
mofojed: Actually after inspecting further, there is some overlap occurring when I'm testing in the styleguide:

image

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -1113,13 +1133,21 @@ class IrisGridTableModel extends IrisGridModel {
...
       return false;
     }
     return !this.isKeyColumn(x);
   }
 
+  isColumnFrozen(x) {
+    if (this.layoutHints?.frozenColumns?.includes(this.columns[x].name)) {
+      return true;
+    }
+    return false;
```
mofojed, line 1148: Can just return directly, no need for if:
```suggestion
    return this.layoutHints?.frozenColumns?.includes(this.columns[x].name) ?? false;
```

mofojed (changes requested):

mofojed:
`MockIrisGridTreeModel.getColumnIndexByName` needs to be fixed - shouldn't be calling this.model.getCol (since that's a GridModel which doesn't have that)

packages/grid/src/GridRenderer.js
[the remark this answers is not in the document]
spasovski: I really can't repro this. We might need to screen-share...I'll post the rest of the review fixes in the meantime.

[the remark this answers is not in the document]
mofojed: Looks like it's only reproducible when the row headers are visible, which we have in the style guide. I set a breakpoint in the drawColumnHeaders when it was filling in the background to figure out the floating background it was drawing was a bit off:

image

Just needed to offset the background drawing by `gridX` instead of `0`

packages/grid/src/GridRenderer.js
```diff
@@ -1142,16 +1155,67 @@ class GridRenderer {
...
+    }
+
+    if (containsFrozenColumns) {
+      floatingLeftColumnsWidth =
+        visibleColumnXs.get(floatingLeftColumnCount - 1) +
+        visibleColumnWidths.get(floatingLeftColumnCount - 1);
+
+      // Frozen columns' background
+      context.fillStyle = headerBackgroundColor;
+      context.fillRect(0, 0, floatingLeftColumnsWidth, columnHeaderHeight);
```
mofojed, line 1175: As mentioned in my other comment, this is drawing the column background in the incorrect location if the row headers are visible. You can test this using the style guide, go to http://localhost:4000/styleguide... Need to set `floatingLeftColumnCount` in MockGridModel to something other than 0 as well temporarily.

```suggestion
      context.fillRect(gridX, 0, floatingLeftColumnsWidth, columnHeaderHeight);
```

mofojed (changes requested):

mofojed (approved):

PR #273, opened 2021-11-02
Add a label check action

mofojed:
- Checks if one of the lerna-changelog tags is present

You can see a failed check here: https://github.com/deephaven/web-client-ui/runs/4081157392?check_suite_focus=true

I like how it tells you how to correct the problem:
> Error: Label check failed: required 1 of bug, enhancement, internal, documentation, but found 0.

We should consider adding that to `check-doc-labels.sh` in `deephaven-core` as well, as the current error "No documentation requirements found" and "Conflicting documentation requirements" is vague/requires prior knowledge of which labels satisfy that condition.

### DIFF
--- a/.github/scripts/label-check-one-of.sh
+++ b/.github/scripts/label-check-one-of.sh
@@ -21,3 +21,4 @@ if [ $COUNT_TRUE -eq 1 ] ; then
 fi
 
 >&2 echo "Expected only one of the following labels, instead found ${COUNT_TRUE}:${ALL_LABELS}"
+exit 1
--- a/.github/scripts/label-check-one-of.sh
+++ b/.github/scripts/label-check-one-of.sh
@@ -16,9 +16,7 @@ for label in ${!LABEL_*}; do
   fi
 done
 
-if [ $COUNT_TRUE -eq 1 ] ; then
-  exit 0
+if [ $COUNT_TRUE -ne 1 ] ; then
+  >&2 echo "Expected only one of the following labels, instead found ${COUNT_TRUE}: ${ALL_LABELS:2}"
+  exit 1
 fi
-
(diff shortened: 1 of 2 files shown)


dsmmcken:
Can you do a gitbot comment with the error, so you don't have to dig? Could have something like "One of the following labels are required: `internal`, `enhancement`, etc.. see [link to guideline section on labels] for more details." or something.

```
 - name: 'Comment PR'
        uses: actions/github-script@v3
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          script: |
            const { issue: { number: issue_number }, repo: { owner, repo }  } = context;
            github.issues.createComment({ issue_number, owner, repo, body: `comment goes here, markdown works` }
 ```

devinrsmith:
Does this error out if you have *more* than one of a label? I'm not against simplifying the deephaven-core side label check w/ better action if it's applicable and easy to understand.

mofojed:
Yeaaa it probably shouldn't take a minute. I used this action because it does have the `one_of` or `any_of` or `all_of` operations, so it would fail if there are more than one of a label. But could implement similar to deephaven-core to make it simpler

mofojed:
@devinrsmith I wrote a `label-check-one-of.sh` script to handle this, now it takes ~4s instead of ~1m. The script checks all `LABEL_` prefixed environment variables and ensures exactly 1 is true; in all other cases it lists off how many were true and the list of labels it can only be one of.

This can be used in deephaven-core as well, the error message there doesn't tell you what labels you need.

Note that this method doesn't work with labels that have spaces (since env vars can't have spaces)... you could just sub in underscores and I'm sure people would figure it out

devinrsmith (approved):

PR #275, opened 2021-11-03
do not collapse console errors on mouseUp

### LINKED ISSUE #192: Click in an expanded exception in the Console should not collapse it
**Description**

If I click anywhere in an expanded exception, it minimizes the exception.
I feel like it should only expand/collapse if you click the caret on the left.

**Steps to reproduce**
1. Type any command resulting in an exception in the console, i.e. `test`
2. Click on the red collapsed exception text to expand it
3. Click anywhere in the expanded exception text

**Expected results**

The exception stays expanded.

**Actual results**

The exception gets collapsed.

spasovski:
fixes #192 

This was clearly intentionally allowed before but I agree the new behaviour is better.

### DIFF
--- a/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
+++ b/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
@@ -64,7 +64,12 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
     }
   }
 
-  handleMouseUp() {
+  handleMouseUp(event) {
+    // We don't want to expand/collapse the error if user is holding shift or an alt key
+    // They may be trying to adjust their selection
+    if (this.isClicking && !event.shiftKey && !event.metaKey && !event.altKey) {
+      this.handleToggleError();
+    }
     this.mouseX = null;
     this.mouseY = null;
     this.isClicking = false;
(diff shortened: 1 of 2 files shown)


mofojed (changes requested):
Don reviewed this for UX? You can't select the first line of the exception, we need to be able to select the full exception.

dsmmcken:
true, not sure it's any better, it just shifted the problem from the body to the button.

spasovski:
The button text is selectable (though not easily) - I'd float a copy to clipboard button to the far right of the error expand button and we can manually write the full text to the clipboard.

mofojed:
You can't select the button/top line of text at all right now? A floating copy button might be handy for copying a particular exception, but you should also just be able to copy the console output in general (eg. if you have a bunch of logging lines, then an exception, more logging, you should be able to select all that).

packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
```diff
@@ -112,16 +108,18 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
...
-          tabIndex="0"
-          className="error-content"
-          onKeyPress={this.handleKeyPress}
-          onMouseDown={this.handleMouseDown}
-          onMouseMove={this.handleMouseMove}
-          onMouseUp={this.handleMouseUp}
-        >
-          {message}
+        <div className="error-content">
+          <button
```
mofojed, line 112: Keep this in a div so you can select text.

Check behaviour in Firefox to ensure right-click does not trigger.

mofojed (changes requested):

packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
```diff
@@ -99,7 +113,7 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
             <button
               type="button"
               onClick={this.handleToggleError}
-              className="error-btn-link"
+              className={arrowBtnClasses}
```
mofojed, line 116: This looks much better! Only issue is that the button doesn't highlight when you are hovering over it itself.

Could add this here:
```
            onMouseEnter={this.handleMouseEnter}
            onMouseLeave={this.handleMouseLeave}
```
Or just add a `error-message.btn:hover` styling as well

mofojed (changes requested):

mofojed (approved):

mofojed (approved):

PR #276, opened 2021-11-04
Convert base redux to TS

### LINKED ISSUE #274: Convert base redux package to typescript
Helpful for #176. Since shortcuts are already in TS, it seems counterproductive to create typings for redux JS files instead of just swapping redux to TS for selectors.

Selectors are still imported from the package which creates them (redux, dashboard-plugins, or code-studio) and thus each package which creates selectors should only need to worry about the store elements it creates.

mattrunyon:
Closes #274 

A few things are marked as unknown and there are almost definitely some extra props missing from the `RootState`, but nothing that is consumed by TS. Once other things get converted to TS then the extra props can be added and corrected as needed.


packages/redux/src/store.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export interface RootState {
+  user: User;
+  storage: Storage;
+  workspace: Workspace;
+  activeTool: string;
+}
+
+// TODO #70: Separate all reducers into their respective modules, register from there
```
mofojed, line 63: Can take out this TODO, that ticket is already complete 
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #278, opened 2021-11-08
Fix text truncation at very narrow widths

mofojed:
- Fixed binary search loop - was checking `hi > lo` instead of `hi >= lo`
- Add a check to handle case where the text never fits
- Set lo to be lower just in case our estimate wasn't low enough
- Return empty string instead of null for consistent return type
- Fix mock iris grid to send an update (styleguide just kept showing the spinner since it was waiting for an update event)

### DIFF
--- a/packages/grid/src/GridRenderer.js
+++ b/packages/grid/src/GridRenderer.js
@@ -97,6 +97,9 @@ class GridRenderer {
       return '';
     }
 
+    // Estimate the possible low and high boundaries for truncating the text
+    // Use the width of the space divided by the estimated width of each character,
+    // and take half that as the low (minus 5 just to be extra safe), and double that as the high.
     const lo = Math.min(
       Math.max(0, Math.floor(width / fontWidth / 2) - 5),
       str.length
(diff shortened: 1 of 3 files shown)


packages/grid/src/GridRenderer.js
```diff
@@ -89,10 +94,13 @@ class GridRenderer {
     fontWidth = GridRenderer.DEFAULT_FONT_WIDTH
   ) {
     if (width <= 0 || str.length <= 0) {
-      return null;
+      return '';
     }
 
-    const lo = Math.min(Math.floor(width / fontWidth / 2), str.length);
+    const lo = Math.min(
+      Math.max(0, Math.floor(width / fontWidth / 2) - 5),
```
spasovski, line 101: Guessing the 5 is a pixel offset - might be better to declare it somewhere for readability.

packages/grid/src/GridRenderer.js
```diff
@@ -56,7 +56,7 @@ class GridRenderer {
     let lo = start;
     let hi = Math.min(str.length - 1, end);
     let result = str;
-    while (hi > lo) {
+    while (hi >= lo) {
```
spasovski, line 59: Can't believe this found a place outside a tech interview.

spasovski (approved):
The magic number is a nitpick - it's fine to merge without it.

spasovski (approved):

PR #285, opened 2021-11-11
Ability to copy column header from ui

dsmmcken:
> This is a small annoyance, but I would love to have "copy header" on the header right click menu.I can get it from the copying a selection with headers, but it would be nice to just copy the header.Just one. I'm copying it and using it a text. Either pasting into some Javascript code or putting it in an Excel Spreadsheet.

fixes [DH-11843](https://deephaven.atlassian.net/browse/DH-11843)

Added to column menu and column tooltip.

### DIFF
--- a/packages/iris-grid/src/ColumnStatistics.jsx
+++ b/packages/iris-grid/src/ColumnStatistics.jsx
@@ -2,8 +2,12 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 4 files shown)


mofojed:
Should we allow copying the column type somehow as well? Text is currently not selectable in the column header:

image

mofojed (approved):

dsmmcken:
> Should we allow copying the column type somehow as well? Text is currently not selectable in the column header:

No, I don't think that's something likely that you would need to do. I think we intentionally made it not selectable so when you double click that text you get just the name.

mofojed (approved):

PR #296, opened 2021-11-15
dependabot alerts

mofojed:
- Run npm audit fix in root
- Run `./node_modules/.bin/lerna exec npm install -- --package-lock-only`


vbabich:
I'm getting changes in `packages/jsapi-shim/package-lock.json` and `packages/log/package-lock.json` when I run `npm run clean; npm i`. Is this something with my local config?

mofojed:
@vbabich No, it wasn't just your local config... it looks like running the `package-lock-only` command changes the package locks, but then doing a clean install changes it back. I think it's more correct doing the clean/install, as that's what devs are going to do.

vbabich (approved):

PR #297, opened 2021-11-16
DH-11661 refresh viewport on IrisGridTableMode.handleTotalsUpdate

spasovski:
Not sure if this is the ideal path but the event detail and copyTotalsData were doing the right thing yet the aggs table wasn't updating until a viewport change happened. Easy review though!


mofojed (changes requested):
I don't think this is the fix we want - doing a quick debug reproducing the issue (prior to this fix), it looks like we _are_ getting a totals update, and even updating with the new colours/formatting, but for some reason the old value is drawn.

If you've got a sec let's screen share and figure it out.

mofojed (approved):

PR #298, opened 2021-11-16
Enable code coverage tests, add thresholds

mofojed:
Started with a baseline threshold just below existing levels for the whole project.
However, since the coverage report only runs on existing changes, it will check your diff and fail if they fall below those levels.
Only using the 'text' type reporter as that will be output in our console or in CI log output and we can see the failures from there.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -20,7 +20,16 @@ module.exports = {
     'jest-watch-typeahead/testname',
     'jest-watch-select-projects',
   ],
-  collectCoverage: false,
+  collectCoverage: true,
   coverageDirectory: '<rootDir>/coverage/', // This is relative to monorepo root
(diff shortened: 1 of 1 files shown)


vbabich:
Running `npm test` locally shows an error:
```
Failed to collect coverage from .../web-client-ui/packages/code-studio/src/styleguide/Editors.stories.jsx
ERROR: babel-jest: Babel ignores packages/code-studio/src/styleguide/Editors.stories.jsx - make sure to include the file in Jest's transformIgnorePatterns as well.
```

mofojed:
@vbabich I'm not getting the same error as you - do you have another flag or enter another command afterwards? Running just `npm test` results in no tests running since there hasn't been any changes...

vbabich (approved):

vbabich (approved):

PR #300, opened 2021-11-17
account for number of floating bottom rows in lastTop

### LINKED ISSUE #299: aggregations/floating rows cover the last visible table row
You will need more than one aggregation to really experience this so in a table add for example a SUM and an ABS_SUM aggregation. If the table has at least a page of row the last row will be covered. This is especially apparent when frozen columns exist.

spasovski:
fixes #299 

This seems to work as expected now.

### DIFF
--- a/packages/grid/src/GridMetricCalculator.js
+++ b/packages/grid/src/GridMetricCalculator.js
@@ -618,7 +618,7 @@ class GridMetricCalculator {
       y += rowHeight;
 
       if (y >= visibleHeight) {
-        return Math.min(lastTop + 1, rowCount - 1);
+        return Math.min(lastTop + floatingBottomRowCount, rowCount - 1);
       }
 
       lastTop -= 1;
--- a/packages/grid/src/GridMetricCalculator.js
+++ b/packages/grid/src/GridMetricCalculator.js
@@ -191,6 +191,11 @@ class GridMetricCalculator {
     const maxX = columnWidthValues.reduce((x, w) => x + w, 0) - leftOffset;
     const maxY = rowHeightValues.reduce((y, h) => y + h, 0) - topOffset;
 
+    const floatingBottomHeight = this.getFloatingBottomHeight(
+      state,
+      visibleRowHeights
+    );
+
     const lastLeft = this.getLastLeft(
       state,
       null,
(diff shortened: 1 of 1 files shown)


packages/grid/src/GridMetricCalculator.js
```diff
@@ -618,7 +618,7 @@ class GridMetricCalculator {
       y += rowHeight;
 
       if (y >= visibleHeight) {
-        return Math.min(lastTop + 1, rowCount - 1);
+        return Math.min(lastTop + floatingBottomRowCount, rowCount - 1);
```
mofojed, line 621: This isn't right... so the params passed in to `getLastTop` include `visibleHeight`, which _should_ be the height of the "visible" viewport (for lack of a better name), ie. everything that's not floating. So visually:
```
|-------------------------------|
| floatingTopRowHeight |
|-------------------------------|
|                                     |
|       visibleHeight          |
|                                     |
|-------------------------------|
| floatingBottomRowHeight |
|-------------------------------|
```

This algorithm is looking for the `lastTop`, which is the row that is at the top when scrolled so the last row is visible.
When we have floating bottom rows, we need to make sure the last non-floating row is visible, so we start at `let lastTop = Math.max(0, rowCount - floatingBottomRowCount - 1);` (last non-floating row), and iterate backward adding the height of each row until we exceed the `visibleHeight`, and then return the row after that (eg. `lastTop + 1` in this case). Adding `floatingBottomRowCount` doesn't make sense - if `floatingBottomRowCount` is `0`, then you're `lastTop` is going to be one row higher than it should be, and you won't be able to scroll to the bottom when there's no `floatingBottomRowCount`. You can observe this by checking the style guide (http://localhost:4000/styleguide)

I _think_ this algorithm is correct here (before the change), so the issue may be with the `visibleHeight` that's passed in...

mofojed (changes requested):

packages/grid/src/GridMetricCalculator.js
mofojed: Yea - check where `getLastTop` is called from initializing `lastTop`:
```
    const lastTop = this.getLastTop(
      state,
      null,
      height - gridY - scrollBarSize
    );
```

That `height - gridY - scrollBarSize` isn't taking into account the `floatingBottomHeight` at all. Need to take that into account. I don't _think_ you need to worry about the `floatingTopRowHeight` because the top scrolled row is scrolled underneath the floating top rows, but try playing around with MockGridModel by changing up floatingTop/BottomRowCount and testing in the style guide (would be nice to add controls straight into the style guide to modify stuff, but meh)

mofojed (approved):

PR #302, opened 2021-11-18
Convert GridMetricCalculator, GridRange, GridUtils to TypeScript

mofojed:
- Convert GridMetricCalculator, GridUtils, GridRange to TypeScript
- Added JSDocs to each method
- Partial #21


packages/grid/src/GridRenderer.js
```diff
@@ -1011,7 +1011,7 @@ class GridRenderer {
   }
 
   drawTreeMarker(context, state, columnX, rowY, treeBox, color, isExpanded) {
-    const [x1, y1, x2, y2] = treeBox;
+    const { x1, y1, x2, y2 } = treeBox;
```
spasovski, line 1014: This should still be an array as far as I can tell.

packages/grid/src/GridUtils.ts
```diff
@@ -451,29 +620,45 @@ class GridUtils {
...
+    from: Index,
+    to: Index,
+    oldMovedItems: MoveOperation[] = []
+  ): MoveOperation[] {
     if (from === to) {
       return oldMovedItems;
     }
 
-    const movedItems = [].concat(oldMovedItems);
+    const movedItems: MoveOperation[] = [...oldMovedItems];
```
spasovski, line 661: 👍

packages/grid/src/GridRenderer.js
mofojed: I changed this to an object everywhere, to the new `BoxCoordinates` type. I figured that was more descriptive than an array, and there was no real reason for it to be an array anyway.

spasovski (approved):
Looks good. Glad this is moving forward.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1234,10 +1690,10 @@ class GridMetricCalculator {
 
   /**
    * Sets the width for the specified column
-   * @param {Number} column The column model index to set
-   * @param {Number} size The size to set it to, or null to reset the column size
+   * @param column The column model index to set
+   * @param size The size to set it to, or null to reset the column size
    */
-  setColumnWidth(column, size) {
+  setColumnWidth(column: ModelIndex, size?: number): void {
```
mattrunyon, line 1696: Comment says "or null to reset". Types defined here wouldn't allow that (would allow undefined or omitted)

Actually the logic says setting to null would delete the width from the map

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1250,19 +1706,19 @@ class GridMetricCalculator {
...
 
   /**
    * Sets the width for the specified row
-   * @param {Number} row The row model index to set
-   * @param {Number} size The size to set it to, or null to reset the row size
+   * @param row The row model index to set
+   * @param size The size to set it to, or null to reset the row size
    */
-  setRowHeight(row, size) {
+  setRowHeight(row: ModelIndex, size?: number): void {
```
mattrunyon, line 1721: Same issue as above w/ setting to null not being allowed by types and actually deleting the value

packages/grid/src/GridMetrics.ts
```diff
@@ -0,0 +1,153 @@
...
+/** Coordinates of a box */
+export type BoxCoordinates = {
+  x1: Coordinate;
+  y1: Coordinate;
+  x2: Coordinate;
+  y2: Coordinate;
+};
+
+/** The visible index of the item, eg. after moves are applied */
+export type Index = number;
```
mattrunyon, line 13: Maybe rename to `VisibleIndex`? VSCode rename should handle it automagically?

packages/grid/src/GridUtils.ts
```diff
@@ -39,27 +96,43 @@ class GridUtils {
 
   /**
    * Iterate through all floating items in increasing order, starting with the top items.
-   * @param {number} start Count of start floating rows, eg. floatingTopRowCount
-   * @param {number} end Count of end floating rows, eg. floatingBottomRowCount
-   * @param {number} total Total number of items
-   * @param {(itemIndex:number) => any | undefined} callback Callback called for each value, stopping the iterating and returning the value if one is returned
+   * @param start Count of start floating rows, eg. floatingTopRowCount
+   * @param end Count of end floating rows, eg. floatingBottomRowCount
```
mattrunyon, line 100: e.g.

packages/grid/src/GridUtils.ts
```diff
@@ -85,21 +158,45 @@ class GridUtils {
...
 
-  static isInItem(itemIndex, itemXs, itemSizes, x) {
-    const itemX = itemXs.get(itemIndex);
-    const itemSize = itemSizes.get(itemIndex);
+  /**
+   * Check if the coordinate is within the item specified in this dimension
+   * @param itemIndex Index of the item to check
+   * @param itemXs Coordinate of all items in this dimension
+   * @param itemSizes Size of all items in this dimension
+   * @param x The coordinate to check
```
mattrunyon, line 166: Rename `itemXs` to `itemCoordinates` and `x` to `coordinate` would be more accurate I think

packages/grid/src/GridUtils.ts
```diff
@@ -337,8 +474,9 @@ class GridUtils {
    * Check if the provided row is a floating row
    * @param {number} row The row index to check
    * @param {GridMetrics} metrics The grid metrics to check against
```
mattrunyon, line 476: Remove types from comment

packages/grid/src/GridUtils.ts
```diff
@@ -370,7 +509,19 @@ class GridUtils {
     );
   }
 
-  static getHiddenItems(itemIndex, visibleSizes, visibleItems) {
+  /**
+   * Get all the items that are hidden under the same Index
+   * Eg. If columns are 1, 2, 3, 4, 5, and column 2, 3, 4 are hidden, and we check for item 4, the return will be [2, 3, 4]
```
mattrunyon, line 514: E.g.

mattrunyon (changes requested):

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -37,44 +82,73 @@ class GridMetricCalculator {
...
-      x += sizeMap.get(i);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      x += sizeMap.get(i)!;
     }
 
     x = max;
     for (let i = 0; i < endCount && totalCount - i - 1 >= 0; i += 1) {
-      x -= sizeMap.get(totalCount - i - 1);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      x -= sizeMap.get(totalCount - i - 1)!;
```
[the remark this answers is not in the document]
mofojed: There's a few places where as a result of converting to TS this started getting flagged. I did add `?? 0` in most places, but really - it _should_ be an error, since if that value doesn't exist in the map, then the grid metric calculations were incorrect, therefore this calculation is going to be invalid.

In any case, I was incorrectly using the non-null assertion here as you pointed out. I expected it to throw, similar to in Swift with forced unwrapping. But it appears this is only a compile time flag, and at runtime it just returns `undefined`. So `?? 0` it is

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -37,44 +82,73 @@ class GridMetricCalculator {
...
+  /** Calculated column widths based on cell contents */
+  private calculatedColumnWidths: ModelSizeMap;
+
+  /** Calculated row heights based on cell contents */
+  private calculatedRowHeights: ModelSizeMap;
+
+  /** Cache of fonts to estimated width of one char */
+  private fontWidths: Map<string, number>;
+
+  /** Map from visible index to model index for rows (eg. reversing movedRows operations) */
```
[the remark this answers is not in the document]
mofojed: Damn I do this everywhere. Why is it `e.g.` and etc is etc and not `et.c.`? We'll never know

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -89,8 +163,12 @@ class GridMetricCalculator {
     this.movedColumns = movedColumns;
   }
 
-  /** update the calculated metrics from the model/canvas that are useful for many functions */
-  getMetrics(state) {
+  /**
+   * Get the metrics for the provided metric state
+   * @params state The state to get metrics for
+   * @returns The updated metrics
```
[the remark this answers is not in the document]
mofojed: Yes

packages/grid/src/GridMetricCalculator.ts
mofojed: Yea it was a loose null check, but was passing `undefined` in. I'll update to clarify.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -99,14 +99,12 @@ class GridMetricCalculator {
     let x = 0;
     for (let i = 0; i < startCount && i < totalCount; i += 1) {
       coordinates.set(i, x);
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      x += sizeMap.get(i)!;
+      x += sizeMap.get(i) ?? 0;
```
mattrunyon, line 102: If you wanted this to error, do

```
const size = sizeMap.get(i);
if (size === undefined) { // or == null?
  throw...
}
x += size;
```

As it was previously written, it wouldn't error. X would just become `NaN` and stay `NaN`

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1368,11 +1383,11 @@ class GridMetricCalculator {
   }
 
   /**
-   * Get the size of the specified item
+   * Get the size (width or height) of the specified item
    * @param modelIndex The model index to get the size for
    * @param userSizes The user set sizes
    * @param calculateSize Method to calculate the size for this item
-   * @returns The size of the specified item
+   * @returns The size (width or height) of the specified item
```
mattrunyon: I'm still confused a bit on this

Is index a row or column model index? How does it differentiate between row 1 and column 1?

Is `userSizes` actually `rowSizes` or `columnSizes` and the index passed in corresponds to the size map you passed in?

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1691,11 +1712,11 @@ class GridMetricCalculator {
   /**
    * Sets the width for the specified column
    * @param column The column model index to set
-   * @param size The size to set it to, or null to reset the column size
+   * @param size The size to set it to, or undefined to reset the column size
    */
   setColumnWidth(column: ModelIndex, size?: number): void {
     const userColumnWidths = new Map(this.userColumnWidths);
-    if (size != null) {
+    if (size !== undefined) {
```
mattrunyon, line 1719: Did you check usages from JS code to make sure none explicitly pass `null`? Could leave this as `!= null` and TS shouldn't complain I think. That way any JS stuff that hasn't been converted yet still works. Those files would still complain if you were calling w/ explicit null once it's converted to TS.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1716,11 +1737,11 @@ class GridMetricCalculator {
   /**
    * Sets the width for the specified row
    * @param row The row model index to set
-   * @param size The size to set it to, or null to reset the row size
+   * @param size The size to set it to, or undefined to reset the row size
    */
   setRowHeight(row: ModelIndex, size?: number): void {
     const userRowHeights = new Map(this.userRowHeights);
-    if (size != null) {
+    if (size !== undefined) {
```
mattrunyon, line 1744: Same as above

mattrunyon (changes requested):

packages/grid/src/GridMetricCalculator.ts
mofojed: I'm going to add a `mapGet` method that throws if the value is undefined.

packages/grid/src/GridMetricCalculator.ts
mofojed: Changed it to say "from the size map"

packages/grid/src/GridMetricCalculator.ts
mofojed: Yea I'm just going to force calling into `resetRowHeight` instead, and require a number here.

All places that call `setRowHeight`/`setColumnWidth` pass a value

packages/grid/src/GridMetricCalculator.ts
mofojed: Yea I'm just going to force calling into `resetRowHeight` instead, and require a number here.

All places that call `setRowHeight`/`setColumnWidth` pass a value

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -47,6 +47,66 @@ export type GridMetricState = {
...
+  return coordinates;
+};
+
+function mapGet<K, V>(map: Map<K, V>, key: K): V {
+  const value = map.get(key);
+  if (value !== undefined) {
+    return value;
+  }
+
+  throw new Error(`Missing value for key ${key}`);
```
mattrunyon, line 107: I know you just changed this, but do we actually want to throw an error here? It would stop any future code in that function from running, so I guess it's just break earlier rather than later?

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -312,8 +323,9 @@ class GridMetricCalculator {
           )
         : 0;
 
-    const leftColumnWidth = visibleColumnWidths.get(left) ?? 0;
-    const topRowHeight = visibleRowHeights.get(top) ?? 0;
+    const leftColumnWidth =
+      columnCount > 0 ? mapGet(visibleColumnWidths, left) : 0;
+    const topRowHeight = rowCount > 0 ? mapGet(visibleRowHeights, top) : 0;
```
mattrunyon, line 328: `mapGet` makes TS more opaque in this case. If someone were to write something similar but forget to check the size, then `mapGet` would throw, but only when it's in use as opposed to the static compile time error of the regular `get` possibly being undefined

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1714,14 +1718,10 @@ class GridMetricCalculator {
    * @param column The column model index to set
    * @param size The size to set it to, or undefined to reset the column size
```
mattrunyon, line 1719: Remove "or undefined" part of comment

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1739,14 +1741,10 @@ class GridMetricCalculator {
    * @param row The row model index to set
    * @param size The size to set it to, or undefined to reset the row size
```
mattrunyon, line 1742: Remove "or undefined" part of comment

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1755,7 +1753,9 @@ class GridMetricCalculator {
    * @param row The row model index to reset
    */
   resetRowHeight(row: ModelIndex): void {
-    this.setRowHeight(row, undefined);
+    const userRowHeights = new Map(this.userRowHeights);
```
mattrunyon, line 1756: Curious why the reset and set height/width methods create a new map. If it were in a component then it would be needed to trigger update lifecycles, but not sure why in this case. If there's a reason, we should document it so that somebody in the future doesn't come through and remove it as an unnecessary clone

mattrunyon (changes requested):

packages/grid/src/GridMetricCalculator.ts
mofojed: I'll switch it to only throw in dev mode. The idea is that you use this function when the value _must_ be there

packages/grid/src/GridMetricCalculator.ts
mofojed: Yea it's because a consumer of the metrics may memoize based on column widths/row heights (like we do when saving the state in IrisGridPanel). Otherwise there's no quick way to tell if the values have changed/stayed the same

packages/grid/src/GridMetricCalculator.ts
mofojed: Actually that just makes it complicated. This function should be used when you _know_ the value is there, we fail fast and hard if the value isn't there.
Calling map.get could return undefined, and rather than having to check that everywhere or provide a default value (which would mask the error), just fail hard

packages/grid/src/GridMetricCalculator.ts
mofojed: Yes, we want mapGet to fail fast when used incorrectly.

mattrunyon (changes requested):
Change `mapGet` to be called something like `getOrThrow` or `mapGetOrThrow` to make it clear to anybody in the future that it can throw

Remove default value if you want

mattrunyon (approved):

PR #311, opened 2021-11-24
re-calc lastTop

### LINKED ISSUE #307: floating bottom rows - keyboard navigation results in wrong scrollbar position
Create a sample table (with enough rows to cause a scrollbar). Add some aggregation rows on the bottom. Click anywhere in the table. Use the keyboard down arrow to reach the bottom row, notice the scrollbar on the right is not quite at the end position.

spasovski:
fixes #307 

Noticed lastTop needs to get recalculated. Seems to fix the scrollbar position issue but not sure if this is the right approach.

### DIFF
--- a/packages/grid/src/GridMetricCalculator.js
+++ b/packages/grid/src/GridMetricCalculator.js
@@ -201,7 +201,8 @@ class GridMetricCalculator {
       null,
       width - gridX - scrollBarSize - rowFooterWidth
     );
-    const lastTop = this.getLastTop(
+
+    let lastTop = this.getLastTop(
       state,
       null,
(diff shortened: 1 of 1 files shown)


packages/grid/src/GridMetricCalculator.js
```diff
@@ -271,6 +274,12 @@ class GridMetricCalculator {
       ]);
     }
 
+    lastTop = this.getLastTop(
```
mofojed, line 277: Hmmm why does this need to be re-calculated?

mofojed:
Let's wait til #302 gets merged before merging this one I think

PR #312, opened 2021-11-24
Enable codecov support 

mofojed:
- Create codecov.yml, with the carryforward flag for partial reports. Already uploaded the initial full report from current `main`
- Add badge to readme
- Enable only on pull requests

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Test (Pull Request)
         if: ${{ github.event_name == 'pull_request' }}
         run: |
-          npm run test:ci -- --changedSince origin/${{ github.event.pull_request.base.ref }}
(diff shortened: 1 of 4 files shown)


mattrunyon:
Will this fail ever for coverage that is too low? And are you planning to set it up w/ a bot comment/link for the report? Just the report might be fine so we can see changes and determine if we should be adding tests

mofojed:
Yeah I forgot to remove the `coverageThresholds`, they're removed now.

I've requested to install the GitHub app which will supposedly comment as codecov, need authorization from Devin first to enable it. Then it should comment automatically on all PRs, with links back to things.

mofojed:
Here's an example of a report on my repo: https://github.com/mofojed/web-client-ui/pull/1

Once authorized here it should post on here as `codecov-commenter`. I _think_ the carry forward flag will work correctly once the codecov.yml is merged into main. We'll need to make sure that works.

mattrunyon (approved):

PR #315, opened 2021-11-26
Fix unusable filter set manager

vbabich:
Update FilterSetManagetPanel and IrisGridPanel to use the new hydrate/dehydrate methods for quick and advanced filters


mofojed (approved):
Can't wait to TypeScript'ify everything...

Issue #308, opened 2021-11-22
floating bottom rows - background on row labels does not fill in


spasovski:
This isn't 100% reproducible but play around with the table panel position and size and it should eventually happen. Create a table with enough rows to cause a vertical scroll. Add some aggregation bottom rows...while scrolling at any point in the table except the very bottom the label rows do not have a background colour filled in.

PR #317, opened 2021-11-26
rerevert grid cleanup

mofojed:
- Convert GridMetricCalculator, GridRange, GridUtils to TypeScript (#302)
- Fix the calculation for visibleRowHeights/visibleColumnWidths


mofojed:
@mattrunyon @spasovski It would be easiest to review this just by looking at the second change.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -241,19 +241,28 @@ class GridMetricCalculator {
...
-    ]);
+    let visibleRowHeights = this.getVisibleRowHeights(state);
+    let visibleColumnWidths = this.getVisibleColumnWidths(
+      state,
+      firstColumn,
+      treePaddingX
+    );
 
     // Calculate the metrics for the main grid
     const visibleRows = Array.from(visibleRowHeights.keys());
```
mofojed, line 252: The problem was the `visibleRows` map was including floatingRows which is wrong. I have a ticket to make things make more sense, but I figure making it make more sense is a breaking change so leaving it in another change is safest.

mattrunyon (approved):

spasovski (approved):
Looks good and fixes #307 and #308

PR #319, opened 2021-11-29
Update DateTime reference to new package location

mofojed:
DBDateTime is moving due to an engine change, update to use the new name.

Engine PR: https://github.com/deephaven/deephaven-core/pull/1473


vbabich (approved):

mofojed:
Going to wait for @rcaudy 's change to be approved to ensure the final rename.

vbabich (approved):

mofojed:
> There's also one `com.illumon.iris.db.tables.utils.DBDateTime` reference in the API mock in DHE.

That's fine, I don't think this will change on the DHE side.

PR #321, opened 2021-11-29
freeze columns via context menu

spasovski:


packages/iris-grid/src/IrisGridProxyModel.js
```diff
@@ -286,6 +286,10 @@ class IrisGridProxyModel extends IrisGridModel {
     return this.model.layoutHints;
   }
 
+  set uiFrozenColumns(columns) {
+    this.model.contextFrozenColumns = columns;
```
mofojed, line 290: Should be `this.model.uiFrozenColumns = columns`

packages/iris-grid/src/IrisGridModelUpdater.jsx
```diff
@@ -97,6 +98,9 @@ const IrisGridModelUpdater = React.memo(
     useEffect(() => {
       model.pendingDataMap = pendingDataMap;
     }, [model, pendingDataMap]);
+    useEffect(() => {
+      model.contextFrozenColumns = contextFrozenColumns;
```
mofojed, line 102: In IrisGridModel, you called it `uiFrozenColumns`. This should be `model.uiFrozenColumns`

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,41 +460,48 @@ class IrisGridTableModel extends IrisGridModel {
...
+        frozenColumns.length !== (hints?.frozenColumns?.length ?? 0)
       ) {
         throw new Error(
           'Layout hints are invalid (contain invalid column names)'
         );
       }
 
+      // Handle frozen columns by user UI action.
+      if (uiFrozenColumns.length) {
+        frozenColumns = uiFrozenColumns
```
mofojed, line 500: We need to be able to freeze in addition to any existing hints, as well as unfreeze any columns from hints. Not sure how we do it with this data construct?

mofojed (changes requested):

packages/iris-grid/src/IrisGridModel.js
```diff
@@ -186,6 +186,13 @@ class IrisGridModel extends GridModel {
     throw new Error('set customColumns not implemented');
   }
 
+  /**
+   * @param {String[]} columns The columns to treat as frozen
```
mofojed, line 190: We should default to `undefined`, and that means just use the layout hints (if they're there), otherwise if we pass an array then those override the layout hints.
```suggestion
   * @param {String[]|undefined columns The columns to treat as frozen, or undefined to just use the layout hints
```

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -74,6 +74,8 @@ class IrisGridTableModel extends IrisGridModel {
     // These rows can be sparse, so using a map instead of an array.
     this.pendingNewDataMap = new Map();
     this.pendingNewRowCount = 0;
+
+    this.frozenColumns = [];
```
mofojed, line 78: Think of the `frozenColumns` array more of a `userFrozenColumns`, we only want to set it when the user explicitly sets it. So we should just initialize it to `this.frozenColumns = null`. Otherwise, we would save the layoutHints frozen columns into the workspace, and then when reloading we would always use that, even if the layout hints have changed.

One easy way to deal with this could be having a function like:

```
get frozenColumns() {
    return this.userFrozenColumns ?? this.layoutHints?.frozenColumns ?? [];
}
```
Or something like that

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -460,84 +460,72 @@ class IrisGridTableModel extends IrisGridModel {
...
+      if (hints?.backColumns) {
+        backColumns = hints.backColumns
+          .map(name => columnMap.get(name))
+          .filter(Boolean);
+      }
+      if (userFrozenColumns.length) {
+        frozenColumns = userFrozenColumns
+          .map(name => columnMap.get(name))
+          .filter(Boolean);
+      }
```
mofojed, line 486: If userFrozenColumns isn't defined, then you should be looking at the layout hints frozen columns

packages/iris-grid/src/IrisGridUtils.js
```diff
@@ -166,8 +168,8 @@ class IrisGridUtils {
       selectedSearchColumns,
       invertSearchColumns = true,
       pendingDataMap = [],
+      frozenColumns = [],
```
mofojed, line 171: Shouldn't initialize to an empty array, we don't want to save frozenColumns unless user has set them

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -3037,6 +3059,7 @@ IrisGrid.propTypes = {
 
   pendingDataMap: PropTypes.instanceOf(Map),
   getDownloadWorker: PropTypes.func,
+  frozenColumns: PropTypes.arrayOf(PropTypes.string).isRequired,
```
mofojed, line 3062: Shouldn't be `.isRequired`, should default to null

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1371,6 +1373,24 @@ export class IrisGrid extends Component {
     this.grid.forceUpdate();
   }
 
+  freezeColumnByColumnName(columnName) {
+    const { frozenColumns } = this.state;
+    log.debug2('freezing column', columnName);
```
mofojed, line 1378: If `frozenColumns` is null here, then you need to pull the frozen columns from the layoutHints in the model then add/remove from that array and set it

packages/dashboard-core-plugins/src/panels/IrisGridPanel.jsx
```diff
@@ -123,6 +123,7 @@ export class IrisGridPanel extends PureComponent {
       pluginFetchColumns: [],
       modelQueue: [],
       pendingDataMap: new Map(),
+      frozenColumns: [],
```
mofojed, line 126: We should be initializing to the `frozenColumns` prop, which should be `null` by default.

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1371,6 +1374,34 @@ export class IrisGrid extends Component {
...
   }
 
+  freezeColumnByColumnName(columnName) {
+    const { frozenColumns } = this.state;
+    const { model } = this.props;
+    log.debug2('freezing column', columnName);
+
+    let allFrozenColumns = new Set([...frozenColumns, columnName]);
+
+    if (frozenColumns.length === 0) {
```
mofojed, line 1384: This is not the check we want - we only want to pull from the model layout hints if the frozenColumns are `null`. We need to do that to differentiate between two different cases:
1) User hasn't set frozen columns (value is `null`)
2) User has explicitly unfrozen all columns (value is `[]`)

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,35 +460,34 @@ class IrisGridTableModel extends IrisGridModel {
     return '';
   }
 
-  getColumnsWithLayoutHints = memoize((columns, hints) => {
-    if (hints) {
+  getColumnsWithLayoutHints = memoize((columns, hints, userFrozenColumns) => {
+    if (hints || userFrozenColumns.length) {
```
mofojed, line 464: Should be `userFrozenColumns?.length` in case userFrozenColumns are null

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,35 +460,34 @@ class IrisGridTableModel extends IrisGridModel {
...
       }
-      if (hints.backColumns) {
+      if (hints?.backColumns) {
         backColumns = hints.backColumns
           .map(name => columnMap.get(name))
           .filter(Boolean);
       }
-      if (hints.frozenColumns) {
-        frozenColumns = hints.frozenColumns
+      if (userFrozenColumns) {
```
mofojed, line 482: You're still not pulling from the layoutHints frozen columns here.

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -516,16 +517,25 @@ class IrisGridTableModel extends IrisGridModel {
...
     return this.getColumnsWithLayoutHints(
       this.table.columns,
-      this.table.layoutHints
+      this.layoutHints,
+      this.frozenColumns
     );
   }
 
+  get frozenColumns() {
+    if (this.userFrozenColumns && this.userFrozenColumns.length) {
```
mofojed, line 526: Should just be checking `this.userFrozenColumns != null`. Again, two cases - `[]` for when they explicitly unfreeze all columns from the UI, and `null` when they haven't frozen/unfrozen anything (only layoutHints are set)

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/IrisGridPanel.jsx
spasovski: Agreed.

packages/iris-grid/src/IrisGridTableModel.js
spasovski: We're passing in `this.frozenColumns` here, bad variable naming I think - so it would get LH.frozenColumns.

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -458,35 +460,34 @@ class IrisGridTableModel extends IrisGridModel {
     return '';
   }
 
-  getColumnsWithLayoutHints = memoize((columns, hints) => {
-    if (hints) {
+  getColumnsWithLayoutHints = memoize((columns, hints, userFrozenColumns) => {
```
mofojed, line 463: I see why I was getting so confused... this is still called `userFrozenColumns` here, but you're passing in `this.frozenColumns`, which is the userFrozenColumns OR the layoutHints frozen columns.
You could just rename this var to `frozenColumnsParam` and be done with it... I think what would be even clearer though would be changing this signature, e.g.:
```
getColumnsWithLayoutHints = memoize((columnMap, frontColumns, backColumns, frozenColumns) => {
  ...
}
```

Then `get columns()` changes to:
```
return this.getColumnsWithLayoutHints(this.columnMap, this.frontColumns, this.backColumns, this.frozenColumns)
```
Then you add memoized functions for columnMap, frontColumns, backColumns similar to frozenColumns, eg.
```
getMemoizedColumnMap = memoize((tableColumns) => {
  const columnMap = new Map();
  columns.forEach(col => columnMap.set(col.name, col));
  return columnMap;
})

get columnMap() {
  return this.getColumnMap(this.table.columns);
}

getMemoizedFrontColumns = memoize((layoutHintsFrontColumns) => {
   return layoutHintsFrontColumns ?? [];
});

get frontColumns() {
  return this.getMemoizedFrontColumns(this.layoutHints?.frontColumns);
}

getMemoizedBackColumns = memoize((layoutHintsBackColumns) => {
   return layoutHintsBackColumns ?? [];
});

get backColumns() {
  return this.getMemoizedFrontColumns(this.layoutHints?.backColumns);
}

getMemoizedFrozenColumns = memoize((layoutHintsFrozenColumns, userFrozenColumns) => {
   return userFrozenColumns ?? layoutHintsFrozenColumns ?? [];
});

get frozenColumns() {
  return this.getMemoizedFrozenColumns(this.layoutHints?.frozenColumns, this.userFrozenColumns);
}
```

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -516,16 +517,25 @@ class IrisGridTableModel extends IrisGridModel {
...
+      this.frozenColumns
     );
   }
 
+  get frozenColumns() {
+    if (this.userFrozenColumns && this.userFrozenColumns != null) {
+      return this.userFrozenColumns;
+    }
+
+    return this.layoutHints?.frozenColumns ?? [];
```
mofojed, line 530: We should either memoize this (similar to `get columns()`, e.g. `getFrozenColumnsWithLayoutHints`), or have a static empty array created that is used here. Otherwise any UI looking at `model.frozenColumns` may think it's changed if just doing a shallow equals

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -516,16 +517,25 @@ class IrisGridTableModel extends IrisGridModel {
...
     return this.getColumnsWithLayoutHints(
       this.table.columns,
-      this.table.layoutHints
+      this.layoutHints,
+      this.frozenColumns
     );
   }
 
+  get frozenColumns() {
+    if (this.userFrozenColumns && this.userFrozenColumns != null) {
```
mofojed, line 526: Only need the != null check:

```suggestion
    if (this.userFrozenColumns != null) {
```

mofojed: Could even just do `this.userFrozenColumns ?? this.layoutHints?.frozenColumns`... but see my other comment about memoizing all these functions.
Right now if there are no hints, we're getting a new frozenColumns array on each call, which causes the `get columns()` call to not memoize correctly, which could cause many unnecessary updates/renders.

mofojed (changes requested):
Need some memoization for frozen columns at least or we'll be causing more rendering

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -460,76 +460,114 @@ class IrisGridTableModel extends IrisGridModel {
...
-        frontColumns.length !== (hints?.frontColumns?.length ?? 0) ||
-        backColumns.length !== (hints?.backColumns?.length ?? 0)
-      ) {
-        throw new Error(
-          'Layout hints are invalid (contain invalid column names)'
-        );
-      }
+        if (
+          finalFrontColumns.length !== frontColumns.length ||
+          finalBackColumns.length !== backColumns.length
```
mofojed, line 488: I think could also throw in a finalFrozenColumns check here

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -460,76 +460,114 @@ class IrisGridTableModel extends IrisGridModel {
...
-        ...backColumns,
-      ];
+        return [
+          ...finalFrozenColumns,
+          ...finalFrontColumns,
+          ...middleColumns,
+          ...finalBackColumns,
+        ];
+      }
+      return this.table.columns;
```
mofojed, line 512: Should probably return `[...columnMap.values()]` to be safe. Could just initialize `const columns = [...columnMap.values();]` at the top and use here and when calculating `middleColumns`

packages/iris-grid/src/IrisGridTableModel.js
```diff
@@ -460,76 +460,114 @@ class IrisGridTableModel extends IrisGridModel {
...
+  }
+
+  // get frozenColumns() {
+  //   if (this.userFrozenColumns && this.userFrozenColumns != null) {
+  //     return this.userFrozenColumns;
+  //   }
+
+  //   return this.layoutHints?.frozenColumns ?? [];
+  // }
+
```
mofojed, line 570: Delete unused code

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1371,6 +1374,34 @@ export class IrisGrid extends Component {
...
+
+    this.setState({
+      frozenColumns: [...allFrozenColumns],
+    });
+  }
+
+  unFreezeColumnByColumnName(columnName) {
+    const { frozenColumns } = this.state;
+    log.debug2('unfreezing column', columnName);
+
```
mofojed, line 1399: You need the `allFrozenColumns` thing like you do in `freezeColumnByColumnName`. Otherwise if the first operation is an unfreeze on a table with frozen columns in layout hints, boom.

mofojed (changes requested):

mofojed (approved):

mofojed (approved):

PR #322, opened 2021-11-30
Add UI version to support log data

### LINKED ISSUE #196: Export Version information with support logs
As a user, I want my exported support logs to include the version I'm running and browser I'm using so that support can debug my issue more effectively.

mattrunyon:
Fixes #196 

Adds a file `metadata.json` to the support log zip which currently just contains `uiVersion`

### DIFF
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -86,7 +86,7 @@ export async function exportLogs(
   const folder = zip.folder(fileNamePrefix) as JSZip;
   folder.file('console.txt', logHistory.getFormattedHistory());
   folder.file('redux.json', getReduxDataString());
-  folder.file('metadata.json', JSON.stringify(metadata));
+  folder.file('metadata.json', JSON.stringify(metadata, null, 2));
 
   const blob = await zip.generateAsync({ type: 'blob' });
   const link = document.createElement('a');
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -61,7 +61,7 @@ function makeSafeToStringify(
   return output;
 }
 
-function getReduxDataString() {
+function getReduxDataString(): string {
   const reduxData = store.getState();
   return JSON.stringify(
     makeSafeToStringify(reduxData),
@@ -72,9 +72,11 @@ function getReduxDataString() {
 
 /**
  * Export support logs with the given name.
+ * @param metadata Any extra metadata to be written to a metadata file
  * @param fileNamePrefix The zip file name without the .zip extension. Ex: test will be saved as test.zip
  */
 export async function exportLogs(
+  metadata = {},
   fileNamePrefix = `${dh.i18n.DateTimeFormat.format(
     FILENAME_DATE_FORMAT,
     new Date()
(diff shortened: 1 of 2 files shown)


packages/code-studio/src/settings/SettingsMenu.jsx
```diff
@@ -24,8 +24,15 @@ export class SettingsMenu extends Component {
...
   }
 
+  static getVersion() {
+    return process.env.REACT_APP_VERSION;
+  }
+
   static handleExportSupportLogs() {
-    exportLogs();
+    const metadata = {
+      uiVersion: SettingsMenu.getVersion(),
```
dsmmcken, line 33:
```suggestion
      uiVersion: SettingsMenu.getVersion(),
      userAgent: navigator.userAgent,
```

I didn't test that, but grabbing the user agent would be handy.

packages/code-studio/src/log/LogExport.ts
```diff
@@ -84,6 +86,7 @@ export async function exportLogs(
   const folder = zip.folder(fileNamePrefix) as JSZip;
   folder.file('console.txt', logHistory.getFormattedHistory());
   folder.file('redux.json', getReduxDataString());
+  folder.file('metadata', JSON.stringify(metadata));
```
mofojed, line 89:
```suggestion
  folder.file('metadata.json', JSON.stringify(metadata));
```

packages/code-studio/src/log/LogExport.ts
```diff
@@ -72,9 +72,11 @@ function getReduxDataString() {
 
 /**
  * Export support logs with the given name.
+ * @param metadata Any extra metadata to be written to a metadata file
  * @param fileNamePrefix The zip file name without the .zip extension. Ex: test will be saved as test.zip
  */
 export async function exportLogs(
+  metadata = {},
```
mofojed, line 79: Seems a little strange to require the metadata to be passed in separately, when everything else is pulled from redux/global stores here. Why not just have a `getMetadata()` function defined in LogExport, then the SettingsMenu doesn't need to change as well?

mofojed (changes requested):

packages/code-studio/src/log/LogExport.ts
mattrunyon: I made that change before creating the PR, but I guess didn't push the commit

packages/code-studio/src/log/LogExport.ts
mattrunyon: My thought on this was that the log export doesn't need to be duplicated in Enterprise. It could be part of some shared package. Then Enterprise can add other metadata but the logic for saving the logs is the same.

Might not work though if we add more Enterprise specific data to the logs

mofojed: I think there will always be different stuff - I could see us logging console sessions in both, but perhaps the list of PQs in the active dashboard and their statuses or something like that. I think we had a bunch of ideas on the Enterprise ticket.
In other words, I think LogExport is unique to each and there may be a bit of overlap but not enough to try and share design.

mofojed: Alternatively, pass in all the files to export, eg:
```
exportZip({
  'metadata.json': { uiVersion, userAgent },
  'console.txt': logHistory.getFormattedHistory(),
  'redux.json': getReduxDataString(),
})
```
and then that iterates through all properties and creates the files, checking the type, if it's an object then do the safe stringify thing, etc.

mattrunyon: Just put it all in LogExport for now

mofojed (approved):

PR #324, opened 2021-12-02
Fix the isolated linker

mofojed:
Was setting `undefined` but checking against `null`. Instead check against `undefined`.


vbabich (changes requested):
There's also a couple of `isolatedLinkerPanelId ` `null` checks in `Linker.tsx` left.

vbabich (approved):

PR #329, opened 2021-12-07
Better console history handling of multi-line commands

mattrunyon:
Fixes DH-10208

Up arrow only takes you to previous command if at top line. Keeps focus on top line
Down arrow takes to next command if at bottom line. Keeps focus on bottom line
Doesn't affect commands double clicked from history. Those are. viewed as edited commands and don't cycle through history

### DIFF
--- a/packages/console/src/ConsoleInput.jsx
+++ b/packages/console/src/ConsoleInput.jsx
@@ -144,29 +144,33 @@ export class ConsoleInput extends PureComponent {
      * Can't do it in `onDidChangeModelContent` either, since we want to stop the Enter action from modifying the command.
      */
     this.commandEditor.onKeyDown(keyEvent => {
-      if (
-        !this.isCommandModified ||
-        this.commandEditor.getModel().getValueLength() === 0
-      ) {
-        if (keyEvent.keyCode === monaco.KeyCode.UpArrow) {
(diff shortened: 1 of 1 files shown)


packages/console/src/ConsoleInput.jsx
```diff
@@ -144,29 +144,33 @@ export class ConsoleInput extends PureComponent {
...
-      if (
-        !this.isCommandModified ||
-        this.commandEditor.getModel().getValueLength() === 0
-      ) {
-        if (keyEvent.keyCode === monaco.KeyCode.UpArrow) {
-          const { commandHistoryIndex } = this;
+      const { commandEditor, commandHistoryIndex } = this;
+      const { lineNumber } = commandEditor.getPosition();
+      const model = commandEditor.getModel();
+      if (!this.isCommandModified || model.getValueLength() === 0) {
```
mofojed, line 150: I don't think we want this if check anymore? Chrome doesn't do it, and you can get into a state where you can't cycle through commands again until you fully delete a command you've been editing, and then it's still in the point in history you were at... probably better to just remove this check now that you have the correct top/bottom behaviour

mofojed (changes requested):

mofojed (approved):

PR #332, opened 2021-12-07
Cross column search does not restrict to selected columns

vbabich:
Fix the issue with cross-column search where it doesn't properly restrict to selected columns. Tested manually, unit tests pass.


mofojed:
This looks good but we're going to merge the codecov fix first

mofojed (approved):

PR #335, opened 2021-12-07
Remove cursor position marker in console input bar when scrollbar is hidden

mattrunyon:
DH-11364


mofojed (changes requested):
Still seeing some artifacts when there's no overflow when on a line with parentheses:

image

PR #336, opened 2021-12-07
Fix custom column deletion

mofojed:
API changed custom columns to an object, but they can (and should) be stringified for saving.

Fix for Enterprise https://deephaven.atlassian.net/browse/DH-11906

### DIFF
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -1922,7 +1922,10 @@ export class IrisGrid extends Component {
       selectDistinctColumns,
     } = this.state;
 
-    const { columns, customColumns: oldCustomColumns } = model;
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1922,7 +1922,10 @@ export class IrisGrid extends Component {
       selectDistinctColumns,
     } = this.state;
 
-    const { columns, customColumns: oldCustomColumns } = model;
+    const { columns } = model;
+    const oldCustomColumns = model.customColumns.map(
+      customColumn => `${customColumn}`
```
vbabich, line 1927: Nice, I didn't know we can serialize custom columns like this!

vbabich (approved):

PR #337, opened 2021-12-07
tree table snapshots

spasovski:
missing formatting and I don't think the intersection is quite right but it's quite the snack

### DIFF
--- a/packages/iris-grid/src/IrisGridTreeTableModel.js
+++ b/packages/iris-grid/src/IrisGridTreeTableModel.js
@@ -1,11 +1,7 @@
 /* eslint class-methods-use-this: "off" */
 import memoize from 'memoize-one';
-import Log from '@deephaven/log';
 import { GridRange } from '@deephaven/grid';
 import IrisGridTableModel from './IrisGridTableModel';
-import IrisGridUtils from './IrisGridUtils';
-
-const log = Log.module('IrisGridTreeTableModel');
 
 class IrisGridTreeTableModel extends IrisGridTableModel {
   applyBufferedViewport(viewportTop, viewportBottom, columns) {
--- a/packages/iris-grid/src/IrisGridTreeTableModel.js
+++ b/packages/iris-grid/src/IrisGridTreeTableModel.js
@@ -1,6 +1,11 @@
 /* eslint class-methods-use-this: "off" */
 import memoize from 'memoize-one';
+import Log from '@deephaven/log';
+import { GridRange } from '@deephaven/grid';
 import IrisGridTableModel from './IrisGridTableModel';
+import IrisGridUtils from './IrisGridUtils';
+
+const log = Log.module('IrisGridTreeTableModel');
 
 class IrisGridTreeTableModel extends IrisGridTableModel {
   applyBufferedViewport(viewportTop, viewportBottom, columns) {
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGridTreeTableModel.js
```diff
@@ -32,6 +37,35 @@ class IrisGridTreeTableModel extends IrisGridTableModel {
...
+    const viewportRange = new GridRange(
+      0,
+      this.viewportData.offset,
+      columns.length,
+      this.viewportData.offset + this.viewportData.rows.length
+    );
+    const intersection = GridRange.intersection(viewportRange, ranges);
+
+    for (let i = intersection.startRow; i < intersection.endRow; i += 1) {
+      const row = this.viewportData.rows[i];
```
mofojed, line 52: You'll need to iterate through the columns as well, e.g.
```
  const resultRow = [];
  const viewportRow = this.viewportData.rows[i - this.viewportData.offset];
  for (let j = intersection.startColumn; j <= intersection.endColumn; j += 1) {
    resultRow.push(formatValue(viewportRow.data.get(j).value, this.columns[j]));
  }
  result.push(resultRow);
}
```

packages/iris-grid/src/IrisGridTreeTableModel.js
```diff
@@ -32,6 +37,35 @@ class IrisGridTreeTableModel extends IrisGridTableModel {
...
+    const result = [];
+    const viewportRange = new GridRange(
+      0,
+      this.viewportData.offset,
+      columns.length,
+      this.viewportData.offset + this.viewportData.rows.length
+    );
+    const intersection = GridRange.intersection(viewportRange, ranges);
+
+    for (let i = intersection.startRow; i < intersection.endRow; i += 1) {
```
mofojed, line 51: Needs to be `i <= intersection.endRow`, as the index is inclusive

packages/iris-grid/src/IrisGridTreeTableModel.js
```diff
@@ -32,6 +37,35 @@ class IrisGridTreeTableModel extends IrisGridTableModel {
     };
   }
 
+  snapshot(ranges) {
```
mofojed, line 40: This still needs to be async, it needs to match the signature of the super IrisGridTableModel
Side note IrisGridModel should update the signature to match IrisGridTableModel as well

packages/iris-grid/src/IrisGridTreeTableModel.js
```diff
@@ -32,6 +37,35 @@ class IrisGridTreeTableModel extends IrisGridTableModel {
...
+    const intersection = GridRange.intersection(viewportRange, ranges);
+
+    for (let i = intersection.startRow; i < intersection.endRow; i += 1) {
+      const row = this.viewportData.rows[i];
+      result.push([...row.data.values()].map(rowVal => rowVal.value));
+    }
+    return result;
+  }
+
+  textSnapshot(
```
mofojed, line 58: I don't think you need to override `textSnapshot` if you're override `snapshot`

mofojed (changes requested):

packages/iris-grid/src/IrisGridTreeTableModel.js
```diff
@@ -32,6 +33,40 @@ class IrisGridTreeTableModel extends IrisGridTableModel {
...
+
+    for (let i = 0; i < ranges.length; i += 1) {
+      const intersection = GridRange.intersection(viewportRange, ranges[i]);
+
+      for (let r = intersection.startRow; r <= intersection.endRow; r += 1) {
+        const resultRow = [];
+        const viewportRow = this.viewportData.rows[
+          r - this.viewportData.offset
+        ];
+        for (
```
spasovski, line 54: It's linear time just trust me...

mofojed (changes requested):

mofojed (approved):

PR #339, opened 2021-12-08
Fix unit tests not running on dashboard and core-plugins packages

vbabich:
Rename jest.config.js to .cjs, fix baseConfig path


mofojed (approved):
...Whoops!

PR #341, opened 2021-12-09
Fix Send to Active Notebook functionality

mofojed:
- Added `getLastUsedPanelOfTypes` to PanelManager (will be needed for Enterprise)
- Clean up some of the TypeScript types

### DIFF
--- a/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
+++ b/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
@@ -3,17 +3,12 @@ import {
   DashboardPluginComponentProps,
   LayoutUtils,
   PanelComponent,
+  PanelComponentType,
   useListener,
 } from '@deephaven/dashboard';
 import { FileUtils } from '@deephaven/file-explorer';
 import Log from '@deephaven/log';
-import React, {
-  ComponentType,
-  useCallback,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useDispatch } from 'react-redux';
 import shortid from 'shortid';
 import { ConsoleEvent, NotebookEvent } from './events';
@@ -442,26 +437,14 @@ export const ConsolePlugin = ({
 
   useEffect(() => {
     const cleanups = [
-      registerComponent(
-        ConsolePanel.COMPONENT,
-        (ConsolePanel as unknown) as ComponentType
-      ),
-      registerComponent(
-        CommandHistoryPanel.COMPONENT,
-        (CommandHistoryPanel as unknown) as ComponentType
-      ),
+      registerComponent(ConsolePanel.COMPONENT, ConsolePanel),
+      registerComponent(CommandHistoryPanel.COMPONENT, CommandHistoryPanel),
       registerComponent(
         FileExplorerPanel.COMPONENT,
-        (FileExplorerPanel as unknown) as ComponentType
+        (FileExplorerPanel as unknown) as PanelComponentType
       ),
-      registerComponent(
-        LogPanel.COMPONENT,
-        (LogPanel as unknown) as ComponentType
-      ),
-      registerComponent(
-        NotebookPanel.COMPONENT,
-        (NotebookPanel as unknown) as ComponentType
-      ),
+      registerComponent(LogPanel.COMPONENT, LogPanel),
+      registerComponent(NotebookPanel.COMPONENT, NotebookPanel),
     ];
 
     return () => {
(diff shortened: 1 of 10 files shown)


packages/dashboard-core-plugins/src/ConsolePlugin.tsx
```diff
@@ -442,26 +437,14 @@ export const ConsolePlugin = ({
...
-        (FileExplorerPanel as unknown) as ComponentType
-      ),
-      registerComponent(
-        LogPanel.COMPONENT,
-        (LogPanel as unknown) as ComponentType
-      ),
-      registerComponent(
-        NotebookPanel.COMPONENT,
-        (NotebookPanel as unknown) as ComponentType
+        (FileExplorerPanel as unknown) as PanelComponentType
```
mofojed, line 444: @mattrunyon any idea how to get TypeScript typings correct so we don't need to explicitly convert this one? I'm having a hard time getting TypeScript to co-operate here.

mofojed: The error:
```
Argument of type 'ConnectedComponent<typeof FileExplorerPanel, Omit<ClassAttributes<FileExplorerPanel> & PanelProps & { localDashboardId: string; metadata?: Record<...> | undefined; } & StateProps & DispatchProp<...>, "dispatch" | ... 2 more ... | "fileStorage"> & PanelProps & { ...; }>' is not assignable to parameter of type 'PanelComponentType<ComponentClass<{}, any>, Omit<ClassAttributes<FileExplorerPanel> & PanelProps & { localDashboardId: string; metadata?: Record<...> | undefined; } & StateProps & DispatchProp<...>, "dispatch" | ... 2 more ... | "fileStorage"> & PanelProps & { ...; }>'.
  Type 'ConnectedComponent<typeof FileExplorerPanel, Omit<ClassAttributes<FileExplorerPanel> & PanelProps & { localDashboardId: string; metadata?: Record<...> | undefined; } & StateProps & DispatchProp<...>, "dispatch" | ... 2 more ... | "fileStorage"> & PanelProps & { ...; }>' is not assignable to type 'FunctionComponent<{}>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type '{ children?: ReactNode; }' is not assignable to type 'Omit<ClassAttributes<FileExplorerPanel> & PanelProps & { localDashboardId: string; metadata?: Record<string, unknown> | undefined; } & StateProps & DispatchProp<...>, "dispatch" | ... 2 more ... | "fileStorage"> & PanelProps & { ...; }'.
        Type '{ children?: ReactNode; }' is missing the following properties from type 'Omit<ClassAttributes<FileExplorerPanel> & PanelProps & { localDashboardId: string; metadata?: Record<string, unknown> | undefined; } & StateProps & DispatchProp<...>, "dispatch" | ... 2 more ... | "fileStorage">': glContainer, glEventHub, localDashboardIdts(2345)
```

packages/dashboard/src/PanelManager.test.ts
```diff
@@ -72,3 +72,24 @@ it('gets last used panel of type properly', () => {
...
+  const panelB = new TestComponentB();
+  layout.eventHub.emit(PanelEvent.MOUNT, panelA);
+  layout.eventHub.emit(PanelEvent.MOUNT, panelB);
+
+  expect(panelManager.getLastUsedPanelOfTypes([TestComponentA])).toBe(panelA);
+  expect(
+    panelManager.getLastUsedPanelOfTypes([TestComponentA, TestComponentB])
+  ).toBe(panelB);
+  expect(
+    panelManager.getLastUsedPanelOfTypes([TestComponentB, TestComponentB])
```
vbabich, line 90: I made a typo in the original code, this was supposed to be
```suggestion
    panelManager.getLastUsedPanelOfTypes([TestComponentB, TestComponentA])
```

packages/dashboard/src/PanelManager.ts
```diff
@@ -162,8 +162,22 @@ class PanelManager {
...
+  }
+
+  getLastUsedPanelOfTypes(
+    types: PanelComponentType[]
+  ): PanelComponent | undefined {
+    return this.getLastUsedPanel(panel =>
+      types.some(
+        type =>
+          panel instanceof type ||
+          (isWrappedComponent(type) &&
```
vbabich, line 176: Can we get away with a simple `panel => panel instanceof type` check here? just pass unwrapped components to `getLastUsedPanelOfTypes` and skip the `isWrappedComponent` check? I didn't think of this solution when I added this on the enterprise side, then I saw you had it simplified in DHC. Then I guess remove `| WrappedComponentType<C, P>` from the `PanelComponentType` definition too.

vbabich (changes requested):

vbabich (approved):
Nice!

packages/dashboard/src/PanelManager.ts
mofojed: I've fixed this, but for posterity, we still need the WrappedComponentType for registerComponent anyway, so may as well make it work right!

PR #342, opened 2021-12-09
Fix autoresizetextarea default height in Firefox

dsmmcken:
Fixes difference in default size in FF, chrome was fine with either.

Firefox calculates scrollHeight differently then chrome with height auto. Setting height 0 instead makes them both behave the same. Component is setting height to 0, getting the scrollHeight, then setting component to that height.
image

Screenshot shows chrome left/firefox right. After this change they both behave like chrome screenshot.

### DIFF
--- a/packages/components/src/AutoResizeTextarea.tsx
+++ b/packages/components/src/AutoResizeTextarea.tsx
@@ -62,7 +62,7 @@ const AutoResizeTextarea = ({
     if (!element.current) {
       return;
     }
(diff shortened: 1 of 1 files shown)


mofojed:
@dsmmcken Need a little more context on this ticket - what's the issue this is solving? What scenario was this problem observed?

dsmmcken:
Forgot to attach screenshot. I even took it earlier.

mofojed (approved):

PR #344, opened 2021-12-09
Don't trigger console csv upload overlay if dragging text

dsmmcken:
- You can select text and drag it in a browser (kind == "string"), this shouldn't trigger the csv overlay
- Fix Button component proptype error with "type" other than button requiring an onClick function ex. (csv overlay bar, type=submit)
- Multiple file error should return, as filetype error can replace it

### DIFF
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -29,7 +29,7 @@ interface BaseButtonProps {
   icon?: IconDefinition | JSX.Element;
   disabled?: boolean;
   active?: boolean;
-  onClick: React.MouseEventHandler<HTMLButtonElement>;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
   children?: React.ReactNode;
   className?: string;
   style?: React.CSSProperties;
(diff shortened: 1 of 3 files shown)


packages/components/src/Button.tsx
```diff
@@ -29,7 +29,7 @@ interface BaseButtonProps {
   icon?: IconDefinition | JSX.Element;
   disabled?: boolean;
   active?: boolean;
-  onClick: React.MouseEventHandler<HTMLButtonElement>;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
```
mofojed, line 32: Right now you could pass `type='button'` and `onClick=undefined` and TypeScript wouldn't report an error - you'd only get an error at run time with prop types.

You'd need to add another interface that extends this, and then use that with `ButtonWithChildren` and `IconOnlyButton` I think. Eg.

```
interface ButtonButtonProps extends BaseButtonProps {
  type?: 'button';
  onClick: React.MouseEventHandler<HTMLButtonElement>;
}

interface SubmitButtonProps extends BaseButtonProps {
  type: 'reset' | 'submit';
  onClick: undefined;  // I think? Or omit if we still want to allow onClick with submit/reset types
}

interface ButtonWithTypeProps = ButtonButtonProps | SubmitButtonProps;

interface ButtonWithChildren extends ButtonWithTypeProps { ... }
interface IconOnlyButton extends ButtonWithTypeProps { ... }

```

mofojed (changes requested):

dsmmcken:
Uh, just trying to make both defaultProps and typescript to work together for the type field, and extend the default react button types for everything else.

mofojed (approved):

PR #350, opened 2021-12-13
fix console error message for single line errors

### LINKED ISSUE #348: Last character gets chopped off if error is only one line
**Description**

If an error is only one line, the last character gets truncated, and you can still click to expand/collapse it even though there is no arrow

**Steps to reproduce**
1. Upload a csv file with a string column (drag into console).
2. Look at the error

**Expected results**
2. Error should read `java.lang.IllegalArgumentException: No support for handling data of type string`, should not be expandable

**Actual results**
2. Error reads `java.lang.IllegalArgumentException: No support for handling data of type strin` without the g, and is "expandable"

If applicable, add any additional screenshots, logs, or other attachments to help explain your problem.

**Versions**
 - Deephaven: Version 0.7.0

spasovski:
fixes #348

### DIFF
--- a/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
+++ b/packages/console/src/console-history/ConsoleHistoryResultErrorMessage.jsx
@@ -95,7 +95,10 @@ class ConsoleHistoryResultErrorMessage extends PureComponent {
     const { message: messageProp } = this.props;
     const lineBreakIndex = messageProp.indexOf('\n');
     const isMultiline = lineBreakIndex > -1;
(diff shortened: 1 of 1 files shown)


mofojed (approved):
I was thrown for a bit of loop because I was testing with this code to print an error message:
```
import sys

def eprint(*args, **kwargs):
    print(*args, file=sys.stderr, **kwargs)

eprint("Hello")
```
But the last char is always a `\n`. We could possibly trim extra whitespace off error messages, but doesn't have to be in this change.

PR #356, opened 2021-12-16
Make redefined shortcuts work in monaco

### LINKED ISSUE #95: Make redefined shortcuts work in Monaco
Currently within the script editor we use Monaco's `addAction` to define the Run and Run Selected shortcuts.  As a result, changing those shortcuts doesn't actually work. Figure out how to adjust those to change with the shortcuts when a user changes these shortcuts.

Steps to reproduce
1. Change the notebook shortcut for "Run" or "Run all"

Expected
2. The new shortcut works within the notebook.

Actual
2. The new shortcut does not work within the notebook. Only the default shortcut will ever work.

Until this is fixed, those shortcuts will be marked as not editable by users

mattrunyon:
Fixes #95 

This only applies to the 2 shortcuts we add to monaco (run and run selected).

They are now able to be changed in settings panel. Tested with multiple open notebooks and the context menu + tooltip on buttons updated properly in addition to the shortcut working

Unit tests needed some config massaging so that other tests didn't import and transpile all of monaco to CJS (because Jest doesn't handle ESM well)


packages/code-studio/src/main/AppMainContainer.jsx
```diff
@@ -225,6 +227,10 @@ export class AppMainContainer extends Component {
     this.setState({ isSettingsMenuShown: true });
   }
 
+  handleShortcutUpdate() {
+    this.emitLayoutEvent(NotebookEvent.UPDATE_SHORTCUTS);
```
mofojed, line 231: As I mentioned on our call... I think the event should be emitted through the `Shortcut` or `ShortcutRegistry` itself. It seems logical that any consumer of Shortcuts in the @deephaven/components project may want to listen to changes whenever they occur to shortcuts. So something to trigger in the Shortcut.setKeyState I think, and ability to add/remove listeners.

mofojed (changes requested):

packages/code-studio/src/main/AppMainContainer.jsx
mattrunyon: I currently made it to where the changes will emit an update event from the registry w/ an array of changed shortcuts. Let me know if I should switch it to be emitted from the shortcut class. I thought that might be a bit more annoying since then there's an event for each shortcut and you could just listen to the registry event and filter for any shortcuts you care about.

Could also have each shortcut's update method emit an event to the registry to then emit an "any shortcut changed" event?

packages/components/src/shortcuts/ShortcutRegistry.ts
```diff
@@ -1,41 +1,56 @@
...
   /**
    * Adds a shortcut to the registry. Throws if a shortcut with the same ID already exists
    * @param shortcut Shortcut to add to the registry
    */
-  static add(shortcut: Shortcut): void {
-    if (ShortcutRegistry.shortcutMap.has(shortcut.id)) {
-      throw new Error(
+  add(shortcut: Shortcut): void {
+    if (this.shortcutMap.has(shortcut.id)) {
+      log.warn(
```
mofojed, line 35: Why was this changed from a throw to a warning?

mofojed (approved):

packages/components/src/shortcuts/ShortcutRegistry.ts
mattrunyon: It was getting really annoying in dev b/c hot reload would cause it to throw errors

Also I figured it would be preferable to just have the shortcut potentially broken instead of breaking other things if there's some shortcut registry version inconsistency?

Could change `createAndAdd` to return the shortcut in the registry instead of the created shortcut that isn't added if it's a duplicate ID

mofojed: Warn and return existing shortcut with the same ID

mofojed (approved):

PR #357, opened 2021-12-17
Make jsapi-shim more robust

mattrunyon:
This is getting very annoying to work around in docs since SSR has no window object and can't be defined globally it seems due to module scope/whatever environment the SSR on docs runs in. The global workaround I've used for Node in other places doesn't work w/ the docs SSR.

This shouldn't affect anything for our app and just make the shim function in docs SSR. The API isn't actually used in SSR, but the utils import the shim which then errors out since window is undefined

### DIFF
--- a/packages/jsapi-shim/src/dh.ts
+++ b/packages/jsapi-shim/src/dh.ts
@@ -1,24 +1,12 @@
 // The Deephaven API script isn't packaged as a module (yet), and is just included in index.html, exported to the global namespace
 // This include file is simply a wrapper so that it behaves like a module, and can be mocked easily for unit tests.
 // https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#using-global-variables
-/* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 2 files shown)


packages/jsapi-shim/src/dh.ts
```diff
@@ -0,0 +1,24 @@
...
+declare global {
+  interface Window {
+    dh: any;
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, import/no-mutable-exports
+let dh: any;
+try {
+  dh = window.dh;
```
mofojed, line 16: `window` may not exist in SSR, did you try with `self` instead?

mattrunyon: I switched to `globalThis` instead. Available in all contexts (since Node 12 at least and in every modern browser for the last few years)

It's still possible for the API to be undefined, but at least the shim will just return undefined instead of crashing.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

PR #362, opened 2021-12-21
Add support for updating the default decimal/integer formats

mofojed:
- Prerequisite for DH-10087
- Added IrisGrid and Chart support for using global decimal/integer formatter settings
- Add section in settings to update the Decimal/Integer default settings
- Add unit tests for FormattingSectionContent
- BREAKING CHANGE - updated the `.makeFormat` function to re-order the params to match TableColumnFormatter


packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -718,6 +780,54 @@ export class FormattingSectionContent extends PureComponent {
...
+                )}
+                data-lpignore
+                placeholder={DecimalColumnFormatter.DEFAULT_FORMAT_STRING}
+                type="text"
+                value={defaultDecimalFormatString}
+                onChange={this.handleDefaultDecimalFormatChange}
+              />
+            </div>
+          </div>
+          <div className="form-row mb-2">
```
vbabich, line 807:
```suggestion
          <div className="form-row mb-3">
```
And change `mb-3` to `mb-2` on line 773.

packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -380,21 +411,44 @@ export class FormattingSectionContent extends PureComponent {
...
       ...settings,
       formatter,
       defaultDateTimeFormat,
       showTimeZone,
       showTSeparator,
       timeZone,
-    });
+    };
+    if (
+      defaultDecimalFormatOptions?.defaultFormatString &&
```
vbabich, line 432:
```suggestion
      defaultDecimalFormatOptions?.defaultFormatString != null &&
```
I think we need to either accept empty string as a valid input here, or mark it as invalid in `render`:
- Open a table with decimals, open settings sidebar
- Change default decimal format to '#abc', confirm the decimal columns get formatted with the new format
- Clear the decimal format input. Now it shows the default placeholder, no red input border, and the columns are still formatted with the previous `#abc` format

packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -380,21 +411,44 @@ export class FormattingSectionContent extends PureComponent {
...
+      DecimalColumnFormatter.isValid(
+        DecimalColumnFormatter.makeCustomFormat(
+          defaultDecimalFormatOptions?.defaultFormatString
+        )
+      )
+    ) {
+      newSettings.defaultDecimalFormatOptions = defaultDecimalFormatOptions;
+    }
+    if (
+      defaultIntegerFormatOptions?.defaultFormatString &&
```
vbabich, line 442: Same here

vbabich (changes requested):

packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -621,8 +675,8 @@ export class FormattingSectionContent extends PureComponent {
           const selectedFormat = DecimalColumnFormatter.makeFormat(
             '',
             e.target.value,
-            null,
-            DecimalColumnFormatter.TYPE_GLOBAL
+            DecimalColumnFormatter.TYPE_GLOBAL,
```
vbabich, line 678: Need to update the `makeFormat` call on line 650. Github won't let me comment on that line directly.

packages/code-studio/src/settings/FormattingSectionContent.jsx
mofojed: I think empty should be invalid, just like the other rules.

packages/code-studio/src/settings/FormattingSectionContent.jsx
mofojed: Good catch

vbabich (approved):

PR #364, opened 2021-12-21
Add types for JS API

mattrunyon:
Adds some TS to the JS API

Need to add overrides and extended types in enterprise so the enterprise shim doesn't cause TS errors

### DIFF
--- a/packages/jsapi-shim/src/dh.types.ts
+++ b/packages/jsapi-shim/src/dh.types.ts
@@ -1,0 +1,682 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/ban-types */
+/* eslint-disable @typescript-eslint/no-empty-interface */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable lines-between-class-members */
+/* eslint-disable max-classes-per-file */
+export default dh;
+
+export interface dh {
+  IdeConnection: IdeConnectionConstructor;
+  Session: IdeSession;
+  VariableType: typeof VariableType;
+  i18n: any;
+  plot: Plot;
+}
+
+const VariableType = {
+  FIGURE: 'Figure',
+  OTHERWIDGET: 'OtherWidget',
+  PANDAS: 'Pandas',
+  TABLE: 'Table',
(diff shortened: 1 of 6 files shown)


packages/jsapi-shim/src/dh.types.ts
```diff
@@ -0,0 +1,682 @@
...
+export interface FilterCondition {
+  invoke(method: string, ...args: FilterValue[]): FilterCondition;
+  search(value: FilterValue, columns?: FilterValue[]): FilterCondition;
+
+  not(): FilterCondition;
+  and(first: FilterCondition, ...rest: FilterCondition[]): FilterCondition;
+  or(first: FilterCondition, ...rest: FilterCondition[]): FilterCondition;
+
+  toString(): string;
+  // columns: Column[] //doesnt work
```
mofojed, line 412: Just take this out, I don't know why it's even in the code, it always returns an empty array.
```suggestion
```

packages/jsapi-shim/src/dh.types.ts
```diff
@@ -0,0 +1,682 @@
...
+  readonly EVENT_SORTCHANGED: string;
+  readonly EVENT_FILTERCHANGED: string;
+  readonly EVENT_CUSTOMCOLUMNSCHANGED: string;
+  readonly EVENT_DISCONNECT: string;
+  readonly EVENT_RECONNECT: string;
+  readonly EVENT_RECONNECTFAILED: string;
+
+  readonly SIZE_UNCOALESCED: number;
+
+  reverse(): Sort;
```
mofojed, line 441: Kinda weird this is in the middle of all the props/readonly stuff, should just have it after

mofojed (changes requested):

packages/jsapi-shim/src/dh.types.ts
```diff
@@ -469,7 +468,6 @@ export interface Table extends Evented {
   getViewportData(): Promise<TableData>;
 
   subscribe(columns: Column[]): TableSubscription;
-  // inputTable(): Promise<InputTable>
```
mofojed: `inputTable()` method _does_ exist... we can add later though

mofojed (approved):

PR #369, opened 2021-12-28
Deep linking notebooks

### LINKED ISSUE #269: Allow deep-linking to open a notebook in the demo environment
As a Demo user, I want to be able to click a demo link that opens to a specific notebook example so that I can see everything Deephaven has to offer

mattrunyon:
Resolves #269 

Adds routes `/notebook/:notebookPath` which will open the provided notebook path if it is not open in the workspace and focus it within its stack if it is open in the workspace. The paths can contain slashes to denote directories so `/notebook/myFolder/myFile.py` would open `/myFolder/myFile.py` from the file explorer

If the file does not exist, it will open a panel which will then fail to load the file and give an error in the panel about the file not existing/unable to be loaded.

Tested locally setting `REACT_APP_ROUTER_BASE_NAME=/ide/` in my `.env.development.local`

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.jsx
+++ b/packages/code-studio/src/main/AppMainContainer.jsx
@@ -3,6 +3,7 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import { CSSTransition } from 'react-transition-group';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
 import shortid from 'shortid';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
@@ -16,6 +17,7 @@ import Dashboard, {
   DEFAULT_DASHBOARD_ID,
   getDashboardData,
   updateDashboardData as updateDashboardDataAction,
+  LayoutUtils,
 } from '@deephaven/dashboard';
 import {
   ChartEvent,
@@ -27,6 +29,7 @@ import {
   IrisGridEvent,
   LinkerPlugin,
   MarkdownEvent,
+  NotebookEvent,
   MarkdownPlugin,
   PandasPlugin,
   getDashboardSessionWrapper,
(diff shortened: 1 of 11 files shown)


packages/code-studio/src/main/AppMainContainer.jsx
```diff
@@ -203,6 +207,53 @@ export class AppMainContainer extends Component {
...
+          NotebookEvent.SELECT_NOTEBOOK,
+          session,
+          language,
+          notebookSettings,
+          fileMetadata,
+          true
+        );
+      } else {
+        // Focus the notebook within its stack if it is already open
+        const notebookPanel = LayoutUtils.getContentItemInStack(
```
vbabich, line 248: This doesn't change existing preview tabs into editable notebooks - the tab stays open as a preview.
I think the `NotebookEvent.SELECT_NOTEBOOK` event should be able to handle both new and existing notebook tabs.

mattrunyon: Calling `NotebookEvent.SELECT_NOTEBOOK` if the notebook is already open causes it to duplicate the notebook. The map of open notebooks isn't populated at the time this is called. The panel is created, but the notebook isn't opened yet which doesn't add it to the map. So `SELECT_NOTEBOOK` finds no open file and says to open a 2nd copy of it

I updated the comment to explain this and was able to set the props on the component before it mounts to make sure it isn't a preview

vbabich: Yea, I think the problem is the `onLayoutInitialized` handler gets called too early before the layout is actually initialized. Might need to rename it to something like 'onConfigLoaded`.

Or... can you try calling `onLayoutInitialized` on the first `stateChanged` event in `DashboardLayout.tsx`? This might work for what we need.

vbabich (approved):

vbabich (approved):

PR #370, opened 2021-12-30
Run tests without building most packages

### LINKED ISSUE #351: Run tests without requiring build first
As a developer, I want be able to run `npm test` without first needing to run `npm start` or `npm run build` so that I can test in parallel or test without having to build first.

mattrunyon:
Fixes #351 (mostly)

This still needs to build icons and golden-layout, but those have been added to postinstall and pretest/pretest:ci scripts. They take 3-4 seconds to build for me

Most of this is achieved through using tsconfig paths. As long as we maintain imports from just the root of packages, we shouldn't need to add paths if packages are added (as long as the folder corresponds to the package name)

Also uses Typescript references in the package tsconfigs which allows for faster type checking and smarter rebuilds. By having the root tsconfig list all packages as references, we can use a single `tsc -b` command to build all types. This is ~4x faster (CPU time) and ~2x faster (wall time) than currently w/ lerna. Lerna parallelizes 2x more it seems (400% CPU w/ lerna vs 200% w/ tsc). Followed guidance from https://www.typescriptlang.org/docs/handbook/project-references.html#guidance

This has the added benefit of improving go to definition in the editor. It will take you to the source code now and also types should update near instantly. Also before there were some issues when running the dev server where types would just disappear from imported packages which I believe was due to VSCode looking for the types in the dist folders. Since those were getting overwritten by the dev server they would occasionally take a while for the editor to pick up the new types.

One issue I have still noticed is our tests are ignored by tsconfig which can cause VSCode to not show intellisense on imports from `@deephaven/package` if the packages aren't built.

I also switched to using babel-jest locally instead of ts-jest because ts-jest was giving me issues with scss imports and extending the global type for the JS API shim. The only downside of this is when running tests there aren't type checks now, but types will be checked on build since `tsc` is run.

Added `clean:cache` scripts to clean the eslint cache and jest cache. Specifically the jest cache will not rerun failed linter tests (and probably successful ones) if the eslint config changes but the file contents do not. Might be related to the `jest-runner-eslint.config.js` setting `cache: true`


mofojed:
Pulling Vlad off this review to balance reviewers a bit (since he'll need to review my PRs)

packages/code-studio/src/index.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
 import 'fira';
+// eslint-disable-next-line import/no-unresolved
```
mofojed, line 5: I don't think disabling this line is necessary?
```suggestion
```

mofojed (approved):

packages/code-studio/src/index.jsx
mattrunyon: I think it's required b/c otherwise the scss from components needs to be built for the import to resolve. I can't do the paths imports that I did for the few imports from `iris-grid/dist` because it's a css file

This is the only place that imports css from another package and it should remain that way

mofojed (approved):

PR #379, opened 2022-01-12
Upgrade to TypeScript 4.5

### LINKED ISSUE #352: Update TypeScript version
Update TypeScript to v4.5. Also enforce VS Code to use the workspace version of TypeScript, so we don't get discrepancies between workspace and tests.

mattrunyon:
Fixes #352 

Also moves the TS dependency to the root only since the packages all depend on the root TS version unless otherwise specified (which is no other packages right now)


mofojed (approved):

mattrunyon:
We could add a util method like `asError(e: unknown): Error` that turns any non-Error value into the message of the Error returned. If it's already an error just return as is. Then make sure the `onError` always calls w/ an error by using the util

mofojed:
@mattrunyon The message of the error returned? So like `${e}` or something?

mattrunyon:
@mofojed Ya I guess it would be something like

```typescript
asError(e: unknown): Error {
  if (e instanceof Error) {
    return e;
  }
  return new Error(`${e}`);
}
```

packages/utils/src/ErrorUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Converts a value to an Error if it is not an Error
+ * The value is stringified to the Error message
+ * Returns the value as is if it is already an Error
+ *
+ * Used for the TS 4.4 change which sets caught values to unknown by default
+ * It is legal to throw any value in JS, not just Error
+ * @param e Possible error or any other thrown value
+ * @returns An Error with the param as a message if it is not an Error
+ */
+export function asError(e: unknown): Error {
```
mofojed, line 12: Just put this in Grid.tsx for now, do not add the dep between @deephaven/grid and @deephaven/utils

mofojed (changes requested):

mofojed (approved):

PR #380, opened 2022-01-12
Add vscode workspace shared settings

mattrunyon:
Adds the core settings for vscode that everyone should be using. Also added the extensions we all use as workspace recommended extensions. Now a new contributor just needs to open VSCode and install workspace recommended extensions to be setup in VSCode.

We should add debug configs to this in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,12 @@
 /dist
 
 # misc
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+!.vscode/*.code-snippets
 .DS_Store
 .env.local
 .env.development.local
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -11,7 +11,6 @@
 /dist
 
 # misc
-.vscode
```
mofojed, line 14: Should still .gitignore other vscode files, eg:

```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
!.vscode/*.code-snippets
```
(lifted from https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore)

mofojed (changes requested):

mofojed (approved):

PR #387, opened 2022-01-14
hide search bar toggle if serverConfig showSearch is false

spasovski:
This is pretty standard but I don't like the name similarity with the showSearchBar prop. Suggestions are welcomed.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -257,7 +257,6 @@ export class IrisGrid extends Component {
     };
     this.contextActions = [
       this.toggleFilterBarAction,
-      this.toggleSearchBarAction,
       this.discardAction,
       this.commitAction,
     ];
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -3209,6 +3219,8 @@ IrisGrid.propTypes = {
 
   // Theme override for IrisGridTheme
   theme: PropTypes.shape({}),
+
+  isSearchBarToggleVisible: PropTypes.bool.isRequired,
```
mofojed, line 3223: Don't make this `.isRequired`, just add a default to `true`

mofojed: May as well name it `canToggleSearch` to conform with `canCopy`/`canDownloadCsv`

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -285,8 +284,14 @@ export class IrisGrid extends Component {
       pendingDataMap,
       canCopy,
       frozenColumns,
+      isSearchBarToggleVisible,
     } = props;
 
+    // Hide search bar is serverConfig.showSearch is false.
```
mofojed, line 290: Don't mention `serverConfig.showSearch` here, since this package shouldn't require a serverConfig or anything like that.
We're just disallowing the action if `canToggleSearch` flag is false.

Also - should put a check in the `toggleSearchBar()` action itself, and return if the action isn't available. If you do that, you don't really need to do this part and not register the action, since the keyboard shortcut just won't do anything then.

mofojed (changes requested):

mofojed (approved):

PR #404, opened 2022-01-28
DH-12099 IrisGrid canToggleSearch should allow override by layoutHints

spasovski:
I'll explain the whole flow for reference.

From enterprise IrisGridPanel we are passed `canToggleSearch` which considers the boolean server config setting `ui.show.search`. Tables can have this overridden however via Layout Hints. This PR handles the override case where if a LH is found we simply ignore the server config and use the LH value.

The API support for this work is deployed on the BHS, `davor-searchbar-api` if you wish to test it in action.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -1679,7 +1679,7 @@ export class IrisGrid extends Component {
 
   toggleSearchBar() {
     const { showSearchBar } = this.state;
-    if (!this.isToggleSearchAvailable()) {
+    if (!this.isTableSearchAvailable()) {
       return;
     }
 
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -2830,7 +2830,7 @@ export class IrisGrid extends Component {
       isFilterBarShown,
       showSearchBar,
       canDownloadCsv,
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1661,10 +1661,27 @@ export class IrisGrid extends Component {
...
+    const searchDisplayMode = model?.layoutHints?.searchDisplayMode;
+
+    if (searchDisplayMode) {
+      if (searchDisplayMode === dh.SearchDisplayMode.SEARCH_DISPLAY_HIDE) {
+        return false;
+      }
+      if (searchDisplayMode === dh.SearchDisplayMode.SEARCH_DISPLAY_SHOW) {
+        return true;
+      }
+    }
```
mofojed, line 1675: The outside if is unnecessary, could just be:
```suggestion
    if (searchDisplayMode === dh.SearchDisplayMode.SEARCH_DISPLAY_HIDE) {
      return false;
    }
    if (searchDisplayMode === dh.SearchDisplayMode.SEARCH_DISPLAY_SHOW) {
      return true;
    }
```

mofojed (approved):

mofojed (approved):

mofojed (changes requested):
Need to fix the unit tests. Might need to update `__mocks__/dh-core.js`

PR #408, opened 2022-01-28
Visually distinguish null vs blank string and char

mattrunyon:
Addresses DH-12038

Null values for char and string columns will now display a gray, italic null instead of blank. They will still copy/export as blank which is the existing behavior


mattrunyon:
There's a flash for new tables that needs to be fixed before this merges

packages/grid/src/GridRenderer.ts
```diff
@@ -1060,7 +1060,8 @@ export class GridRenderer {
     context: CanvasRenderingContext2D,
     state: GridRenderState,
     column: VisibleIndex,
-    row: VisibleIndex
+    row: VisibleIndex,
+    textOverride?: string
```
mofojed, line 1064: I think having a `drawCellText` method may make more sense... so something like:

```
drawCellContent(context, state, column, row) {
  drawCellText(context, state, column, row, text);
}

drawCellText(context, state, column, row, text) {
   // ... all the contents of drawCellContent, except you shouldn't need the model row/column or to fetch anything from the model
}
```

Let me know if that makes sense

packages/iris-grid/src/Formatter.ts
```diff
@@ -176,9 +178,9 @@ class Formatter {
     columnName = '',
     formatOverride?: TableColumnFormat
   ): string {
-    if (value == null) {
-      return '';
-    }
+    // if (value == null) {
+    //   return '';
+    // }
```
mofojed, line 183: Remove

```suggestion
```

packages/iris-grid/src/IrisGridRenderer.js
```diff
@@ -81,6 +81,25 @@ class IrisGridRenderer extends GridRenderer {
...
   }
 
+  drawCellContent(context, state, column, row) {
+    const { metrics, model } = state;
+    const { modelColumns, modelRows } = metrics;
+    const modelRow = modelRows.get(row);
+    const modelColumn = modelColumns.get(column);
+    const value = model.valueForCell(modelColumn, modelRow);
+    if (
+      value == null &&
```
mofojed, line 91: I think just change this to `===` instead of `==` and that will get rid of the flashing.

mofojed (changes requested):

mofojed:
Tested with Groovy snippet:

```
char [] charValues = new char[]{' ', '\0', io.deephaven.util.QueryConstants.NULL_CHAR, '\n', 'A'}
double [] doubleValues = new double[]{0, Double.NaN, Double.POSITIVE_INFINITY, io.deephaven.util.QueryConstants.NULL_DOUBLE, 1.0}
charEx=newTable(charCol("C", charValues), doubleCol("D", doubleValues), stringCol("S", "A", "null", "C", null, "(null)"))
```

packages/grid/src/GridRenderer.ts
mattrunyon: I don't think this will make any difference/improvements really. There's a lot of variables used in `drawCellText` if it's split out

packages/iris-grid/src/Formatter.ts
mattrunyon: This should actually be left in (uncommented) for copy/export

mattrunyon:
The flash was caused by the API returning undefined instead of null for values in string columns. Fixed it by adjusting `valueForCell` in `IrisGridTableModel`.

https://github.com/deephaven/deephaven-core/issues/1917

mofojed (approved):

mofojed (approved):

PR #410, opened 2022-02-01
Iris-grid Typescript Chunk 2

mattrunyon:
Just another chunk of TS for iris-grid to wrap up everything I had worked on converting previously


packages/grid/src/index.ts
```diff
@@ -6,16 +6,21 @@ export * from './GridMetricCalculator';
...
+export * from './GridMouseHandler';
 export { default as GridRange } from './GridRange';
+export * from './GridRange';
 export { default as GridRenderer } from './GridRenderer';
+export * from './GridRenderer';
 export { default as GridTestUtils } from './GridTestUtils';
 export { default as GridTheme } from './GridTheme';
 export type { GridTheme as GridThemeType } from './GridTheme';
 export { default as GridUtils } from './GridUtils';
+export * from './GridUtils';
```
mofojed, line 18: Could change GridUtils to `export class GridUtils` and avoid the default import here

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -0,0 +1,502 @@
...
+    // no-op
+  }
+
+  /**
+   * Gets the columns for this model
+   * @returns All columns in the model
+   */
+  get columns(): Column[] {
+    throw new Error('get columns not implemented');
+  }
```
mofojed, line 118: `IrisGridModel` should ultimately be an abstract class, then we don't provide an impl for these at all. Right now it's not clear which methods an implementing class needs to override (e.g. `MyIrisGridModel extends IrisGridModel` won't get an error for not overriding `get columns()` even though it's required)

packages/iris-grid/src/mousehandlers/IrisGridColumnSelectMouseHandler.ts
```diff
@@ -1,23 +1,31 @@
 /* eslint class-methods-use-this: "off" */
-import { GridMouseHandler } from '@deephaven/grid';
+import {
+  GridMouseHandler,
+  GridPoint,
+  EventHandlerResult,
+} from '@deephaven/grid';
+import type { Column } from '@deephaven/jsapi-shim';
+import type { IrisGrid } from '../IrisGrid';
```
mofojed, line 8: This makes an import cycle - I think ultimately the IrisGrid type should be separated? Or perhaps the type passed into the constructor is limited to the props/state this mouse handler needs? Thoughts?

packages/iris-grid/src/mousehandlers/IrisGridColumnSelectMouseHandler.ts
```diff
@@ -1,23 +1,31 @@
...
-  isActive() {
+  irisGrid: IrisGrid;
+
+  isActive(): boolean {
     const { isSelectingColumn } = this.irisGrid.props;
     return isSelectingColumn;
   }
 
-  isValidColumn(tableColumn) {
+  isValidColumn(tableColumn: Column | null): boolean {
```
mofojed, line 28: Could be private

```suggestion
  private isValidColumn(tableColumn: Column | null): boolean {
```

packages/iris-grid/src/mousehandlers/IrisGridColumnSelectMouseHandler.ts
```diff
@@ -1,23 +1,31 @@
...
     super();
 
     this.irisGrid = irisGrid;
     this.cursor = null;
   }
 
-  isActive() {
+  irisGrid: IrisGrid;
+
+  isActive(): boolean {
```
mofojed, line 23: Could be private
```suggestion
  private isActive(): boolean {
```

packages/iris-grid/src/mousehandlers/IrisGridColumnTooltipMouseHandler.ts
```diff
@@ -1,46 +1,54 @@
...
+  constructor(irisGrid: IrisGrid) {
     super();
 
     this.irisGrid = irisGrid;
   }
 
-  destroyColumnTooltip() {
+  irisGrid: IrisGrid;
+
+  destroyColumnTooltip(): void {
```
mofojed, line 21:
```suggestion
  private destroyColumnTooltip(): void {
```

packages/iris-grid/src/mousehandlers/IrisGridColumnTooltipMouseHandler.ts
```diff
@@ -1,46 +1,54 @@
...
     const { shownColumnTooltip } = this.irisGrid.state;
     if (shownColumnTooltip != null) {
       this.irisGrid.setState({ shownColumnTooltip: null });
     }
   }
 
-  hideColumnTooltip() {
+  hideColumnTooltip(): void {
```
mofojed, line 28:
```suggestion
  private hideColumnTooltip(): void {
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridModel.ts
mattrunyon: I did make the class abstract, but I'm not sure what methods should be abstract and which shouldn't

packages/iris-grid/src/mousehandlers/IrisGridColumnSelectMouseHandler.ts
mattrunyon: `import type` is completely stripped from the output, so it's not forbidden. But also not necessarily the best design.

Limiting to the props on the mouse handler might be better overall as it doesn't need an `IrisGrid` specifically. Some other component could have the same props. Although I'm not sure how TS will handle this when `IrisGrid` is actually TS (it's JS now). The import type might just be `JSX.Element` which isn't helpful

packages/grid/src/index.ts
```diff
@@ -6,16 +6,21 @@ export * from './GridMetricCalculator';
 export * from './GridMetrics';
 export { default as GridModel } from './GridModel';
 export { default as GridMouseHandler } from './GridMouseHandler';
+export * from './GridMouseHandler';
 export { default as GridRange } from './GridRange';
+export * from './GridRange';
 export { default as GridRenderer } from './GridRenderer';
+export * from './GridRenderer';
```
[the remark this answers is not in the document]
mattrunyon: Should I remove the `export default` from each class as well?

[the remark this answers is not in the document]
mofojed: Na

packages/iris-grid/src/IrisGridModel.ts
mofojed: Any of the methods that throw and do nothing else should be abstract

packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
```diff
@@ -235,6 +242,62 @@ class MockIrisGridTreeModel
...
+
+  get rollupConfig(): RollupConfig {
+    throw new Error('Not defined in mock');
+  }
+
+  set rollupConfig(rollupConfig: RollupConfig) {
+    throw new Error('Not defined in mock');
+  }
+
+  get selectDistinctColumns(): string[] {
```
mofojed, line 270: Can't the return here be `never[]` since it always throws?

mofojed (approved):
We can worry about the import cycle when converting IrisGrid to TS. Maybe an interface defined and used by the mouse handlers that IrisGrid can implement.

packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
mattrunyon: Yep. I changed all the methods that throw in the mock to return `never` or `Promise<never>` as needed.

mofojed (approved):

PR #411, opened 2022-02-01
upgrade react to v17

### LINKED ISSUE #385: Update React to v17
As a developer, I want React to be the latest stable version so that we're up to date on the latest.

spasovski:
fixes #385 

I expect this to be benign based on the release notes but I welcome additional testing by anyone else.


mofojed (changes requested):
code-studio/package.json was not updated to the correct version of react. After updating that, be sure to run an `npm run clean`, `npm install` to make sure all the package-locks are correct.

Also need to update `@types/react` across all packages to be v17

packages/chart/package.json
```diff
@@ -42,19 +42,19 @@
   },
   "peerDependencies": {
     "@deephaven/log": "^0.3.1",
-    "react": "^16.x"
+    "react": "^17.x"
```
mofojed, line 45: Not caused by you, but we should probably be consistent with our peer dependencies format. In some places we do `^17.x`, in others we do `^17.0.0` ... I think I prefer the latter, but I leave that as an exercise for another time.

mofojed (approved):

vbabich (approved):
Unit tests run without errors, smoke-tested the IU functionality - all looks good. 👍

PR #414, opened 2022-02-02
Fix pouchdb deprecation warning when starting up UI

### LINKED ISSUE #368: Remove deprecated PouchDB db.type calls
When loading DHC, there are warnings that `db.type() is deprecated and will be removed in a future version of PouchDB` emitted from `CommandHistoryPanel`. We should update those uses to whatever is appropriate and not deprecated

mofojed:
Fixes #368

### DIFF
--- a/packages/code-studio/src/storage/PouchStorageTable.ts
+++ b/packages/code-studio/src/storage/PouchStorageTable.ts
@@ -138,6 +138,10 @@ export class PouchStorageTable<T extends StorageItem = StorageItem>
   constructor(databaseName: string) {
     this.db = new PouchDB<T & PouchStorageItem>(`${DB_PREFIX}${databaseName}`);
 
+    // Need to set `_remote` to false to remove deprecation warnings: https://github.com/pouchdb/pouchdb/issues/6106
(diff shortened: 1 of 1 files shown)


packages/code-studio/src/storage/PouchStorageTable.ts
```diff
@@ -138,6 +138,10 @@ export class PouchStorageTable<T extends StorageItem = StorageItem>
   constructor(databaseName: string) {
     this.db = new PouchDB<T & PouchStorageItem>(`${DB_PREFIX}${databaseName}`);
 
+    // Need to set `_remote` to false to remove deprecation warnings: https://github.com/pouchdb/pouchdb/issues/6106
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, no-underscore-dangle
+    (this.db as any)._remote = false;
```
mattrunyon, line 143: Weird this needs to be cast to any

mattrunyon (approved):

packages/code-studio/src/storage/PouchStorageTable.ts
mofojed: @types/pouchdb-browser seems to be a few versions behind :/

PR #421, opened 2022-02-02
Fix the objects not being loaded with console

### LINKED ISSUE #405: Console table dropdown doesnt populate on opening DHC console from existing session
Steps to reproduce:

1. Start deephaven core, open a browser window to connect
2. Create a new table of some kind
3. Open a new browser tab and connect to the same server
4. Click the console table/widget pickers and try to see existing objects
image

Expected:
The existing tables will be available there.

Actual:
The menus are grayed out, no options are available.

Workaround:
Open the Panels menu instead, and select the missing objects from there.

mofojed:
- In Community, ConsolePanel is always loaded, so remove transition/isLoaded flags
- Keep track of objects in the ConsolePanel and pass it into Console
Fixes #405

### DIFF
--- a/packages/console/src/Console.jsx
+++ b/packages/console/src/Console.jsx
@@ -936,7 +936,7 @@ Console.defaultProps = {
   actions: [],
   disconnectedChildren: null,
   timeZone: 'America/New_York',
-  objectMap: undefined,
+  objectMap: new Map(),
 };
 
 export default Console;
--- a/packages/console/src/Console.jsx
+++ b/packages/console/src/Console.jsx
@@ -98,7 +98,7 @@ export class Console extends PureComponent {
     this.pending = new Pending();
     this.queuedLogMessages = [];
 
-    const { settings } = this.props;
+    const { objectMap, settings } = this.props;
 
     this.state = {
       // Need separate histories as console history has stdout/stderr output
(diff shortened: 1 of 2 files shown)


packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
```diff
@@ -96,6 +98,29 @@ class ConsolePanel extends PureComponent {
...
+
+  subscribeToFieldUpdates() {
+    const { sessionWrapper } = this.props;
+    const { session } = sessionWrapper;
+
+    this.objectSubscriptionCleanup = session.connection.subscribeToFieldUpdates(
+      updates => {
+        log.debug('Got updates', updates);
+
+        const { updated, created } = updates;
```
vbabich, line 112: Should we handle deletion as well?

packages/console/src/Console.jsx
```diff
@@ -453,6 +457,11 @@ export class Console extends PureComponent {
     });
   }
 
+  updateObjectMap() {
+    const { objectMap } = this.props;
+    this.setState({ objectMap });
```
vbabich, line 462: `objectMap` is optional according to `PropTypes`. Either do `{ objectMap: new Map(objectMap) }` here or update `PropTypes` (then drop `new Map()` in the `constructor` too to match `updateObjectMap`)

vbabich (changes requested):

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
mofojed: Good call, that's my bad. Fixed

vbabich (approved):

PR #427, opened 2022-02-03
Improve command history tooltip layout

### LINKED ISSUE #415: Multi-line commands don't show elapsed time in history
**Description**

**Steps to reproduce**
1. Run a large multi-line command, such as from a notebook
2. In the Command History, hover over the item for that long command until the tooltip appears

**Expected results**

The command appears (abbreviated, probably, so the tooltip is reasonably sized), including the runtime

**Actual results**

The tooltip has scrollbars that must be used to scroll down and find the elapsed time of the command

screenshot775

dsmmcken:
Resolves #415

Makes separate scroll panes for code and result.

### DIFF
--- a/packages/console/src/command-history/CommandHistoryItemTooltip.scss
+++ b/packages/console/src/command-history/CommandHistoryItemTooltip.scss
@@ -21,10 +21,10 @@
       padding: $spacer-1 $spacer-2;
       overflow: auto;
       max-height: 25vh;
-      .time-string {
+      .time-wrapper {
         text-align: right;
         color: $text-muted;
-        span {
+        .time-string {
           color: $foreground;
         }
       }
--- a/packages/console/src/command-history/CommandHistoryItemTooltip.scss
+++ b/packages/console/src/command-history/CommandHistoryItemTooltip.scss
@@ -1,23 +1,36 @@
 @import '../../../components/scss/custom.scss';
 .command-history-item-popper {
   padding: 0;
+  .popper-content .tooltip-content {
+    padding: 0;
+  }
   .command-history-item-tooltip {
+    display: flex;
+    flex-direction: column;
     max-height: 50vh;
     max-width: 50vw;
-    overflow: auto;
     text-align: left;
-    padding: $tooltip-padding-y $tooltip-padding-x;
+    .scroll-container {
+      padding: $spacer-2;
+      overflow: auto;
+      flex: 0 1 auto;
+    }
     .result-info {
-      text-align: right;
-    }
-    .time-string {
-      color: $text-muted;
-    }
-    hr {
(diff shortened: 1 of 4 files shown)


packages/console/src/command-history/CommandHistoryItemTooltip.scss
```diff
@@ -1,23 +1,36 @@
...
+      overflow: auto;
+      flex: 0 1 auto;
     }
-    hr {
-      background-color: $gray-400;
-    }
-    .error-message {
-      color: $danger;
+    .result-info {
+      flex: 1 0 auto;
```
mattrunyon, line 19: Why should this grow, but not the content? Shouldn't the popper have no empty space to fill?

mattrunyon (changes requested):
Fix failing tests. Looks like the display value changed from `5s` to `Elapsed time: 5s` and needs to be updated

packages/console/src/command-history/CommandHistoryItemTooltip.scss
dsmmcken: Test it visually and see if any combo looks weird. It should look good for all cases should be good if long messaga + short error (DHC), long message + long error (DHE), long message + no error, short message + long error, short message + short error (DHC)

.scroll-container will still claim space until .result + the parent height hits the 50vh max-height, then it will get scroll. There should still be no empty space to fill.

flex-grow only controls the case when the parent max-height is reached, at that point, however we don't want them do grow proportionally, we want the .result to grow to claim as much space as required for the .error to a max of 25vh.

Ideally .result should be able to grow the full height if scroll-container is only 1 line, but I couldn't figure out a way.

mattrunyon: flex-grow should only have an affect if the parent is bigger than the content of the children, no? I changed it to `0 0 auto` and didn't notice any differences

Doesn't really affect anything, I was just curious

mattrunyon (dismissed):
Up to you, but I think it would look better w/ error and time elapsed swapped. On wider tooltips you get a space in the bottom right corner w/ our typical error message

If you want to keep it as is then merge it

image

packages/console/src/command-history/CommandHistoryItemTooltip.scss
dsmmcken: You are correct, technically I just need the max-height on the result, and the flex-shrink, the rest can be left as default.

dsmmcken:
For long error messages (DHE), I still want the elapsed time accessible without scroll. I could pin it separately, but that seems unnecessary.

image

How about just adding something for balance:
image

mattrunyon (approved):

PR #429, opened 2022-02-04
Fix uncaught error from #404

mattrunyon:
The constants do not exist on DHC JS API, so we need to check if they exist first

Fixes bug from #404


spasovski (approved):
Great catch thanks!

PR #432, opened 2022-02-07
JS plugins loading from the server

mofojed:
Expects a manifest.json file of a format similar to the following:

```
{
  "plugins": [
    { "name": "@deephaven/matplotlib-js-plugin", "main": "@deephaven/matplotlib-js-plugin/dist/index.js" }
  ]
}
```
 
To help test, follow the instructions in https://github.com/mofojed/deephaven-core/tree/js-remote-plugin-2
It will load the matplotlib plugin published from https://github.com/mofojed/deephaven-js-dashboard-plugin/tree/matplotlib

BREAKING CHANGES:
- Now just use PanelEvent.OPEN/CLOSE, deprecate IrisGridEvent.OPEN/CLOSE, ChartEvent.OPEN/CLOSE, PandasEvent.OPEN/CLOSE.
- Now ConsolePanel/AppMainContainer just emit a PanelEvent.OPEN for all widget types, and registered plugins listen for and open based on those events.
- IrisGridModelFactory - Remove uncoalesced table check - in community, no tables are from a PQ so that check does not apply. It should be checked outside of IrisGridModelFactory (in the `fetch` function in `openWidget`)
- New `IrisGridEvent.CREATE_CHART` event. Previously `IrisGridPanel` (and subsequently the `GridPlugin`) had a dependency on `@deephaven/chart`. Now it no longer requires a dependency on `@deephaven/chart`. There is now the `ChartBuilderPlugin` that is the glue listening for the CREATE_CHART event and opening the proper ChartPanel. At some point we should instead pass the entire ChartBuilder UI in to the IrisGrid menu, so that it is completely abstracted out (think different charting libraries supporting different types of charts).

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.jsx
+++ b/packages/code-studio/src/main/AppMainContainer.jsx
@@ -85,6 +85,7 @@ export class AppMainContainer extends Component {
     this.handleExportLayoutClick = this.handleExportLayoutClick.bind(this);
     this.handleImportLayoutClick = this.handleImportLayoutClick.bind(this);
     this.handleImportLayoutFiles = this.handleImportLayoutFiles.bind(this);
+    this.handleLoadTablePlugin = this.handleLoadTablePlugin.bind(this);
     this.handleResetLayoutClick = this.handleResetLayoutClick.bind(this);
     this.handleWidgetMenuClick = this.handleWidgetMenuClick.bind(this);
     this.handleWidgetsMenuClose = this.handleWidgetsMenuClose.bind(this);
@@ -479,12 +480,29 @@ export class AppMainContainer extends Component {
     }
   }
 
+  /**
+   * Load a Table plugin specified by a table
+   * @param {string} pluginName The name of the plugin to load
+   * @returns {JSX.Element} An element from the plugin
+   */
+  handleLoadTablePlugin(pluginName) {
+    const { plugins } = this.props;
+
+    // First check if we have any plugin modules loaded that match the TablePlugin.
+    const pluginModule = plugins.get(pluginName);
+    if (pluginModule.TablePlugin) {
+      return pluginModule.TablePlugin;
+    }
+
+    return PluginUtils.loadComponentPlugin(pluginName);
+  }
+
   hydrateGrid(props, id) {
     const { session } = this.props;
     return {
       ...props,
       getDownloadWorker: DownloadServiceWorkerUtils.getServiceWorker,
-      loadPlugin: PluginUtils.loadComponentPlugin,
+      loadPlugin: this.handleLoadTablePlugin,
       localDashboardId: id,
       makeModel: () => createGridModel(session, props.metadata),
     };
@@ -519,9 +537,7 @@ export class AppMainContainer extends Component {
   getDashboardPlugins = memoize(plugins =>
     [...plugins.entries()]
       .filter(([, { DashboardPlugin }]) => DashboardPlugin)
-      .map(([name, { DashboardPlugin }]) =>
-        DashboardPlugin ? <DashboardPlugin key={name} /> : null
-      )
+      .map(([name, { DashboardPlugin }]) => <DashboardPlugin key={name} />)
   );
 
   render() {
@@ -625,7 +641,7 @@ export class AppMainContainer extends Component {
           <GridPlugin
             hydrate={this.hydrateGrid}
             getDownloadWorker={DownloadServiceWorkerUtils.getServiceWorker}
-            loadPlugin={PluginUtils.loadComponentPlugin}
+            loadPlugin={this.handleLoadTablePlugin}
           />
           <ChartPlugin hydrate={this.hydrateChart} />
           <ConsolePlugin />
--- a/packages/code-studio/src/main/AppMainContainer.jsx
+++ b/packages/code-studio/src/main/AppMainContainer.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
+import memoize from 'memoize-one';
 import { CSSTransition } from 'react-transition-group';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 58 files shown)


mofojed:
I'm getting an error with plotting. jQuery is not being set as expected:
```
[Chart] TypeError: jQuery is not a function
    at Object.triggerHandler (events.js:92:1)
    at Object._doPlot (plot_api.js:72:1)
    at Object.newPlot (plot_api.js:566:1)
    at Object.react (plot_api.js:2623:1)
    at factory.js:82:1
```
Debugging, it looks like initially, window.jQuery is set to a function as expected, but then by the time Chart is loaded, it's set to the module instead:
Initially:
```
window.jQuery
00:17:47.010 ƒ (selector, context) {
    // The jQuery object is actually just the init constructor 'enhanced'
    // Need init if jQuery is called (just allow error to be thrown if not included)
    return new jQu…
```
After:
```
Module {__esModule: true, Symbol(Symbol.toStringTag): 'Module'}
```

So something is changing the value?
Apparently this kind of thing can be fixed with webpack config, but we can't set that when not ejected:
```
new webpack.ProvidePlugin({
      $: 'jquery',
      jQuery: 'jquery',
      'window.jQuery': 'jquery',
    }),
```

Not sure why the problem is appearing now

mofojed:
Fixed the error related to jQuery. Was an issue with the way Webpack was used to package the plugin. Fixed with this change, adding config to `webpack.config.js` in the package: https://github.com/mofojed/deephaven-js-dashboard-plugin/commit/db32a6a6d5852b09f23aab5644e56332061a14d4

```
  plugins: [
    new webpack.ProvidePlugin({
      $: "jquery",
      jQuery: "jquery",
      "window.jQuery": "jquery",
    }),
  ],
```

packages/dashboard-core-plugins/src/PandasPlugin.tsx
```diff
@@ -66,6 +97,7 @@ export const PandasPlugin = (props: PandasPluginProps): JSX.Element => {
 
   useListener(layout.eventHub, PandasEvent.OPEN, handleOpen);
   useListener(layout.eventHub, PandasEvent.CLOSE, handleClose);
```
vbabich, line 99: Do we still need `PandasEvent.OPEN` and `PandasEvent.CLOSE`?

packages/dashboard-core-plugins/src/ChartPlugin.tsx
```diff
@@ -63,8 +69,7 @@ export const ChartPlugin = (props: ChartPluginProps): JSX.Element => {
     };
   }, [hydrate, registerComponent]);
 
-  useListener(layout.eventHub, ChartEvent.OPEN, handleOpen);
-  useListener(layout.eventHub, ChartEvent.CLOSE, handleClose);
+  useListener(layout.eventHub, PanelEvent.OPEN, handlePanelOpen);
```
vbabich, line 72: Looks like `IrisGridPanel` still emits `ChartEvent.OPEN` in `handleCreateChart`

packages/dashboard-core-plugins/src/events/IrisGridEvent.js
```diff
@@ -1,6 +1,8 @@
 class IrisGridEvent {
+  /** @deprecated Use PanelEvent.OPEN instead */
```
vbabich, line 2: Add similar comments to `ChartEvent`

packages/dashboard-core-plugins/src/GridPlugin.tsx
```diff
@@ -29,14 +30,26 @@ export const GridPlugin = (props: GridPluginProps): JSX.Element => {
...
-    ) => {
+      widget,
+    }: {
+      dragEvent?: DragEvent;
+      fetch: () => Promise<Table>;
+      panelId?: string;
+      widget: VariableDefinition;
+    }) => {
+      const { name, type } = widget;
+      if (type !== dh.VariableType.TABLE) {
```
vbabich, line 46: I see there's also `dh.VariableType.TREETABLE` type, should this plugin handle it as well?

vbabich (changes requested):

packages/code-studio/src/main/AppMainContainer.jsx
```diff
@@ -482,11 +485,49 @@ export class AppMainContainer extends Component {
...
+  handleLoadTablePlugin(pluginName) {
+    const { plugins } = this.props;
+
+    // First check if we have any plugin modules loaded that match the TablePlugin.
+    const pluginModule = plugins.get(pluginName);
+    if (pluginModule?.TablePlugin) {
+      return pluginModule.TablePlugin;
+    }
+
+    return PluginUtils.loadComponentPlugin(pluginName);
```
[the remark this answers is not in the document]
mofojed: It should be memoized in the @paciolan/remote-module-loader layer: https://github.com/Paciolan/remote-module-loader/blob/a2fab0a11b0d963051abca47b98442b2339cfe64/src/lib/loadRemoteModule.ts#L38

packages/dashboard-core-plugins/src/PandasPlugin.tsx
mofojed: Nope, this can be removed.

packages/dashboard-core-plugins/src/ChartPlugin.tsx
mofojed: I've created a new `IrisGridEvent.CREATE_CHART` to remove the dependency of grid onto chart. In the future, we should think of passing Chart Builder functionality in to the Iris Grid Menu so it can be defined completely externally.

packages/dashboard-core-plugins/src/GridPlugin.tsx
mofojed: We don't have tree tables yet but I should be listening for it in anticipation.

vbabich:
Should there be new commits to review? I got the request but don't see any new commits.

vbabich (approved):

PR #454, opened 2022-02-25
add component mount test for styleguide

### LINKED ISSUE #431: create jest test for styleguide
Can be just a "mounts without error" - we tend to break this via changes to IrisGrid.

spasovski:
fixes #431

### DIFF
--- a/packages/code-studio/src/styleguide/StyleGuide.test.jsx
+++ b/packages/code-studio/src/styleguide/StyleGuide.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import StyleGuide from './StyleGuide';
 
 class MockPath2D {}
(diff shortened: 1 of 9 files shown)


packages/code-studio/src/styleguide/StyleGuide.test.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { render, screen } from '@testing-library/react';
+import StyleGuide from './StyleGuide';
+
+class MockPath2D {}
+
+window.Path2D = MockPath2D;
+
+describe('<StyleGuide /> mounts', () => {
+  test('h1 text of StyleGuide renders', () => {
+    render(<StyleGuide />);
```
mofojed, line 11: I think you can do:
```
expect(() => render(<StyleGuide />)).not.toThrow();
```

mofojed (changes requested):

packages/code-studio/src/styleguide/StyleGuide.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { render } from '@testing-library/react';
```
mofojed, line 2: I think this needs to be added as a dev dependency for code-studio

```
./node_modules/.bin/lerna add @testing-library/react --scope=@deephaven/code-studio --dev
```

mofojed (changes requested):

Issue #445, opened 2022-02-18
EPIC: Improve Documentation


mofojed:
Our documentation needs to be beefed up in many areas to encourage usage.

- [ ] Docs for @deephaven/log
- [ ] Docs for @deephaven/grid
- [ ] Docs for @deephaven/iris-grid
- [ ] Docs for @deephaven/components

mofojed:
Documentation push should start in March

PR #460, opened 2022-03-01
Add the package directory to each packages package.json

mofojed:
- Manually updated the repository information for each package.json to specify the directory that package is in
- Should make relative links work correctly when packages are published
- Partial #445 
Implemented to match https://docs.npmjs.com/cli/v8/configuring-npm/package-json#repository

### DIFF
--- a/packages/babel-preset/package.json
+++ b/packages/babel-preset/package.json
@@ -2,7 +2,11 @@
   "name": "@deephaven/babel-preset",
   "version": "0.11.0",
   "description": "Deephaven Babel preset",
-  "repository": "https://github.com/deephaven/web-client-ui",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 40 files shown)


mofojed:
Kicked off an alpha build to test published links: https://github.com/deephaven/web-client-ui/actions/runs/1917076585

mofojed:
Created PR to address this in `marky-markdown`: https://github.com/npm/marky-markdown/pull/455

Not sure it will actually be updated though, since that package hasn't been updated in a while. Seems odd that _nobody_ has tried to fix this before, and the workaround is to just provide absolute links.

spasovski (dismissed):
Neat. Monorepo support.

spasovski (approved):
💯

PR #463, opened 2022-03-02
DH-10082 Fix unexpected time input behaviour

mofojed:
Now MaskedInput allows a function to be passed to provide the preferred replacement string. Check the current selection and the replacement char to see if it falls in a specific case for TimeInput, and provide a different string if necessary


mattrunyon (approved):
Works in the styleguide (only place we seem to use time inputs in DHC)

MaskedInput seems like it does too much/is too complex for a base component, but that's an issue for another day. And if it ain't broke don't fix it

PR #467, opened 2022-03-02
Remove storybook

### LINKED ISSUE #459: Remove storybook and stories
Remove our dependencies on storybook, remove all stories. We never really ended up using it, and we're just using StyleGuide to handle it.

mofojed:
We never really used it.
Fixes #459


packages/grid/package.json
```diff
@@ -35,6 +35,7 @@
   "devDependencies": {
     "@babel/cli": "^7.16.0",
     "@deephaven/tsconfig": "^0.11.0",
+    "@types/color-convert": "^2.0.0",
```
mattrunyon, line 38: Where did this come from? Also it shouldn't be in regular dependencies if this needs to stay (it's in both dev and regular right now)

mattrunyon (changes requested):

packages/grid/package.json
mofojed: I was getting a build error:
```
packages/grid/src/GridColorUtils.ts:1:21 - error TS7016: Could not find a declaration file for module 'color-convert'. '/home/bender/dev/deephaven/iris-oss/web-client-ui/main/node_modules/color-convert/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/color-convert` if it exists or add a new declaration (.d.ts) file containing `declare module 'color-convert';`

1 import convert from 'color-convert';
                      ~~~~~~~~~~~~~~~

packages/grid/src/GridColorUtils.ts:2:21 - error TS7016: Could not find a declaration file for module 'color-convert/conversions'. '/home/bender/dev/deephaven/iris-oss/web-client-ui/main/node_modules/color-convert/conversions.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/color-convert` if it exists or add a new declaration (.d.ts) file containing `declare module 'color-convert/conversions';`

2 import { HEX } from 'color-convert/conversions';
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

Will remove from dependencies

mofojed: It was required by `@storybooks/components`, so that's why it was getting installed before

mattrunyon (approved):
I didn't find any more references/files w/ story or stories

PR #475, opened 2022-03-08
DH-9053 expose chart titles when specified

spasovski:
I used the same query from the ticket:

```
sampleTable2 = db.t("LearnDeephaven", "EODTrades").where("ImportDate=`2017-11-01`")
 .where("Ticker=`CSC`")

samplePlot2 = plot("Price Trend", sampleTable2, "EODTimestamp", "Close")
 .chartTitle(sampleTable2, "Ticker")
 .show()
```

### DIFF
--- a/packages/chart/src/Chart.jsx
+++ b/packages/chart/src/Chart.jsx
@@ -10,6 +10,7 @@ import Plot from './plotly/Plot';
 
 import ChartModel from './ChartModel';
 import ChartUtils from './ChartUtils';
+import ChartTheme from './ChartTheme';
 import './Chart.scss';
 
 const log = Log.module('Chart');
@@ -182,11 +183,21 @@ export class Chart extends Component {
   initData() {
     const { model } = this.props;
     const { layout } = this.state;
+    const data = model.getData();
+    const title = data[0].name;
     this.setState({
-      data: model.getData(),
+      data,
       layout: {
         ...layout,
         ...model.getLayout(),
+        title: {
+          text: title,
+          font: {
+            color: ChartTheme.title_color,
+            size: 14,
+          },
+          y: 0.95, // relative to bottom of chart with range [0-1]
+        },
       },
     });
   }


packages/chart/src/Chart.jsx
```diff
@@ -182,11 +183,21 @@ export class Chart extends Component {
   initData() {
     const { model } = this.props;
     const { layout } = this.state;
+    const data = model.getData();
+    const title = data[0].name;
```
mofojed, line 187: Right now you're pulling the title from the first series, which is not what the ticket is asking for. The ticket is asking for the dynamic title (in the screenshot, it is "CSC". "Price Trend" is in the legend, showing the name of the series, and is not the title of the chart). 

Also, retrieval of the title should be done in `FigureChartModel.getDefaultTitle()`. 

Looking into it a bit though, it looks like this dynamic title doesn't come through the API, so this may actually require an API change... in `FigureWidgetTranslator.java` it's calling `clientChart.setTitle(chart.getTitle())`, which should work... but I think it doesn't have the title yet because the table data isn't initialized yet? As in the API needs to listen for the title change, and update the title and notify us when it does update... and I just don't think that's wired up.

A better example to show the dynamic title nature of this:
```
from deephaven.TableTools import timeTable
from deephaven.Plot import plot
tt = timeTable("00:00:01").update("X=i", "Y=i*i").reverse()
pp = plot("Test Table", tt, "X", "Y").chartTitle(tt, "Y").show()
```

That produces a chart where the title is the last Y value (video below). I've added these details to the ticket. But yea this requires an API change, and may or may not require a UI change afterward... so not really a snack.

https://user-images.githubusercontent.com/4505624/157460406-c6524651-daaf-4347-bf7b-2d2267d85eac.mp4

mofojed (changes requested):

spasovski:
Yea this is definitely the wrong approach...will close.

PR #479, opened 2022-03-11
DH-12233 Save rollups, aggregations, selectDistinct in filter sets

vbabich:


packages/dashboard-core-plugins/src/panels/IrisGridPanel.jsx
```diff
@@ -681,12 +669,29 @@ export class IrisGridPanel extends PureComponent {
...
     }
 
     try {
       const { gridState, irisGridState, irisGridPanelState } = panelState;
+      const {
+        quickFilters: indexedQuickFilters,
+        advancedFilters: indexedAdvancedFilters,
+      } = irisGridStateOverrides;
+      if (indexedQuickFilters) {
+        irisGridStateOverrides.quickFilters = IrisGridUtils.changeFilterColumnNamesToIndexes(
```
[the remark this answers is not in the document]
vbabich: Hmm... not sure how this happened. The update wasn't in the diff, VSCode showed it as 'unsaved' when I opened the file, but it still worked for me locally? Saved and pushed.

packages/dashboard-core-plugins/src/panels/IrisGridPanel.jsx
```diff
@@ -681,12 +669,29 @@ export class IrisGridPanel extends PureComponent {
 
   loadPanelState(model) {
     const { panelState } = this.props;
+    const { irisGridStateOverrides } = this.state;
     if (panelState == null) {
```
mofojed, line 673: This check causes it to return if you create a new panel and add filters and save them to the filter set manager without closing/re-opening the dashboard first. That's because the `panelState` provided in the props is undefined.

I think you should check both `panelState == null && Object.keys(irisGridStateOverrides).length === 0` and return then, then if `panelState == null`, use `panelState` from `this.state` instead.

mofojed (changes requested):

mofojed (approved):

packages/dashboard-core-plugins/src/panels/FilterSetManagerPanel.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function renderFilterSetManagerPanel({
+  filterSets = [],
+  localId = 'TEST_ID',
+  panelMap = new Map(),
+  tableMap = new Map(),
+  setFilterSets = jest.fn(),
+} = {}) {
+  render(
```
mofojed, line 32: You can do `return render`, then use `getByText`, instead of `screen.getByText`

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/FilterSetManagerPanel.test.tsx
vbabich: Good point, updated!

packages/dashboard-core-plugins/src/panels/FilterSetManagerPanel.test.tsx
```diff
@@ -1,9 +1,13 @@
...
+import { render, cleanup } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-
+// import { LayoutUtils } from '@deephaven/dashboard';
 import { FilterSetManagerPanel } from './FilterSetManagerPanel';
 
+// jest.mock('LayoutUtils', () => ({
+//   getPanelId: () => 'Grid1',
+// }));
+
```
mofojed, line 10: Can remove this commented out text

packages/dashboard-core-plugins/src/panels/FilterSetManagerPanel.test.tsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, cleanup } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-
+// import { LayoutUtils } from '@deephaven/dashboard';
```
mofojed, line 4: Remove commented out text

mofojed (changes requested):

mofojed (approved):

PR #486, opened 2022-03-15
rm enzyme: FormattingSectionContent refactor

spasovski:
Part of issue 440. This also adds some missed `for` attributes on a few labels.

### DIFF
--- a/packages/code-studio/src/settings/FormattingSectionContent.jsx
+++ b/packages/code-studio/src/settings/FormattingSectionContent.jsx
@@ -820,12 +820,18 @@ export class FormattingSectionContent extends PureComponent {
             Applies a formatting rule to all columns of a set type.
           </div>
           <div className="form-row mb-2">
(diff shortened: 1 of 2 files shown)


packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -820,12 +820,18 @@ export class FormattingSectionContent extends PureComponent {
...
+              htmlFor="select-reset-timezone"
+            >
+              Time zone
+            </label>
             <div className="col pr-0">
               <select
                 className="custom-select"
                 value={timeZone}
                 onChange={this.handleTimeZoneChange}
+                id="select-reset-timezone"
```
mofojed, line 834: Fine here since there's only ever one `FormattingSectionContent`, but as a side note if you need to add the `id` anywhere else - if it's in any component that could be re-used, you should give it a unique ID (generate a shortid on component creation and use it in the element ID)

PR #487, opened 2022-03-15
DH-12166: Allow truncating number columns with pound symbol

mattrunyon:
This adds the ability to truncate number columns with ### instead of cutting and adding an ellipsis at the end. There is a new setting in the settings panel to toggle the option.

Tested with this python snippet

```python
from deephaven.TableTools import newTable, intCol, doubleCol

table = newTable(
    intCol("Int", 123456789, 234567890, 42, 8675309, 1),
    doubleCol("Double", 123.456, 555555.5555, 1234567890.1234567890, 42.42, 1.0)
)
```

### DIFF
--- a/packages/code-studio/src/settings/FormattingSectionContent.jsx
+++ b/packages/code-studio/src/settings/FormattingSectionContent.jsx
@@ -191,6 +191,9 @@ export class FormattingSectionContent extends PureComponent {
     this.handleResetDecimalFormat = this.handleResetDecimalFormat.bind(this);
     this.handleResetIntegerFormat = this.handleResetIntegerFormat.bind(this);
     this.handleResetTimeZone = this.handleResetTimeZone.bind(this);
+    this.handletruncateNumbersWithPoundChange = this.handletruncateNumbersWithPoundChange.bind(
+      this
+    );
 
     const {
       formatter,
@@ -470,6 +473,17 @@ export class FormattingSectionContent extends PureComponent {
     );
   }
 
+  handletruncateNumbersWithPoundChange() {
+    this.setState(
+      state => ({
+        truncateNumbersWithPound: !state.truncateNumbersWithPound,
+      }),
+      () => {
+        this.debouncedCommitChanges();
+      }
+    );
+  }
+
   handleFormatRuleEntered(elem) {
     this.scrollToFormatBlockBottom();
     FormattingSectionContent.focusFirstInputInContainer(elem);
@@ -484,6 +498,7 @@ export class FormattingSectionContent extends PureComponent {
       timeZone,
       defaultDecimalFormatOptions,
       defaultIntegerFormatOptions,
+      truncateNumbersWithPound,
     } = this.state;
 
     const formatter = formatSettings
@@ -498,6 +513,7 @@ export class FormattingSectionContent extends PureComponent {
       showTimeZone,
       showTSeparator,
       timeZone,
+      truncateNumbersWithPound,
     };
     if (
       FormattingSectionContent.isValidFormat(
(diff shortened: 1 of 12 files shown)


packages/code-studio/src/settings/FormattingSectionContent.jsx
```diff
@@ -191,6 +191,9 @@ export class FormattingSectionContent extends PureComponent {
     this.handleResetDecimalFormat = this.handleResetDecimalFormat.bind(this);
     this.handleResetIntegerFormat = this.handleResetIntegerFormat.bind(this);
     this.handleResetTimeZone = this.handleResetTimeZone.bind(this);
+    this.handletruncateNumbersWithPoundChange = this.handletruncateNumbersWithPoundChange.bind(
```
mofojed, line 195: Change to `handleTruncateNumbersWithPoundChange` (casing on `t`)

mofojed, line 195: Change to `handleTruncateNumbersWithPoundChange` (casing on `t`)

packages/grid/src/GridModel.ts
```diff
@@ -50,6 +50,19 @@ abstract class GridModel<
    */
   abstract textForCell(column: ModelIndex, row: ModelIndex): string;
 
+  /**
+   * Get the truncation character for the specified cell. Can be undefined
```
mofojed, line 54:
```suggestion
   * Get the character to replace text when truncated for a specific cell. Leave undefined to show text truncated with ellipses
```

mofojed, line 54:
```suggestion
   * Get the character to replace text when truncated for a specific cell. Leave undefined to show text truncated with ellipses
```

packages/redux/src/store.ts
```diff
@@ -37,6 +37,7 @@ export interface WorkspaceSettings {
   timeZone: string;
   showTimeZone: boolean;
   showTSeparator: boolean;
+  truncateNumbersWithPound: boolean;
```
mofojed, line 40: While this does work... and I know we talked about it a bit... but since we're writing to workspace data I want to make sure we make it flexible enough we don't have to do a data migration later. I think it would be better to save it as `truncationChar?: string;` in `defaultDecimalFormatOptions`, `defaultIntegerFormatOptions`. 
The check box in settings will need to trigger to update both, if for some reason they're out of sync then just show the intermediate checkbox state (they shouldn't get out of sync though).
Having just this option kind of boxes us in.

mofojed, line 40: While this does work... and I know we talked about it a bit... but since we're writing to workspace data I want to make sure we make it flexible enough we don't have to do a data migration later. I think it would be better to save it as `truncationChar?: string;` in `defaultDecimalFormatOptions`, `defaultIntegerFormatOptions`. 
The check box in settings will need to trigger to update both, if for some reason they're out of sync then just show the intermediate checkbox state (they shouldn't get out of sync though).
Having just this option kind of boxes us in.

mofojed (changes requested):

packages/code-studio/src/settings/FormattingSectionContent.jsx
mattrunyon: Good catch

packages/redux/src/store.ts
mattrunyon: Per our call today, we'll just leave it as is since there aren't any current requests to make this a per-column option. If that's the case then we'd want to spec out what other truncation styles are needed (if any) and if this setting should be changed to something else or if it can be left as the global setting and then have the model account for specific formatters based on column name.

mofojed (approved):

PR #491, opened 2022-03-18
rm enzyme: refactor ChartBuilder tests to use RTL

spasovski:
RTL does not have a way to access component state intentionally. The guiding principle being to test user-facing functionality rather than internal implementation details. This definitely makes some of these tests a bit weaker but more resilient to component updates.

### DIFF
--- a/packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
+++ b/packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
@@ -48,7 +48,9 @@ it('has x-axis selection with the proper columns', () => {
   expect(xAxisSelectElm.querySelectorAll('option').length).toBe(
     COLUMN_NAMES.length
   );
-  expect(xAxisSelectElm.value).toBe(COLUMN_NAMES[0]);
+  userEvent.selectOptions(xAxisSelectElm, COLUMN_NAMES[1]);
+
+  expect(xAxisSelectElm.value).toBe(COLUMN_NAMES[1]);
   unmount();
 });
 
@@ -60,6 +62,8 @@ it('updates series selection with the proper columns', () => {
     COLUMN_NAMES.length
   );
   expect(seriesSelectElm.value).toBe(COLUMN_NAMES[1]);
+  userEvent.selectOptions(seriesSelectElm, COLUMN_NAMES[2]);
+  expect(seriesSelectElm.value).toBe(COLUMN_NAMES[2]);
 
   unmount();
 });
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -12,152 +13,114 @@ function makeChartBuilderWrapper({
...
+  const xAxisSelectElm = container.querySelector('.select-x-axis');
 
-  expect(wrapper.find('.select-x-axis').children().length).toBe(
+  expect(xAxisSelectElm.querySelectorAll('option').length).toBe(
     COLUMN_NAMES.length
   );
-  expect(wrapper.find('.select-x-axis').instance().value).toBe(COLUMN_NAMES[0]);
-  expect(wrapper.state('xAxis')).toBe(COLUMN_NAMES[0]);
-
-  wrapper.find('.select-x-axis').childAt(1).simulate('change');
```
mofojed, line 53: We should still have the change happen as a user event, see if the column names get updated

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -12,152 +13,114 @@ function makeChartBuilderWrapper({
...
+  expect(seriesSelectElm.querySelectorAll('option').length).toBe(
     COLUMN_NAMES.length
   );
-  expect(wrapper.find('.select-series').instance().value).toBe(COLUMN_NAMES[1]);
-  expect(wrapper.state('seriesItems').map(item => item.value)).toEqual([
-    COLUMN_NAMES[1],
-  ]);
+  expect(seriesSelectElm.value).toBe(COLUMN_NAMES[1]);
 
-  wrapper.find('.select-series').childAt(2).simulate('change');
```
mofojed, line 70: Still need to change

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -12,152 +13,114 @@ function makeChartBuilderWrapper({
...
-
-  wrapper.find('.btn-delete-series').at(0).simulate('click');
-
-  expect(wrapper.state('seriesItems').map(item => item.value)).toEqual([
-    COLUMN_NAMES[3],
-    COLUMN_NAMES[2],
-  ]);
-});
-
-it('updates linked state', () => {
```
mofojed, line 105: I think we still need this check. Instead of looking at `isLinked` in the state, we can listen for `onChange` prop to get called with the new `isLinked` value.

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -12,152 +13,114 @@ function makeChartBuilderWrapper({
...
+  expect(container.querySelectorAll('.form-series-item').length).toBe(1);
 
-  expect(wrapper.state('seriesItems').length).toBe(3);
+  userEvent.click(addSeriesItemBtn);
+  userEvent.click(addSeriesItemBtn);
 
-  wrapper.find('.select-series').at(1).childAt(3).simulate('change');
-  wrapper.find('.select-series').at(2).childAt(2).simulate('change');
-
-  expect(wrapper.state('seriesItems').map(item => item.value)).toEqual([
```
mofojed, line 91: Check that the series items match by querying the proper selector

mofojed (changes requested):

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
spasovski: couldn't quite do it that way but I did add a check for the input's value

mofojed:
Superceded by #505

PR #493, opened 2022-03-18
Re-enable Codecov reporting

### LINKED ISSUE #394: Fix Code Coverage Reporting
**Description**

Codecov reports are only getting the diff currently when running reports. The reports are therefore useless. Set up the reports to either correctly use the diff (is this even possible?), run the full tests when making a report, or just use a different code coverage reporting tool.

mattrunyon:
Fixes #394 

The Jest tests have been changed to run all unit tests and only lint changed files in CI. Local `npm test` still does changed only in watch mode.

Updated caching to utilize jest cache (helps schedule tests for faster runs) and webpack/terser caches. These are stored in `node_modules/.cache` so they need to be restored after `npm ci` which removes `node_modules` if it exists.

The cache for PRs should hit the previous tested commit in the PR or the latest target branch cache entry. On pushes to main, the cache should always hit the previous commit.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -93,15 +93,15 @@ jobs:
       - name: Test (Pull Request)
         if: ${{ github.event_name == 'pull_request' }}
         run: |
-          npm run test:unit -- --maxWorkers=2 --cacheDirectory node_modules/.cache/jest
-          npm run test:lint -- --maxWorkers=2 --cacheDirectory node_modules/.cache/jest --changedSince origin/${{ github.event.pull_request.base.ref }}
+          npm run test:unit -- --cacheDirectory node_modules/.cache/jest
+          npm run test:lint -- --cacheDirectory node_modules/.cache/jest --changedSince origin/${{ github.event.pull_request.base.ref }}
           npm run test:golden-layout
 
       - name: Test (Push)
         if: ${{ github.event_name == 'push' }}
         run: |
-          npm run test:unit -- --maxWorkers=2 --cacheDirectory node_modules/.cache/jest
-          npm run test:lint -- --maxWorkers=2 --cacheDirectory node_modules/.cache/jest --lastCommit
+          npm run test:unit -- --cacheDirectory node_modules/.cache/jest
+          npm run test:lint -- --cacheDirectory node_modules/.cache/jest --lastCommit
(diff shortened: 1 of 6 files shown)


.github/workflows/node.js.yml
```diff
@@ -12,13 +12,17 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.ref }}
```
mofojed, line 16: I think for safety we should include the workflow name as well: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow
Concurrency isn't used in any other workflows yet, but we do have other workflows

.github/workflows/node.js.yml
```diff
@@ -30,55 +34,79 @@ jobs:
...
+            ${{ runner.os }}-build-cache-
 
       - name: Build
         run: npm run build
 
       - name: Test (Pull Request)
         if: ${{ github.event_name == 'pull_request' }}
         run: |
-          npm run test:ci -- --changedSince origin/${{ github.event.pull_request.base.ref }}
+          npm run test:unit -- --cacheDirectory node_modules/.cache/jest
```
mofojed, line 96: Could use a comment here explaining why we run all the unit tests every time but only linting the main tests.

mofojed (dismissed):
Minor suggested comments

.github/workflows/node.js.yml
```diff
@@ -90,18 +90,20 @@ jobs:
...
         run: npm run build
 
+      # Run unit tests on all files for accurate code coverage
+      # Lint only changed files so linting is faster
       - name: Test (Pull Request)
         if: ${{ github.event_name == 'pull_request' }}
         run: |
-          npm run test:unit -- --cacheDirectory node_modules/.cache/jest
-          npm run test:lint -- --cacheDirectory node_modules/.cache/jest --changedSince origin/${{ github.event.pull_request.base.ref }}
+          npm run test:unit -- --maxWorkers=50% --cacheDirectory node_modules/.cache/jest
```
mofojed, line 98: This option doesn't seem to have much of an impact, so I suggest we omit the option so we don't have to explain it later

mofojed (approved):

PR #499, opened 2022-03-25
Update package-lock.json with new v0.11.3 version

mofojed:
npm install was failing... now it's working?


mofojed:
Replaced with #502

PR #505, opened 2022-03-31
Remove enzyme from ChartBuilder tests

mofojed:
- rm enzyme: refactor ChartBuilder tests to use RTL
- Send an event on isLinked update - before we weren't sending it, which wasn't really an issue because we didn't have live previews, but it was inconsistent.
- Replaces #491

### DIFF
--- a/packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
+++ b/packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import ChartBuilder from './ChartBuilder';
 import IrisGridTestUtils from '../IrisGridTestUtils';
@@ -24,19 +24,18 @@ it('renders without crashing', () => {
 });
 
 it('updates the chart type', () => {
-  const { container, unmount } = makeChartBuilderWrapper();
+  const { unmount } = makeChartBuilderWrapper();
 
-  let chartButtons = container.querySelectorAll('.btn-chart-type');
+  const lineButton = screen.getByText('Line');
+  const barButton = screen.getByText('Bar');
 
-  expect(chartButtons[0].classList.contains('active')).toBe(true);
-  expect(chartButtons[1].classList.contains('active')).toBe(false);
+  expect(lineButton.classList.contains('active')).toBe(true);
+  expect(barButton.classList.contains('active')).toBe(false);
 
-  userEvent.click(chartButtons[1]);
+  userEvent.click(barButton);
 
-  chartButtons = container.querySelectorAll('.btn-chart-type');
-
-  expect(chartButtons[0].classList.contains('active')).toBe(false);
-  expect(chartButtons[1].classList.contains('active')).toBe(true);
+  expect(lineButton.classList.contains('active')).toBe(false);
+  expect(barButton.classList.contains('active')).toBe(true);
 
   unmount();
 });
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -12,152 +13,153 @@ function makeChartBuilderWrapper({
...
 
 it('handles form submission', () => {
   const onSubmit = jest.fn();
-  const wrapper = makeChartBuilderWrapper({ onSubmit });
+  const { container, unmount } = makeChartBuilderWrapper({ onSubmit });
 
-  wrapper.find('.btn-submit').simulate('submit');
+  userEvent.click(container.querySelector('.btn-submit'));
 
   expect(onSubmit).toHaveBeenCalled();
```
mattrunyon, line 138: Should we `toHaveBeenCalledTimes(1)` to be a bit more specific?

mattrunyon (changes requested):

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -24,19 +24,18 @@ it('renders without crashing', () => {
...
 it('updates the chart type', () => {
-  const { container, unmount } = makeChartBuilderWrapper();
-
-  let chartButtons = container.querySelectorAll('.btn-chart-type');
+  const { unmount } = makeChartBuilderWrapper();
 
-  expect(chartButtons[0].classList.contains('active')).toBe(true);
-  expect(chartButtons[1].classList.contains('active')).toBe(false);
+  const lineButton = screen.getByText('Line');
+  const barButton = screen.getByText('Bar');
```
mattrunyon, line 30: I don't think you have to `import screen` here

It looks like render returns queries with their argument bound to the base element [See here](https://testing-library.com/docs/react-testing-library/api/#queries-1)

So instead you can just `const { unmount, getByText } = makeChartBuilderWrapper()`

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -24,19 +24,18 @@ it('renders without crashing', () => {
...
 
-  chartButtons = container.querySelectorAll('.btn-chart-type');
+  userEvent.click(barButton);
 
-  expect(chartButtons[0].classList.contains('active')).toBe(false);
-  expect(chartButtons[1].classList.contains('active')).toBe(true);
+  expect(lineButton.classList.contains('active')).toBe(false);
+  expect(barButton.classList.contains('active')).toBe(true);
 
   unmount();
```
mattrunyon, line 40: We might not need to call `unmount`. It looks like it might automatically be called by RTL since we're using Jest. https://testing-library.com/docs/react-testing-library/api#cleanup

See if it works without calling `unmount`. If you want to leave `unmount` for being explicit that's fine, but I think it's better to omit it and just let the `afterEach` handle it

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -69,24 +68,30 @@ it('updates series selection with the proper columns', () => {
 });
 
 it('add and deletes series items', () => {
-  const { container } = makeChartBuilderWrapper();
-  const addSeriesItemBtn = container.querySelector('.btn-add-series');
+  const { unmount } = makeChartBuilderWrapper();
+  const addSeriesItemBtn = screen.getByText('Add Series');
 
-  expect(container.querySelectorAll('.form-series-item').length).toBe(1);
+  expect(screen.getAllByTestId(/form-series-item-./).length).toBe(1);
```
mattrunyon, line 74: Same comment about taking these from the `render` result instead of `screen`

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -69,24 +68,30 @@ it('updates series selection with the proper columns', () => {
...
 
   userEvent.click(addSeriesItemBtn);
   userEvent.click(addSeriesItemBtn);
 
-  const seriesItems = container.querySelectorAll('.select-series');
-  userEvent.selectOptions(seriesItems[1], COLUMN_NAMES[3]);
-  userEvent.selectOptions(seriesItems[2], COLUMN_NAMES[2]);
+  const seriesItem1 = screen.getByTestId('select-series-item-1');
+  const seriesItem2 = screen.getByTestId('select-series-item-2');
```
mattrunyon, line 80: Use `render` result, not `screen`

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -69,24 +68,30 @@ it('updates series selection with the proper columns', () => {
...
+  userEvent.selectOptions(seriesItem1, COLUMN_NAMES[3]);
+  userEvent.selectOptions(seriesItem2, COLUMN_NAMES[2]);
 
-  expect(seriesItems[1].value).toBe(COLUMN_NAMES[3]);
-  expect(seriesItems[2].value).toBe(COLUMN_NAMES[2]);
+  expect(seriesItem1.value).toBe(COLUMN_NAMES[3]);
+  expect(seriesItem2.value).toBe(COLUMN_NAMES[2]);
 
-  userEvent.click(container.querySelector('.btn-delete-series'));
+  userEvent.click(screen.getByTestId('delete-series-1'));
```
mattrunyon, line 87: Use `render` result, not `screen`

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -69,24 +68,30 @@ it('updates series selection with the proper columns', () => {
...
 
-  expect(container.querySelectorAll('.form-series-item').length).toBe(2);
+  expect(screen.getAllByTestId(/form-series-item-./).length).toBe(2);
+  expect(screen.getByTestId('select-series-item-1').value).toBe(
```
mattrunyon, line 90: Use `render` result, not `screen`

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
```diff
@@ -69,24 +68,30 @@ it('updates series selection with the proper columns', () => {
...
+    COLUMN_NAMES[2]
+  );
+
+  unmount();
```
mattrunyon, line 94: Can probably remove `unmount`. See previous comment

mattrunyon (changes requested):

packages/iris-grid/src/sidebar/ChartBuilder.test.jsx
mofojed: Nice good call!

mattrunyon (approved):

PR #509, opened 2022-04-01
Fix flakey test

### LINKED ISSUE #508: CodeCov sometimes reports changes in ContextMenu
CodeCov sometimes reports changes in ContextMenu even though it hasn't changed. An example PR: https://github.com/deephaven/web-client-ui/pull/506

The steps run seem to be in a `requestAnimationFrame` block: https://app.codecov.io/gh/deephaven/web-client-ui/compare/506/changes
It's possible the `requestAnimationFrame` is causing some indeterminate behaviour. Mock the `requestAnimationFrame` to provide certainty.

dsmmcken:
Codecov reporting the test was changing coverage amount. I think it might be the requestAnimationFrame, and mocking that should fix it.

Resolves #508

### DIFF
--- a/packages/components/src/context-actions/ContextActions.test.tsx
+++ b/packages/components/src/context-actions/ContextActions.test.tsx
@@ -4,6 +4,20 @@ import { TestUtils } from '@deephaven/utils';
 import ContextActionUtils from './ContextActionUtils';
 import ContextMenuRoot from './ContextMenuRoot';
 
+let spy: jest.SpyInstance;
+beforeEach(() => {
+  spy = jest.spyOn(window, 'requestAnimationFrame').mockImplementation(cb => {
+    // mock request animation frame
+    // only safe to mock like this if RAF is non-recursive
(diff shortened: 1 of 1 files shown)


packages/components/src/context-actions/ContextActions.test.tsx
```diff
@@ -4,6 +4,20 @@ import { TestUtils } from '@deephaven/utils';
 import ContextActionUtils from './ContextActionUtils';
 import ContextMenuRoot from './ContextMenuRoot';
 
+let spy: jest.SpyInstance;
+beforeEach(() => {
+  spy = jest.spyOn(window, 'requestAnimationFrame').mockImplementation(cb => {
```
mattrunyon, line 9: I would maybe add a comment pointing to the issue on github or an explanation that this causes inconsistent code coverage because the RAF call can quit early if the test runs super fast

mattrunyon (approved):

PR #512, opened 2022-04-04
Add data-testid Attributes for CSV Download

dgodinez-dh:
Add data-testid attributes to allow test automation to use CSV Download.

### DIFF
--- a/packages/components/src/RadioItem.tsx
+++ b/packages/components/src/RadioItem.tsx
@@ -32,6 +32,9 @@ type RadioItemProps = {
 
   /** The value to associate with this radio item */
   value: string;
+
+  /** An id used for test */
+  dataTestid?: string;
 };
 
 /**
@@ -50,6 +53,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
       name,
       onChange,
       value,
+      dataTestid,
     } = props;
 
     const [id] = useState(shortid());
@@ -59,6 +63,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
         <input
           type="radio"
           id={id}
+          data-testid={dataTestid}
           name={name}
           ref={ref}
           className={classNames('custom-control-input', inputClassName, {
@@ -91,6 +96,7 @@ RadioItem.defaultProps = {
   labelClassName: '',
   name: undefined,
   onChange: undefined,
+  dataTestid: undefined,
 };
 
 export default RadioItem;
--- a/packages/components/src/RadioItem.tsx
+++ b/packages/components/src/RadioItem.tsx
@@ -34,7 +34,7 @@ type RadioItemProps = {
   value: string;
 
   /** An id used for test */
-  dataTestid?: string;
+  'data-testid'?: string;
 };
 
 /**
@@ -53,7 +53,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
       name,
       onChange,
       value,
-      dataTestid,
+      'data-testid': dataTestId,
     } = props;
 
     const [id] = useState(shortid());
@@ -63,7 +63,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
         <input
           type="radio"
           id={id}
-          data-testid={dataTestid}
+          data-testid={dataTestId}
           name={name}
           ref={ref}
           className={classNames('custom-control-input', inputClassName, {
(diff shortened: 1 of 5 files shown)


packages/components/src/RadioItem.tsx
```diff
@@ -32,6 +32,9 @@ type RadioItemProps = {
 
   /** The value to associate with this radio item */
   value: string;
+
+  /** An id used for test */
+  dataTestid?: string;
```
mofojed, line 37: I think this is a case where we want to just allow the `data-testid` to be passed just like that:
```suggestion
  'data-testid'?: string;
```
I'm suggesting that because I was looking at other common attributes, such as the `aria-label`, and those are defined in the types using the kebab-case names: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b92cbb35c8cb01cc2150736ec8b8640c6af3b085/types/react/index.d.ts#L1558

It is an exception to how props are normally defined in JSX, but given the number of aria-* props and data-* props, I think we should stick with just naming it `data-testid`. 

We should probably define a type `type DataAttributes = { 'data-testid'?: string }` and then use that in pretty much all of our components (in addition to extending `InputHTMLAttributes<HTMLInputElement>` in this case).

I'm going to ask the team as well to ensure we're all on the same page for this.

packages/components/src/RadioItem.tsx
```diff
@@ -50,6 +53,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
       name,
       onChange,
       value,
+      dataTestid,
```
mofojed, line 56: As per my above comment,
```suggestion
      'data-testid': dataTestId,
```

packages/components/src/RadioItem.tsx
```diff
@@ -59,6 +63,7 @@ const RadioItem = React.forwardRef<HTMLInputElement, RadioItemProps>(
         <input
           type="radio"
           id={id}
+          data-testid={dataTestid}
```
mofojed, line 66:
```suggestion
          data-testid={dataTestId}
```

packages/components/src/RadioItem.tsx
```diff
@@ -91,6 +96,7 @@ RadioItem.defaultProps = {
   labelClassName: '',
   name: undefined,
   onChange: undefined,
+  dataTestid: undefined,
```
mofojed, line 99:
```suggestion
  'data-testid': undefined,
```

packages/components/src/navigation/Page.tsx
```diff
@@ -26,6 +26,7 @@ export const Page = ({
         {onBack !== undefined && (
           <button
             className="btn btn-link btn-back"
+            data-testid="page-back-button"
```
mofojed, line 29: Any reason to not use similar naming as the class? E.g.
```suggestion
            data-testid="btn-page-back"
```

packages/components/src/navigation/Page.tsx
```diff
@@ -39,6 +40,7 @@ export const Page = ({
         {onClose !== undefined && (
           <button
             className="btn btn-link btn-link-icon btn-close px-2 m-1"
+            data-testid="page-close-button"
```
mofojed, line 43:
```suggestion
            data-testid="btn-page-close"
```

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -3295,6 +3295,7 @@ export class IrisGrid extends Component {
             >
               <button
                 type="button"
+                data-testid={`${name}-iris-grid-table-menu-button`}
```
mofojed, line 3298:
```suggestion
                data-testid={`btn-iris-grid-settings-button-${name}`}
```

packages/iris-grid/src/sidebar/TableCsvExporter.jsx
```diff
@@ -226,15 +226,21 @@ class TableCsvExporter extends Component {
             value={downloadOption}
             disabled={isDownloading}
           >
-            <RadioItem value={TableCsvExporter.DOWNLOAD_OPTIONS.ALL_ROWS}>
+            <RadioItem
+              value={TableCsvExporter.DOWNLOAD_OPTIONS.ALL_ROWS}
+              dataTestid="csv-exporter-download-all-radio"
```
mofojed, line 231: As mentioned above prefer the input type to be first in the naming convention, similar to how bootstrap suggests: https://codeguide.co/#css-classes

mofojed (changes requested):

mofojed (approved):

PR #517, opened 2022-04-06
Add grid ThemeContext for providing the theme in a context

### LINKED ISSUE #516: Allow passing in theme to Grid from context
As a developer, I want to be able to provide the theme using a context wrapper so that I don't have to pass the theme in as a prop on each Grid instance on my page.

This could be handy for our Grid docs.

mofojed:
- Added theme context to styleguide as well
- Fixes #516

### DIFF
--- a/packages/grid/src/ThemeContext.ts
+++ b/packages/grid/src/ThemeContext.ts
@@ -1,0 +1,8 @@
+import React from 'react';
+import GridTheme, { GridTheme as GridThemeType } from './GridTheme';
+
+export const ThemeContext: React.Context<
+  Partial<GridThemeType>
+> = React.createContext(GridTheme as Partial<GridThemeType>);
+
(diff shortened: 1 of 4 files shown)


packages/grid/src/ThemeContext.ts
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import GridTheme, { GridTheme as GridThemeType } from './GridTheme';
+
+export const ThemeContext: React.Context<
+  Partial<GridThemeType>
+> = React.createContext(GridTheme as Partial<GridThemeType>);
```
vbabich, line 6: Change the default to an empty object, since we merge the value with `GridTheme` in `Grid.getTheme` anyway.

vbabich (changes requested):

vbabich (approved):

PR #518, opened 2022-04-06
Add @testing-library/jest-dom, jest-canvas-mock

### LINKED ISSUE #511: Add @testing-library/jest-dom
Add the @testing-library/jest-dom package: https://github.com/testing-library/jest-dom
This will help us write tests more succinctly. I.e. Instead of writing:
```
expect(chartButtons[0].classList.contains('active')).toBe(true);
```

We would just write:
```
expect(chartButtons[0]).toHaveClass('active)
```

mofojed:
- Added @testing-library/jest-dom
- Added jest-canvas-mock
- Migrated jest.setup.js to jest.setup.ts
- Cleaned up the unit tests using .classList.contains checks to use .toHaveClass instead
- Fixes #511


packages/eslint-config/index.js
```diff
@@ -79,6 +79,7 @@ module.exports = {
           'error',
           { ignoreFunctionalComponents: true },
         ],
+        'import/no-extraneous-dependencies': 'off',
```
mattrunyon, line 82: What was this causing a lint error on? I don't think we should turn this off for all TS files

If it's the `jest.setup.ts` then just add a 1 time ignore there. Test files ignore this error because test util packages should be dev dependencies since they aren't required for prod.

mattrunyon (changes requested):

jest.setup.ts
```diff
@@ -1,12 +1,12 @@
 import 'regenerator-runtime/runtime';
+import '@testing-library/jest-dom';
+import 'jest-canvas-mock';
 import { configure } from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
-import { TestUtils } from './packages/utils/src';
```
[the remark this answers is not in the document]
mofojed: It was being used previously to set `getContext` below. But we don't need that anymore with `jest-canvas-mock` being imported.

packages/eslint-config/index.js
mofojed: It was actually the unused `embed-grid/setupTests.ts` file, I've just deleted that and added this rule back.

That being said - a few lines above the rule is already applied to all js/jsx/ts/tsx files - if I add that rule back, there's about 30 failures in test files. That can be saved for another ticket.

mattrunyon: That rule is only applied to `test.{js,jsx,ts,tsx}` files which I think is fine. It triggers test failures because we import dev dependencies for tests sometimes. But those dependencies should be left as dev since they aren't used in prod code

mattrunyon (approved):

packages/eslint-config/index.js
mofojed: Ah. My dumb ass just can't read a simple regex. My bad! Good call.

PR #535, opened 2022-04-27
Fix vertical scroll bar when FileExplorer tab shown

mofojed:
- There was a vertical scroll bar shown when the FileExplorer tab was the full height of the dashboard
- Add height/width 100% so it fills the container

### DIFF
--- a/packages/file-explorer/src/FileExplorer.scss
+++ b/packages/file-explorer/src/FileExplorer.scss
@@ -6,8 +6,6 @@ $file-explorer-bg: $gray-900;
   background: $file-explorer-bg;
   position: relative;
   flex-grow: 1;
-  width: 100%;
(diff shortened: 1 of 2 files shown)


dsmmcken:
This doesn't fix the issue.

packages/file-explorer/src/FileExplorer.scss
```diff
@@ -6,6 +6,8 @@ $file-explorer-bg: $gray-900;
   background: $file-explorer-bg;
   position: relative;
   flex-grow: 1;
+  width: 100%;
+  height: 100%;
```
dsmmcken, line 10:
```suggestion
```

I don't think this is needed anymore

dsmmcken (approved):

PR #538, opened 2022-04-28
DH-10208 Fix arrow up at top of history loading empty history commands

### LINKED ISSUE #537: Arrow up past the top of command history creates temp empty commands
**Description**

If you arrow up several times past the end of the command history, you must arrow down several times to get back to your first command. I believe this broke in #386 

**Steps to reproduce**
1. Arrow up to the first command in your command history
2. Press arrow up 3 more times
3. Arrow down

**Expected results**

First command (or really 2nd. I don't think we should go past the top)

**Actual results**

A blank temp command. Must arrow down 2 more times to reach the top command in the history

**Versions**
 - Deephaven: 0.11

mattrunyon:
Fixes #537. This was discovered as part of DH-10208 testing which is tangentially related to where the bug was actually introduced.

Now at the top of history in the console input, pressing arrow up should not move up to other commands

I tried to write a test for this w/ RTL, but was hitting obstacles w/ Monaco so I just skipped it for now

### DIFF
--- a/packages/console/src/ConsoleInput.jsx
+++ b/packages/console/src/ConsoleInput.jsx
@@ -311,18 +311,20 @@ export class ConsoleInput extends PureComponent {
    * @param {number | null} index The index to load. Null to load command started in the editor and not in the history
    */
   loadCommand(index) {
-    let value = '';
-    if (index !== null && index < this.history.length) {
-      value = this.history[this.history.length - index - 1];
-
-      if (index > this.history.length - BUFFER_SIZE) {
-        this.loadMoreHistory();
-      }
+    if (index >= this.history.length) {
+      return;
     }
 
+    const modifiedValue = this.modifiedCommands.get(index);
+    const historyValue =
+      index === null ? '' : this.history[this.history.length - index - 1];
+
     this.commandHistoryIndex = index;
-    value = this.modifiedCommands.get(index) ?? value;
-    this.commandEditor.getModel().setValue(value);
+    this.commandEditor.getModel().setValue(modifiedValue ?? historyValue);
+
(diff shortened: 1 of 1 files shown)


packages/console/src/ConsoleInput.jsx
```diff
@@ -311,18 +311,20 @@ export class ConsoleInput extends PureComponent {
...
    */
   loadCommand(index) {
-    let value = '';
-    if (index !== null && index < this.history.length) {
-      value = this.history[this.history.length - index - 1];
-
-      if (index > this.history.length - BUFFER_SIZE) {
-        this.loadMoreHistory();
-      }
+    if (index >= this.history.length) {
```
mofojed, line 314:
```suggestion
    if (index !== null && index >= this.history.length) {
```

packages/console/src/ConsoleInput.jsx
```diff
@@ -311,18 +311,20 @@ export class ConsoleInput extends PureComponent {
...
+    const modifiedValue = this.modifiedCommands.get(index);
+    const historyValue =
+      index === null ? '' : this.history[this.history.length - index - 1];
+
     this.commandHistoryIndex = index;
-    value = this.modifiedCommands.get(index) ?? value;
-    this.commandEditor.getModel().setValue(value);
+    this.commandEditor.getModel().setValue(modifiedValue ?? historyValue);
+
+    if (index > this.history.length - BUFFER_SIZE) {
```
mofojed, line 325:
```suggestion
    if (index !== null && index > this.history.length - BUFFER_SIZE) {
```

mofojed (changes requested):

mofojed (approved):

PR #543, opened 2022-04-29
DH-10111 Fix missing horizontal scroll bar for single column tables

### LINKED ISSUE #446: Horizontal scroll never appears for single column table
**Description**

With a single column table, it is not possible to trigger the horizontal scroll bar even if the column extends past the edge of the container

**Steps to reproduce**
1. Create a single column table
```python
from deephaven.TableTools import newTable, stringCol

source = newTable(
    stringCol("Long", "A"*50),
)
```
2. Resize such that the column extends past the edge of the container

**Expected results**

Horizontal scroll appears

**Actual results**

No horizontal scroll

This only appears to be a problem for tables w/ 1 column. I believe the problem is in `GridMetricCalculator` specifically where `hasHorizontalBar` is calculated. This also technically affects tables w/ a single row and vertical scroll probably, but a panel smaller than a row's height is pretty useless to scroll

**Versions**
 - Deephaven: 0.9.3

mattrunyon:
Fixes #446 and related to DH-10111 test failure

This adds a horizontal scroll bar for any single column table which does not fit in the panel fully

Example Python to test with
```python
import random
import string

letters = string.printable
strings = [''.join(random.choice(letters) for i in range(random.randint(0, 1000))) for j in range(0, 100)]
table1 = new_table([string_col("Random", strings)])
```

Also tested with changing `IrisGridTheme.rowHeaderWidth` to be non-zero

### DIFF
--- a/packages/grid/src/GridMetricCalculator.ts
+++ b/packages/grid/src/GridMetricCalculator.ts
@@ -303,7 +303,7 @@ export class GridMetricCalculator {
     const maxY = rowHeightValues.reduce((y, h) => y + h, 0) - topOffset;
 
     // Visible space available in the canvas viewport
-    const viewportWidth = width - gridX - rowHeaderWidth;
+    const viewportWidth = width - gridX;
 
     // How much total space the content will take
     const contentWidth = leftOffset + maxX + rowFooterWidth;
--- a/packages/grid/src/GridMetricCalculator.ts
+++ b/packages/grid/src/GridMetricCalculator.ts
@@ -1,3 +1,4 @@
+import clamp from 'lodash.clamp';
 import GridModel from './GridModel';
 import type {
   GridMetrics,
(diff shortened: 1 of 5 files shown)


packages/grid/src/GridMetricCalculator.ts
```diff
@@ -319,47 +325,54 @@ export class GridMetricCalculator {
...
-      lastTop > 0
-        ? Math.min(
-            Math.max(
-              minScrollHandleSize,
-              barHeight * ((rowCount - lastTop) / rowCount)
-            ),
-            barHeight - 1
-          )
-        : 0;
+    // How big the horizontal is relative to the horizontal bar
```
mofojed, line 335: We should be doing this for both horizontal and vertical scroll bars. Row height can be resized to be larger than the viewport as well (though not a use case we currently have).

mofojed (changes requested):
Shift+Wheel (which should scroll horizontally) isn't working. I think need to update Grid.tsx `handleWheel` function

packages/grid/src/GridMetricCalculator.ts
mattrunyon: We are doing this for vertical bar currently, I just didn't rename it like I did horizontal. Do you want me to rename vertical too so it's more clear?

mattrunyon:
Updated so vertical scroll uses the same logic as horizontal

Tested by setting a big `rowHeight` in `IrisGridTheme.module.scss`. Also fixed/tested horizontal scroll with `scrollSnapToColumn: true` in `IrisGridTheme.ts`

PR #550, opened 2022-05-03
Added data-testid

### LINKED ISSUE #513: Add data-testid attribute to all @deephaven/components
In our @deephaven/components library, we should add the `data-testid` as a prop to each component. This will allow for passing in a testid for cases where we need to be more explicit in testing.

SparkyFnay:
Fixes #513,
Added data-testid to a few components in ./components

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,17 +1,14 @@
 {
+  "editor.guides.bracketPairs": "active",
   "typescript.tsdk": "node_modules/typescript/lib",
   "editor.formatOnSave": true,
   "editor.tabSize": 2,
+  "editor.bracketPairColorization.enabled": true,
   "eslint.workingDirectories": [
     {
       "mode": "auto"
     }
   ],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "stylelint.validate": [
-    "css",
-    "html",
-    "sass",
-    "scss"
-  ]
+  "stylelint.validate": ["css", "html", "sass", "scss"]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,7 @@
 {
-  "editor.guides.bracketPairs": "active",
   "typescript.tsdk": "node_modules/typescript/lib",
   "editor.formatOnSave": true,
   "editor.tabSize": 2,
-  "editor.bracketPairColorization.enabled": true,
   "eslint.workingDirectories": [
     {
       "mode": "auto"
--- a/packages/components/src/UISwitch.test.tsx
+++ b/packages/components/src/UISwitch.test.tsx
@@ -18,17 +18,14 @@ it('mounts and unmounts properly', () => {
 });
 
 it('get element by data-testid works', () => {
-  let clicked = false;
-  const onClick = () => {
-    clicked = true;
-  };
+  const onClick = jest.fn();
   const testId = 'test id for UISwitch';
   const testSwitch = makeUISwitch({ 'data-testid': testId, onClick });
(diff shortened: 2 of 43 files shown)


SparkyFnay:

packages/components/src/UISwitch.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+it('mounts and unmounts properly', () => {
+  function click() {
+    return true;
+  }
+
+  const testId = 'test id for UISwitch';
+  const testSwitch = makeUISwitch(testId, click);
+  testSwitch.unmount();
```
mofojed, line 17: With the above change, you don't need to create any parameters as you'll just use the defaults.
We also don't need to explicitly call `unmount()`, since that happens automatically during the `cleanup` phase with `testing-library/react`: https://testing-library.com/docs/react-testing-library/api/#cleanup

So then this test simply becomes:
```suggestion
  makeUISwitch();
```

packages/components/src/navigation/MenuItem.tsx
```diff
@@ -63,7 +65,7 @@ export const MenuItem = ({
       )}
       <div className="title">{title}</div>
       {subtitle && <div className="shortcut">{subtitle}</div>}
-      <div className="accessory">
+      <div className="accessory" data-testid={dataTestId}>
```
mofojed, line 68: Put the test id on the outermost div (`.btn-navigation-menu-item`)

mofojed (changes requested):

.vscode/settings.json
```diff
@@ -1,17 +1,14 @@
 {
+  "editor.guides.bracketPairs": "active",
```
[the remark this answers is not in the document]
mattrunyon: Only if it's a change we want enabled by default for our team. It should probably go in user settings instead.

Unfortunately VSCode doesn't have support for shared workspace settings and custom workspace settings. It's just workspace and global. There's a longstanding issue to support it https://github.com/microsoft/vscode/issues/40233

Basically some teams don't commit `.vscode/settings.json` because the devs want all settings within their vscode separated by project. And some do because they have a default list of settings that is useful (like ours)

packages/components/src/UISwitch.test.tsx
```diff
@@ -1,34 +1,34 @@
...
+  const testSwitch = makeUISwitch({ 'data-testid': testId, onClick });
   const elements = testSwitch.getAllByTestId(testId);
   expect(elements.length).toBe(1);
   const button = elements[0];
   expect(button instanceof HTMLButtonElement).toBe(true);
-  expect(result).toBe(false);
+  expect(clicked).toBe(false);
   userEvent.click(button);
-  expect(result).toBe(true);
+  expect(clicked).toBe(true);
```
mofojed, line 33: Reading this again - all you're really doing here is checking that the method is called when the button is clicked. You could simply check that the jest fn was called instead of having your own intermediate variable, for example:
```
it('get element by data-testid works', () => {
  const onClick = jest.fn();
  const testId = 'test id for UISwitch';
  const testSwitch = makeUISwitch({ 'data-testid': testId, onClick });
  const elements = testSwitch.getAllByTestId(testId);
  expect(elements.length).toBe(1);
  const button = elements[0];
  expect(button instanceof HTMLButtonElement).toBe(true);
  expect(onClick).not.toHaveBeenCalled();
  userEvent.click(button);
  expect(onClick).toHaveBeenCalledTimes(1);
});
```

mofojed (changes requested):

mofojed (approved):

PR #566, opened 2022-05-11
Fix various layout hint issues

### LINKED ISSUE #560: Freezing a column resizes other columns incorrectly
**Description**

Freezing a column after a table has already been fetched will incorrectly resize other columns. This only happens to columns between the start and where the frozen column was initially. Example video where the short string column expands to be the width of the long string column.

https://user-images.githubusercontent.com/8619789/167085431-7fd2e2b6-bdc9-43d7-a613-4c113e1d23ec.mov

**Steps to reproduce**
1. Run this Python
```python
from deephaven.column import string_col, char_col, int_col
import random
import string

table2 = new_table([
    string_col("Long", ["A"*50 , "B"*500, "C"*5000, "D"*49900]),
    char_col("Char", "ABCD"),
    int_col("Int", [123456789, 234567890, 42, 8675309]),
])
```
2. Right click on the "Char" column header and freeze column.

**Expected results**

Column is effectively reordered and floating at the front. No resizes

**Actual results**

Column floats to the front, but columns in between update to the width of the previous columns in their location (or recalculates to larger)

I believe the issue is that the `IrisGridTableModel.columns` getter was changed to not return the table columns. This is problematic because the table columns is the model order and should never be changed. The engine does nothing to change the model for frozen columns. So after the freeze we're calling model column 2 (the Char column) model row 0, but only within the table model. The API still correctly says the Char column is index 2.

This should maybe be patched in 0.6.x for Bard since I believe this functionality shipped in Bard.

**Versions**
 - Deephaven: 0.11.8

mattrunyon:
Fixes #560. Also disables context menu freeze option on other locked columns (front/back). Added a check for columns listed in multiple layout hints which will display an error. 

Instead of modifying the model columns (which caused the initial issue), the model now defines a starting list of moved columns. The list is derived from the layout hints. If a user makes more changes, they are simply added to the moved columns list. The model column order is left untouched. The model base moved columns will be useful for column groups via layout hints

When a user unfreezes a column, it is moved to just after the remaining frozen columns + front locked columns. 

This may cause odd behavior in DHE PQs if the underlying table changes (but the original implementation would too). We should probably have some way of versioning PQ tables like we do for shared dashboards (if we don't already), because all other user modifications to a table would be incorrect if the underlying table changed.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -1173,29 +1173,26 @@ export class IrisGrid extends Component {
     (
       alwaysFetchColumns,
       columns,
+      movedColumns,
       floatingLeftColumnCount,
       floatingRightColumnCount
     ) => {
-      let floatingLeftColumns = [];
-      let floatingRightColumns = [];
+      const floatingColumns = [];
 
-      if (floatingLeftColumnCount) {
-        floatingLeftColumns = columns
-          .slice(0, floatingLeftColumnCount)
-          .map(col => col.name);
+      for (let i = 0; i < floatingLeftColumnCount; i++) {
+        floatingColumns.push(
+          columns[GridUtils.getModelIndex(i, movedColumns)].name
+        );
       }
 
-      if (floatingRightColumnCount) {
-        floatingRightColumns = columns
-          .slice(-floatingRightColumnCount)
-          .map(col => col.name);
+      for (let i = 0; i < floatingRightColumnCount; i++) {
+        floatingColumns.push(
+          columns[GridUtils.getModelIndex(columns.length - 1 - i, movedColumns)]
+            .name
+        );
       }
 
-      const columnSet = new Set([
-        ...alwaysFetchColumns,
-        ...floatingLeftColumns,
-        ...floatingRightColumns,
-      ]);
+      const columnSet = new Set([...alwaysFetchColumns, ...floatingColumns]);
 
       return [...columnSet];
     }
@@ -3316,6 +3313,7 @@ export class IrisGrid extends Component {
                 alwaysFetchColumns={this.getAlwaysFetchColumns(
                   alwaysFetchColumns,
                   model.columns,
+                  movedColumns,
                   model.floatingLeftColumnCount,
                   model.floatingRightColumnCount
                 )}
--- a/packages/iris-grid/src/IrisGrid.jsx
+++ b/packages/iris-grid/src/IrisGrid.jsx
@@ -309,8 +309,8 @@ export class IrisGrid extends Component {
       isFilterBarShown,
       isSelectingPartition,
       model,
-      movedColumns,
-      movedRows,
+      movedColumns: movedColumnsProp,
+      movedRows: movedRowsProp,
       partition,
       partitionColumn,
       rollupConfig,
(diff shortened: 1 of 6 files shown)


mattrunyon:
Fixed an issue where the frozen columns were not being fetched if there were enough columns in the table that not all data is fetched when scrolling right

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -131,6 +132,16 @@ abstract class IrisGridModel<
     return this.columns;
   }
 
+  /** List of column movements defined by the model. Used as initial movements for IrisGrid */
+  get movedColumns(): MoveOperation[] {
```
mofojed, line 136: I'm going to need to think about this a bit... but essentially `movedColumns` doesn't belong in IrisGridModel, since it doesn't know about visible stuff.
Before, it applied all the layoutHint stuffs on the underlying model, translating the columns before upper layers consumed.

Now, I think it more appropriate if you just pass up the layout hints, and IrisGrid translates those to the appropriate `movedColumns`. Since IrisGrid and Grid concern themselves with the layout, whereas the models are just access to the data.

mofojed: Of course, upon reflection, I realize `floatingLeftCount` and all those options are dependent on the visual order... even those should just be at the Grid/IrisGrid level instead of GridModel/IrisGridModel. What's the right path going forward? Going to need to think about that one a bit.

mattrunyon: It might be time to start designing Grid V2 or at least documenting what goes where with reasons. I think the component/model/renderer/metrics are slowly starting to overlap somewhat and the boundary is becoming less clear.

In my opinion the actual component shouldn't do much more than pull in all the parts and tie them together w/ the JS API

packages/iris-grid/src/IrisGrid.jsx
```diff
@@ -1536,42 +1539,56 @@ export class IrisGrid extends Component {
...
   unFreezeColumnByColumnName(columnName) {
-    const { frozenColumns } = this.state;
+    const { frozenColumns, movedColumns } = this.state;
     const { model } = this.props;
     log.debug2('unfreezing column', columnName);
 
-    let allFrozenColumns;
+    const allFrozenColumns =
+      frozenColumns == null
+        ? new Set(model.layoutHints?.frozenColumns)
```
mofojed, line 1574: Should just be `model.frozenColumns`, as the result is then memoized:
```suggestion
        ? new Set(model.frozenColumns)
```

packages/iris-grid/src/IrisGridModel.ts
mofojed: Agreed, the division is not clear. Especially with things like the floating columns and floating rows.

In terms of Grid v2 and responding to some of your other ideas:
- Treating floating grids as separate grids I think is a great idea, would just need to make sure the grids stay in sync.
- Treating events kind of like it is a DOM (such as subscribing to a `headerEvent` or something). The raw mouse event handlers already have the coordinate/cell that's clicked on, perhaps some more info could be added to that. Exposing the raw mouse events/handlers that we do right now on just the grid canvas is quite flexible, but it gets sticky whenever we mix actual DOM in though (such as for advanced filters and input tables). In reality I think we should simplify out just the `GridCanvas` component that just draws the data and emits the raw events. Then at a higher level, you can add a mouse handler that and combine it for the `headerEvent` (or other such events); and with floating rows/columns, they'd be separate `GridCanvas` (going back to the first point).

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.jsx
mattrunyon: Good catch. This is how the previous version accessed the model frozen columns.

I also switched the same call in freeze by name ~25 lines up from this one

mofojed (approved):

mofojed:
Discussion for V2 of Grid here: https://github.com/deephaven/web-client-ui/discussions/587

PR #574, opened 2022-05-12
Use session type provided by the server instead of hardcoding python

### LINKED ISSUE #573: allow iframes to work with groovy
Currently when trying to get an iframe from groovy, get the following error:

```
Error: session type 'python' is not supported
```

mofojed:
Now you can embed grids from Groovy sessions as well

Fixes #573


dsmmcken:
> Session type could be also selectable via url params

url params can't set the session type though, so I don't see a point to doing it that way.

mofojed:
I mean aside from the point that for this grid it's language agnostic and doesn't care if it's python OR groovy. Would rather there was some way to connect and get a table without specifying a language if you're not going to be running commands.

PR #575, opened 2022-05-12
Fix floating column border not touching top and bottom

### LINKED ISSUE #556: Frozen column double border on right edge does not touch top or bottom
**Description**

There is a double border on the right edge of frozen columns. It is persistent when scrolled and ~~causes a gap in row hover/select styles~~ the hover/select styles go up to the right edge of the border

image

**Versions**
 - Deephaven: 0.11.8

mattrunyon:
Fixes #556

### DIFF
--- a/packages/grid/src/GridRenderer.ts
+++ b/packages/grid/src/GridRenderer.ts
@@ -499,15 +499,15 @@ export class GridRenderer {
         getOrThrow(visibleRowYs, floatingTopRowCount - 1) +
         getOrThrow(visibleRowHeights, floatingTopRowCount - 1) +
         0.5;
-      context.moveTo(0.5, y);
-      context.lineTo(maxX - 0.5, y);
+      context.moveTo(0, y);
+      context.lineTo(maxX, y);
     }
 
     if (floatingBottomRowCount > 0) {
       const y =
         getOrThrow(visibleRowYs, rowCount - floatingBottomRowCount) - 0.5;
-      context.moveTo(0.5, y);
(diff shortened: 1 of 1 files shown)


mofojed (commented):
Hmm why do we not need to offset y by 0.5 anymore but we still need to offset x? I wonder what's changed

mattrunyon:
Not sure. This was added pre DHC so we'd have to dig into some old commits. Maybe the header bottom border wasn't always half overlapping?

mofojed (approved):
I'd like to know why we need to offset x but not y, but :shrug:

dsmmcken:
Let me review this more before merging, half pixels might have been wrong here, but they might also be wrong elsewhere.

dsmmcken (changes requested):
Looks like it was always like that https://gitlab.deephaven.io/illumon/iris/-/commit/8a659bf0efbd0a319e625808709af4883050b083 but based on the size of that review was likely just missed.

Half-pixels only go on the cross axis. So on the x position for vertical lines, and on the y for horizontal lines.

The half-pixels are correct in the const x/y declarations, but all the other ones are also wrong and should be removed as well.

mattrunyon:
@dsmmcken Can you clarify what other half pixels? Are you just talking about in this specific function where there's `0.5, y` and `maxX - 0.5, y`?

dsmmcken:
You fixed the incorrect half pixels on Y for the vertical lines in drawFloatingBorders, but also should fix the same for X on the horizontal lines. All within that function.

dsmmcken (approved):

PR #576, opened 2022-05-12
BREAKING: Clean up iris-grid dependencies

### LINKED ISSUE #549: Remove chart dependency on iris-grid
Current `@deephaven/chart` depends on `@deephaven/iris-grid`, solely for importing the following classes:
- Formatter
- FormatterUtils
- DateUtils
- TableUtils

Pull these into their own package, jsapi-utils, so they can be used without requiring all of iris-grid and it's dependencies. This should clean up some other dependencies (such as the input filter plugins).

mofojed:
- Move formatters, DateUtils, and TableUtils out from iris-grid into @deephaven/jsapi-utils
- Move Filter type definitions to @deephaven/filters
- No more `iris-grid/dist` imports
- @deephaven/chart no longer depends on @deephaven/iris-grid
- Fixes #549


packages/code-studio/src/styleguide/Grids.jsx
```diff
@@ -5,7 +5,7 @@ import {
   MockTreeGridModel,
   ThemeContext,
 } from '@deephaven/grid';
-import IrisGrid from '@deephaven/iris-grid/dist/IrisGrid';
+import IrisGrid from '@deephaven/iris-grid';
```
mattrunyon, line 8: I think there's preference for named exports over defaults in ESM now. Supposedly the defaults were mostly for CJS compatibility and named only make some difference to potential tree shaking and refactoring. Also auto importing works better w/ named imports.

Could be a separate ticket if it's something we want to enable
[eslint rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-default-export.md)
[some reasons to avoid default](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad)
[some more reasons to avoid default](https://rajeshnaroth.medium.com/avoid-es6-default-exports-a24142978a7a)

Looks like 1 exception is on code splitting in React

packages/filters/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "babel": "babel ./src --out-dir ./dist --extensions \".ts,.tsx,.js,.jsx\" --source-maps --root-mode upward",
+    "build:babel": "npm run babel",
+    "watch": "run-p watch:*",
+    "watch:babel": "npm run babel -- -w --skip-initial-build",
+    "prestart": "npm run build-dev",
+    "start": "cross-env NODE_ENV=development npm run watch"
+  },
+  "dependencies": {
+    "prop-types": "^15.7.2"
+  },
```
mattrunyon, line 31: I don't think this dependency is used

packages/jsapi-utils/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "build-dev": "cross-env NODE_ENV=development run-p build:*",
+    "babel": "babel ./src --out-dir ./dist --extensions \".ts,.tsx,.js,.jsx\" --source-maps --root-mode upward",
+    "build:babel": "npm run babel",
+    "watch": "run-p watch:*",
+    "watch:babel": "npm run babel -- -w --skip-initial-build",
+    "prestart": "npm run build-dev",
+    "start": "cross-env NODE_ENV=development npm run watch"
+  },
+  "dependencies": {
+    "prop-types": "^15.7.2",
```
mattrunyon, line 30: Unused dependency

packages/redux/package.json
```diff
@@ -27,6 +27,7 @@
     "start": "cross-env NODE_ENV=development npm run watch"
   },
   "dependencies": {
+    "@deephaven/jsapi-utils": "file:../jsapi-utils",
```
mattrunyon, line 30: The other `@deephaven` package dependencies list a version and not the local file

packages/storage/package.json
```diff
@@ -30,10 +30,12 @@
     "lodash.throttle": "^4.1.1"
   },
   "peerDependencies": {
+    "@deephaven/filters": "^0.11.7",
     "@deephaven/log": "^0.11.2",
     "react": "^17.0.0"
   },
   "devDependencies": {
+    "@deephaven/filters": "file:../filters",
```
mattrunyon, line 38: I think the file path is fine for dev deps though. Since you would have the entire monorepo downloaded for development, but not for installing as a normal dep in a project

mattrunyon (dismissed):

packages/code-studio/src/styleguide/Grids.jsx
mofojed: Created a ticket for this: https://github.com/deephaven/web-client-ui/issues/585

packages/redux/package.json
mofojed: I think we use `file` everywhere now.

mofojed: Just to clarify - we use `file` in all `dependencies` and `devDependencies`, but not for `peerDependencies`.
When the package is published, the `package.json` generated converts the `file:....` to the actual version number. You can confirm this by doing an `npm pack @deephaven/redux@latest` and checking the `package.json`. It has the actual version number of the deps instead of keeping the file tag.
I don't quite understand why it converts deps and devDeps but does not convert peerDeps.

packages/storage/package.json
mofojed: See my above comment - it is fine for `devDependencies` and `dependencies`, does not appear to be fine for `peerDependencies` (which we should probably remove from most of our packages anyway.

packages/redux/package.json
mofojed: Another note - I converted to `file:...` because on version bump, lerna would bump these versions and then `npm install` wouldn't work - because the version hadn't been published yet!

vbabich (approved):

vbabich (approved):

PR #583, opened 2022-05-17
Improved clipboard write handling

### LINKED ISSUE #376: Improve clipboard write handling
In `ContextActionUtils.copyToClipboard()` we currently check for the `clipboard-write` permission. This isn't on the w3c spec any more and in Firefox results in a rejected promise. When that happens we fall back to the old/deprecated style of writing to the clipboard.

According to MDN, the `clipboard-write` permission is given to any active tab [link](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)

MDN is somewhat inconsistent b/c the [Permissions API](https://developer.mozilla.org/en-US/docs/Web/API/Permissions) is not finalized yet.

We should just try writing to the clipboard async and then on rejected promise assume we don't have the permission and fallback to the old method. `writeText` exists in all browsers but IE for over a year now.

SparkyFnay:
Fixes #376

### DIFF
--- a/packages/components/src/context-actions/ContextActionUtils.ts
+++ b/packages/components/src/context-actions/ContextActionUtils.ts
@@ -115,8 +115,8 @@ class ContextActionUtils {
    * @param text The text to copy
    * @returns Promise Resolved on success, rejected on failure
    */
-  static async copyToClipboard(text: string): Promise<void> {
-    await navigator.clipboard.writeText(text).catch(err => {
+  static copyToClipboard(text: string): Promise<void> {
+    return navigator.clipboard.writeText(text).catch(() => {
       ContextActionUtils.copyToClipboardExecCommand(text);
     });
   }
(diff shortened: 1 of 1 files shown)


packages/components/src/context-actions/ContextActionUtils.ts
```diff
@@ -116,20 +116,9 @@ class ContextActionUtils {
...
-    try {
-      const permissions = await navigator.permissions.query({
-        name: 'clipboard-write' as PermissionName,
-      });
-      if (permissions.state !== 'granted' && permissions.state !== 'prompt') {
-        throw new Error('Invalid permissions for clipboard-write');
-      }
-    } catch (e) {
-      // Fallback if we can't get permissions for some reason
+    await navigator.clipboard.writeText(text).catch(err => {
```
mattrunyon, line 119: You shouldn't need the `await` here any more or the `async` in the function name. Just `return navigator.clipboard...`

mattrunyon (changes requested):

mattrunyon (approved):
For future PRs, it's helpful to put things you tested in the description. Like for this PR I would put something like "Tested copy row and copy cell via context menu and keyboard shortcut"

PR #588, opened 2022-05-18
Converted styleguide to typescript

### LINKED ISSUE #584: Convert styleguide to TypeScript

SparkyFnay:
Fixes #584


packages/code-studio/src/styleguide/DropdownMenus.tsx
```diff
@@ -16,17 +16,37 @@ import {
...
+  isShown: boolean;
+}
+
+interface Actions {
+  title: string;
+  icon: IconDefinition;
+  action: () => void;
+  shortcut: Shortcut;
+  group?: number;
+}
```
mofojed, line 35: Instead of declaring `Actions` here, we should export the type from `DropdownMenu`. So at the top of DropdownMenu, export a type and use it here, e.g.

```

export type DropdownAction = ContextAction & { actions?: never };

export type DropdownActions = DropdownAction | DropdownAction[];

type DropdownMenuProps = {
  actions: DropdownActions;
  ...
```

Will need to add it to `components/src/index.ts` as well.

And here then do `as DropdownAction[]` instead of `as Actions[]`

mofojed (changes requested):

packages/code-studio/src/styleguide/DropdownMenus.tsx
```diff
@@ -26,13 +26,13 @@ interface DropdownMenusState {
...
-  shortcut: Shortcut;
-  group?: number;
-}
+// interface Actions {
+//   title: string;
+//   icon: IconDefinition;
+//   action: () => void;
+//   shortcut: Shortcut;
+//   group?: number;
+// }
```
mofojed, line 35: Delete unused code
```suggestion
```

mofojed (changes requested):

packages/code-studio/src/storage/WorkspaceStorage.ts
```diff
@@ -1,12 +1,8 @@
...
-  columnType: string;
-  columnName: string;
-  format: string;
-};
+import { FormattingRule } from '@deephaven/iris-grid';
 
 export type WorkspaceSettings = {
   defaultDateTimeFormat: string;
-  formatter: WorkspaceFormattingRule[];
+  formatter: FormattingRule[];
```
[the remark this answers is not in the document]
mofojed: Created #592 for followup.

mofojed (approved):

PR #598, opened 2022-05-31
Refactor GridUtils move items

mattrunyon:
Adds model -> visible range transform methods and refactors the moved item methods to utilize the same logic w/ an optional reverse flag for converting from visible -> model

The model -> visible transform is needed for column header groups

### DIFF
--- a/packages/grid/src/GridUtils.test.ts
+++ b/packages/grid/src/GridUtils.test.ts
@@ -1,6 +1,6 @@
 import GridMetrics, { ModelIndex, MoveOperation } from './GridMetrics';
 import GridRange, { GridRangeIndex } from './GridRange';
-import GridUtils, { AxisRange } from './GridUtils';
+import GridUtils from './GridUtils';
 
 function expectModelIndexes(
   movedItems: MoveOperation[],
@@ -166,18 +166,16 @@ describe('floating item checks', () => {
   });
 });
 
-describe('start/end range adjustment in one dimension', () => {
+describe('start/end range adjustment in one dimension visible to model', () => {
   function testRange(
     start: GridRangeIndex,
     end: GridRangeIndex,
     movedItems: MoveOperation[] = [],
     expectedResult = [[start, end]]
   ) {
-    expect(
-      GridUtils.getModelRangeIndexes(start, end, movedItems).sort(
-        (a, b) => (a as AxisRange)[0] - (b as AxisRange)[0]
-      )
-    ).toEqual(expectedResult);
+    expect(GridUtils.getModelRangeIndexes(start, end, movedItems)).toEqual(
+      expectedResult
+    );
   }
   it('handles no transforms', () => {
     testRange(0, 0);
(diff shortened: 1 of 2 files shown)


packages/grid/src/GridUtils.test.ts
```diff
@@ -190,15 +188,38 @@ describe('start/end range adjustment in one dimension', () => {
     testRange(10, 20, movedItems);
   });
   it('handles items moved into the range', () => {
-    const movedItems = GridUtils.moveItem(100, 7);
-    testRange(5, 10, movedItems, [
+    testRange(5, 10, GridUtils.moveItem(1, 7), [
+      [6, 10],
+      [1, 1],
```
mofojed, line 193: I would've thought this would be `[6,6], [1,1], [7,10]` based on our discussion yesterday... If we're not guaranteeing the order (which we may not need to - if it's important then you can just getModelRangeIndexes with the specific indices that matter for you), then may as well sort the results so test cases are more consistent.

packages/grid/src/GridUtils.test.ts
```diff
@@ -226,9 +246,159 @@ describe('start/end range adjustment in one dimension', () => {
       [52, 52],
     ]);
   });
+  it('handles moves within the range', () => {
+    testRange(5, 10, GridUtils.moveItem(9, 6), [
+      [5, 5],
+      [9, 9],
+      [6, 8],
+      [10, 10],
```
mofojed, line 254: See this implies the order is maintained, unlike the test case above. If we don't care, we should sort/consolidate. If we do care, then the order should be specific in the first case above as well.

packages/grid/src/GridUtils.ts
```diff
@@ -745,90 +745,107 @@ export class GridUtils {
...
 
   /**
-   * Translate the provided UI start/end indexes to the model start/end indexes by applying the `movedItems` transformations.
-   * Since moved items can split apart a range, multiple pairs of indexes are returned
-   *
-   * @param start Start item in one dimension
-   * @param end End item in one dimension
-   * @param movedItems Moved item pairs in this dimension
-   * @returns Array of start/end pairs of the indexes after transformations applied.
+   * Applies the items moves to the 1D range
```
mofojed, line 759: 1D range, also been calling it an AxisRange

packages/grid/src/GridUtils.ts
```diff
@@ -745,90 +745,107 @@ export class GridUtils {
...
+
+  private static applyItemMoves(
+    start: GridRangeIndex,
+    end: GridRangeIndex,
+    movedItems: MoveOperation[],
+    reverse = false
   ): GridAxisRange[] {
+    if (typeof start === 'string') {
+      return [];
+    }
```
mofojed, line 789: Why is this case here? There should be no case where the start/end is a string

packages/grid/src/GridUtils.ts
```diff
@@ -745,90 +745,107 @@ export class GridUtils {
...
     if (start == null || end == null) {
       return [[start, end]];
     }
```
mofojed, line 792: Should add a test case for `null`s... as I think this is incorrect. If the range is `[5, null]` and we have move `[1, 10]`, then I would expect it to be `[[6, 9], [1, 1], [10, null]]`... (`null` meaning selecting to the end of that range).

packages/grid/src/GridUtils.ts
```diff
@@ -745,90 +745,107 @@ export class GridUtils {
...
+  ): [number, number][];
+
+  // Without this duplicate signature, TS will complain
+  private static applyItemMoves(
     start: GridRangeIndex,
     end: GridRangeIndex,
-    movedItems: MoveOperation[]
+    movedItems: MoveOperation[],
+    reverse?: boolean
+  ): GridAxisRange[];
```
mofojed, line 779: I was trying to think of a way to write this with generics instead, but couldn't figure it out :(

mofojed (changes requested):

packages/grid/src/GridUtils.ts
mattrunyon: I was testing if the method overload problems I was having were due to the types being subsets of each other (`number` vs `number | null`) or if it was some weirder TS issue. I switched `number` to `string` and forgot to remove this check after determining it was a weirder TS issue where I needed to duplicate the method signature

packages/grid/src/GridUtils.ts
mattrunyon: I got something to work. Requires a cast at the return, but the rest of it seems to work

The reason I added the overloads was the `getVisibleIndex` and `getModelIndex` method calls `applyItemMoves` w/ a known non-null value. But previously this always returned possibly null values in the array. So I needed a null check or assertion that _should_ never fail

mofojed (dismissed):

PR #605, opened 2022-06-07
Convert Iris-Grid from Javascript to Typescript

### LINKED ISSUE #590: Convert @deephaven/iris-grid package to TypeScript

SparkyFnay:
Fixes #590


mofojed (changes requested):
The `dist` folders (`filters/dist`, `jsapi-utils/dist`) should not be included as part of this review - they should be gitignored files. I think if you just delete those folders and commit those changes it should resolve the issue.

packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -495,17 +585,19 @@ class AdvancedFilterCreator extends PureComponent {
...
-                  table={valuesTable}
-                  onChange={this.handleSelectValueChange}
-                  invertSelection={invertSelection}
-                  selectedValues={selectedValues}
-                  formatter={formatter}
-                  showSearch={!isDateType}
-                  timeZone={formatter.timeZone}
-                />
-              </div>
+              {valuesTable && (
```
mofojed, line 588: We should _probably_ put a spinner here while the valuesTable is loading... something like:
```
{ !valuesTable && <LoadingSpinner ... />}
```

Not necessary as part of this change though.

mofojed (changes requested):
Not done yet - leaving some comments for the parts I've scanned so far.

packages/iris-grid/src/AdvancedFilterCreatorSelectValue.tsx
```diff
@@ -1,17 +1,47 @@
...
+  selectedValues: T[];
+  table: Table;
+  formatter: Formatter;
+  onChange: (selectedValues: T[], invertSelection: boolean) => void;
+  showSearch: boolean;
+  timeZone: string;
+}
+
+interface AdvancedFilterCreatorSelectValueState<T> {
+  error: unknown;
```
mofojed, line 22: In this class, we know the `error` is going to be a `string`. Make it optional here:
```suggestion
  error?: string;
```

packages/iris-grid/src/AdvancedFilterCreatorSelectValue.tsx
```diff
@@ -73,15 +106,19 @@ class AdvancedFilterCreatorSelectValue extends PureComponent {
     this.stopUpdateFilterTimer();
   }
 
-  getColumnName() {
+  searchTablePromise: Promise<Table> | null;
+
+  updateFilterTimer: NodeJS.Timeout | null;
```
mofojed, line 111: Prefer optional `?` over `null`.

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
```diff
@@ -85,14 +119,16 @@ class AdvancedFilterCreatorSelectValueList extends PureComponent {
...
   }
 
-  handleSelect(itemIndex, value) {
+  list: SelectValueList<T> | null;
+
+  handleSelect(itemIndex: number, value: T): void {
     const { invertSelection } = this.props;
     let { selectedValues } = this.state;
-    selectedValues = [].concat(selectedValues);
+    selectedValues = ([] as T[]).concat(selectedValues);
```
mofojed, line 131: Just do `[...selectedValues]`
```suggestion
    selectedValues = [...selectedValues];
```

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
```diff
@@ -172,28 +208,28 @@ class AdvancedFilterCreatorSelectValueList extends PureComponent {
...
+  stopListening(table?: Table): void {
     if (table != null) {
       table.removeEventListener(dh.Table.EVENT_UPDATED, this.handleTableUpdate);
     }
   }
 
-  resetViewport() {
-    if (this.list) {
+  resetViewport(): void {
+    if (this.list && this.list.topRow && this.list.bottomRow) {
```
mofojed, line 224: Check `topRow != null` and `bottomRow != null` instead of the truthy value here. Right now the number `0` will not pass, even though it should.

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
```diff
@@ -234,7 +270,9 @@ class AdvancedFilterCreatorSelectValueList extends PureComponent {
           onSelect={this.handleSelect}
           onViewportChange={this.handleViewportChange}
           ref={list => {
-            this.list = list;
+            if (list) {
```
mofojed, line 273: I don't think you need to wrap this in `if (list)`, we can update the ref to `null` if it is `null`

packages/iris-grid/src/ColumnStatistics.tsx
```diff
@@ -128,7 +159,7 @@ class ColumnStatistics extends Component {
     onStatistics();
   }
 
-  handleError(error) {
+  handleError(error: CanceledPromiseError): void {
     if (error && error.isCanceled) {
```
mofojed, line 163: Should just be type `Error`, then check if `PromiseUtils.isCanceled(error)`

packages/iris-grid/src/FormatterUtils.ts
```diff
@@ -3,30 +3,37 @@ import Formatter from './Formatter';
...
-  }): FormattingRule[] {
-    const { formatter } = settings;
-    return formatter;
+  static getColumnFormats(
+    settings?:
+      | {
+          formatter?: FormattingRule[];
+        }
+      | null
+      | undefined
```
mofojed, line 12: Shouldn't need `| null | undefined` here, the `settings?` is already optional.
```suggestion
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -116,6 +168,16 @@ function isEmptyConfig({
...
   selectDistinctColumns,
   sorts,
+}: {
+  advancedFilters: Map<number, AdvancedFilter>;
+  aggregationSettings: AggregationSettings;
+  customColumns: string[];
+  quickFilters: Map<number, QuickFilter>;
+  reverseType: ReverseType;
+  rollupConfig: UIRollupConfig | null;
+  searchFilter: FilterCondition | null;
```
mofojed, line 178: Prefer optionals:
```suggestion
  rollupConfig?: UIRollupConfig;
  searchFilter?: FilterCondition;
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -130,14 +192,317 @@ function isEmptyConfig({
...
+    defaultFormatString?: string;
+  };
+  formatter?: FormattingRule[];
+  timeZone?: string;
+  showTimeZone?: boolean;
+  showTSeparator?: boolean;
+  truncateNumbersWithPound?: boolean;
+  disableMoveConfirmation?: boolean;
+  showSystemBadge?: boolean;
+};
```
mofojed, line 210: We should only need to define the settings that IrisGrid actually uses here. So I think we can drop `disableMoveConfirmation` and `showSystemBadge` for example.
```suggestion
type Settings = {
  defaultDateTimeFormat?: string;
  defaultDecimalFormatOptions?: {
    defaultFormatString?: string;
  };
  defaultIntegerFormatOptions?: {
    defaultFormatString?: string;
  };
  formatter?: FormattingRule[];
  timeZone?: string;
  showTimeZone?: boolean;
  showTSeparator?: boolean;
  truncateNumbersWithPound?: boolean;
};
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -130,14 +192,317 @@ function isEmptyConfig({
...
+  onError: (error: unknown) => void;
+  onDataSelected: (index: number, map: unknown) => void;
+  onStateChange: (irisGridState: IrisGridState, gridState: GridState) => void;
+  onPartitionAppend: (partitionColumn: Column, value: string) => void;
+  onAdvancedSettingsChange: AdvancedSettingsMenuCallback;
+  partition: string;
+  partitionColumn: Column;
+  sorts: Sort[];
+  reverseType: ReverseType;
+  quickFilters: Map<number, QuickFilter> | null;
```
mofojed, line 270:
```suggestion
  quickFilters?: Map<ModelIndex, QuickFilter>;
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -130,14 +192,317 @@ function isEmptyConfig({
...
+
+export interface UITotalsTableConfig extends TotalsTableConfig {
+  operationOrder: AggregationOperation[];
+  showOnTop: boolean;
+}
+
+export type InputFilter = { name: string; type: string; value: string };
+export interface IrisGridProps {
+  children: React.ReactNode;
+  advancedFilters: Map<number, AdvancedFilter>;
```
mofojed, line 244:
```suggestion
  advancedFilters: Map<ModelIndex, AdvancedFilter>;
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -130,14 +192,317 @@ function isEmptyConfig({
...
+  advancedSettings: Map<AdvancedSettingsType, boolean>;
+  alwaysFetchColumns: string[];
+  isFilterBarShown: boolean;
+  applyInputFiltersOnInit: boolean;
+  conditionalFormats: SidebarFormattingRule[];
+  customColumnFormatMap: Map<string, FormattingRule>;
+  movedColumns: { from: number; to: number }[];
+  movedRows: { from: number; to: number }[];
```
mofojed, line 252:
```suggestion
  movedColumns: MoveOperation[];
  movedRows: MoveOperation[];
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -130,14 +192,317 @@ function isEmptyConfig({
...
+  inputFilters: InputFilter[];
+  customFilters: unknown[];
+  model: IrisGridProxyModel;
+  onCreateChart: (
+    settings: ChartBuilderSettings,
+    model: IrisGridProxyModel
+  ) => void;
+  onColumnSelected: (column: Column) => void;
+  onError: (error: unknown) => void;
+  onDataSelected: (index: number, map: unknown) => void;
```
mofojed, line 262: We should introduce a `type ColumnName = string` so that we can be more descriptive in places where we expect a `ColumnName`.
```suggestion
  onDataSelected: (index: ModelIndex, map: Record<ColumnName, unknown>) => void;
```

mofojed (changes requested):
WIP, only halfway through IrisGrid.tsx

mattrunyon (changes requested):
Made it through AdvancedFilterCreatorSelectValueList

I started this before Bender's most recent review, so some items may overlap or be the same comment on slightly different lines

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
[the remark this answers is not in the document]
mattrunyon: I opened #610 for adding this linter rule

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -0,0 +1,610 @@
...
+    this.model.isRowMovable(...args);
+
+  isColumnMovable: IrisGridTableModel['isColumnMovable'] = (...args) =>
+    this.model.isColumnMovable(...args);
+
+  isColumnFrozen(x: number): boolean {
+    if (TableUtils.isTreeTable(this.table)) {
+      throw new Error('TreeTable cannot be frozen');
+    }
+    return (this.model as IrisGridTableModel).isColumnFrozen(x);
```
mofojed, line 261: `isColumnFrozen` should be defined on `IrisGridModel`, and then `IrisGridTreeTable` model can just return `false`.

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -0,0 +1,610 @@
...
+    return false;
+  }
+
+  isRowExpandable: IrisGridTreeTableModel['isRowExpandable'] = (...args) => {
+    if (TableUtils.isTreeTable(this.model.table)) {
+      return (this.model as IrisGridTreeTableModel).isRowExpandable(...args);
+    }
+    throw Error(
+      'Function isRowExpandable does not exist on IrisGridTableModel'
+    );
```
mofojed, line 277: Just return `false` if it's not a tree table model instead of throwing.

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -0,0 +1,610 @@
...
+  setValues: IrisGridTableModel['setValues'] = (...args) =>
+    this.model.setValues(...args);
+
+  isValidForCell: IrisGridTableModel['isValidForCell'] = (...args) =>
+    this.model.isValidForCell(...args);
+
+  delete: IrisGridTableModel['delete'] = (...args) =>
+    this.model.delete(...args);
+
+  get pendingDataMap(): PendingDataMap<UIRow | UITreeRow> {
```
mofojed, line 582: Can just be `UIRow` here, `UITreeRow` extends `UIRow`

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -379,27 +440,27 @@ class IrisGridTableModel extends IrisGridModel {
...
+        formatOverride ?? undefined
       );
       this.cacheFormattedValue(x, y, text);
     }
 
     return this.formattedStringData[x][y];
   }
 
-  textForCell(x, y) {
+  textForCell(x: ModelIndex, y: ModelIndex): string | null {
```
mofojed, line 451: Shouldn't allow `null` to be returned here, add `return text ?? ''` at the bottom.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -379,27 +440,27 @@ class IrisGridTableModel extends IrisGridModel {
...
     const text = this.textValueForCell(x, y);
     if (text == null && this.isKeyColumn(x)) {
       const pendingRow = this.pendingRow(y);
-      if (this.pendingDataMap.has(pendingRow)) {
+      if (pendingRow && this.pendingDataMap.has(pendingRow)) {
```
mofojed, line 455: `pendingRow != null`
```suggestion
      if (pendingRow != null && this.pendingDataMap.has(pendingRow)) {
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -379,27 +440,27 @@ class IrisGridTableModel extends IrisGridModel {
...
         // Asterisk to show a value is required for a key column on a row that has some data entered
         return '*';
       }
     }
     return text;
   }
 
-  truncationCharForCell(x) {
+  truncationCharForCell(x: number): '#' | undefined {
```
mofojed, line 463: Could be a char other than `'#'`
```suggestion
  truncationCharForCell(x: number): string | undefined {
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -757,29 +724,32 @@ class IrisGridTableModel extends IrisGridModel {
     log.debug2('copyTotalsData', dataMap);
 
     this.totalsDataMap = dataMap;
+    //  Map<string, {data: Map<number, {value: unknown, format: Format}>}>
```
mofojed, line 727: Remove unused comment

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -757,29 +724,32 @@ class IrisGridTableModel extends IrisGridModel {
...
     return this.getColumnIndicesByNameMap(this.columns).get(name);
   }
 
-  getColumnIndicesByNameMap = memoize(columns => {
-    const indices = new Map();
-    columns.forEach(({ name }, i) => indices.set(name, i));
-    return indices;
-  });
+  getColumnIndicesByNameMap = memoize(
+    (columns: Column[]): Map<string, number> => {
```
mofojed, line 740:
```suggestion
    (columns: Column[]): Map<ColumnName, ModelIndex> => {
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -757,29 +724,32 @@ class IrisGridTableModel extends IrisGridModel {
...
+      const indices = new Map();
+      columns.forEach(({ name }, i) => indices.set(name, i));
+      return indices;
+    }
+  );
 
   /**
    * Copies all the viewport data into an object that we can reference later.
    * @param {ViewportData} data The data to copy from
```
mofojed, line 749: Remove params from comments

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1039,26 +974,31 @@ class IrisGridTableModel extends IrisGridModel {
...
+      throw new Error(`Invalid snapshot ranges ${ranges}`);
     }
 
     // Need to separate out the floating ranges as they're from a different source
-    const topFloatingRowsSet = new Set();
-    const tableRanges = [];
-    const bottomFloatingRowsSet = new Set();
+    const topFloatingRowsSet = new Set() as Set<number>;
+    const tableRanges = [] as GridRange[];
+    const bottomFloatingRowsSet = new Set() as Set<number>;
```
mofojed, line 994:
```suggestion
    const topFloatingRowsSet: Set<number> = new Set();
    const tableRanges: GridRange[] = [];
    const bottomFloatingRowsSet: Set<number> = new Set();
```

mofojed (changes requested):
Didn't get through all the files, but I think there's enough comments to address now, and some apply to all files. e.g.
- Declare variables with the type instead of explicitly casting to the type, e.g. prefer const styles: CSSProperties = { ... } instead of const styles = { ... } as CSSProperties
- A lot of places you're doing a truthy check when you should just be doing a null check, e.g. if (modelIndex)  instead of if (modelIndex != null). This will cause bugs in cases of 0
- Prefer to use custom types instead of primitive types. E.g. you can use VisibleIndex or ModelIndex instead of just number to be clear; should use a custom type ColumnName instead of just string to be clear
- undefined vs. null - prefer to use undefined in optional cases, or cases where it's just an uninitialized variable. Should only use null if it's an intentional setting of the absence of a value
- Remove parameter types from JS Docs (since they're in the function signature now).

SparkyFnay:
Breaking change: onContextMenu no long takes in a table argument

packages/iris-grid/src/sidebar/TableSaver.tsx
```diff
@@ -142,29 +142,29 @@ export default class TableSaver extends PureComponent<
     if (this.snapshotHandlerTimeout) clearTimeout(this.snapshotHandlerTimeout);
   }
 
-  sw: ServiceWorker | null;
+  sw: ServiceWorker | undefined;
```
mofojed, line 145: Prefer to define as optional instead of `| undefined`
```suggestion
  sw?: ServiceWorker;
```
Same with all the others below

mofojed (changes requested):

packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -102,6 +103,14 @@ class AdvancedFilterCreator extends PureComponent<
     return { key: shortid() };
   }
 
+  static assertFilterOperatorValue(
```
mofojed, line 106: In terms of code structure, this would make more sense in `FilterOperator.ts` here instead of as a static function on AdvancedFilterCreator.

packages/iris-grid/src/CommonTypes.tsx
```diff
@@ -0,0 +1,9 @@
+import { ModelIndex } from '@deephaven/grid';
+import AggregationOperation from './sidebar/aggregations/AggregationOperation';
+import { AdvancedFilter, QuickFilter } from './IrisGrid';
```
mofojed, line 3: `AdvancedFilter`, `QuickFilter`, `AdvancedFilterOptions`, `AggregationOperation`, basically all those types should come into here, as they are all used across many files within iris-grid.

Right now this is importing from IrisGrid, which imports from AdvancedFilterCreator, which imports from AdvancedFilterCreatorSelectValue, which imports from CommonTypes, got a cycle.

Pretty much all the types currently defined and exported in IrisGrid.tsx should just be in CommonTypes

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -117,6 +114,12 @@ class IrisGridTableModelTemplate<
 
   static COLUMN_BUFFER_PAGES = 0;
 
+  static isIrisGridTableModelTemplate(
```
mofojed, line 117: Instead of making it a static, export it as it's own function defined outside of the class.

packages/iris-grid/src/asserts.tsx
```diff
@@ -0,0 +1,15 @@
...
+  value: T | undefined
+): asserts value is T {
+  if (value === undefined) throw new Error('Value is undefined');
+}
+export function assertNotNullNorUndefined<T>(
+  value: T | null | undefined
+): asserts value is T {
+  assertNotNull(value);
+  assertNotUndefined(value);
+}
```
mofojed, line 15: I don't think you need this. `assertNotNull` is already verifying the value is not undefined (since it's doing `value == null` instead of `value === null`).

mofojed (changes requested):

packages/iris-grid/src/asserts.tsx
SparkyFnay: Yes, assertNotNull is verifying that value is not undefined, but TypeScript can't recognize that, so I need an explicit assertNotUndefined

packages/iris-grid/src/asserts.tsx
```diff
@@ -1,15 +1,5 @@
...
-  if (value == null) throw new Error('Value is null');
-}
-
-export function assertNotUndefined<T>(
-  value: T | undefined
-): asserts value is T {
-  if (value === undefined) throw new Error('Value is undefined');
-}
-export function assertNotNullNorUndefined<T>(
+export default function assertNotNull<T>(
```
mofojed, line 1: Should either name the file `assertNotNull.ts` or change the export to `export function assertNotNull` (may need to suppress a warning about preferring default with one export).

It doesn't make sense for the file to be called `asserts` and only have one function exported as default.

Also - it makes sense to move this to `@deephaven/utils` package. I think it will come in handy outside of iris-grid.

packages/iris-grid/src/CommonTypes.tsx
```diff
@@ -1,9 +1,64 @@
...
+  TotalsTableConfig,
+  FilterCondition,
+  Format,
+} from '@deephaven/jsapi-shim';
+import { Shortcut } from '@deephaven/components';
+import { IconDefinition } from '@deephaven/icons';
 import AggregationOperation from './sidebar/aggregations/AggregationOperation';
-import { AdvancedFilter, QuickFilter } from './IrisGrid';
+import { AdvancedFilterOptions } from './AdvancedFilterCreator';
+import { OptionType } from './sidebar';
```
mofojed, line 11: As mentioned in my previous comment, I think `AdvancedFilterOptions` and `AggregationOperation` should be in here as well. Should bring `OptionType` in too, so this file doesn't have any dependencies on other files within `iris-grid`: https://github.com/deephaven/web-client-ui/pull/605#discussion_r902761491

packages/utils/src/Asserts.tsx
```diff
@@ -1,4 +1,5 @@
-export default function assertNotNull<T>(
+// eslint-disable-next-line import/prefer-default-export
+export function assertNotNull<T>(
```
mofojed, line 2: This can just be a `.ts` file instead of a `.tsx` file, as it does not contain any React/JSX

mofojed (approved):

mofojed (approved):

PR #611, opened 2022-06-09
Create custom Modal component to remove dependency on reactstrap

### LINKED ISSUE #338: Remove dependency on reactstrap
We only use reactstrap for modals. Remove the dependency and replace with our own Modal components

SparkyFnay:
Fixes #338


packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const getCentered = (): string => {
+    if (centered) {
+      return 'modal-dialog-centered';
+    }
+    return '';
+  };
+
+  return isOpen ? (
+    ReactDOM.createPortal(
+      <div className={`modal  ${getCentered()}`} onClick={toggle} role="dialog">
```
mofojed, line 90: Use `classNames` like we do elsewhere for this, get rid of the `getCentered` method:
```suggestion
      <div className={classNames('modal', { 'modal-dialog-centered': centered })} onClick={toggle} role="dialog">
```

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <div className={`modal-dialog ${className}`} ref={outerDivRef}>
```
mofojed, line 91:
```suggestion
        <div className={classNames('modal-dialog', className)} ref={outerDivRef}>
```

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,109 @@
...
+            className="modal-content"
+            onClick={e => e.stopPropagation()}
+            data-testid={dataTestId}
+            role="dialog"
+          >
+            {children}
+          </div>
+        </div>
+      </div>,
+      document.getElementsByTagName('BODY')[0]
```
mofojed, line 102: I think can just use `document.body` here...

mofojed (changes requested):
I'll let Don review style

mattrunyon:
It looks like this will create a new portal for each modal and not clean up the portal after it closes. We should either re-use 1 portal created at the root or make sure to clean up after close animation

dsmmcken:
Unable to test due to build failure.

packages/components/src/modal/Modal.scss
```diff
@@ -0,0 +1,100 @@
...
+@keyframes backdropFadeIn {
+  0% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
+
+.modal {
```
dsmmcken, line 21: you probably don't need any or hardly any of this css, it's already part of bootstrap

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      }}
+      timeout={ThemeExport.transitionMs}
+    >
+      <div
+        className="modal fade"
+        onClick={toggle}
+        role="dialog"
+        style={{ zIndex: 1050, display: 'block', paddingRight: '15px' }}
+      >
+        <div className={classNames('modal-backdrop fade show')} />
```
mofojed, line 95: You noticed that your transition is a little bit faster than the native bootstrap transition. If you look at the bootstrap animation at 0.1x speed, you can see that it fades in the backdrop entirely first, then when that transition is complete, the inner modal starts fading in:
https://user-images.githubusercontent.com/4505624/174670002-b1be43a2-fd50-4d8a-a5cf-f3ccf4d788f3.mp4

It looks like reactstrap starts fading/showing all at the same time. I don't know if @dsmmcken has a particular opinion about if the backdrop should fade in first, but if it should, then need to change it up a bit... Put `modal-backdrop` as a sibling of `modal`, wrap both backdrop/modal in CSSTransition, and only put the `onEnterDone` on the wrap around modal (so it only triggers the animation at the end of the transition) - so the overall structure would be something like:
```
React.createPortal(
<>
  <CSSTransition classNames={{ onEnterActive: 'show', onEnterDone: 'show' }}>
    <div className="modal-backdrop fade"></div>
  </CSSTransition>
  <CSSTransition classNames={{ onEnterDone: 'show' }} timeout={ThemeExport.transitionMs}>
    <div className="modal fade">
    </div>
  </CSSTransition>
</>
)
```

mofojed (changes requested):

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+  useEffect(
+    function open() {
+      if (!isOpen && onClosed) {
+        onClosed();
+      }
+    },
+    [onClosed, isOpen]
+  );
+
+  return ReactDOM.createPortal(
```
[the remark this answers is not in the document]
mattrunyon: If using react transition group, it might be better to use the transition lifecycle methods like onExited instead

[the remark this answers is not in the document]
mofojed: Yes, use onExited to ’setIsOpenInternal(false)’, but still need the state variable

packages/components/src/modal/Modal.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+it('does not render when isOpen is false', () => {
+  render(makeModal({ isOpen: false }));
+  expect(screen.queryByRole('dialog')).toBeNull();
+});
+
+it('renders when isOpen is true', async () => {
+  render(makeModal({ isOpen: true }));
+  const dialogs = screen.getAllByRole('dialog');
+  expect(dialogs.length).not.toBe(0);
```
mofojed, line 52: There should only be one `dialog`, clean up Modal.tsx so you can update all these tests to just be `expect(screen.queryByRole('dialog')).toBeNull()` or `expect(screen.queryByRole('dialog')).not.toBeNull()`

packages/components/src/modal/Modal.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+it('renders when isOpen changes', async () => {
+  const { rerender } = render(makeModal({ isOpen: false }));
+
+  expect(screen.queryByRole('dialog')).toBeNull();
+  rerender(makeModal({ isOpen: true }));
+
+  expect(screen.getAllByRole('dialog').length).not.toBe(0);
+  rerender(makeModal({ isOpen: false }));
+
+  await waitFor(() => expect(screen.queryByRole('dialog')).toBeNull());
```
mofojed, line 64: Put a comment here about how it's waiting for the animation to complete.

Or - another way you could write this is by using fake timers. `waitFor` works but it polls every 50ms, which if we had a lot of these kinds of tests it could cause them to run for a long time. We don't want to _actually_ wait the 150ms it takes for the animation to complete to know if the tests pass or not.

Instead, use fake timers, and then run all pending timers prior to checking. That will ensure the tests run as quickly as possible:

```
it('renders when isOpen changes', () => {
  jest.useFakeTimers();

  const { rerender } = render(makeModal({ isOpen: false }));

  expect(screen.queryByRole('dialog')).toBeNull();
  rerender(makeModal({ isOpen: true }));

  expect(screen.getAllByRole('dialog').length).not.toBe(0);
  rerender(makeModal({ isOpen: false }));

  // It will fade out, so it will still be visible until the animation is complete
  expect(screen.queryAllByRole('dialog').length).not.toBe(0);
  jest.runAllTimers();
  expect(screen.queryAllByRole('dialog').length).toBe(0);

  jest.useRealTimers();
});
```

packages/components/src/modal/Modal.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+  userEvent.click(screen.getAllByRole('dialog')[1]);
+  expect(toggle).toBeCalledTimes(0);
+});
+
+it('calls onOpen when opens', () => {
+  const onOpened = jest.fn();
+  const toggle = jest.fn();
+  const { rerender } = render(makeModal({ isOpen: false, toggle, onOpened }));
+  expect(onOpened).not.toBeCalled();
+  rerender(<Modal isOpen onOpened={onOpened} />);
```
mofojed, line 105: Why not call `makeModal` here?

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const element = useRef<HTMLElement>();
+
+  const handleKeyDown = useCallback(
+    (event: KeyboardEvent): void => {
+      switch (event.key) {
+        case 'Escape':
+          if (toggle) {
+            toggle();
+          }
```
mofojed, line 51: Can write this on one line with optionals:
```suggestion
          toggle?.();
```

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+        );
+        element.current.setAttribute('role', 'modal-container');
+        document.body.appendChild(element.current);
+        setIsPortalOpen(true);
+      }
+    },
+    [isOpen]
+  );
+
+  const onExited = () => {
```
mofojed, line 116: Any reason not to wrap it in a `useCallback` for a stable reference?

Also I don't think you need `isPortalOpen` at all - simply remove the element if it exists in `onExited`

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  useEffect(
+    function open() {
+      if (isOpen && !element.current) {
+        element.current = document.createElement('div');
+        element.current.setAttribute(
+          'style',
+          'z-index: 1050; padding-right: 15px; display: block'
+        );
+        element.current.setAttribute('role', 'modal-container');
```
mofojed, line 108: This isn't a valid role - I don't think you need one? If you do, I would pick [presentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role).

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+        >
+          <div className={classNames('modal-backdrop fade')} />
+        </CSSTransition>
+        <CSSTransition
+          appear
+          mountOnEnter
+          unmountOnExit
+          in={isOpen}
+          classNames={{
+            enterActive: 'show',
```
mofojed, line 143: I think for content, you only want to specify `enterDone` - that way the content fades in after the background fade transition.

mofojed (changes requested):

package.json
```diff
@@ -62,6 +62,7 @@
     "@svgr/cli": "^6.2.1",
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^12.1.3",
+    "@testing-library/user-event": "^12.8.3",
```
mofojed, line 65: You didn't change this, but... right now our `@types/react-transition-group` version is `v.4.4.0`, whereas the actually version of `react-transition-group` we install is `v2.9.0`. We should update them both to the latest `v4.4.2` (there's no breaking changes that affect us). May as well do that as part of this change, so find/replace all `react-transition-group` version specs in all package.jsons you can find with the latest version then run `npm install` again to update package lock.

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+            enterDone: 'show',
+          }}
+          timeout={ThemeExport.transitionLongMs}
```
mofojed, line 129: This should still just be `ThemeExport.transitionMs`

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+          />
+        </CSSTransition>
+        <CSSTransition
+          appear
+          mountOnEnter
+          unmountOnExit
+          in={isOpen}
+          classNames={{
+            enterActive: 'show',
+            enterDone: 'show',
```
mofojed, line 127: Playing with it a bit more, and noticed a couple things:
- `enterActive` class doesn't seem to be getting applied at all (not sure why??)
- `enterDone` class gets applied after the `timeout` transition has completed
- `appearActive` appears to apply right away, then `appearDone` gets applied after the transition

Based on that information, you should change the dialog `classNames` to only specify `appearDone` (so the `show` class only gets applied when the transition is completed; which will _actually_ kick off the CSS transition), and change the `enterActive`/`enterDone` classNames on the background to be `appearActive` and `appearDone` instead (right now it's just delaying the transition by 150ms and then applying the backdrop. You can try with a really long timeout (like 3000) to see what I mean).
```suggestion
            appearDone: 'show',
```

packages/components/src/modal/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    ReactDOM.createPortal(
+      <>
+        <CSSTransition
+          appear
+          mountOnEnter
+          unmountOnExit
+          in={isOpen}
+          classNames={{
+            enterActive: 'show',
+            enterDone: 'show',
```
mofojed, line 110:
```suggestion
            appearActive: 'show',
            appearDone: 'show',
```

mofojed (changes requested):

packages/components/src/modal/Modal.tsx
SparkyFnay: I need transitionLongMs for the animation delay, simply removing enterActive from the second transition doesn't do anything

SparkyFnay: Ah nvm, after changing to appear active, I don't need the longer timeout anymore

mofojed (approved):

dsmmcken (approved):
Animation looks good :shipit:

PR #613, opened 2022-06-13
Add support for moving ranges within grids

mattrunyon:
This adds support for moving an entire range within a grid

The operation is to slice the range from the current grid, re-index the remaining columns as if the range was not there, and then insert with the first element in the range at the position specified

### DIFF
--- a/packages/grid/src/GridUtils.ts
+++ b/packages/grid/src/GridUtils.ts
@@ -741,6 +741,11 @@ export class GridUtils {
 
   /**
    * Move a visible range in the grid
+   *
+   * This will effectively slice the range out of the grid,
+   * re-index the remaining columns,
+   * then insert the range with the first element at the provided index
+   *
    * @param from The visible axis range to move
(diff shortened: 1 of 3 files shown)


packages/grid/src/GridUtils.ts
```diff
@@ -739,6 +739,46 @@ export class GridUtils {
...
+  ): MoveOperation[] {
+    if (from[0] === to) {
+      return oldMovedItems;
+    }
+
+    const movedItems: MoveOperation[] = [...oldMovedItems];
+    const lastMovedItem = movedItems[movedItems.length - 1];
+    if (
+      lastMovedItem &&
+      Array.isArray(lastMovedItem.from) &&
```
mofojed, line 767: Might be worth having an `isAxisRange` predicate (not just an array but an array of length 2)

mofojed (approved):

mofojed (approved):

PR #615, opened 2022-06-15
Scrollbar Selection Ticks

### LINKED ISSUE #175: Scrollbar selection ticks
Inspired by monaco, we should add ticks in the scrollbar marking the users active selection. 

This will make it easier for users to be able to return to their scroll position in a large table, and make the user aware of any offscreen selections. If the grid is in row selection mode, only show ticks on the vertical scrollbar. Minimum tick size is 1px high, but otherwise they should be spaced in the scrollbar proportionally to the area the selection represents. Show an extra “cursor” representing where the ‘active’ keyboard row is, it should be shown as a 2px bar, with extra alpha.

mockup:
image

SparkyFnay:
Fixes #175


packages/grid/src/GridRenderer.ts
```diff
@@ -2539,6 +2696,80 @@ export class GridRenderer {
         vScrollBarSize - scrollBarCasingWidth,
         handleHeight
       );
+
+      if (!autoSelectColumn && scrollBarSelectionTickColor != null) {
```
mofojed, line 2700: The algorithm for rows should be the same as columns

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+            const rightTickSize =
+              (barWidth - (x + handleWidth)) /
+              (columnCount - rightVisibleIndex + remainderRight);
+
+            return Math.round(
+              barWidth - (columnCount - index + 1) * rightTickSize
+            );
+          }
+          return Math.round(
+            (getOrThrow(visibleColumnXs, index) / width) * handleWidth + x
```
mofojed, line 2591: `barWidth` instead of `width`.

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+        // Scrollbar Selection Tick
+        const { selectedRanges, model, cursorColumn } = state;
+        const { visibleColumns, visibleColumnWidths } = metrics;
+        const { columnCount } = model;
+
+        const leftVisibleIndex = Math.min(...visibleColumns.values());
+        const rightVisibleIndex = Math.max(...visibleColumns.values());
+
+        const getTickX = (index: number) => {
+          if (index < leftVisibleIndex) {
```
mofojed, line 2572: `<=`

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+            return Math.round((index / columnCount) * barWidth);
+          }
+          if (index > rightVisibleIndex) {
```
mofojed, line 2575: `>=`

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+            mergedRanges[mergedRanges.length - 1].endColumn
+          ) {
+            mergedRanges[mergedRanges.length - 1].endColumn = Math.max(
+              mergedRanges[mergedRanges.length - 1].endColumn,
+              endColumn
+            );
+          } else {
+            mergedRanges.push({ startColumn, endColumn });
+          }
+        }
```
mofojed, line 2619: Use `AxisRange`, and put something like this in `GridUtils` (similar to `consolidate` method in `GridRange`)

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+              y + scrollBarCasingWidth + Math.round(trackHeight / 3),
+              tickWidth,
+              Math.round(trackHeight / 3)
+            );
+          }
+        }
+
+        // Current Active Tick
+        if (cursorColumn != null) {
+          // const tickX = Math.round((cursorColumn / columnCount) * barWidth);
```
mofojed, line 2650: Unused comment

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+          const tickX = getTickX(cursorColumn);
+          const tickWidth = 2;
```
mofojed, line 2652: You shouldn't be hardcoding the tick width, you should be drawing for the "width" of the column, e.g.:

```
const startTickX = getTickX(cursorColumn);
const endTickX = getTickX(cursorColumn+1);
```

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+          ) {
+            const tickX = Math.round(
+              (range.startColumn / columnCount) * barWidth
+            );
+            const tickWidth = Math.max(
+              1,
+              Math.round(
+                ((range.endColumn + 1) / columnCount) * barWidth - tickX
+              )
+            );
```
mofojed, line 2637: Should be calculating the tickX the same as the active selection, e.g.:
```suggestion
            const startTickX = getTickX(range.startColumn);
            const endTickX = getTickX(range.endColumn + 1);
```

packages/grid/src/GridRenderer.ts
```diff
@@ -2503,6 +2557,109 @@ export class GridRenderer {
...
+        context.fillStyle = scrollBarSelectionTickColor;
+        // Scrollbar Selection Tick
+        const { selectedRanges, model, cursorColumn } = state;
+        const { visibleColumns, visibleColumnWidths } = metrics;
+        const { columnCount } = model;
+
+        const leftVisibleIndex = Math.min(...visibleColumns.values());
+        const rightVisibleIndex = Math.max(...visibleColumns.values());
+
+        const getTickX = (index: number) => {
```
mofojed, line 2571: Something still seems off with this... looking at Tree Grid in the style guide for testing, and if all columns are the same width, then we shouldn't have the ticks moving around at all - they should be correctly and consistently positioned no matter where you scroll. It's only in cases where there's variable column sizes where it should move around a bit.

It may also be that the scroll bar thumb width isn't the "correct" size - the thumb within the scroll bar should match the viewport compared to the whole grid view.

mofojed (changes requested):

packages/grid/src/GridRenderer.ts
SparkyFnay: Don wanted active tick to be 2px always

packages/grid/src/GridUtils.ts
```diff
@@ -18,6 +18,7 @@ import { GridWheelEvent } from './GridMouseHandler';
 type Range<T> = [start: T, end: T];
 
 export type AxisRange = Range<GridRangeIndex>;
+export type NoneNullAxisRange = Range<number>;
```
mofojed, line 21: Just use `BoundedAxisRange` instead of `NoneNullAxisRange`.

mofojed (changes requested):

dsmmcken (changes requested):
1. Hmm, select all on a table looks odd with the gap at the bottom:

image

image

2. I think this should have a boolean setting in grid theme for enabling/disabling this feature entirely.

mofojed (approved):

dsmmcken (approved):

PR #621, opened 2022-06-24
update fontawesome packages

### LINKED ISSUE #582: Update to FontAwesome 6.1.1
We should update our FontAwesome deps to 6.1.1 for both [@fortawesome/fontawesome-common-types](https://www.npmjs.com/package/@fortawesome/fontawesome-common-types) and [@fortawesome/fontawesome-svg-core](https://www.npmjs.com/package/@fortawesome/fontawesome-svg-core).
They previously published v0.3.0 and v1.3.0 which broke some dependencies (we had to submit fix #435  for it) before switching to v6.x.x versioning. Those builds are now deprecated with the note to update to 6.x: https://www.npmjs.com/package/@fortawesome/fontawesome-svg-core/v/1.3.0

dsmmcken:
Fixes #582

Don't see any breaking changes that impact us. Also visually checked and didn't notice anything out of order.


mofojed (changes requested):
Need to update snapshots and add label

mofojed (approved):

PR #627, opened 2022-06-28
Fix column tooltip error when side menu is opened

### LINKED ISSUE #626: Column tooltip error when opening side menu
**Description**

When opening the table side menu if a column tooltip is active, iris grid throws an error at an assertion for IrisGrid.tsx:3396 which is for `shownColumnTooltipX`

**Steps to reproduce**
1. Open styleguide
2. On the iris-grid example, hover over the rightmost column
3. Move cursor slightly over to the menu button and click

**Expected results**

Menu opens

**Actual results**

Grid goes to an error state

The issue occurs because the tooltip position is not in metrics since it is fully obscured when the menu opens. The iris-grid TS conversion added several assertions which throw errors we may have just been eating before

https://user-images.githubusercontent.com/8619789/176206814-4698bd4a-59d1-4bf3-bcd6-bc69bcd177a5.mov

**Versions**
 - Deephaven: 0.14

mattrunyon:
Fixes #626 

We could modify the hover handler to know where the button is, but this fix seems easier/more robust to future issues

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -2531,7 +2531,8 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.updateFormatter({ customColumnFormatMap });
   }
 
-  handleMenu(): void {
+  handleMenu(e: React.MouseEvent<HTMLButtonElement>): void {
+    e.stopPropagation();
     this.setState({ isMenuShown: true });
   }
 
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3170,6 +3170,106 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
   );
 
+  getColumnTooltip = memoize(
```
dsmmcken, line 3173: Does model/metrics return new objects making this memorize correctly?

mofojed (dismissed):

dsmmcken:
Nice, this originally had an onLeave and I removed it to handle the case when the popper is above the grid, this is better.

PR #634, opened 2022-06-29
DH-12796: fixed redux workspace types redeclared in code-studio error

### LINKED ISSUE #592: Workspace Storage redeclares types already defined in Redux
**Description**

Types in Workspace Storage don't match the types in Redux.
Workspace should import from redux workspace instead of code-studio workspace.

Code-studio Workspace: https://github.dev/deephaven/web-client-ui/blob/b82791e64062d6563c924b448e788878222df2ef/packages/code-studio/src/storage/WorkspaceStorage.ts#L26

Redux Workspace: https://github.dev/deephaven/web-client-ui/blob/67c02cb4ea6cb8c66d81bcc62a4185ba9f85edd9/packages/redux/src/store.ts#L56

**Steps to reproduce**
1. Go to PR #588 
2. Click on (https://github.com/deephaven/web-client-ui/pull/588/files#diff-930839a7a608d6655fc1141ff3c85d5d5997feb9edcce649011ffd877311e8b6)
4. Scroll down to `setWorkspace(LocalWorkspaceStorage.makeDefaultWorkspace() as Workspace);`

**Expected results**

No need to cast

**Actual results**

Need to cast from Code Studio Workspace to Redux Workspace

SparkyFnay:
Fixes #592 
DH-12796 for fixing breaking change


packages/redux/src/store.ts
```diff
@@ -37,9 +37,9 @@ export interface WorkspaceSettings {
   timeZone: string;
   showTimeZone: boolean;
   showTSeparator: boolean;
-  truncateNumbersWithPound: boolean;
+  truncateNumbersWithPound?: boolean;
   disableMoveConfirmation: boolean;
-  showSystemBadge: boolean;
+  showSystemBadge?: boolean;
```
mattrunyon, line 42: Why are these optional now? I think they should always exist in the workspace settings

mattrunyon (changes requested):

packages/redux/src/store.ts
SparkyFnay: `truncateNumbersWithPound` and `showSystemBadge` were missing from code-studio `WorkspaceSetting` under `LocalWorkspaceStorage.ts`. I ended up remove the optional from truncateNumbersWithPound and set it to false in `LocalWorkspaceStorage`, but I don't think `showSystemBadge` is used anywhere, so I made it optional

mofojed: Yes it is only used in Enterprise... so the proper way to do this would be to remove `showSystemBadge` from here, then we need to extend these types in Enterprise.

mattrunyon (approved):

PR #638, opened 2022-06-30
Added Goto row feature in Iris Grid

### LINKED ISSUE #616: Go to row
As user it is sometimes helpful to be able to navigate to a specific row number for viewing data at a specific location. This feature adds a go to row behavior to iris grid.

image

SparkyFnay:
Fixes #616


packages/components/src/shortcuts/GlobalShortcuts.ts
```diff
@@ -43,6 +43,13 @@ const GLOBAL_SHORTCUTS = {
     macShortcut: [KEY.ESCAPE],
     isEditable: false,
   }),
+  GOTO: ShortcutRegistry.createAndAdd({
```
mofojed, line 46: This should be in `IrisGridShortcuts`, not `GlobalShortcuts`

packages/grid/src/Grid.tsx
```diff
@@ -549,6 +560,49 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+    const {
+      rowHeight,
+      bottom,
+      visibleRowYs,
+      visibleRowHeights,
+      columnHeaderHeight,
+    } = this.metrics;
+    const arr: [number, number][] = [];
+    userRowHeights.forEach((x, number) => arr.push([x, number]));
+    if (updateTop) {
```
mofojed, line 573: Instead of putting this in `componentDidUpdate`, this logic should either be in the `moveSelection` method (with some sort of `viewportPosition` parameter added) or a new method. Then IrisGrid calls `grid.moveSelection`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -933,6 +952,11 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     { max: 100 }
   );
 
+  getGotoRowOpenedHandler = memoize(
+    cellInfo => this.handleGotoRowOpened.bind(this, cellInfo),
+    { max: 100 }
+  );
+
```
mofojed, line 959: Doesn't seem to be used

```suggestion
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3171,6 +3218,12 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
   );
 
+  onGotoRowSelectedRowNumberChanged(rowValue: string): void {
```
mofojed, line 3221: I'd name this method `handleGotoRowSelectedRowNumberChanged` instead of `on*`

mofojed (changes requested):

packages/grid/src/Grid.tsx
```diff
@@ -53,6 +53,7 @@ import {
 import { EventHandlerResultOptions } from './EventHandlerResult';
 import { assertIsDefined } from './errors';
 import ThemeContext from './ThemeContext';
+import { getOrThrow } from '.';
```
mofojed, line 56: Should import from the file directly instead of from `'.'`, this causes a circular import

packages/grid/src/Grid.tsx
```diff
@@ -551,6 +552,58 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+      visibleRowHeights,
+      columnHeaderHeight,
+      userRowHeights,
+    } = this.metrics;
+
+    const arr: [number, number][] = [];
+    userRowHeights.forEach((x, number) => arr.push([x, number]));
+    let currentTop = top;
+    const tableHeight =
+      (this.canvas?.clientHeight ?? columnHeaderHeight) - columnHeaderHeight;
```
mofojed, line 584: Just use `height` from metrics instead of `this.canvas?.clientHeight`. Also should use `gridY` instead of `columnHeaderHeight`

packages/grid/src/Grid.tsx
```diff
@@ -551,6 +552,58 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+
+    const halfViewportHeight =
+      Math.round(tableHeight / 2) - getOrThrow(visibleRowHeights, top);
+    const bottomToFill = Math.max(
+      0,
+      tableHeight -
+        rowHeight * 2 -
```
mofojed, line 591: So - `Grid` has adjustable row heights, whereas `IrisGrid` does not. Since this is in `Grid`, need to take the row heights into consideration.

Right now, you're setting the top to where you want the focus to be, and then you're moving on the next tick after the metric calculator has updated. Instead, what you should do is just use `metricCalculator.getLastTop` to calculate it, something like:
```
const top = metricCalculator.getLastTop(state, focusRow, halfTableHeight)
```

I think that should get you where you want.

mofojed (changes requested):

mofojed (approved):

PR #645, opened 2022-07-11
Convert Log package to typescript

### LINKED ISSUE #644: Convert @deephaven/log to TypeScript

SparkyFnay:
Fixes #644


packages/log/src/Logger.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  debug: (...data: unknown[]) => void;
+
+  debug2: (...data: unknown[]) => void;
+
+  setLogLevel(level?: number): void {
+    this.level = level;
+
+    this.error =
+      level != null && ERROR <= level
```
mattrunyon, line 43: Minor comment that I think helps the readability of the code and is a good thing to think about when doing comparisons (I'm aware you were just copying this logic here)

When comparing to a constant, put the variable first in the comparison. It's minor, but helps the readability. I think most people would say "if the log level is greater or equal to the error level" which corresponds to `level >= ERROR` and not "if the error level is less or equal to the current level" like written

Functionally the same, but makes the code just ever so slightly easier to read. For these cases it's not complicated really, but if you have 4 or 6 boolean checks in a condition, it can make the conditional easier to understand

packages/log/src/DefaultLogger.ts
```diff
@@ -45,15 +45,17 @@ class DefaultLogger extends Logger {
...
+  module(name: string): Logger {
     if (!this.modules[name]) {
       this.modules[name] = new Logger(name, this.level);
     }
 
     return this.modules[name];
   }
 
-  setLogLevel(level) {
+  setLogLevel(level: number | undefined): void {
```
mofojed, line 58: I think this should always be a `number` - why allow `undefined` here?

mofojed (changes requested):

packages/log/src/DefaultLogger.ts
SparkyFnay: The constructor can be called with an undefined level, and setLogLevel is called there. But now that I'm "manually" defaulting the log functions to silent, I don't think I have to call setLogLevel anymore if level is undefined?

packages/log/src/Logger.ts
```diff
@@ -15,7 +15,9 @@ class Logger {
     this.debug = silent;
     this.debug2 = silent;
 
-    this.setLogLevel(level);
+    if (level !== undefined) {
```
mofojed, line 18: I don't think you should allow `level` to be `undefined` at all, just allow a `number` in the constructor.

Really the `name` should be optional, and should be the second parameter, then level is optional and defaults to the defaultLogLevel()

mofojed (changes requested):

mofojed (approved):

PR #646, opened 2022-07-11
Convert Console to typescript

### LINKED ISSUE #641: Convert @deephaven/console package to TypeScript

SparkyFnay:
Fixes #641


packages/console/src/common/ConsoleUtils.test.ts
```diff
@@ -40,3 +40,8 @@ describe('parsing shell arguments from text', () => {
     testStr('foo # bar', ['foo']);
   });
 });
+
+describe('predicates', () => {
```
mofojed, line 44: May as well add tests for `hasPattern` and `hasOp` as well

packages/console/src/common/ConsoleUtils.ts
```diff
@@ -1,25 +1,45 @@
...
+    return ShellQuote.parse(str as string)
+      .filter(arg => !this.hasComment(arg))
+      .map(arg => {
+        let ret = `${arg}`;
+        if (this.hasPattern(arg)) {
+          ret = arg.pattern;
+        } else if (this.hasOp(arg)) {
+          ret = arg.op;
+        }
+        return ret;
```
mofojed, line 38: I'd rather write it:
```suggestion
        if (this.hasPattern(arg)) {
          return arg.pattern;
        } else if (this.hasOp(arg)) {
          return arg.op;
        }
        return `${arg}`;
```

packages/console/src/common/ConsoleUtils.ts
```diff
@@ -1,25 +1,45 @@
...
+  }
+
   /**
    * Given the provided text, parse out arguments using shell quoting rules.
    * @param str The text to parse.
    * @return string[] of the arguments. Empty if no arguments found.
    */
-  static parseArguments(str) {
+  static parseArguments(str: unknown): string[] {
     if (!str || !(typeof str === 'string' || str instanceof String)) {
```
mofojed, line 23: Not because of your change, but this seems weird to me - we return here if `str` is null or if `str` is not a string; but then below we do the filtering and check `hasComment`, `hasPattern`, etc. How is that possible?

packages/console/src/common/ConsoleUtils.ts
```diff
@@ -32,38 +52,39 @@ class ConsoleUtils {
...
   }
 
-  static defaultHost() {
+  static defaultHost(): string {
     let defaultHost = null;
-    try {
-      const url = new URL(process.env.REACT_APP_CORE_API_URL);
+    const apiUrl = process.env.REACT_APP_CORE_API_URL;
+    if (apiUrl != null) {
+      const url = new URL(apiUrl);
```
mofojed, line 59: This is a change in behaviour - I think we still want to wrap a try/catch around the URL creation, since it can throw, eg. `new URL('__mocks__')` (which it will get with unit tests).

I'd write it something like:

```
let defaultHost = window.location.hostname;
if (apiUrl != null) {
  try {
      const url = new URL(apiUrl);
      apiUrl = url.hostname;
  } catch (error) { }

```

packages/console/src/common/ConsoleUtils.ts
```diff
@@ -32,38 +52,39 @@ class ConsoleUtils {
...
       defaultHost = url.hostname;
-    } catch (error) {
+    } else {
       defaultHost = window.location.hostname;
     }
     return defaultHost;
   }
 
-  static isTableType(type) {
+  static isTableType(type: unknown): boolean {
```
mofojed, line 67: I think you know `type` is a `string`?

packages/console/src/csv/CsvOverlay.tsx
```diff
@@ -1,19 +1,45 @@
...
+  onPaste: (clipText: string) => void;
+  clearDragError: () => void;
+  dragError: string | null;
+  onError: (e: unknown) => void;
+  uploadInProgress: boolean;
+}
+
+interface CsvOverlayState {
+  selectedFileName: string;
+  dropError: string | null;
```
mofojed, line 41: Optional
```suggestion
  dropError?: string;
```

packages/console/src/csv/CsvOverlay.tsx
```diff
@@ -1,19 +1,45 @@
...
   vsTrash,
   vsWarning,
 } from '@deephaven/icons';
 import './CsvOverlay.scss';
-import { TextUtils } from '@deephaven/utils';
+import { assertNotNull, TextUtils } from '@deephaven/utils';
+
+interface CsvOverlayProps {
+  allowZip: boolean;
+  onFileOpened: (file: File | null) => void;
```
mofojed, line 31: As mentioned in another comment, add a new prop `onCancel` that is used instead of `onFileOpened(null)`

packages/console/src/log/LogView.tsx
```diff
@@ -96,15 +103,29 @@ class LogView extends PureComponent {
...
 
     window.removeEventListener('resize', this.handleResize);
   }
 
-  getMenuActions(shownLogLevels) {
+  cancelListener?: () => void | null;
+
+  editor?: monaco.editor.IStandaloneCodeEditor;
+
+  editorContainer: HTMLDivElement | null;
```
mofojed, line 118:
```suggestion
  editorContainer?: HTMLDivElement;
```

packages/console/src/command-history/CommandHistoryTypes.tsx
```diff
@@ -0,0 +1,21 @@
+import { Shortcut } from '@deephaven/components';
+import { IconDefinition } from '@deephaven/icons';
+
+export type ItemAction = {
```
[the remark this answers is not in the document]
SparkyFnay: A issue I ran into is that all of the properties in `ContextAction` are optional, but that's not the case with `ItemAction` here. So code wise, the only thing I can simplify is `shortcut?: Shortcut`.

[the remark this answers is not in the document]
SparkyFnay: So adding the extension would only be for communicating that it's also a `ContextAction`

packages/console/src/common/Code.tsx
```diff
@@ -1,19 +1,25 @@
...
     this.container = null;
   }
 
-  componentDidMount() {
+  componentDidMount(): void {
     this.colorize();
   }
 
-  colorize() {
+  container: HTMLDivElement | null;
```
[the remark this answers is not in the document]
SparkyFnay: container is a ref, and when it's assigned a value, the assigned value can be null

packages/console/src/log/LogView.tsx
```diff
@@ -96,15 +103,29 @@ class LogView extends PureComponent {
...
+
+  logLevelMenuItems: Record<string, LogLevelMenuItem | null>;
```
mofojed, line 120: I don't think need to track `null` here. Just check if `element != null` before setting.
```suggestion
  logLevelMenuItems: Record<string, LogLevelMenuItem>;
```

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -183,21 +197,35 @@ class MonacoUtils {
...
     monaco.editor.setTheme('dh-dark');
 
-    registerLanguages([DbLang, PyLang, GroovyLang, LogLang, ScalaLang]);
+    registerLanguages(([
+      DbLang,
+      PyLang,
+      GroovyLang,
+      LogLang,
+      ScalaLang,
+    ] as unknown[]) as Language[]);
```
mofojed, line 206: Instead of doing this - in the language files, declare the language like `const lang: Language = { id, conf, language }; export default lang`.

Will need to move the Language type to another file (probably just Language.ts)

Then you shouldn't need to cast here, it'll just be a `Language[]`.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -207,9 +235,11 @@ class MonacoUtils {
...
         const language = languageParam;
         log.debug2('Overriding default language loader:', language.id);
-        language.loader = () => ({
-          then: () => {},
-        });
+
+        // @ts-ignore
+        // language.loader = () => ({
+        //   then: () => undefined,
+        // });
```
mofojed, line 242: If this isn't needed (console still works, tests work), then just remove these lines.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -183,21 +197,35 @@ class MonacoUtils {
...
+   * @param color The hex color string to remove the hashtag from, eg. '#ffffff'
    */
-  static removeHashtag(color) {
+  static removeHashtag(color: string): string {
     return color.substring(1);
   }
 
-  static registerLanguages(languages) {
+  static registerLanguages(
+    languages: {
```
mofojed, line 221: Use the `Language` type you created above.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -218,7 +248,13 @@ class MonacoUtils {
     });
   }
 
-  static registerLanguage(language) {
+  static registerLanguage(language: {
```
mofojed, line 251: Use the `Language` type you created above.

packages/console/src/common/ConsoleUtils.ts
SparkyFnay: because in between, we are calling `ShellQuote.parse(str)`, thus turning `str` into `ParseEntry[]`. This ParseEntry type can be string, or an object with hasComment, etc.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -363,7 +415,9 @@ class MonacoUtils {
       keybindings.forEach(keybinding =>
         MonacoUtils.removeKeybinding(
```
mofojed, line 416: Check if the keybinding is empty before calling `removeKeyBinding` ... then you don't need the check for empty str in `MonacoUtils.removeKeybinding`.

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -1,19 +1,51 @@
...
 
-class ScriptEditor extends Component {
-  constructor(props) {
+interface ScriptEditorProps {
+  error?: { message?: string };
+  isLoading: boolean;
+  isLoaded: boolean;
+  focusOnMount?: boolean;
+  onChange: (e: editor.IModelContentChangedEvent) => void;
+  onRunCommand: (command: string | null) => void;
```
mofojed, line 23: Don't allow `null` here
```suggestion
  onRunCommand: (command: string) => void;
```

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -22,24 +54,24 @@ class ScriptEditor extends Component {
...
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps: ScriptEditorProps): void {
     const { sessionLanguage, settings } = this.props;
-    const { language } = settings;
-
-    const languageChanged = language !== prevProps.settings.language;
+    let language = '';
+    if (settings) {
+      language = settings.language;
+    }
```
mofojed, line 73: Why not just set it right away?
```suggestion
    const language = settings?.language;
```

mofojed (changes requested):

packages/console/src/log/LogView.tsx
SparkyFnay: EditorContainer is a ref, so it's value can be null

packages/console/src/notebook/ScriptEditor.tsx
SparkyFnay: The actual function passed in does accept values of null though

packages/console/src/monaco/MonacoUtils.ts
SparkyFnay: `const lang: Language = { id, conf, language };` 
this line causes lots of type errors. For example, the `Language` type require `blockComment` property to be of type `[string, string]`, but the language has type string[] for `blockComment`. Using as const also doesn't work since `Language` requires the array to be mutable. I think the easiest solution is to cast it

packages/console/src/command-history/CommandHistoryItemTooltip.tsx
```diff
@@ -1,57 +1,81 @@
...
+export class CommandHistoryItemTooltip extends Component<
+  CommandHistoryItemTooltipProps,
+  CommandHistoryItemTooltipState
+> {
+  static defaultProps = {
+    onUpdate: (): void => undefined,
+  };
+
+  static getTimeString(
+    startTime: string | undefined,
```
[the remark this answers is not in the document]
mofojed: Okay

packages/console/src/common/Code.tsx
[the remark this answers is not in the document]
mofojed: Sounds good. We could also be updating our refs to be ReactRef type, but that's not really in the scope of this change.

packages/console/src/Console.tsx
```diff
@@ -391,52 +504,61 @@ export class Console extends PureComponent {
...
+      const disableOldObject = (
+        object: VariableDefinition,
+        isRemoved = false
+      ) => {
+        const { title } = object;
+        assertNotNull(title);
+        const oldIndex = objectHistoryMap.get(title);
         // oldIndex can be -1 if a object is active but doesn't have a command in consoleHistory
         // this can happen after clearing the console using 'clear' or 'cls' command
+        assertNotNull(oldIndex);
```
mofojed, line 516: Instead of assertNotNull here, just check if `oldIndex != null && oldIndex >= 0` in the if statement below. We may at some point truncate the console history, in which case the objectHistoryMap may return null.

packages/console/src/Console.tsx
```diff
@@ -391,52 +504,61 @@ export class Console extends PureComponent {
...
         if (oldIndex >= 0) {
           // disable outdated object variable in the old consoleHistory item
           history[oldIndex].disabledObjects = history[
             oldIndex
-          ].disabledObjects.concat(name);
+          ].disabledObjects?.concat(title);
```
mofojed, line 521: Should handle this case better if `disabledObjects` is null... Something like:
```
          history[oldIndex].disabledObjects = [ ...(history[
            oldIndex
          ].disabledObjects ?? []), title ];
```

packages/console/src/monaco/MonacoUtils.ts
mofojed: What's the issue with `blockComment`? All of the `blockComment` we defined are of type `[string, string]`

mofojed (changes requested):

packages/console/src/monaco/lang/db.ts
```diff
@@ -42,7 +45,9 @@ const conf = {
   },
 };
 
-const language = {
+const language:
+  | monaco.languages.IMonarchLanguage
+  | monaco.Thenable<monaco.languages.IMonarchLanguage> = {
```
mofojed, line 50: You can just declare this as `monaco.languages.IMonarchLanguage`. `monaco.Thenable` is for async functions.
```suggestion
const language: monaco.languages.IMonarchLanguage = {
```

packages/console/src/monaco/lang/groovy.ts
```diff
@@ -41,7 +44,9 @@ const conf = {
   },
 };
 
-const language = {
+const language:
+  | monaco.languages.IMonarchLanguage
+  | monaco.Thenable<monaco.languages.IMonarchLanguage> = {
```
mofojed, line 49:
```suggestion
const language: monaco.languages.IMonarchLanguage = {
```

packages/console/src/monaco/lang/log.ts
```diff
@@ -1,10 +1,14 @@
...
+import { Language } from './Language';
 
 const id = 'log';
 
 const conf = {};
 
-const language = {
+const language:
+  | monaco.languages.IMonarchLanguage
+  | monaco.Thenable<monaco.languages.IMonarchLanguage> = {
```
mofojed, line 11:
```suggestion
const language: monaco.languages.IMonarchLanguage = {
```

packages/console/src/monaco/lang/python.ts
```diff
@@ -41,7 +44,9 @@ const conf = {
   },
 };
 
-const language = {
+const language:
+  | monaco.languages.IMonarchLanguage
+  | monaco.Thenable<monaco.languages.IMonarchLanguage> = {
```
mofojed, line 49:
```suggestion
const language: monaco.languages.IMonarchLanguage = {
```

packages/console/src/monaco/lang/scala.ts
```diff
@@ -67,7 +70,9 @@ const conf = {
   },
 };
 
-const language = {
+const language:
+  | monaco.languages.IMonarchLanguage
+  | monaco.Thenable<monaco.languages.IMonarchLanguage> = {
```
mofojed, line 75:
```suggestion
const language: monaco.languages.IMonarchLanguage = {
```

mofojed (changes requested):

mofojed (approved):

PR #657, opened 2022-07-15
Go to row UX cleanup

dsmmcken:
I missed reviewing PR #638 before merge, here is some minor cleanup

- Simplified CSS using flex box, allowing for better wrapping and narrow panel sizes
- Use $spacer variables instead of fixed px sizes for spacing, ensures globally consistent paddings.
- Removed icon from context menu (per spec), and changed behaviour to toggle
- Focus on open, return focus on grid

Issue I need help with:
1. ctrl-g
2. type a row like 100
3. ctrl-g to close, which now correctly returns keyboard focus to grid
4. arrow down.

expected:
row is on 101
actual:
row jumps to row 2

So not sure what's being set incorrectly?

Assigning @SparkyFnay to fix the remaining issues:
Todo:
- [x] fix the above value
- [x] clear value on close
- [x] when opened via context menu, populate the input with the row that contextmenu was opened from

### DIFF
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -19,7 +19,6 @@ const DEFAULT_FORMAT_STRING = '###,##0';
 interface GotoRowProps {
   gotoRow: string;
   gotoRowError: string;
-  selectGotoRowInput: boolean;
   onSubmit: () => void;
   model: IrisGridModel;
   onGotoRowNumberChanged: (event: ChangeEvent<HTMLInputElement>) => void;
@@ -34,7 +33,6 @@ interface GotoRowProps {
 const GotoRow = ({
   gotoRow,
   gotoRowError,
-  selectGotoRowInput,
   onSubmit,
   isShown,
   onEntering,
@@ -52,10 +50,12 @@ const GotoRow = ({
   const { rowCount } = model;
 
   useEffect(() => {
-    if (selectGotoRowInput) {
+    // when row changes without focus (i.e. via context menu), re-select input
+    if (document.activeElement !== inputRef.current) {
       inputRef.current?.select();
     }
-  }, [selectGotoRowInput]);
+  }, [gotoRow]);
+
   return (
     <IrisGridBottomBar
       isShown={isShown}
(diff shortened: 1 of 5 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2330,9 +2341,23 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     );
   }
 
-  toggleGotoRow(): void {
+  toggleGotoRow(row: string): void {
     const { isGotoRowShown } = this.state;
-    this.setState({ isGotoRowShown: !isGotoRowShown });
+    if (!isGotoRowShown) {
+      this.setState({
+        isGotoRowShown: !isGotoRowShown,
```
mofojed, line 2348: This is kind of confusing... in both cases you're setting the same values, except for `selectGotoRowInput`. You could write it just as one setState:
```
this.setState({
  isGotoRowShown: !isGotoRowShown,
  gotoRow: row,
  gotoRowError: '',
  selectGotoRowInput: !isGotoRowShown,
});
```

mofojed (commented):
Don gets last approval

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -37,67 +45,58 @@ const GotoRow = ({
...
-          <h6>
+        <div className="goto-row-wrapper">
+          <div className="goto-row-text">Go to row</div>
+          <div className="goto-row-input">
+            <input
+              ref={inputRef}
+              type="number"
+              onKeyDown={e => {
+                if (e.key === 'Enter') {
+                  onSubmit();
```
mofojed, line 80: You should consume this event
```suggestion
                  e.stopPropagation();
                  e.preventDefault();
                  onSubmit();
```

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
dsmmcken: I modified behavior slightly. After playing with it a bit, I actually preferred it to stay open on subsequent context menus calls instead of toggle, but still update the row value with the row it was invoked from.

I also removed the need for `selectGotoRowInput` by just checking document.activeElement to determine if it needs to be selected.

SparkyFnay (approved):

mofojed (approved):

PR #679, opened 2022-07-26
Clean up treemap in web UI

mofojed:
- Automatically fill in labels with IDs if labels are blank
- Fix a possible null axis check

Tested with https://github.com/deephaven/deephaven-core/pull/2678
Was able to open and interact with Treemaps correctly

### DIFF
--- a/packages/chart/src/FigureChartModel.js
+++ b/packages/chart/src/FigureChartModel.js
@@ -566,6 +566,13 @@ class FigureChartModel extends ChartModel {
       if (yLow && yHigh && yLow !== y) {
         seriesData.error_y = ChartUtils.getPlotlyErrorBars(y, yLow, yHigh);
       }
+    } else if (plotStyle === dh.plot.SeriesPlotStyle.TREEMAP) {
+      const { ids, labels } = seriesData;
+      if (ids && !labels) {
+        // If the user only provided IDs, we assign the IDs to the labels property as well automatically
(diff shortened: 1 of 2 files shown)


packages/chart/src/FigureChartModel.js
```diff
@@ -566,6 +566,13 @@ class FigureChartModel extends ChartModel {
       if (yLow && yHigh && yLow !== y) {
         seriesData.error_y = ChartUtils.getPlotlyErrorBars(y, yLow, yHigh);
       }
+    } else if (plotStyle === dh.plot.SeriesPlotStyle.TREEMAP) {
+      const { ids, labels } = seriesData;
+      if (ids && !labels) {
```
vbabich, line 571: What's the type of `labels`? If it's an array, is it enough to only check for truthy/falsy value here?

mofojed: Yes, it's an array. You're right I should be checking for undefined instead.

vbabich: Actually, I was thinking about the situation when the `labels` array is truthy but empty. Is it possible we get an empty array from the API?

mofojed: Right - if we get an empty array, we should be keeping that. It's _possible_, but would also mean there's nothing to plot. Basically means you created a plot with an empty data set.

vbabich (approved):

PR #683, opened 2022-07-27
Add "Copy" button to cell overflow preview modal

### LINKED ISSUE #637: Add "Copy" button to full cell contents preview
As a user viewing data, I want to be able to click a button to easily copy the contents of a cell when I've opened it in preview mode so that I don't have to click and drag trying to select it all.

image (33)

dsmmcken:
Resolves #637


mattrunyon (changes requested):
It will only copy the original string (which I think is fine, we only pretty print JSON)

Can you add some feedback that copy happened? Like with the "Copy column name" button on tables changing to the check icon

mattrunyon (approved):

PR #693, opened 2022-07-29
dashboard-core-plugins typescript conversion

### LINKED ISSUE #671: Convert packages/dashboard-core-plugins to typescript

SparkyFnay:
Fixes #671


packages/chart/src/isFigureChartModel.ts
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
```
mofojed, line 2: Get rid of these compiler warnings. There's an easy way to get `FigureChartModel` to conform to `ChartModel` without having to convert everything immediately.

packages/chart/src/ChartModel.ts
```diff
@@ -24,122 +29,131 @@ class ChartModel {
...
+
+  formatter?: Formatter;
+
+  rect?: DOMRect;
+
+  isDownsamplingDisabled: boolean;
+
+  title?: string;
+
+  getData(): [] {
```
mofojed, line 45: Need to have a type for the array returned - right now this is always specifying an empty array, which is why `FigureChartModel` doesn't conform (and you're having to specify `ChartModel | FigureChartModel` everywhere instead of just `ChartModel`).

You can just put `unknown[]` for now, unless you have a more specific type (I think we can fill that in when converting `FigureChartModel` though.
```suggestion
  getData(): unknown[] {
```

packages/code-studio/src/main/WidgetUtils.ts
```diff
@@ -47,10 +50,10 @@ export type GridPanelMetadata = {
 };
 
 export const createChartModel = async (
-  session: DhSession,
+  session: IdeSession,
   metadata: ChartPanelMetadata,
   panelState?: ChartPanelPanelState
-): Promise<ChartModel> => {
+): Promise<ChartModel | FigureChartModel> => {
```
mofojed, line 56:
```suggestion
): Promise<ChartModel> => {
```

packages/code-studio/src/styleguide/ContextMenus.tsx
```diff
@@ -129,7 +130,9 @@ class ContextMenus extends Component {
           }}
         >
           Right Click Me
-          <ContextActions actions={delayedActions} />
+          <ContextActions
+            actions={(delayedActions as unknown) as ResolvableContextAction[]}
```
mofojed, line 134: I know this is just the style guide... but there's a better way to fix this:
- In `ContextActions.tsx`, change `actions` in `ContextActionsProps` to allow for `ResolvableContextAction | ResolvableContextAction[]` (we check all through file if that prop is an array or not)
- Declare `delayedActions` as a `ResolvableContextAction`, get rid of casting as `Promise<ContextAction[]>` (e.g. `const delayedActions: ResolvableContextAction = ...`
- Remove casting here

packages/components/src/context-actions/ContextActions.tsx
```diff
@@ -3,18 +3,18 @@
...
 import './ContextActions.scss';
 
 const log = Log.module('ContextActions');
 
 interface ContextActionsProps {
-  actions:
-    | ContextAction[]
-    | (() => ContextAction[])
-    | (() => Promise<ContextAction[]>);
+  actions: ResolvableContextAction[];
```
mofojed, line 17: Should be `ResolvableContextAction | ResolvableContextAction[]`

packages/console/src/command-history/CommandHistory.tsx
```diff
@@ -32,14 +32,17 @@ import { ItemAction, HistoryAction } from './CommandHistoryTypes';
...
 
-type Settings = {
+export type CommandHistorySettings = {
   value: string;
-  language: string;
+  language?: string;
 };
 interface CommandHistoryProps {
-  language: string;
+  language?: string;
```
mofojed, line 40: `language` should still be required in `@deephaven/console`, I think you changed it to optional in a few places. `CommandHistoryPanel` shouldn't be rendering `CommandHistory` without a language.

mofojed (changes requested):

packages/dashboard-core-plugins/src/Plugin.ts
```diff
@@ -0,0 +1,18 @@
+import type ChartBuilderPlugin from './ChartBuilderPlugin';
+import type ChartPlugin from './ChartPlugin';
+import type ConsolePlugin from './ConsolePlugin';
+import type FilterPlugin from './FilterPlugin';
+import type GridPlugin from './GridPlugin';
+import type LinkerPlugin from './LinkerPlugin';
+import type MarkdownPlugin from './MarkdownPlugin';
+import type PandasPlugin from './PandasPlugin';
+
+export type Plugin =
```
mofojed, line 10: We shouldn't be defining `Plugin` like this - is this file even necessary?

mofojed (changes requested):

packages/chart/src/Chart.tsx
```diff
@@ -325,30 +425,36 @@ export class Chart extends Component {
...
   }
 
-  handleRelayout(changes) {
+  handleRelayout(changes: { hiddenlabels?: string[] }): void {
     log.debug('handleRelayout', changes);
     if (Object.keys(changes).includes('hiddenlabels')) {
       const { onSettingsChanged } = this.props;
       // Pie charts store series visibility in layout.hiddenlabels and trigger relayout on changes
       // Series visibility for other types of charts is handled in handleRestyle
+      assertNotNull(changes.hiddenlabels);
```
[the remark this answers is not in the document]
SparkyFnay: I thought the outer if already checks that `changes.hiddenlabels` is not undefined by checking that `hiddenlabels` is a key in `changes`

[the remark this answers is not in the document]
mofojed: The `Object.keys(changes).includes('hiddenLabels')` checks that the key is defined, but the value could still theoretically be `null` and then this assert would throw (e.g. if the passed in value for `changes` was `{ hiddenLabels: null}` or even `{ hiddenLabels: undefined }` - the key is still defined in the object, but that doesn't mean anything about the value itself).

packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx
```diff
@@ -1,29 +1,64 @@
...
+class CommandHistoryPanel extends Component<
+  CommandHistoryPanelProps,
+  CommandHistoryPanelState
+> {
+  static defaultProps = {
+    panelState: null,
+    session: null,
+    sessionId: null,
+    language: null,
+  };
```
mofojed, line 51: All these props should just stay as `undefined` if not set, meaning you don't need to define any `defaultProps`:
```suggestion
```

packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx
```diff
@@ -1,29 +1,64 @@
...
 const log = Log.module('CommandHistoryPanel');
 
-class CommandHistoryPanel extends Component {
+interface CommandHistoryPanelProps {
+  glContainer: Container;
+  glEventHub: EventEmitter;
+  panelState: Record<string, never>;
+  session: IdeSession;
+  sessionId: string;
+  language: string;
```
mofojed, line 29: All of these props should be optional.

At the Panel level, these props are optional, and when they are set then it renders the appropriate `CommandHistory` component (where these props are required).

```suggestion
  panelState?: Record<string, never>;
  session?: IdeSession;
  sessionId?: string;
  language?: string;
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -24,12 +34,62 @@ const DEFAULT_PANEL_STATE = Object.freeze({
...
+  commandHistoryStorage: CommandHistoryStorage;
+  glContainer: Container;
+  glEventHub: EventEmitter;
+
+  panelState: PanelState;
+
+  sessionWrapper: SessionWrapper;
+
+  timeZone: string;
+  unzip: (file: File) => Promise<JSZipObject[]>;
```
mofojed, line 65: `panelState` and `unzip` should be optional params:
```suggestion
  panelState?: PanelState;

  sessionWrapper: SessionWrapper;

  timeZone: string;
  unzip?: (file: File) => Promise<JSZipObject[]>;
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -24,12 +34,62 @@ const DEFAULT_PANEL_STATE = Object.freeze({
...
+
+export class ConsolePanel extends PureComponent<
+  ConsolePanelProps,
+  ConsolePanelState
+> {
+  static defaultProps = {
+    panelState: null,
+    unzip: null,
+  };
+
```
mofojed, line 87: Don't need `defaultProps`, they should just be `undefined` if not set.
```suggestion
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -143,19 +207,21 @@ export class ConsolePanel extends PureComponent {
...
+  handleTabBlur(): void {
+    if (this.consoleRef && this.consoleRef.current) {
+      ((this.consoleRef.current as unknown) as HTMLElement).blur();
     }
   }
 
-  handleTabFocus() {
+  handleTabFocus(): void {
     if (this.consoleRef) {
-      this.consoleRef.focus();
```
mofojed, line 154: Whoops, guess this just didn't work before :/

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -143,19 +207,21 @@ export class ConsolePanel extends PureComponent {
     return id;
   }
 
-  handleTabBlur() {
-    if (this.consoleRef) {
-      this.consoleRef.blur();
+  handleTabBlur(): void {
+    if (this.consoleRef && this.consoleRef.current) {
+      ((this.consoleRef.current as unknown) as HTMLElement).blur();
```
mofojed, line 212: This doesn't make sense - the ref is for `Console`, which is not an `HTMLElement`. It does not have a `blur()` function.
I think just remove this entire `handleTabBlur` method and all refs to it in `ConsolePanel`. I believe we had a use for this on Enterprise but it's not in Community and it simply wasn't wired up.

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -25,15 +28,75 @@ import {
...
+  type?: string;
+  value?: string;
+  isValueShown?: boolean;
+  timestamp?: number;
+}
+
+interface DropdownFilterPanelProps {
+  glContainer: Container;
+  glEventHub: EventEmitter;
+  panelState: PanelState;
```
mofojed, line 50: Optional:
```suggestion
  panelState?: PanelState;
```

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -25,15 +28,75 @@ import {
...
+  isLinkerActive: boolean;
+  columns: Column[];
+  columnSelectionValidator: (
```
mofojed, line 53: Optional:
```suggestion
  columnSelectionValidator?: (
```

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -25,15 +28,75 @@ import {
...
+
+class DropdownFilterPanel extends Component<
+  DropdownFilterPanelProps,
+  DropdownFilterPanelState
+> {
+  static defaultProps = {
+    columnSelectionValidator: null,
+    panelState: null,
+  };
+
```
mofojed, line 97: No need to set `defaultProps`, just leave them as `undefined`.
```suggestion
```

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -131,7 +196,7 @@ class DropdownFilterPanel extends Component {
       prevState.valuesTable !== null
     ) {
       log.debug('Table in state modified, closing the old table.');
-      prevState.valuesTable.close();
+      prevState.valuesTable?.close();
```
mofojed, line 199: Shouldn't need this.
The outer `if` check should check if it `!== undefined`, not `!== null` (due to the changes you've made `valuesTable` can be `TableTemplate | undefined`, it can't be `null`).

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -160,25 +225,43 @@ class DropdownFilterPanel extends Component {
...
+    if (valuesTable != null) {
       valuesTable.close();
     }
   }
 
-  getCachedValues = memoize((rawValues, { type, name }, formatter) => {
-    if (TableUtils.isDateType(type)) {
+  dropdownFilterRef: RefObject<DropdownFilter> | null;
+
+  panelContainer: RefObject<HTMLDivElement> | null;
```
mofojed, line 235: These can never be null, the RefObject is set in the constructor:
```suggestion
  dropdownFilterRef: RefObject<DropdownFilter>;

  panelContainer: RefObject<HTMLDivElement>;
```

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -160,25 +225,43 @@ class DropdownFilterPanel extends Component {
...
-      // Skip formatting for nulls so they don't get converted to ''
-      value != null ? formatter.getFormattedString(value, type, name) : null
-    );
-  });
+  );
 
-  getCoordinateForColumn() {
-    if (!this.panelContainer.current) {
+  getCoordinateForColumn(): [number, number] | null {
+    if (!this.panelContainer?.current) {
```
mofojed, line 264: `this.panelContainer` is never null, no need to do an optional check.
That being said, we should be more explicit about these checks instead of relying on the truthy/falsy value anyway:
```suggestion
    if (this.panelContainer.current == null) {
```

packages/dashboard-core-plugins/src/panels/DropdownFilterPanel.tsx
```diff
@@ -25,15 +28,75 @@ import {
...
+}
+
+interface DropdownFilterPanelState {
+  column?: DropdownFilterColumn;
+  formatter: Formatter;
+  valuesTable?: TableTemplate;
+  valuesColumn?: Column;
+  sourceSize: number;
+  value: string;
+  timestamp: number | null;
```
mofojed, line 74: I think you could make this optional (then it'd match `PanelState`):
```suggestion
  timestamp?: number;
```

mofojed (changes requested):

packages/dashboard-core-plugins/src/controls/markdown/MarkdownEditor.tsx
```diff
@@ -1,15 +1,31 @@
...
 import Markdown from 'react-markdown';
+import { CodeComponent } from 'react-markdown/src/ast-to-react';
 import { Code, Editor } from '@deephaven/console';
+import type * as monaco from 'monaco-editor/esm/vs/editor/editor.api.js';
 
-export default class MarkdownEditor extends PureComponent {
-  constructor(props) {
+interface MarkdownEditorProps {
+  isEditing: boolean;
+  content: string;
```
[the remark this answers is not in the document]
SparkyFnay: For these, when you have defaultProps set, typescript automatically recognizes that these props are optional when the component is used. So we should keep the types in `MarkdownEditorProps` not optional since inside the component, they are never optional.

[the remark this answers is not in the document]
mofojed: Good call, my bad

packages/dashboard-core-plugins/src/panels/LogPanel.tsx
```diff
@@ -3,21 +3,36 @@
...
 import Panel from './Panel';
 import { getDashboardSessionWrapper } from '../redux';
 
 const log = Log.module('LogPanel');
 
-class LogPanel extends PureComponent {
+interface LogPanelProps {
+  glContainer: Container;
+  glEventHub: EventEmitter;
+  session: IdeSession;
```
mofojed, line 19: Technically this should be optional.

Doesn't matter except on the Enterprise side - we show a "Waiting for session connection..." message while the user has not yet connected.

packages/dashboard-core-plugins/src/panels/MarkdownNotebook.tsx
```diff
@@ -1,15 +1,55 @@
...
+export class MarkdownNotebook extends PureComponent<
+  MarkdownNotebookProps,
+  MarkdownNotebookState
+> {
+  static defaultProps = {
+    content: '',
+    onLinkClick: undefined,
+    onRunCode: (): void => undefined,
+    transformImageUri: undefined,
+    transformLinkUri: undefined,
```
mofojed, line 49: No need to set `defaultProps` to `undefined` on optional params:
```suggestion
    onRunCode: (): void => undefined,
```

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -0,0 +1,115 @@
...
+import { Link } from '../linker/LinkerUtils';
+
+/**
+ * Set the session wrapper for the dashboard specified
+ * @param id The ID of the dashboard to set the session for
+ * @param sessionWrapper The session wrapper object to set for the dashboard
+ */
+export const setDashboardSessionWrapper = (
+  id: string,
+  sessionWrapper: unknown
```
mofojed, line 17: We should be able to set the type for `sessionWrapper` here. 
Also need to delete the old `actions.js`, and convert `selectors.js` to TS as well.

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -0,0 +1,115 @@
...
+  dispatch(updateDashboardData(id, { isolatedLinkerPanelId }));
+
+/**
+ * Set the column selection validator for a dashboard
+ * @param id The ID of the dashboard to set the column selection validator on
+ * @param columnSelectionValidator The column selection validator to set
+ */
+export const setDashboardColumnSelectionValidator = (
+  id: string,
+  columnSelectionValidator: unknown
```
mofojed, line 91: Should be able to define the function type for `columnSelectionValidator`

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -0,0 +1,115 @@
...
+  dispatch(updateDashboardData(id, { columnSelectionValidator }));
+
+/**
+ * Set the console settings for a dashboard
+ * @param {string} id The ID of the dashboard to set the console settings on
+ * @param {ConsoleSettings} consoleSettings The console settings to set for the dashboard
+ */
+export const setDashboardConsoleSettings = (
+  id: string,
+  consoleSettings: unknown
```
mofojed, line 102: Should be able to define a type for `consoleSettings`

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -0,0 +1,115 @@
...
+export const setDashboardColumnSelectionValidator = (
+  id: string,
+  columnSelectionValidator: unknown
+): ThunkAction<unknown, RootState, undefined, Action<unknown>> => dispatch =>
+  dispatch(updateDashboardData(id, { columnSelectionValidator }));
+
+/**
+ * Set the console settings for a dashboard
+ * @param {string} id The ID of the dashboard to set the console settings on
+ * @param {ConsoleSettings} consoleSettings The console settings to set for the dashboard
```
mofojed, line 98: Get rid of param type in comments:

```suggestion
 * @param id The ID of the dashboard to set the console settings on
 * @param consoleSettings The console settings to set for the dashboard
```

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -0,0 +1,115 @@
...
+export const setDashboardConsoleSettings = (
+  id: string,
+  consoleSettings: unknown
+): ThunkAction<unknown, RootState, undefined, Action<unknown>> => dispatch =>
+  dispatch(updateDashboardData(id, { consoleSettings }));
+
+/**
+ * Set the filter sets for a specific dashboard
+ * @param {string} id The ID of the dashboard to set the filter sets for
+ * @param {FilterSet[]} filterSets The filter sets to set
```
mofojed, line 109:
```suggestion
 * @param id The ID of the dashboard to set the filter sets for
 * @param filterSets The filter sets to set
```

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -288,7 +312,26 @@ class IrisGridUtils {
...
+    | 'rollupConfig'
+    | 'showSearchBar'
+    | 'searchValue'
+    | 'selectDistinctColumns'
+    | 'selectedSearchColumns'
+    | 'invertSearchColumns'
+    | 'pendingDataMap'
+    | 'frozenColumns'
+    | 'aggregationSettings'
+  > & { userColumnWidths: ModelSizeMap; userRowHeights: ModelSizeMap } {
```
mofojed, line 334: Can make this a bit cleaner - I think the only discrepancy is around `metrics` and the `userColumnWidths` and `userRowHeights`:
```suggestion
  ): Omit<HydratedIrisGridState, 'metrics'> & {
    userColumnWidths: ModelSizeMap;
    userRowHeights: ModelSizeMap;
  } {
```

packages/jsapi-utils/src/TableUtils.ts
```diff
@@ -334,7 +334,7 @@ export class TableUtils {
     }
   }
 
-  static isDateType(columnType: string): boolean {
+  static isDateType(columnType: string | undefined): boolean {
```
mofojed, line 337: I don't think ` | undefined` should be allowed here.

mofojed (changes requested):
There's still a few test files that are still in `jsx` format - was there difficulty in converting these?
```
packages/dashboard-core-plugins/src/redux/actions.js (should be deleted, .ts file already exists)
packages/dashboard-core-plugins/src/redux/selectors.js
packages/dashboard-core-plugins/src/panels/Panel.test.jsx
packages/dashboard-core-plugins/src/panels/IrisGridPanel.test.jsx
packages/dashboard-core-plugins/src/panels/PanelContextMenu.test.jsx
packages/dashboard-core-plugins/src/panels/ConsolePanel.test.jsx
packages/dashboard-core-plugins/src/panels/MarkdownPanel.test.jsx
```

packages/dashboard-core-plugins/src/panels/ColumnSelectionValidator.ts
```diff
@@ -0,0 +1,7 @@
+import { PanelComponent } from '@deephaven/dashboard';
+import { LinkColumn } from '../linker/LinkerUtils';
+
+export type ColumnSelectionValidator = (
```
mofojed, line 4: This file should probably be in the `linker` folder... or just even defined in `LinkerUtils`.

packages/dashboard-core-plugins/src/panels/ColumnSelectionValidator.ts
```diff
@@ -0,0 +1,7 @@
...
+  panel: PanelComponent,
+  tableColumn?: Partial<LinkColumn>
```
mofojed, line 6: I think instead of using `Partial<LinkColumn>`, you want to define a new type, `LinkColumnSelection = { name: string, type?: string }` and use that.

With `Partial`, you then also have to check that name is defined, which it always will be.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -550,7 +551,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       case ToolType.LINKER:
         setDashboardColumnSelectionValidator(
           localDashboardId,
-          this.isColumnSelectionValid
+          this.isColumnSelectionValid as ColumnSelectionValidator
```
mofojed, line 554: Instead I think `isColumnSelectionValid` should also take a `Partial<LinkColumn>` rather than just `LinkColumn` - you'll need to put the appropriate checks for undefineds in there as well.

In particular - in LinkerUtils, there's a check on the `columnType`, checking for `null`:
```
    // Null columnType in ending link point allows linking to any type
    const isCompatibleType =
      endColumnType === null ||
      TableUtils.isCompatibleType(startColumnType, endColumnType);
```

mofojed (changes requested):

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -354,7 +353,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
           'endPanel no longer exists, ignoring unsetFilterValue',
           panelId
         );
-      } else if (isLinkablePanel(endPanel)) {
+      } else if (isLinkablePanel(endPanel) && columnType != null) {
```
mofojed, line 356: `unsetFilterValue` should just allow null columnType instead. By checking for `null` here you're changing one of the code paths.
```suggestion
      } else if (isLinkablePanel(endPanel)) {
```

mofojed (changes requested):

mofojed (approved):

PR #702, opened 2022-08-02
Remove environment variables from packages

### LINKED ISSUE #700: Remove env variable references in packages
Only code-studio or other applications should have references to env variables. Currently, the following reference environment variables but are not part of code-studio.

- ConsoleUtils
- ConsoleConstants
- NotebookPanel
- DefaultLogger

These should be removed and the info from the env variables should be passed in from the consuming application

mattrunyon:
Fixes #700 

Removes all environment variables from packages and instead only passes them from applications such as code-studio or embed-grid

Also exposes Log to the browser globals as `DHLog` so we can change log level on the fly if we want via `DHLog.setLogLevel(4)` or any other log level

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -4,7 +4,11 @@ import 'jest-canvas-mock';
 import './__mocks__/dh-core';
 import Log from '@deephaven/log';
 
-Log.setLogLevel(-1);
+let logLevel = parseInt(process.env.DH_LOG_LEVEL ?? '', 10);
+if (!Number.isFinite(logLevel)) {
+  logLevel = -1;
+}
+Log.setLogLevel(logLevel);
 
 // disable annoying dnd-react warnings
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -2,6 +2,9 @@ import 'regenerator-runtime/runtime';
 import '@testing-library/jest-dom';
 import 'jest-canvas-mock';
 import './__mocks__/dh-core';
+import Log from '@deephaven/log';
+
+Log.setLogLevel(-1);
 
 // disable annoying dnd-react warnings
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/code-studio/README.md
+++ b/packages/code-studio/README.md
@@ -30,18 +30,18 @@ One common setup is to override the API server URL to point to another server. F
 REACT_APP_CORE_API_URL=https://www.myserver.com/notebooks
 ```
 
-### DEEPHAVEN_LOG_LEVEL
+### REACT_APP_LOG_LEVEL
 
-Printing detailed logs when debugging can be handy. To enable the highest level of logging, set the log level in your `.env.development.local` file:
+Printing detailed logs when debugging can be handy. The highest level of logging is already set in your `.env.development` file. You can change it in `.env.development.local` if desired.
 
 ```shell
-DEEPHAVEN_LOG_LEVEL=4
+REACT_APP_LOG_LEVEL=2 # Warn/Error
 ```
 
-When running the unit tests, you often do not want any logs to be printed out. To disable the logs while running the unit tests, set the log level in your `.env.test.local` file:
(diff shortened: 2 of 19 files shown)


packages/console/src/common/ConsoleUtils.ts
```diff
@@ -55,20 +52,6 @@ class ConsoleUtils {
...
-      try {
-        const url = new URL(apiUrl);
-        defaultHost = url.hostname;
-      } catch (error: unknown) {
-        log.error('API_URL failed', error);
-      }
-    }
-    return defaultHost;
-  }
-
```
mattrunyon, line 71: I couldn't find any uses of this in DHC or DHE

jest.setup.ts
```diff
@@ -2,6 +2,9 @@ import 'regenerator-runtime/runtime';
 import '@testing-library/jest-dom';
 import 'jest-canvas-mock';
 import './__mocks__/dh-core';
+import Log from '@deephaven/log';
+
+Log.setLogLevel(-1);
```
mofojed, line 7: Tests should always have the env variables, no? It's sometimes useful to enable logs in tests, and this would negate that. Should read REACT_APP_LOG_LEVEL ohhhh wait that's only in code-studio. Hm.
Either way, should allow it to read from an env var, default to -1 if not provided... so then you could do something like `DEEPHAVEN_LOG_LEVEL=4 npm test` or something.

```
let logLevel = parseInt(process.env.DEEPHAVEN_LOG_LEVEL ?? '', 10);
if (!Number.isFinite(logLevel)) {
  logLevel = -1;
}
Log.setLogLevel(logLevel);
```

packages/console/src/common/ConsoleUtils.ts
mofojed: Good to delete then!

packages/console/src/common/ConsoleUtils.ts
```diff
@@ -1,8 +1,5 @@
 import ShellQuote, { ParseEntry, ControlOperator } from 'shell-quote';
 import dh, { VariableTypeUnion } from '@deephaven/jsapi-shim';
-import Log from '@deephaven/log';
-
-const log = Log.module('FigureChartModel');
```
mofojed, line 5: Whoooops 😬

packages/code-studio/src/log/LogInit.ts
```diff
@@ -12,13 +12,15 @@ export const logProxy = new LogProxy();
 export const logHistory = new LogHistory(logProxy);
 
 export default function logInit(): void {
+  Log.setLogLevel(parseInt(process.env.REACT_APP_LOG_LEVEL ?? '', 10));
```
mofojed, line 15: When we do a `parseInt`, we should be checking if it's a valid input... for example, `parseInt('', 10)` returns `NaN`, in which case we should fall back to a default value.

EDIT: Nevermind, it already ignores it in `setLogLevel`

packages/code-studio/README.md
```diff
@@ -30,19 +30,15 @@ One common setup is to override the API server URL to point to another server. F
...
-DEEPHAVEN_LOG_LEVEL=4
+REACT_APP_LOG_LEVEL=2 # Warn/Error
 ```
 
-When running the unit tests, you often do not want any logs to be printed out. To disable the logs while running the unit tests, set the log level in your `.env.test.local` file:
-
-```shell
-DEEPHAVEN_LOG_LEVEL=-1
-```
+When running the unit tests, you often do not want any logs to be printed out. Logging is disabled in unit tests in the root `jest.setup.ts`.
```
mofojed, line 41: We should probably include this information in the top level `README.md`, and have it link to the jest.setup.ts file.

mofojed (changes requested):

packages/code-studio/src/log/LogInit.ts
mattrunyon: Ya I chose to ignore it in `setLogLevel` since `typeof NaN === 'number'` so TS won't complain if you pass `NaN`.

jest.setup.ts
mattrunyon: Yep the env variables aren't read since tests are run from root. I'll add an env variable

mofojed (approved):

PR #707, opened 2022-08-04
Create @deephaven/embed-chart

mofojed:
- Starts up a server on http://localhost:4020 that can be used in an iframe to embed a chart in another page
- E.g. Load figure 'foo' by navigating to http://localhost:4020?name=foo
- Embed in another page with <iframe src="http://localhost:4020?name=foo"></iframe>

Testing Done:
1) In IDE (http://localhost:4000), executed the following snippet to create a figure:
```
from deephaven import read_csv

# source the data
source = read_csv("https://media.githubusercontent.com/media/deephaven/examples/main/StandardAndPoors500Financials/csv/s_and_p_500_market_cap.csv")

# plot the data
s_and_p_marketcap = Figure().plot_treemap(series_name = "S&P 500 Market Cap", t = source, id = "Label", parent = "Parent", label = "Label", value = "Value", hover_text = "HoverText", color = "Color").show()
```
2) Open the newly created figure by navigating to http://localhost:4020/?name=s_and_p_marketcap


packages/embed-chart/jest.config.cjs
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  ...baseConfig,
+  displayName: packageJson.name,
+  resetMocks: false,
+  transform: {
+    // Using ts-jest (even with its babelConfig option set) causes
+    // problems with some code-studio tests
+    '.(ts|tsx|js|jsx)': 'babel-jest',
+  },
```
mattrunyon, line 12: This can be removed. I noticed when I was switching to Vite that it's not needed since we switched the default from `ts-jest` (tests w/ type check) to `babel-jest` at the root.

packages/embed-chart/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "@types/react-dom": "^17.0.9",
+    "classnames": "^2.3.1",
+    "deep-equal": "^2.0.5",
+    "fira": "mozilla/fira#4.202",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-scripts": "4.0.3",
+    "sass": "^1.39.0",
+    "shortid": "^2.2.16",
+    "typescript": "~4.5.4"
```
mattrunyon, line 40: Technically these are all dev dependencies. Since we bundle applications, nothing except the bundle is needed. If you were to install w/ `npm install --production`, things would be fine if none of these dependencies were downloaded and you only got the build/dist folder

At the very least, babel, types, react-scripts, sass, typescript are dev deps

packages/embed-chart/public/index.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html>
```
mattrunyon, line 1: I can't comment directly b/c they're binary files, but I think the `public/favicon.ico` and `public/logo` are the react-scripts generated files and can be deleted

packages/embed-chart/src/App.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [model, setModel] = useState<ChartModel>();
+  const [error, setError] = useState<string>();
+  const [isLoading, setIsLoading] = useState(true);
+  const searchParams = useMemo(
+    () => new URLSearchParams(window.location.search),
+    []
+  );
+
+  const initApp = useCallback(async () => {
+    try {
```
mattrunyon, line 41: This could just go inside the `useEffect` for init instead of in a separate function like this

mattrunyon (changes requested):
Maybe wait for Vite too?

mofojed:
Waiting for #711 to merge first, as that will change the build system and subsequently change how we serve up embed-grid/embed-chart.

packages/embed-chart/public/index.html
mofojed: They're actually Deephaven logos, but they're not used so I'll delete them.

mofojed:
@mattrunyon I think we merge this today/tomorrow, before #711 and leave the Vite transition for Sept.
Want to make sure this gets in so I can update Jupyter plugin with plotting, without risking the dev churn/risk by converting to Vite so late in the release cycle.

mattrunyon (approved):

mattrunyon:
Will need a corresponding core update to add this in the web-client-ui dockerfile

PR #711, opened 2022-08-08
Change build system to Vite

### LINKED ISSUE #494: Update to react-script 5.0.0 or Vite
Update to react-scripts 5.0.0

Follow the migration guide: https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#migrating-from-40x-to-500

mattrunyon:
Fixes #494

Adds a launch profile which can be used to attach to Chrome and set breakpoints within VSCode in any of our packages. The launch profile will also create a Chrome profile within your `.vscode` directory so you only need to install React dev tools once.

Adds the web worker for Monaco editor. Didn't include any of the language workers (JS, TS, HTML, CSS, JSON). The worker does not work in dev mode for Firefox, but will work in prod.

Added `npm run preview` which can run the compiled app. I'm using the `.env.local` below to build the app, but still run against `localhost:10000`. The core API URL is used to create the session.

```
VITE_CORE_API_URL=http://localhost:10000/jsapi
VITE_PROXY_URL=http://localhost:10000
```


package.json
```diff
@@ -25,15 +25,13 @@
...
     "build:all": "lerna run build --ignore=@deephaven/{icons,golden-layout} --stream",
     "build:app": "lerna run --scope=@deephaven/code-studio build",
-    "build:embed-grid": "lerna run --scope=@deephaven/embed-grid build",
-    "build:packages": "lerna run --ignore=@deephaven/{code-studio,embed-grid} build --stream",
+    "build:packages": "lerna run --ignore=@deephaven/code-studio build --stream",
     "build:profile": "lerna run build --stream --profile",
     "build:necessary": "lerna run build --scope=@deephaven/{icons,golden-layout}",
-    "prestart": "lerna run prestart --stream",
-    "start": "npm-run-all types --parallel watch:types start:*",
+    "preview": "lerna run --scope=@deephaven/code-studio preview",
```
mofojed, line 31: Add a note to the readme about what `preview` is about/why you'd want to use it

packages/babel-preset/index.js
```diff
@@ -19,6 +21,14 @@ module.exports = api => ({
...
   plugins: [
     api.env('test') ? false : ['babel-plugin-add-import-extension'],
+    api.env('test')
+      ? [
+          // This is needed to replace import.meta w/ process in Jest
+          // Jest does not play nicely w/ ESM and Vite uses import.meta
+          // import.meta is only avaialable in ESM
+          path.resolve(__dirname, 'importMetaEnvPlugin'),
+        ]
+      : false,
```
mofojed, line 31: You could just write this in one ternary instead of two, e.g.
```suggestion
    api.env('test')
      ? [
          // This is needed to replace import.meta w/ process in Jest
          // Jest does not play nicely w/ ESM and Vite uses import.meta
          // import.meta is only avaialable in ESM
          path.resolve(__dirname, 'importMetaEnvPlugin'),
        ]
      : ['babel-plugin-add-import-extension'],
```

Might be worth explaining why we don't need babel-plugin-add-import-extension for Jest

packages/code-studio/package.json
```diff
@@ -70,19 +68,21 @@
...
   "devDependencies": {
     "@deephaven/eslint-config": "file:../eslint-config",
     "@deephaven/mocks": "file:../mocks",
     "@deephaven/prettier-config": "file:../prettier-config",
     "@deephaven/stylelint-config": "file:../stylelint-config",
-    "@deephaven/tsconfig": "file:../tsconfig"
+    "@deephaven/tsconfig": "file:../tsconfig",
+    "@vitejs/plugin-react": "^2.0.0",
+    "autoprefixer": "^10.4.8",
+    "postcss": "^8.4.14",
```
mofojed, line 84: Update the version in the root `package.json` as well (or you may be able to remove it entirely from there now).

packages/code-studio/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
mofojed, line 1: Would it make more sense to simply include `postcss.config.js` in the `ignorePatterns` in `.eslintrc`? Or add a comment here about how this is only used in dev

packages/code-studio/public/electron.js
```diff
@@ -1,26 +1,29 @@
-const electron = require('electron')
+const electron = require('electron');
```
mofojed, line 1: We could proooobably just delete this file... we don't at all.

packages/code-studio/src/embed-grid/index.html
```diff
@@ -15,31 +12,17 @@
...
       homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
     -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-
-    <% if (process.env.NODE_ENV === 'development') { %>
-    <link rel="icon" href="%PUBLIC_URL%/favicon-cc-app-dev.svg" />
-    <% } else { %>
-    <link rel="icon" href="%PUBLIC_URL%/favicon-cc-app.svg" />
-    <% } %>
+    <link rel="manifest" href="/manifest.json" />
```
mofojed, line 15: Hmm. I think still need some sort of public_url here no? Otherwise if embed-grid is on a path then it'll load the wrong manifest.json

packages/code-studio/src/embed-grid/src/App.scss
```diff
@@ -1,5 +1,5 @@
 // Import Deephaven styling so we can use it's variables
-@import '~@deephaven/components/scss/custom_webpack.scss';
+@import '@deephaven/components/scss/custom.scss';
```
mofojed, line 2: You didn't delete `custom_webpack.scss`, I guess we still need it for other consumers of `@deephaven/components`?

packages/code-studio/.env.development
```diff
@@ -1,15 +1,10 @@
...
+VITE_ENABLE_LOG_PROXY=false
+VITE_CORE_API_URL=http://localhost:10000/jsapi
+VITE_FAVICON=/favicon-cc-app-dev.svg
 
-PUBLIC_URL=/
-REACT_APP_INTERNAL_COMPONENT_PLUGINS=ExamplePlugin
+BASE_URL=/
+VITE_INTERNAL_COMPONENT_PLUGINS=ExamplePlugin
 PORT=4000
 REACT_APP_LOG_LEVEL=4
```
mofojed, line 8: I don't think this is needed anymore, but you should define `VITE_LOG_LEVEL`

packages/code-studio/src/main/AppInit.jsx
```diff
@@ -96,14 +97,16 @@ const AppInit = props => {
     log.debug('Loading plugins...');
     try {
       const manifest = await PluginUtils.loadJson(
-        `${process.env.REACT_APP_MODULE_PLUGINS_URL}/manifest.json`
+        `${import.meta.env.VITE_MODULE_PLUGINS_URL}/manifest.json`
```
mofojed, line 100: Have you tested this with plugins? E.g. the matplotlib plugin?

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Launch Code Studio",
+      "request": "launch",
+      "type": "chrome",
```
mofojed, line 10: Haven't debugged yet, but this isn't working for me on Ubuntu. Probably Chromium vs. Chrome or something.

packages/code-studio/src/plugins/PluginUtils.jsx
```diff
@@ -7,17 +7,19 @@ const log = Log.module('PluginUtils');
...
   /**
-   * Load a component plugin either specified in the REACT_APP_COMPONENT_TABLE_PLUGINS environment variable, or from the server if it's not internal.
+   * Load a component plugin either specified in the VITE_INTERNAL_COMPONENT_PLUGINS environment variable, or from the server if it's not internal.
    * @param {string} pluginName Name of the table plugin to load
    * @returns {JSX.Element} A lazily loaded JSX.Element from the plugin
    */
   static loadComponentPlugin(pluginName) {
     if (
-      process.env.REACT_APP_INTERNAL_COMPONENT_PLUGINS.split(',').includes(
+      import.meta.env.VITE_INTERNAL_COMPONENT_PLUGINS.split(',').includes(
```
mofojed, line 16: Tested with internal plugins?

packages/code-studio/vite.config.ts
```diff
@@ -0,0 +1,123 @@
...
+      },
+    },
+  });
+
+  const packagesDir = path.resolve(__dirname, '..');
+
+  // These are paths which should be proxied to the core server
+  // https://vitejs.dev/config/server-options.html#server-proxy
+  const proxy = {};
+  const proxyPaths = [
```
mofojed, line 31: These could be defined within the if block below, they're not used outside of that.

packages/code-studio/vite.config.ts
```diff
@@ -0,0 +1,123 @@
...
+      };
+    });
+  }
+
+  return {
+    root: 'src',
+    publicDir: path.resolve(__dirname, 'public'),
+    envDir: path.resolve(__dirname),
+    server: {
+      port: Number.parseInt(env.PORT, 10) ?? 4000,
```
mofojed, line 51: The optional chaining doesn't make sense here, `Number.parseInt` always returns a `number` (`NaN` if it can't parse it).
Make a method that parses it, then returns `4000` if the parsed number is `<= 0` or `NaN`

packages/code-studio/vite.config.ts
```diff
@@ -0,0 +1,123 @@
...
+      open: true,
+      proxy,
+    },
+    resolve: {
+      dedupe: ['react', 'react-redux', 'redux'],
```
mofojed, line 56: dedupe ehhhh? Do we need to ensure our versions still match across package.jsons?

packages/code-studio/vite.config.ts
```diff
@@ -0,0 +1,123 @@
...
+        {
+          find: /^@deephaven\/icons$/,
+          replacement: `${packagesDir}/icons/dist/index.es.js`,
+        },
+        {
+          find: /^@deephaven\/(.*)/,
+          replacement: `${packagesDir}/$1/src`,
+        },
+      ],
+      // TODO: See if this is better than the last find/replace above
```
mofojed, line 79: Ticket # for follow-up?

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -414,8 +427,12 @@ class MonacoUtils {
    */
   static removeKeybinding(
     editor: monaco.editor.IStandaloneCodeEditor,
-    keybinding: string
+    keybinding: string | undefined
```
mofojed, line 430: Why not just make optional
```suggestion
    keybinding?: string
```

mofojed (changes requested):
I have some changes to package-lock.json after running `npm install` as well, make sure to check that in after a clean install.

packages/code-studio/vite.config.ts
```diff
@@ -0,0 +1,123 @@
...
+    build: {
+      outDir: path.resolve(__dirname, env.VITE_BUILD_PATH),
+      emptyOutDir: true,
+      commonjsOptions: {
+        include: [/node_modules/, /golden-layout/],
+      },
+      rollupOptions: {
+        input: {
+          main: path.resolve(__dirname, 'src/index.html'),
+          embedGrid: path.resolve(__dirname, 'src/embed-grid/index.html'),
```
mofojed, line 94: How do I actually use `embed-grid` now?

packages/code-studio/.env.development
mattrunyon: Good catch. Probably missed this in a rebase

packages/code-studio/vite.config.ts
mattrunyon: I'm not sure we need the dedupe here actually (I added it when trying to fix an issue that was a different part of the config), but I figure it doesn't hurt.

[Vite docs](https://vitejs.dev/config/shared-options.html#resolve-dedupe) says it will always resolve these packages from the project root. So if for some reason our versions got out of sync and the hoisted version of React was not the version used in code-studio, we would still resolve to the version in `code-studio/node_modules/react`. Or if that doesn't exist then it would go to `<root>/node_modules/react`, but if it does not exist in code-studio node_modules, then the hoisted version satisfies what code-studio specified

packages/console/src/monaco/MonacoUtils.ts
mattrunyon: The method makes no sense to call without this parameter, but the value could be undefined (Windows shortcut we remove, but not a mac shortcut).

I think it's better to have required params that can be undefined if the method makes 0 sense to call without a param

packages/code-studio/vite.config.ts
mattrunyon: I'll update some readmes, but now you go to `localhost:4000/embed-grid/` in dev and without the trailing slash in prod. The `embedGrid` key is just what the output JS chunk gets called

packages/babel-preset/index.js
mattrunyon: I don't know why we ignore the add import extnsion plugin in test, but it causes tests to error so I'll add that comment

packages/code-studio/public/electron.js
mattrunyon: I'll delete it

packages/code-studio/vite.config.ts
mattrunyon: Good catch

packages/code-studio/src/embed-grid/index.html
mattrunyon: Vite automatically rewrites URLs in index.html to include the base path in the config. See [docs]('babel-plugin-add-import-extension')

The only thing it doesn't do is if you are creating strings of URLs dynamically. In those spots you need to include the env variable. I changed everywhere we had `PUBLIC_URL` (CRA variable) to `BASE_URL` (Vite variable)

packages/code-studio/src/embed-grid/src/App.scss
mattrunyon: Nope looks like it's not needed any more. I think it was needed before because of CRA and the way its webpack config handled relative paths/imports

.vscode/launch.json
mattrunyon: That's exactly it. And unfortunately vscode doesn't have support for like a `launch.local.json` where you can override/extend our config. You can just set your local workspace config which applies to all of your workspaces

packages/code-studio/src/main/AppInit.jsx
mattrunyon: It "works". The issue is not on the code-studio side, but on the plugin side.

The plugins need to not externalize reactstrap because it's been removed from code-studio.

packages/code-studio/src/plugins/PluginUtils.jsx
mattrunyon: Works. Although the way to attach plugins is not good in the Python API (you have to get the underlying j_table and attach attributes)

packages/code-studio/vite.config.ts
mattrunyon: I'm removing this comment actually. I don't think mainFields would help unless we were in some scenario where `source` might not always be `packages/{name}/src/index`, but it is in our case and I can't see any reason we won't keep that going forward

packages/code-studio/package.json
mattrunyon: I can actually remove it from both b/c Vite includes postcss

mofojed:
Note that I had issues on Chrome in Ubuntu, where it would just hang with a pending request forever when there was a code change. That issue is identified in this thread: https://github.com/vitejs/vite/issues/5310
A couple of things I did to fix it (all from that thread):
- Updated the `DefaultLimitNOFILE` prop: https://github.com/vitejs/vite/issues/5310#issuecomment-949349291
- Updated `max_user_instances` via `sudo sysctl fs.inotify.max_user_instances=8192`: https://github.com/vitejs/vite/issues/5310#issuecomment-1181860523

mofojed (changes requested):
- Verify this works in deephaven-core with an alpha built package
- I'm getting an error when trying out embedGrid locally. Going to http://localhost:4010/?name=source, the error is:
```
Uncaught ReferenceError: global is not defined
    at ../../node_modules/has-bigints/index.js (index.js:3:15)
    at __require2 (chunk-IKJM6GHQ.js?v=39ba4514:18:50)
    at ../../node_modules/is-bigint/index.js (index.js:3:18)
    at __require2 (chunk-IKJM6GHQ.js?v=39ba4514:18:50)
    at ../../node_modules/which-boxed-primitive/index.js (index.js:7:16)
    at __require2 (chunk-IKJM6GHQ.js?v=39ba4514:18:50)
    at ../../node_modules/deep-equal/index.js (index.js:10:27)
    at __require2 (chunk-IKJM6GHQ.js?v=39ba4514:18:50)
    at dep:deep-equal:1:16
```
- The react-error-overlay would be nice. I see there's a ticket in Vite for it: https://github.com/vitejs/vite/issues/2076. Dunno if it's helpful to see how CRA implements: https://github.com/facebook/create-react-app/blob/f0a837c1f07ebd963ddbba2c2937d04fc1b79d40/packages/react-dev-utils/webpackHotDevClient.js

mattrunyon:
Publishing an alpha now to test w/ DHC

Fixed the local dev mode issue. Also you can run `npm run preview` from the root now and it will start all 3 on 4000, 4010, and 4020. You need to set `.env.local` in the 3 files like specified in the root readme

Opened #748 for the error overlay

mattrunyon:
Local `prepareCompose` works for me in DHC w/ version `0.16.1-vite.20` published from the latest commit. Tested code-studio, embed-grid, and embed-chart

mofojed:
@mattrunyon it looks like this doesn't generate source maps, is there an option to generate source maps as well? I think you mentioned something like that.

Was just trying to analyze the package size using `source-map-explorer`.

mattrunyon:
@mofojed You can add `build.sourcemap: true` in `vite.config.ts` and it will generate separate sourcemap files.

Or we can add a rollup plugin to do it like [this CLI analyzer](https://www.npmjs.com/package/rollup-plugin-analyzer#usage) or [this visualizer plugin](https://www.npmjs.com/package/rollup-plugin-visualizer)

Vite config gets modified like this

```ts
// vite.config.ts
import analyze from 'rollup-plugin-analyzer'

export default {
  build: {
    rollupOptions: {
      plugins: [analyze()],
    },
  },
}
```

mofojed:
@mattrunyon I tried `sourcemap: true` but it seems like the source maps are incomplete (or source-map-explorer doesn't recognize them anyway:

```
➜  main git:(vite) ✗ npx source-map-explorer './packages/code-studio/build/assets/*.js'
./packages/code-studio/build/assets/editor.worker.7a4cf0fa.js
  Your source map refers to generated column 1 on line 2, but the source only contains 0 column(s) on that line.
  Check that you are using the correct source map.
./packages/code-studio/build/assets/index.b34304b6.js
  Your source map refers to generated column 1 on line 42, but the source only contains 0 column(s) on that line.
  Check that you are using the correct source map.
./packages/code-studio/build/assets/monaco.bc5092a3.js
  Your source map refers to generated column 1 on line 5, but the source only contains 0 column(s) on that line.
  Check that you are using the correct source map.
./packages/code-studio/build/assets/vendor.85aaeefc.js
  Your source map refers to generated column 1 on line 33, but the source only contains 0 column(s) on that line.
  Check that you are using the correct source map.
./packages/code-studio/build/assets/helpers.162e09a8.js
  Unable to map 597/643 bytes (92.85%)
./packages/code-studio/build/assets/plotly.b80a4841.js
  Unable to map 333/911623 bytes (0.04%)
```

Using the plugin helps with analyzing package size, but doesn't help us for debugging a production build of the app - source maps would help greatly with that.

mattrunyon:
I've made the source map explorer package we were using before work. It doesn't create the file in the local directory (like the rollup plugin does) and opens the file automatically (unlike the rollup plugin). The errors it was having is due to the rollup CJS helpers I think which injects code that doesn't map to a source. The error can be ignored and it mentions that it can't analyze like 500 bytes of each bundle

There is a root level `npm run analyze` that will generate and open a bundle analysis for all 3 bundles (code-studio, embed-grid, embed-chart). You have to have a local build first

mofojed (approved):

PR #714, opened 2022-08-09
Only collect code coverage in CI

### LINKED ISSUE #662: Disable coverage in dev environment, have CI enable it
As a developer, I want coverage collection disabled by default so that my tests run faster

mattrunyon:
Fixes #662 

`npm run test` locally will not collect coverage now. This is nicer because when you start the tests and then want to quit early to run a specific test, Jest does not calculate coverage for files

CI will still collect coverage as normal

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ We are still using node 16.x and npm 8.x. If you are [using nvm](https://github.
 
   Note that log messages from other sources such as react prop types will still be printed since they do not go through our logger.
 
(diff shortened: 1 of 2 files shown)


mofojed (dismissed):

mofojed:
What's the flag if we want to run locally from the command line and force collect coverage?

mattrunyon:
Added the coverage flag/command to the readme

mofojed (approved):

PR #716, opened 2022-08-10
@deephaven/code-studio TypeScript Conversion

### LINKED ISSUE #651: Convert @deephaven/code-studio package to TypeScript

SparkyFnay:
Fixes #651


packages/code-studio/src/plugins/PluginUtils.tsx
```diff
@@ -8,17 +8,20 @@ const log = Log.module('PluginUtils');
 class PluginUtils {
   /**
    * Load a component plugin either specified in the REACT_APP_COMPONENT_TABLE_PLUGINS environment variable, or from the server if it's not internal.
-   * @param {string} pluginName Name of the table plugin to load
-   * @returns {JSX.Element} A lazily loaded JSX.Element from the plugin
+   * @param pluginName Name of the table plugin to load
+   * @return A lazily loaded JSX.Element from the plugin
```
mofojed, line 12: Missing 's' on @returns
```suggestion
   * @returns A lazily loaded JSX.Element from the plugin
```

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
+
   static inputDebounceTime = 250;
 
-  static focusFirstInputInContainer(container) {
-    const input = container.querySelector('input, select, textarea');
+  static focusFirstInputInContainer(container: HTMLElement | null): void {
+    const input = container?.querySelector('input, select, textarea');
     if (input) {
-      input.focus();
+      (input as HTMLElement).focus();
```
mofojed, line 118: Prefer to type on the declaration instead of casting after:
```
const input: HTMLElement | null | undefined = container?.querySelector(
      'input, select, textarea'
    );
```

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
+  ): boolean {
     // Undefined or empty string formats are always invalid
     if (!columnType || !format.formatString) {
       return false;
     }
 
+    const { formatString } = format;
+    assertNotNull(formatString);
+
+    const nullCheckedFormat = { formatString };
```
mofojed, line 161: Instead, create a predicate to check that formatString isn't null, e.g.:
```
isFormatStringFormat(format: Partial<TableColumnFormat>): format is Pick<TableColumnFormat, 'formatString'>
```

Then you don't need to create a new object each time you check

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
         log.warn('Trying to validate format of unknown type');
         return true;
       }
     }
   }
 
-  static removeFormatRuleExtraProps(item) {
+  static removeFormatRuleExtraProps(
+    item: FormatterItem
+  ): Pick<FormatterItem, 'columnName' | 'columnType' | 'format'> {
```
mofojed, line 178: Would be more accurate to use `Omit`:
```suggestion
  ): Omit<FormatterItem, 'id' | 'isNewRule'> {
```

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
     return rest;
   }
 
-  static isFormatRuleValidForSave(rule) {
+  static isFormatRuleValidForSave(rule: FormatterItem): boolean {
     return (
-      FormattingSectionContent.isValidColumnName(rule.columnName) &&
-      FormattingSectionContent.isValidFormat(rule.columnType, rule.format)
+      FormattingSectionContent.isValidFormat(rule.columnType, rule.format) &&
+      FormattingSectionContent.isValidColumnName(rule.columnName)
```
mofojed, line 186: Not sure why you reversed the order of these calls - I'd keep the order the same, since there's no reason to check if the format is valid if the column name isn't valid

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
+  id?: number;
+  isNewRule?: boolean;
+};
+
+type FormatOption = {
+  defaultFormatString?: string;
+};
+
+interface FormattingSectionContentProps {
+  formatter?: FormatterItem[];
```
mofojed, line 58: `formatter` should be required. There's a bunch of props that used to be required that you've marked optional here, such as `defaultDateTimeFormat`, `showTimeZone`, `showTSeparator`, etc:
```
FormattingSectionContent.propTypes = {
  formatter: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
  defaultDateTimeFormat: PropTypes.string.isRequired,
  showTimeZone: PropTypes.bool.isRequired,
  showTSeparator: PropTypes.bool.isRequired,
  timeZone: PropTypes.string.isRequired,
  truncateNumbersWithPound: PropTypes.bool.isRequired,
  settings: PropTypes.shape({}).isRequired,
  saveSettings: PropTypes.func.isRequired,
  scrollTo: PropTypes.func,
  defaultDecimalFormatOptions: PropTypes.shape({
    defaultFormatString: PropTypes.string,
  }).isRequired,
  defaultIntegerFormatOptions: PropTypes.shape({
    defaultFormatString: PropTypes.string,
  }).isRequired,
  defaults: PropTypes.shape({
    defaultDateTimeFormat: PropTypes.string.isRequired,
    defaultDecimalFormatOptions: PropTypes.shape({
      defaultFormatString: PropTypes.string,
    }).isRequired,
    defaultIntegerFormatOptions: PropTypes.shape({
      defaultFormatString: PropTypes.string,
    }).isRequired,
    showTimeZone: PropTypes.bool.isRequired,
    showTSeparator: PropTypes.bool.isRequired,
    timeZone: PropTypes.string.isRequired,
  }),
};
```

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -208,15 +308,15 @@ export class FormattingSectionContent extends PureComponent {
       truncateNumbersWithPound,
     } = props;
 
-    const formatSettings = formatter.map((item, i) => ({
+    const formatSettings = formatter?.map((item, i) => ({
```
mofojed, line 311: Shouldn't need to check as formatter is required

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -208,15 +308,15 @@ export class FormattingSectionContent extends PureComponent {
...
       ...item,
       id: i,
     }));
 
     this.containerRef = React.createRef();
     this.addFormatRuleButtonRef = React.createRef();
 
-    this.lastFormatRuleIndex = formatSettings.length;
+    this.lastFormatRuleIndex = formatSettings?.length ?? 0;
```
mofojed, line 319: Shouldn't need to check since formatSettings should never be null

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -232,27 +332,37 @@ export class FormattingSectionContent extends PureComponent {
...
+
+  isDuplicateRule(rule: FormatterItem): boolean {
     const { formatSettings } = this.state;
-    return formatSettings.some(
-      item =>
-        item.id !== rule.id &&
-        item.columnName === rule.columnName &&
-        item.columnType === rule.columnType
+    return (
+      formatSettings?.some(
```
mofojed, line 356: formatSettings should never be null

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -27,67 +33,161 @@ import {
...
+    defaultDecimalFormatOptions: FormatOption;
+    defaultIntegerFormatOptions: FormatOption;
+    showTimeZone: boolean;
+    showTSeparator: boolean;
+    timeZone: string;
+  };
+}
+
+interface FormattingSectionContentState {
+  formatSettings?: FormatterItem[];
```
mofojed, line 80: formatSettings should never be undefined (along with a few others of these)

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -505,14 +623,15 @@ export class FormattingSectionContent extends PureComponent {
...
-      .filter(FormattingSectionContent.isFormatRuleValidForSave)
-      .map(FormattingSectionContent.removeFormatRuleExtraProps);
+    const formatter =
+      formatSettings
+        ?.filter(FormattingSectionContent.isFormatRuleValidForSave)
+        .map(FormattingSectionContent.removeFormatRuleExtraProps) ?? [];
 
     const { settings, saveSettings } = this.props;
-    const newSettings = {
+    const newSettings: Partial<WorkspaceSettings> = {
```
mofojed, line 632: Should just be `WorkspaceSettings`, not `Partial`.
```suggestion
    const newSettings: WorkspaceSettings = {
```

mofojed (changes requested):

packages/code-studio/src/settings/FormattingSectionContent.tsx
SparkyFnay: The `FormattingSectionContent.tsx` is used in `SettingsMenu.tsx` with only one prop passed in, which is `scrollTo`. Most of these props also don't have default values. This would mean the props are optional?

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -761,4 +874,6 @@ export default connect(mapStateToProps, {
   setActiveTool: setActiveToolAction,
   updateDashboardData: updateDashboardDataAction,
   updateWorkspaceData: updateWorkspaceDataAction,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
mofojed, line 878: As mentioned in #ddl-team-web, you shouldn't need these. Instead declare AppMainContainer to have the RouteComponentProps:
```
export class AppMainContainer extends Component<
  AppMainContainerProps & RouteComponentProps,
  AppMainContainerState
> {
```

Then you'll also need to cast the props from redux correctly, e.g.:
```
const mapStateToProps = (state: RootState) => ({
  activeTool: getActiveTool(state),
  dashboardData: getDashboardData(state, DEFAULT_DASHBOARD_ID) as AppDashboadData,
  layoutStorage: getLayoutStorage(state) as LayoutStorage,
  plugins: getPlugins(state),
  session: getDashboardSessionWrapper(state, DEFAULT_DASHBOARD_ID).session,
  sessionConfig: getDashboardSessionWrapper(state, DEFAULT_DASHBOARD_ID).config,
  user: getUser(state) as User,
  workspace: getWorkspace(state),
});

```

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -49,44 +70,119 @@ import {
...
-  static handleWindowBeforeUnload(event) {
+type InputFileFormat =
+  | string
+  | number[]
+  | Uint8Array
+  | ArrayBuffer
+  | Blob
+  | NodeJS.ReadableStream;
+
+export type DashboadData = {
```
mofojed, line 108: I was so confused why this wasn't conflicting with the `DashboardData` that already exists in redux - it's because it's misspelled `DashboadData` (missing `r`).

I think just name this `AppDashboardData` instead to be clear.

mofojed (changes requested):
I'll have more comments, just wanted these immediate ones to go out

packages/code-studio/src/settings/FormattingSectionContent.test.tsx
```diff
@@ -111,7 +115,10 @@ describe('default decimal formatting', () => {
       }),
     });
 
-    userEvent.click(container.querySelector('.btn-reset-decimal'));
+    const element = container.querySelector('.btn-reset-decimal');
```
mofojed, line 120: Add a `test-id` to the button and then just use `screen.getByTestId('btn-reset-decimal')` - that's preferred to doing this.

packages/code-studio/src/settings/FormattingSectionContent.test.tsx
```diff
@@ -163,7 +172,10 @@ describe('default integer formatting', () => {
       }),
     });
 
-    userEvent.click(container.querySelector('.btn-reset-integer'));
+    const element = container.querySelector('.btn-reset-integer');
```
mofojed, line 177: Add `data-testid`

packages/dashboard-core-plugins/src/panels/index.ts
```diff
@@ -1,4 +1,5 @@
 export { default as ChartPanel } from './ChartPanel';
+export * from './ChartPanel';
```
mofojed, line 2: Just a random note, I'm surprised this works correctly/exports the correct `ChartPanel`.
In `ChartPanel`, we've got `export class ChartPanel`, but then the export at the end is a default export that is a redux connected wrapped component. And for some reason it looks like that takes precedent? Not sure why that works but it seems to.

If we wanted to be really explicit, we'd probably export a named version of the connected component (e.g.
```
export const ConnectedChartPanel = connect(...)(ChartPanel);

export default ConnectedChartPanel;
```
), but not required for this change.

mofojed (changes requested):

packages/code-studio/src/plugins/internal/ExampleAppPlugin.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable */
+// @ts-nocheck
```
mofojed, line 2: These files in `plugins/internal` should not be converted to tsx as we dynamically import them. Remove the `@ts-nocheck`

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -243,7 +253,11 @@ export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
       isDisconnected: false,
       isLoading: false,
       isLoaded: false,
-      isLinked: metadata && metadata.settings && metadata.settings.isLinked,
+      isLinked:
+        metadata &&
+        isChartPanelTableMetadata(metadata) &&
+        metadata.settings &&
```
mofojed, line 259: Don't need to `metadata.settings` check since it's already verified `metadata.settings !== undefined` in `isChartPanelTableMetadata`

mofojed (approved):

PR #718, opened 2022-08-12
Heap size UI

### LINKED ISSUE #715: Display heap usage in console
Implement heap usage indicator in console information bar per spec
image-20220713-175907
note: change 60 sec to 10 minutes for console, 60 sec for Query Monitor, set behaviour as component prop.

Should be built as a component that can accept the worker as an input, so that the same component can be re-used in enterprise for the Query Monitor (https://deephaven.atlassian.net/browse/DH-9304). Note QM doesn't store history and just displays while open.

SparkyFnay:
Fixes #715 
Added Heap size monitoring UI Component
Added Heap Size Display in Console


packages/console/src/console-history/SizeUsageUI.scss
```diff
@@ -0,0 +1,57 @@
+@import '../../../components/scss/custom.scss';
+
+.max-memory {
+  position: relative;
+  background-color: $gray-900;
+  width: 4rem;
+  vertical-align: middle;
+  text-align: center;
+  border: $black;
+  z-index: 1;
```
dsmmcken, line 10: add user-select: none, so we have default cursor, and 4.00GB isn't text-selectable.

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+                  range: [0, 1],
+                  linecolor: ChartTheme.linecolor,
+                  linewidth: 2,
+                  mirror: true,
+                },
+              }}
+            />
+          </div>
+          <div className="heap-utilisation-text">
+            <h6>% utilisation over {Math.round(monitorDuration / 60)} min.</h6>
```
dsmmcken, line 203: Use american spelling of utilization

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+            `${inUseGB.toFixed(decimalPlace)} of ${maxHeapGB.toFixed(
+              decimalPlace
+            )} GB`,
+            true
+          )}
+          {getRow('Free:', `${freeMemoryGB.toFixed(decimalPlace)} GB`)}
+          {getRow('Total:', `${totalHeapGB.toFixed(decimalPlace)} GB`)}
+          {getRow('Max:', `${maxHeapGB.toFixed(decimalPlace)} GB`)}
+          <div className="heap-plot">
+            <Plot
```
dsmmcken, line 160: What is this plotting? Doesn't seem to make sense, it's going down as my usage is going up. Should be in use/max.

image

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
```diff
@@ -302,6 +303,18 @@ export class ConsolePanel extends PureComponent {
...
                   <div>{ConsoleConstants.LANGUAGE_MAP.get(language)}</div>
+                  <div>&nbsp;</div>
+                  <div>
+                    <SizeUsageUI
+                      connection={connection}
+                      UIParams={{
+                        defaultUpdateInterval: 10000,
+                        hoverUpdateInterval: 3000,
+                      }}
+                      monitorDuration={60 * 10}
```
dsmmcken, line 314: Now I wonder if 10 minutes is too long, graph moves so slow. Maybe 5?

Thoughts?

dsmmcken:
Visually, looks great.

dsmmcken (changes requested):
image

css of how you do the border might not be quite right, it's missing the right border when total is 100%

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
SparkyFnay: Windows task manager shows performance over 60s, so maybe we move to 5 minutes and double the poll rate?

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const toDecimalPlace = (num: number, dec: number) =>
+    Math.round(num * 10 ** dec) / 10 ** dec;
+
+  const decimalPlace = 2;
+  const GbToByte = 1024 ** 3;
+
+  const freeMemoryGB = toDecimalPlace(freeMemory / GbToByte, decimalPlace);
+  const totalHeapGB = toDecimalPlace(totalHeapSize / GbToByte, decimalPlace);
+  const maxHeapGB = toDecimalPlace(maxHeapSize / GbToByte, decimalPlace);
+  const inUseGB = totalHeapGB - freeMemoryGB;
```
mattrunyon, line 100: Why are you trimming to 2 decimal places here and then using `toFixed` later? Is it because the `toFixed` doesn't round properly?

You could instead use a number formatter which is built in now. See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) and [stack overflow example](https://stackoverflow.com/questions/1726630/formatting-a-number-with-exactly-two-decimals-in-javascript)

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+      className={classNames(`heap-usage-info-row`, {
+        'heading-bottom-border': bottomBorder,
+      })}
+    >
+      <div>
+        <h6
+          style={{
+            fontWeight: 'bold',
+          }}
+        >
```
mattrunyon, line 113: @dsmmcken Do we have Sass stuff for the font weight? Should we be using `fontWeight: 'bold'` or a numeric value instead?

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+          {text}
+        </h6>
+      </div>
+      <div style={{}}>
```
mattrunyon, line 117: Why the empty style?

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+        <h6>{size}</h6>
```
mattrunyon, line 118: I'm not sure if this should be a `h6` tag as well. While our app isn't the best from an accessibility standpoint (and we don't enforce accessibility things), it is not great to put the same tag for different levels of importance in the dom. Visually it's fine, but semantically it's probably wrong.

Is the styling of h6 even different from a normal div or span?

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+      </div>
+    </div>
+  );
+
+  const timestamps = historyUsage.current.time;
+  const usages = historyUsage.current.usage;
```
mattrunyon, line 124: I would be consistent with your variable names. Basically you should be able to just destructure your object here

```
const { timestamps, usages } = historyUsage.current;
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
```diff
@@ -272,7 +272,8 @@ export class ConsolePanel extends PureComponent {
       unzip,
     } = this.props;
     const { consoleSettings, error, objectMap } = this.state;
-    const { config, session } = sessionWrapper;
+    // eslint-disable-next-line react/prop-types
```
mattrunyon, line 275: Why is this rule disabled here? Seems you should be providing a prop-type for this

packages/console/src/console-history/index.tsx
```diff
@@ -3,3 +3,4 @@ export { default as ConsoleHistoryItem } from './ConsoleHistoryItem';
 export { default as ConsoleHistoryItemResult } from './ConsoleHistoryItemResult';
 export { default as ConsoleHistoryResultErrorMessage } from './ConsoleHistoryResultErrorMessage';
 export { default as ConsoleHistoryResultInProgress } from './ConsoleHistoryResultInProgress';
+export { default as SizeUsageUI } from './SizeUsageUI';
```
mattrunyon, line 6: This isn't part of console history really. More part of status bar. The history is the part between status bar and input

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
mattrunyon, line 314: I would flip the math here for readability. `10 * 60` since it's 10 minutes and then you're multiplying by 60 as a conversion to seconds. So for millis then `10 * 60 * 1000`

Basically the 10 is what we care about, but reading the code we see 60 first

packages/jsapi-shim/src/dh.types.ts
```diff
@@ -865,7 +865,17 @@ export interface TableMap extends Evented {
   getTable(key: object): Promise<Table>;
 }
 
+export interface WorkerHeapInfo {
+  readonly maximumHeapSize: number;
+  readonly freeMemory: number;
+  readonly totalHeapSize: number;
+  getMaximumHeapSize: () => number;
+  getFreeMemory: () => number;
+  getTotalHeapSize: () => number;
```
mattrunyon, line 874: Looking at the API code, these getters don't exist in this format. GWT will convert the Java into a JS getter since it has `@JSProperty` decorator

So it's just the readonly values. If you were to log the result of `getWorkerHeapInfo()` then it should show you the response which I think will just have the 3 properties

mattrunyon (changes requested):

packages/console/src/console-history/SizeUsageUI.scss
```diff
@@ -0,0 +1,57 @@
+@import '../../../components/scss/custom.scss';
+
+.max-memory {
+  position: relative;
+  background-color: $gray-900;
+  width: 4rem;
```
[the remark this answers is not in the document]
SparkyFnay: Yep, it has enough width to display 999.9 GB per spec

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+    setMaxHeapSize(usage.maximumHeapSize);
+    setTotalHeapSize(usage.totalHeapSize);
+  };
+
+  useEffect(
+    function setUsageUpdateInterval() {
+      const fetchAndUpdate = async () => {
+        const newUsage = await connection.getWorkerHeapInfo();
+
+        if (bgMonitoring || hover) {
```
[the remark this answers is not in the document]
SparkyFnay: The usage is not discarded though, it's passed as a attribute to the setState call later. I moved the setState call to the top to be more clear.

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+          const time = Date.now();
+
+          const past = historyUsage.current;
+          while (
+            past.time.length !== 0 &&
+            time - past.time[0] > monitorDuration * 1000 * 1.5
+          ) {
+            past.time.shift();
+            past.usage.shift();
+          }
```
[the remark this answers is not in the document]
SparkyFnay: Yeah, so the most shifting we'll do is 3 at a time, that happens when the user 'unhovers' over the usage info.

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+        const newUsage = await connection.getWorkerHeapInfo();
+
+        if (bgMonitoring || hover) {
+          const usage = newUsage.freeMemory / newUsage.maximumHeapSize;
+          const time = Date.now();
+
+          const past = historyUsage.current;
+          while (
+            past.time.length !== 0 &&
+            time - past.time[0] > monitorDuration * 1000 * 1.5
```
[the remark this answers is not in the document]
SparkyFnay: It's just a buffer in the historyUsage array so that the graph produced isn't cut short at the beginning

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+          past.time.push(time);
+          past.usage.push(usage);
+
+          historyUsage.current = past;
+        } else {
+          historyUsage.current = { time: [], usage: [] };
+        }
+
+        setUsage(newUsage);
```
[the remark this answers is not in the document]
SparkyFnay: Yes, the background monitoring is only for the graph.

packages/console/src/console-history/SizeUsageUI.tsx
```diff
@@ -0,0 +1,211 @@
...
+      };
+      fetchAndUpdate();
+
+      const updateUsage = setInterval(
+        fetchAndUpdate,
+        hover ? hoverUpdateInterval : defaultUpdateInterval
+      );
```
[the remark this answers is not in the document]
SparkyFnay: background monitoring is only for the graph, i.e., the historyUsage array. The memory usage data still needs to be updated at regular intervals regardless of bgmonitoring

packages/dashboard-core-plugins/src/panels/ConsolePanel.jsx
SparkyFnay: So, I'm expecting the dashboard-core-plugins conversion to be merged first, so there wouldn't be a point in adding prop-type when it's just gonna get removed when I merge it

dsmmcken (approved):
Approved for ux only

packages/redux/src/actions.ts
```diff
@@ -42,7 +41,9 @@ export const setWorkspaceStorage: PayloadActionCreator<WorkspaceStorage> = works
   payload: workspaceStorage,
 });
 
-export const setCommandHistoryStorage: PayloadActionCreator<CommandHistoryStorage> = commandHistoryStorage => ({
+export const setCommandHistoryStorage: PayloadActionCreator<
+  Record<string, unknown>
+> = commandHistoryStorage => ({
```
mattrunyon, line 46: Why is this changing in this PR?

mattrunyon (changes requested):

packages/redux/src/actions.ts
dsmmcken: I guess console was creating a circular reference in the tsconfig, so can't be imported.

packages/redux/src/actions.ts
```diff
@@ -42,7 +41,7 @@ export const setWorkspaceStorage: PayloadActionCreator<WorkspaceStorage> = works
   payload: workspaceStorage,
 });
 
-export const setCommandHistoryStorage: PayloadActionCreator<CommandHistoryStorage> = commandHistoryStorage => ({
+export const setCommandHistoryStorage: PayloadActionCreator<unknown> = commandHistoryStorage => ({
```
mattrunyon, line 44: I still don't understand why redux is being changed in this PR. Nothing touches/modifies/uses this code in this PR. If it were broken elsewhere, shouldn't it be causing issues elsewhere?

Looking at this locally, the issue lies with chart. It should not import redux. Redux should only be consumed by our apps I think

This change also removes type safety from this action

mattrunyon (changes requested):

packages/redux/src/actions.ts
mofojed: Make a separate ticket.

`@deephaven/redux` should not have dependencies on other packages is the direction we should go in. In this case, the redux actions registered should be exported by the package that registers them (e.g. `dashboard-core-plugins`, or `dashboard`)

dsmmcken: cool cool cool... so is this approved then or what?

mofojed: Ticket created: https://github.com/deephaven/web-client-ui/issues/730

mattrunyon (approved):

Issue #703, opened 2022-08-03
Configuration to disable front-end data exfiltration


devinrsmith:
For certain use-cases and environments, it may be advantageous to disable the easy ways of getting large amounts of data out via the front-end. Namely, Download to CSV, and Copy operations.

While this is ostensibly a front-end feature, the configuration should likely be at the server, and the server can inform the client about the the capabilities it has. (That said, the server needs to be able to enforce this limitation itself, so this is a broader issue with respect to server authorization and configuration.)

devinrsmith:
In the absence of existing server APIs, the client does have the opportunity to define this configuration for its own purposes via custom object plugins.

devinrsmith:
Strawman:

```python
web_ui_configuration = WebUiConfiguration({
  'csv_download': false,
  'copy': false,
})
```

niloc132:
Probably should be part of https://github.com/deephaven/deephaven-core/issues/2668 to read the config settings, and offer rate limiting/etc examples in the hooks to be written for https://github.com/deephaven/deephaven-core/issues/2563 to limit scraping data.

devinrsmith:
I've got a prototype appmode + object type plugin that I think may be nice. It adds two application fields:

```
Created: typed_ticket {
  ticket {
    ticket: "a/io.deephaven.web.WebApplication/f/config"
  }
  type: "io.deephaven.web.WebConfig"
}
field_name: "config"
application_name: "Web Application"
application_id: "io.deephaven.web.WebApplication"

Created: typed_ticket {
  ticket {
    ticket: "a/io.deephaven.web.WebApplication/f/layout"
  }
  type: "io.deephaven.web.WebLayout"
}
field_name: "layout"
application_name: "Web Application"
application_id: "io.deephaven.web.WebApplication"
```

```
$ fetch-object --app-id io.deephaven.web.WebApplication --app-field config --type "io.deephaven.web.WebConfig"
{
  "csv_download" : true,
  "copy" : true
}
```

```
$ fetch-object --app-id io.deephaven.web.WebApplication --app-field layout --type "io.deephaven.web.WebLayout"
{ "todo": "some layout file" }
```

These values can be customized by the properties:

```properties
deephaven.web.csvDownload=true
deephaven.web.copy=true
deephaven.web.layoutFile=layout.json
```

I think this is potentially an extensible layer where the web UI could allows itself to be configured without needing extensive server review.

devinrsmith:
[...]

PR #719, opened 2022-08-15
WIP: Disable canCopy and canDownloadCsv

devinrsmith:
Explicit workaround for https://github.com/deephaven/web-client-ui/issues/703

Do not merge.


mofojed:
Was fixed with #904, use configuration properties to set:
```
internal.webClient.appInit.canCopy=false
internal.webClient.appInit.canDownloadCsv=false
```

PR #723, opened 2022-08-18
@deephaven/chart TypeScript Conversion

SparkyFnay:
Converted @deephaven/chart package to TypeScript


packages/chart/src/FigureChartModel.ts
```diff
@@ -546,10 +616,10 @@ class FigureChartModel extends ChartModel {
    * value for x.
    * @param {dh.Series} series The series to clean the data for
    */
-  cleanSeries(series) {
+  cleanSeries(series: Series): void {
     const { name, plotStyle } = series;
     const seriesData = this.seriesDataMap.get(name);
-    if (plotStyle === dh.plot.SeriesPlotStyle.HISTOGRAM) {
+    if (seriesData != null && plotStyle === dh.plot.SeriesPlotStyle.HISTOGRAM) {
```
mofojed, line 622: Just return early if seriesData is `null`, then you don't need the assert/if null checks here

```
const seriesData = this.seriesDataMap.get(anem);
if (seriesData == null) {
  return;
}
...
```

packages/chart/src/FigureChartModel.ts
```diff
@@ -576,21 +656,21 @@ class FigureChartModel extends ChartModel {
...
-  getFilterColumnMap() {
+  getFilterColumnMap(): FilterColumnMap {
     return new Map(this.filterColumnMap);
   }
 
-  isFilterRequired() {
+  isFilterRequired(): boolean {
     return (
-      this.oneClicks.find(oneClick => oneClick.requireAllFiltersToDisplay) !=
+      this.oneClicks.find(oneClick => oneClick.isRequireAllFiltersToDisplay) !=
```
mofojed, line 673: Should not have the `is` prefix, that gets removed by the `@JsProperty` in the API code

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -271,10 +272,16 @@ export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
     }
 
     if (settings !== prevState.settings && isFigureChartModel(model)) {
-      model.updateSettings(settings);
       this.updatePanelState();
     }
 
+    if (
+      ChartSettings !== prevProps.metadata.settings &&
```
mofojed, line 279: Not sure where this block came from, this isn't right - `prevProps.metadata.settings` is never going to change after the component is created.
The `model.updateSettings(settings)` should still be in the previous block, and this block should be deleted.

Ah - also, the `settings` in `ChartPanelState` and `PanelState` should both be `ChartModelSettings` type, not `WorkspaceSettings`. Only the `settings` in `ChartPanelProps` should be `WorkspaceSettings` type.

The prop is the workspace settings being set on the panel and is not modifiable, whereas the state settings is for the chart itself and can be modified.

mofojed (changes requested):

mofojed (approved):

mofojed (approved):

PR #737, opened 2022-08-31
Add ValidateLabelInput component

mofojed:
- ValidateLabelInput wraps an input component and provides validation hints for it
- Moved ValidateLabelInput from Enterprise to Community
- Converted ValidateLabelInput to TypeScript
- Converted Inputs in style guide to a functional component
- Added ValidateLabelInput to Inputs in style guide


packages/code-studio/src/styleguide/ValidateInputs.tsx
```diff
@@ -0,0 +1,325 @@
+import React from 'react';
```
vbabich, line 1: Unused, delete this?

packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -10,349 +10,356 @@ import {
...
-                onChange={value => {
-                  this.setState({ autoResizeTextareaValue: value });
-                }}
+        <div className="col">
+          <div className="form-group">
+            <h5>Validate Label</h5>
+            <ValidateLabelInput
+              validationError={!on ? 'Toggle is off' : undefined}
+            >
+              <UISwitch on={on} onClick={handleToggleClick} />
```
vbabich, line 289: The styling is a bit off:
- without the `labelText` prop this renders as an `inline-block` element when valid, and `inline-block` that is followed by a block `p` for the error message when invalid. Looks misaligned on the styleguide page.
- with the `labelText` there's no space between the label and the toggle switch.

vbabich: We don't use `UISwitch` with `ValidateLabelInput` anywhere at the moment, this could be a separate ticket.

vbabich (changes requested):

vbabich (approved):

PR #738, opened 2022-09-01
DH-12503 Fix TimeSlider selection issue when swapping start/end times

vbabich:


packages/components/src/TimeInput.tsx
```diff
@@ -1,9 +1,15 @@
...
+  useRef,
+  useState,
+} from 'react';
 import Log from '@deephaven/log';
 import { TimeUtils } from '@deephaven/utils';
 import MaskedInput, {
   DEFAULT_GET_PREFERRED_REPLACEMENT_STRING,
-  SelectionSegment,
+  SelectionSegment as MaskedInputSelectionSegment,
 } from './MaskedInput';
```
mofojed, line 14: You don't need to rename this export, you can just do a separate export statement after, e.g.:
```suggestion
  SelectionSegment,
} from './MaskedInput';

export type { SelectionSegment } from './MaskedInput';

```

packages/components/src/TimeInput.tsx
```diff
@@ -15,16 +21,24 @@ type TimeInputProps = {
...
   'data-testid'?: string;
 };
 
+export type TimeInputElement = {
+  focus: () => void;
+  setSelection: (newSelection: SelectionSegment) => void;
+};
+
+export type SelectionSegment = MaskedInputSelectionSegment;
+
```
mofojed, line 37: Don't need to do this to re-export, see my other comment above

```suggestion
```

mofojed (changes requested):

dsmmcken:
that's a big snack.

mofojed (approved):

PR #743, opened 2022-09-02
Support for no console in the web UI

mofojed:
- Use `connection` to `getObject` instead of `session` (Added with https://github.com/deephaven/deephaven-core/pull/2773)
- Create an application state as described in this ticket: https://github.com/deephaven/deephaven-core/issues/2779

Tested by setting `deephaven.console.disable` to `true` in `ConsoleServiceGrpcImpl.java`.
- Tested with an existing layout that had a Console to make sure the UI didn't explode, and you could 'reset layout' to get the empty layout
- Tested in an incognito window to ensure an empty layout was loaded when console.disable was true
- Tested in an incognito to ensure a default console layout was loaded when console.disable was false


packages/embed-grid/src/App.tsx
```diff
@@ -62,67 +62,50 @@ function App(): JSX.Element {
...
+            process.env.REACT_APP_CORE_API_URL ?? '',
+            `${window.location}`
+          );
 
-      const types = await connection.getConsoleTypes();
+          const websocketUrl = `${baseUrl.protocol}//${baseUrl.host}`;
 
-      log.debug('Available types:', types);
+          log.debug(`Starting connection...`);
+          const connection = new dh.IdeConnection(websocketUrl);
```
vbabich, line 85: Not calling `SessionUtils.createConnection` here to avoid adding another dependency?

packages/dashboard-core-plugins/src/redux/actions.ts
```diff
@@ -9,14 +9,28 @@ import { FilterSet } from '../panels';
...
-  };
+  config: SessionConfig;
 }
+
+/**
+ * Set the connection for the dashboard specified
+ * @param id The ID of the dashboard to set the connection for
+ * @param connection The connection object to set for the dashboard
+ */
+export const setDashboardConnection = (
```
vbabich, line 28: Are we planning to use separate connections per dashboard at some point?

vbabich (dismissed):

packages/embed-grid/src/App.tsx
mofojed: Yes, SessionUtils is only in code-studio, so not importing to reduce dependencies.

packages/dashboard-core-plugins/src/redux/actions.ts
mofojed: No plan yet.

packages/embed-grid/src/App.tsx
mofojed: I could move it to jsapi-utils though I think. I'll do that.

packages/dashboard-core-plugins/src/redux/actions.ts
vbabich: Just wondering why we store connections per dashboard, not just a global one.

packages/embed-grid/src/App.tsx
mofojed: Actually nevermind - the createConnection basically just gets the websocket URL from env vars:
```
  const baseUrl = new URL(
    import.meta.env.VITE_CORE_API_URL ?? '',
    `${window.location}`
  );
```

We shouldn't do that in a lib or in jsapi-utils, so I will just keep them separate.

packages/dashboard-core-plugins/src/redux/actions.ts
mofojed: Didn't think there was a need to limit it to a global connection when could make it configurable so each dashboard has it's own.

vbabich (approved):

PR #746, opened 2022-09-07
Blank commands not added to history

### LINKED ISSUE #722: DH-12976 Blank commands should not be added to the up/down command history
If you enter blank commands, they do not get added to the command history, but while in the same session, up and down arrows cycle through the blank commands instead of skipping them.

Type a command.

press enter a bunch of times

type another command

use up and down to cycle through history

Look at the Command History panel - the blanks aren’t there.

Restart the console - the blanks are not reloaded.

https://deephaven.atlassian.net/browse/DH-12976

jason-wang24:
Fixes #722

### DIFF
--- a/packages/console/src/ConsoleInput.tsx
+++ b/packages/console/src/ConsoleInput.tsx
@@ -444,7 +444,9 @@ export class ConsoleInput extends PureComponent<
 
     const command = this.commandEditor?.getValue().trim();
     assertNotNull(command);
-    this.history.push(command);
+    if (command !== '') {
+      this.history.push(command);
+    }
     this.commandEditor?.setValue('');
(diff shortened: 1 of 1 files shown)


mofojed (changes requested):
Great work! Code looks good, a few things to change with the actual PR though:
- Add me as Reviewer. It may not always be me that reviews your code, but if you're not sure who should be reviewing it, just add me and I'll route to the correct person
- Add yourself as the Assignee. The idea is that you own the PR and are responsible for making sure it eventually gets merged.
- Add "Fixes #722 " to the description, instead of the title. When it's in the body of the description, GitHub will automatically link it

mofojed (approved):

PR #747, opened 2022-09-08
Update monaco version

mofojed:
- Update monaco to v0.34.0
- Fix up unit tests
- Fix up warning in Console.test.tsx, other mock fixes
- Use the base monaco API instead, it includes all the KeyMod and KeyCodes
- Update monaco mock
- Remove our ConsoleInput mock

We should now be able to write unit tests around Console that actually use monaco input.


jest.setup.ts
```diff
@@ -13,3 +13,20 @@ Log.setLogLevel(logLevel);
 // disable annoying dnd-react warnings
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 window['__react-beautiful-dnd-disable-dev-warnings' as any] = true;
+
+// Define the watchMedia property so we can mock out monaco properly
```
mattrunyon, line 17:
```suggestion
// Define the matchMedia property so we can mock out monaco properly
```

packages/console/package.json
```diff
@@ -43,7 +43,7 @@
     "lodash.throttle": "^4.1.1",
     "memoize-one": "^5.1.1",
     "memoizee": "^0.4.15",
-    "monaco-editor": "^0.27.0",
+    "monaco-editor": "^0.34.0",
```
mattrunyon, line 46: You checked the changelog for any breaking changes between versions?

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -2,70 +2,12 @@
...
-import 'monaco-editor/esm/vs/editor/standalone/browser/toggleHighContrast/toggleHighContrast.js';
-import 'monaco-editor/esm/vs/editor/contrib/toggleTabFocusMode/toggleTabFocusMode.js';
-import 'monaco-editor/esm/vs/editor/contrib/caretOperations/transpose.js';
-import 'monaco-editor/esm/vs/editor/contrib/unusualLineTerminators/unusualLineTerminators.js';
-import 'monaco-editor/esm/vs/editor/contrib/viewportSemanticTokens/viewportSemanticTokens.js';
-import 'monaco-editor/esm/vs/editor/contrib/wordHighlighter/wordHighlighter.js';
-import 'monaco-editor/esm/vs/editor/contrib/wordOperations/wordOperations.js';
-import 'monaco-editor/esm/vs/editor/contrib/wordPartOperations/wordPartOperations.js';
-import 'monaco-editor/esm/vs/basic-languages/typescript/typescript.contribution.js';
+import * as monaco from 'monaco-editor';
```
mattrunyon, line 8: Sweet. Does this have any negative effect on our bundle size?

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -453,20 +395,20 @@ class MonacoUtils {
...
       return (
         // eslint-disable-next-line no-bitwise
-        (keyState.metaKey && KeyMod.CtrlCmd) |
-        (keyState.shiftKey && KeyMod.Shift) |
-        (keyState.altKey && KeyMod.Alt) |
-        (keyState.ctrlKey && KeyMod.WinCtrl) |
+        (keyState.metaKey ? monaco.KeyMod.CtrlCmd : 0) |
+        (keyState.shiftKey ? monaco.KeyMod.Shift : 0) |
+        (keyState.altKey ? monaco.KeyMod.Alt : 0) |
+        (keyState.ctrlKey ? monaco.KeyMod.WinCtrl : 0) |
```
mattrunyon, line 401: Just changing for readability?

mattrunyon (changes requested):

packages/console/package.json
mofojed: I did - there were a few breaking changes...
- `KeyMod` location changed, it was no longer in standaloneEditor
- `getModeId()` changed to `getLanguageId()`
- KeyCodes were renamed from `KEY_A` to `KeyA` etc
I think that was it

packages/console/src/monaco/MonacoUtils.ts
mofojed: Languages are now lazily loaded, so I don't think it should. Also the link to the "features" that I had before no longer works, since monaco-web-plugin-packager has been rolled into the main monaco-editor repo.

That being said, I tried building and then loading it up locally, the JS that gets loaded before:
image

After:
image

So an increase of 5.6MB to 6.1MB. I'll see if I can pare that back down (I'll need to find the full features list in Monaco-editor again since the webpack link is gone).

mattrunyon: I think it's fine unless we have a good reason to spend time shrinking the main app bundle size. 500kB (1 time load since it should be cached) is not bad IMO for our app. Not sure it's worth the dev time to cut that down at the moment and I think the single import for monaco is a nice improvement to the code

dsmmcken: We should only care about bundle size for embedded versions and the js api itself. If it's the full app, it doesn't really matter.

mofojed: Embedded grid has an increase in size as well, about 0.5mb (~15% increase):

Before the update (3.4mb):
image

With this update (3.9mb):
image

I can see about reducing the size but I do like not having to import every feature.

mofojed: @dsmmcken  Looking into it a bit further, I don't know if I can actually reduce the size much - seems like the size increase is simply with the monaco update, not that we're importing a whole bunch of new features.

For example, using `npx source-map-explorer ` to compare, here's before:
image

And after the update:
image

As you can see it is really spread out over all parts of the monaco package; 100kb increase in base, 50kb increase in platform, 100kb increase in contrib, 80kb to common, etc. I'd really need to dig into which areas we can pull out, as I don't see any obvious ones (perhaps the diffEditorWidget, but that was there prior).

dsmmcken: Okay. Something we could consider in future is pulling monaco from embed grid entirely. I think it's just the syntax highlighter for custom column editor.

mofojed: It's also for the cell overflow modal now too (colourizing and displaying the results)

mofojed: Added a ticket: https://github.com/deephaven/web-client-ui/issues/750
We could also lazy load those features and that would split up the package size a bit, so that's an option. E.g. Don't load monaco-editor until you open a cell overflow or custom column editor.

dsmmcken (approved):

mattrunyon (approved):

PR #756, opened 2022-09-15
Date/DateTime inputs

vbabich:
- Add DateTime input
- Update Date input, remove depencency on jsapi-shim
- Update MaskedInput to allow empty value

Posting this MR in v0.13 branch as the current main isn't compatible with DHE


packages/components/src/DateInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+    return (
+      <div className="d-flex flex-row align-items-center">
+        <MaskedInput
+          ref={ref}
+          className={classNames(className)}
+          example={EXAMPLES}
+          getNextSegmentValue={getNextSegmentValue}
+          getPreferredReplacementString={
+            DEFAULT_GET_PREFERRED_REPLACEMENT_STRING
+          }
```
mofojed, line 96: Not necessary, this is the default already set in `MaskedInput`
```suggestion
```

packages/components/src/DateTimeInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return [dateTimeMillis, micros, nanos].filter(v => v !== '').join('\u200B');
+}
+
+const removeSeparators = (value: string) => value.replace(/\u200B/g, '');
+
+const EXAMPLES = [addSeparators(DEFAULT_VALUE_STRING)];
+
+// Forward ref causes a false positive for display-name in eslint:
+// https://github.com/yannickcr/eslint-plugin-react/issues/2269
+// eslint-disable-next-line react/display-name
```
mofojed, line 43:
```suggestion
```

packages/components/src/DateTimeInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+const DateTimeInput = React.forwardRef<HTMLInputElement, DateTimeInputProps>(
+  (props: DateTimeInputProps, ref) => {
+    const {
+      className = '',
+      onChange = () => false,
+      defaultValue = '',
+      onFocus = () => false,
+      onBlur = () => false,
```
mofojed, line 51:
```suggestion
      className = '',
      onChange = () => undefined,
      defaultValue = '',
      onFocus = () => undefined,
      onBlur = () => undefined,
```

packages/components/src/DateTimeInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+        selectionStart === 20 ||
+        selectionStart === 24 ||
+        selectionStart === 28
+      ) {
+        // Milli, micro, and nanosecond input
+        return getNextNumberSegmentValue(delta, segmentValue, 0, 999, 3);
+      }
+
+      return segmentValue;
+    }
```
mofojed, line 110: Pull out of component, and export either from this file or a separate `DateTimeUtils.js` file so this can be shared with `DateTime`

packages/components/src/DateTimeInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+        log.debug('handleChange', newValue);
+        setValue(newValue);
+        onChange(removeSeparators(newValue));
+      },
+      [onChange]
+    );
+
+    function handleSelect(newSelection: SelectionSegment) {
+      setSelection(newSelection);
+    }
```
mofojed, line 123: This function isn't really necessary - just pass `setSelection` in to `onSelect`

packages/components/src/DateTimeInput.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }
+);
+
+DateTimeInput.defaultProps = {
+  className: '',
+  onChange: () => false,
+  defaultValue: '',
+  onFocus: () => false,
+  onBlur: () => false,
+  'data-testid': undefined,
```
mofojed, line 156:
```suggestion
  onChange: () => undefined,
  defaultValue: '',
  onFocus: () => undefined,
  onBlur: () => undefined,
  'data-testid': undefined,
```

packages/components/src/TimeInput.test.tsx
```diff
@@ -149,6 +151,64 @@ describe('select and type', () => {
...
+    input.setSelectionRange(1, 1);
+
+    userEvent.type(input, '{backspace}');
+
+    expect(input.value).toEqual(``);
+
+    unmount();
+  });
+
+  it('existing invalid behaviors that might need to be fixed', () => {
```
mofojed, line 197: Interesting test :P

packages/components/src/TimeInput.test.tsx
```diff
@@ -149,6 +151,64 @@ describe('select and type', () => {
...
+    // Expected: '20:34:56'?
+    testSelectAndType(1, '5{arrowleft}2', `25:34:56`);
```
mofojed, line 199: As mentioned above I think `20:34:56` makes more sense:
- Start with `12:34:56`, selection on the `2`
- Enter `5`, changes to `05:34:56`, selection on 3
- Press left, selection changes back to first segment
- Enter `2`, sees that `25` is not valid, so changes the second value to `20`
Honestly seems like such an edge case though that I'd almost rather not bother.

packages/components/src/TimeInput.test.tsx
```diff
@@ -149,6 +151,64 @@ describe('select and type', () => {
...
+
+    // Fill in with zeros when skipping positions. Expected: '03:34:56'
+    testSelectAndType(0, '{backspace}3', `${FIXED_WIDTH_SPACE}3:34:56`);
```
mofojed, line 202: Same, it's an edge case, but I think `03:34:56` should be more correct.

packages/components/src/TimeInput.test.tsx
```diff
@@ -149,6 +151,64 @@ describe('select and type', () => {
...
+
+    // Not sure it's ok to skip to the next section when the input isn't valid for the current section
+    // Expected: '03:34:56'?
+    testSelectAndType(0, '35', `03:54:56`);
```
mofojed, line 206: This should be `03:54:56`, this is correct. Yes it should jump to the next section if the previous section is complete, which it is in this case (since you put `3`, which can only go in the second spot, finishing off that segment).

packages/components/src/TimeInput.test.tsx
```diff
@@ -149,6 +151,64 @@ describe('select and type', () => {
...
+
+    // Should validate whole value
+    // Expected: '12:34:11'
+    testSelectAndType(9, '11`"();', `12:34:11\`"();`);
```
mofojed, line 210: Why can I even type those special chars after the fact? Yea it should just be `'12:34:11'`

mofojed (changes requested):

packages/components/src/TimeInput.test.tsx
vbabich: Yea, by existing I mean these behaviors were not introduced by my changes. Some need to be fixed, some dismissed. Didn't want to scope creep the original issue.

packages/components/src/TimeInput.test.tsx
vbabich: MaskedInput was checking for a substring match against the validation regex. As long as the value contained a valid date, it treated any additional chars as valid.

packages/components/src/TimeInput.test.tsx
vbabich: Changed to 'existing edge cases'

packages/components/src/MaskedInput.tsx
```diff
@@ -234,7 +234,7 @@ const MaskedInput = React.forwardRef<HTMLInputElement, MaskedInputProps>(
       checkValue: string,
       cursorPosition = checkValue.length
     ): boolean {
-      const patternRegex = new RegExp(pattern);
+      const patternRegex = new RegExp(`^${pattern}$`);
```
mofojed, line 237: Good catch... side note I wonder if we should just allow the regex pattern to be passed in instead of just taking a string, or if we should just include the `^` and `$` in the string that's passed in instead of implicitly adding it here... but I think it's fine

mofojed (approved):

packages/components/src/MaskedInput.tsx
vbabich: I wasn't sure about that too, ended up updating it here to fix the issues with all dependent components at once. If this doesn't work for some use case we can change it in the future.

PR #758, opened 2022-09-16
Add path information and breadcrumb interactivity to NewItemModal

### LINKED ISSUE #751: Add path information of current focused/selected item to FileExplorerPanel
Related to #734. This would give NewItemModal the information necessary to render the selected FileExplorer path rather than the root directory.

jason-wang24:
Fixes #751


dsmmcken:
✅ jason-wang24
❌ jason Wang

Guessing some commits were under a different user email?

jason-wang24:
> ✅ jason-wang24 ❌ jason Wang
> 
> Guessing some commits were under a different user email?

Yeah, my local git config email is jasonwang@deephaven.io. Not sure if that is causing issues or how I should go about signing the CLA

jason-wang24:

packages/dashboard-core-plugins/package.json
```diff
@@ -44,6 +44,7 @@
     "@deephaven/log": "file:../log",
     "@deephaven/redux": "file:../redux",
     "@deephaven/utils": "file:../utils",
+    "@deephaven/storage": "file:../storage",
```
mofojed, line 47: You need to run `npm install` and commit the modifications to `package-lock.json` as a result of this change.

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    userEvent.click(screen.getAllByRole('listitem')[itemIndex], options);
+  }
+
+  async function getNewItemModal(): Promise<HTMLElement> {
+    userEvent.click(screen.getByRole('button', { name: 'New folder' }));
+    const foundModal = await screen.findAllByRole('dialog');
+    expect(foundModal).toHaveLength(1);
+
+    return foundModal[0];
+  }
```
mofojed, line 94: I don't think this actually needs to be async - you can just use `getByRole('dialog')` and it succeeds.

```suggestion
  function getNewItemModal(): HTMLElement {
    userEvent.click(screen.getByRole('button', { name: 'New folder' }));
    return screen.getByRole('dialog');
  }
```

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+  it('selects directory correctly', async () => {
+    clickItem(0);
+    const NewItemModal = await getNewItemModal();
+
+    const foundButtons = await findAllByRole(NewItemModal, 'button');
+    const buttonContent = foundButtons.map(button => button.innerHTML);
+
+    expect(buttonContent).toContain('root');
+    expect(buttonContent).toContain('testdir0');
+    for (let i = 1; i < 5; i += 1) {
```
mofojed, line 116: dirs.length instead of 5
```suggestion
    for (let i = 1; i < dirs.length; i += 1) {
```

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    const foundButtons = await findAllByRole(NewItemModal, 'button');
+    const buttonContent = foundButtons.map(button => button.innerHTML);
+
+    expect(buttonContent).toContain('root');
+    for (let i = 0; i < dirs.length; i += 1) {
+      expect(buttonContent).not.toContain(`testdir${i}`);
+    }
+  });
+
+  it('selects directory correctly through arrow keys', async () => {
```
mofojed, line 147: Nice

packages/file-explorer/src/FileList.tsx
```diff
@@ -441,9 +444,23 @@ export const FileList = (props: FileListProps): JSX.Element => {
...
+          pathGetter?.(selectedItems[0]);
+        } else {
+          const root: FileStorageItem = {
+            type: 'directory',
+            id: '/',
+            filename: '/',
+            basename: '',
+          };
+          pathGetter?.(root);
+        }
```
mofojed, line 460: Instead of making a new prop `pathGetter` and passing that down, you should instead pipe `onSelectionChange` through so it gets back up to `FileExplorerPanel`.
That'll mean adding the prop to `FileListContainer`, and then in `FileListContainer.handleSelectionChange`, calling `onSelectionChange` after it calls `setSelectedItems`. 

Then all this logic goes into `FileExplorerPanel`, where it decides whether to update the path if there's only one selected item and it's a dir.

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
```diff
@@ -169,8 +169,12 @@ export class FileExplorerPanel extends React.Component<
     });
   }
 
-  getPath(file: FileStorageItem): void {
-    const path = FileUtils.makePath(file.filename);
+  handleSelectionChange(selectedItems: FileStorageItem[]): void {
+    let path = '/';
+    if (selectedItems.length === 1 && selectedItems[0].type === 'directory') {
```
mofojed, line 174: I don't think you need to bother checking the type here - assume you want to create a directory in the same directory that is selected, whether that is a dir or a file.

mofojed (changes requested):
Looks good! Just one minor comment.

mofojed (approved):

Issue #698, opened 2022-08-02
Clean up Golden Layout types


mofojed:
Some of the types in Golden Layout are confusing/mismatched. For example, `ComponentConfig` has `componentName`, but `ReactComponentConfig` does not; however, `config.componentName` always seems to be set to `lm-react-component` for react components.
Verify that is the case and clean up the types.

PR #759, opened 2022-09-16
Covert golden-layout to ES Modules

mattrunyon:
This converts golden-layout to ESM which should help with #698

Removes the need for global jquery. Only golden-layout requires jquery and it is now included as a module

Simplifies our configs some since we don't need an exception for golden-layout imports any more


packages/golden-layout/src/config/defaultConfig.js
```diff
@@ -1,4 +1,4 @@
-lm.config.defaultConfig = {
+const defaultConfig = {
```
mofojed, line 1: Freeze on export

packages/golden-layout/src/config/index.js
```diff
@@ -0,0 +1,7 @@
+import defaultConfig from './defaultConfig.js';
+import itemDefaultConfig from './ItemDefaultConfig.js';
+
+export default {
```
mofojed, line 4: Freeze

packages/golden-layout/src/controls/BrowserPopout.js
```diff
@@ -1,3 +1,6 @@
+import $ from 'jquery';
+import utils from '../utils/index.js';
```
mofojed, line 2: Don't need /index.js

```suggestion
import utils from '../utils';
```

packages/golden-layout/src/controls/DragProxy.js
```diff
@@ -1,3 +1,6 @@
+import $ from 'jquery';
+import utils from '../utils/index.js';
```
mofojed, line 2: Don't need `/index.js`
```suggestion
import utils from '../utils';
```

packages/golden-layout/src/errors/ConfigurationError.js
```diff
@@ -0,0 +1,11 @@
...
+  Error.call(this);
+
+  this.name = 'Configuration Error';
+  this.message = message;
+  this.node = node;
+};
+
+ConfigurationError.prototype = new Error();
+
+export default ConfigurationError;
```
mofojed, line 11: Ugh. I guess if it works don't fix it? Would love to clean up/modernize all this code in another ticket (like there's a `utils.extend` function, and this, when we should just do `class ConfigurationError extends Error`

mofojed (approved):
Loking this change!
Many of my comments are around how to clean up things further, which we can probably just put as a separate ticket in the backlog

packages/golden-layout/src/base.js
```diff
@@ -0,0 +1,17 @@
...
+const lm = {
+  config,
+  container,
+  controls,
+  errors,
+  items,
+  utils,
+};
+
+export default lm;
```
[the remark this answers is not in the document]
mattrunyon: The `/index.js` is actually required to be fully ESM compliant. ESM has no bare relative imports. Every import must include an extension unless the import file itself has no extension.

PR #761, opened 2022-09-16
Enabling Strict boolean expressions check

### LINKED ISSUE #610: Enable TS strict boolean expressions linter rule
We should enable [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) in TypeScript because it prevents accidentally doing the following

```ts
function myFunc(a: number): string {
  if (a) {
    return 'Param exists';
  }
  return 'Param undefined';
}
```

This is an error because you could have the number 0 which is falsy, but still defined. Instead, it would enforce you use `a != null` or `a > 0` if you intend that logic

I created a branch on my fork [link](https://github.com/mattrunyon/web-client-ui/tree/strict-boolean-expressions) with the rule enabled, but have not fixed the errors it creates yet

Zhou-Ziheng:
Fixes #610


packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -450,7 +456,10 @@ class IrisGridRenderer extends GridRenderer {
     }
 
     const fontWidth =
-      fontWidths.get(context.font) || GridRenderer.DEFAULT_FONT_WIDTH;
+      fontWidths.get(context.font) !== undefined
+        ? fontWidths.get(context.font)
+        : GridRenderer.DEFAULT_FONT_WIDTH;
+    assertNotNull(fontWidth);
```
mofojed, line 462:
```suggestion
      fontWidths.get(context.font) ?? GridRenderer.DEFAULT_FONT_WIDTH;
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -346,10 +346,14 @@ class IrisGridTableModel extends IrisGridTableModelTemplate<Table, UIRow> {
         const filter = column.filter().eq(filterValue);
         columnFilters.push(filter);
       }
-      return columnFilters.reduce((agg, curr) => (agg ? agg.and(curr) : curr));
+      return columnFilters.reduce((agg, curr) =>
+        agg != null ? agg.and(curr) : curr
+      );
```
mofojed, line 351:
```suggestion
      return columnFilters.reduce((agg, curr) => agg?.and(curr) ?? curr);
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -346,10 +346,14 @@ class IrisGridTableModel extends IrisGridTableModelTemplate<Table, UIRow> {
...
     });
 
-    const filter = filters.reduce((agg, curr) => (agg ? agg.or(curr) : curr));
+    const filter = filters.reduce((agg, curr) =>
+      agg != null ? agg.or(curr) : curr
+    );
```
mofojed, line 356:
```suggestion
    const filter = filters.reduce((agg, curr) => agg?.or(curr) ?? curr);
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -2041,10 +2041,14 @@ class IrisGridTableModelTemplate<
         const filter = column.filter().eq(filterValue);
         columnFilters.push(filter);
       }
-      return columnFilters.reduce((agg, curr) => (agg ? agg.and(curr) : curr));
+      return columnFilters.reduce((agg, curr) =>
+        agg != null ? agg.and(curr) : curr
+      );
```
mofojed, line 2046: Use null coalescing instead of trinary operator

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -2041,10 +2041,14 @@ class IrisGridTableModelTemplate<
...
     });
 
-    const filter = filters.reduce((agg, curr) => (agg ? agg.or(curr) : curr));
+    const filter = filters.reduce((agg, curr) =>
+      agg != null ? agg.or(curr) : curr
+    );
```
mofojed, line 2051: Use null coalescing instead of trinary operator

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -1172,7 +1172,7 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
           {operator
             ? IrisGridContextMenuHandler.getOperatorAsText(operator)
             : ''}{' '}
-          &quot;{valueText || 'null'}&quot;
+          &quot;{valueText != null ? valueText : 'null'}&quot;
```
mofojed, line 1175: Use null coalescing instead

packages/iris-grid/src/sidebar/RollupRows.tsx
```diff
@@ -78,9 +78,11 @@ class RollupRows extends Component<RollupRowsProps, RollupRowsState> {
     isClone?: boolean;
     selectedCount?: number;
   }): ReactElement {
-    const text = item && item.name;
-    const badgeText = isClone ? `${selectedCount}` : undefined;
-    const className = isClone ? 'item-list-item-clone' : '';
+    const text = item != null ? item.name : undefined;
```
mofojed, line 81: I think you can just use optional chaining
```suggestion
    const text = item?.name;
```

packages/iris-grid/src/sidebar/RollupRows.tsx
```diff
@@ -440,22 +442,26 @@ class RollupRows extends Component<RollupRowsProps, RollupRowsState> {
...
-          <Button
-            kind="ghost"
-            className="btn btn-link btn-link-icon btn-delete-grouping float-right"
-            onClick={() => this.handleDeleteClicked(itemIndex)}
-          >
-            <FontAwesomeIcon icon={vsTrash} />
-          </Button>
-        )}
+        {isClone === undefined ||
+          (!isClone && (
```
mofojed, line 456: Right now the case for `isClone === undefined` will no longer render anything. Should do `&&` instead of `||`

packages/iris-grid/src/sidebar/TableSaver.tsx
```diff
@@ -91,6 +91,7 @@ export default class TableSaver extends PureComponent<
     this.cancelableSnapshots = [];
 
     // WritableStream is not supported in Firefox (also IE) yet. use ponyfillWritableStream instead
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
     this.WritableStream = window.WritableStream || ponyfillWritableStream;
```
mofojed, line 95: Just use null coalescing operator instead and don't disable the rule

packages/iris-grid/src/sidebar/TableSaver.tsx
```diff
@@ -165,7 +166,11 @@ export default class TableSaver extends PureComponent<
   cancelableSnapshots: (CancelablePromise<unknown> | null)[];
 
   // WritableStream is not supported in Firefox (also IE) yet. use ponyfillWritableStream instead
+
+  // TODO: Fix type error
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
   WritableStream = window.WritableStream || ponyfillWritableStream;
+  //  WritableStream: typeof window.WritableStream | typeof ponyfillWritableStream;
```
mofojed, line 173: TODO?
And also use null coalescing operator.

packages/iris-grid/src/sidebar/aggregations/Aggregations.tsx
```diff
@@ -234,7 +237,7 @@ const Aggregations = ({
             {DraggableItemList.renderBadge({ text: badgeText })}
             {DraggableItemList.renderHandle()}
           </div>
-          {!isClone && (
+          {isClone !== undefined && !isClone && (
```
mofojed, line 240: Should be `(isClone === undefined || !isClone) && (`

mofojed (changes requested):

packages/chart/src/ChartUtils.ts
```diff
@@ -690,7 +697,7 @@ class ChartUtils {
       // The default histfunc in plotly is 'count', but the data passed up from the API provides explicit x/y values and bins
       // Since it's converted to bar, just set the widths of each bar
       seriesData.width = [];
-      Object.assign(seriesData.marker.line, {
+      Object.assign(seriesData.marker.line ?? {}, {
```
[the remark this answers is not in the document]
SparkyFnay: `Object.assign` requires an object as the first parameter, so if `line` is undefined, then it needs to be converted to `{}`

packages/components/src/Button.tsx
```diff
@@ -133,17 +133,19 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       >
         {icon && iconElem}
         {children}
-        {tooltip && !disabled && tooltipElem}
+        {tooltip != null &&
+          (disabled === undefined || !disabled) &&
```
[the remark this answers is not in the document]
SparkyFnay: I don't think our defaultProps actually works: https://github.com/facebook/react/issues/16653

packages/dashboard-core-plugins/src/ConsolePlugin.tsx
```diff
@@ -217,14 +222,14 @@ export const ConsolePlugin = (
...
-    (fileMetadata, isPreview) => {
-      // Note: unregister event is triggered AFTER new register when switching from preview to edit mode
-      // due to the LayoutUtils implementation (new tab added before deleting the old one)
-      // This doesn't cause any issues because previews and editable files are stored in different maps,
-      // but this situation could be completely avoided by sending an event to the tab
+    (fileMetadata: FileMetadata, isPreview: boolean) => {
+      // Note: unregister event is triggered AFT ER new register when switching from preview to edit mode
+      // due to the LayoutUtils im plementation (new tab added before deleting the old one)
+      // This doesn't cause any issues because p reviews and editable files are stored in different maps,
+      // but this situation co uld be completely avoided by sending an event to the tab
```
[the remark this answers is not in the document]
SparkyFnay: Idk why, but sometimes when I open up a file, eslint basically puts red squiggly lines on the entire file, and the only way to fix it was to use `fix all auto-fixable problems` feature, which sometimes add random space to random parts of the code.

packages/iris-grid/src/sidebar/TableSaver.tsx
SparkyFnay: This is suppose to be the type declaration of `WritableStream`, which was broken from the TypeScript conversion.

packages/dashboard-core-plugins/src/ConsolePlugin.tsx
```diff
@@ -112,15 +117,15 @@ export const ConsolePlugin = (
   }, []);
 
   const getPanelIdForFileMetadata = useCallback(
-    (fileMetadata, createIfNecessary = true) => {
+    (fileMetadata: FileMetadata, createIfNecessary = true) => {
```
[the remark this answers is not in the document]
mofojed: There is an unanswered question related to this: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/57734
The definition for `useCallback` is wrong essentially, as it specifies `any` for the args... I think we could do better but passing a type into the `useCallback`, like `useCallback<(fileMetadata: fileMetadata, createIfNecessary: boolean) => string | null | undefined>( ... )`, but that's also lame.

I think the `@types/react` should be updated to just be `T extends Function`. I believe that would also force us to define types is `useCallback` (which is a good thing). But that requires upstream changes in `@types/react`. For now your work around is okay.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -24,12 +24,12 @@ const log = Log.module('MonacoUtils');
...
   interface Window {
-    MonacoEnvironment: Environment;
+    MonacoEnvironment?: Environment;
   }
 }
 
 window.MonacoEnvironment = {
-  getWorker(workerId, label) {
+  getWorker() {
     return new EditorWorker();
```
mofojed, line 33: This will probably need to be resolved with the latest main - we downgraded monaco-editor and it should just be `return editorWorker()` (I think)

mofojed (approved):
You likely need to merge the latest main and resolve conflicts

PR #762, opened 2022-09-19
DH-12970 TableInput component for Parameterized Queries

vbabich:
- Add TableInput component for displaying dynamic value list from a table column
- Add useTableListener, useTable, useTableColumn hooks
- Add disabled and onBlur props to SearchInput
- Add onBlur prop and scrollIntoView method to SelectValueList


packages/jsapi-utils/src/TableInput.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+  useEffect(() => {
+    if (updatedSelection !== null) {
+      setSelection(updatedSelection);
+      onChange(Array.from(updatedSelection));
+    }
+  }, [onChange, updatedSelection]);
+
+  const initTable = useCallback(async promise => {
+    const resolved = await promise;
```
mofojed: Should wrap in a try/catch, as the promise could reject on unmount.

packages/jsapi-utils/src/TableInput.tsx
```diff
@@ -0,0 +1,319 @@
...
+  defaultIntegerFormatOptions?: IntegerColumnFormatterOptions;
+}
+
+interface TableInputProps {
+  className?: string;
+  columnName: string;
+  settings: Settings;
+  defaultValue: Value[];
+  isInvalid?: boolean;
+  table: Promise<Table>;
```
mofojed: We're wrapping it in a promise anyway, this could be `Promise<Table> | Table`

packages/jsapi-utils/src/TableInput.tsx
```diff
@@ -0,0 +1,319 @@
...
+  onChange(items: Value[]): void;
+  onBlur?: () => void;
+}
+
+const SIZE_LIMIT = 250;
+
+const TableInput = (props: TableInputProps): ReactNode => {
```
mofojed: I know I haven't been consistent on this, but for function components and hooks, I prefer to use the `function` declaration like they do in the docs: https://reactjs.org/docs/components-and-props.html
e.g. it should be `function TableInput(props: TableInputProps): ReactNode` instead of `const TableInput = (props: TableInputProps): ReactNode`

packages/jsapi-utils/src/useTable.ts
```diff
@@ -0,0 +1,70 @@
...
+      const viewportData = columns.map(column =>
+        (detail.rows as Row[]).map(r => r.get(column))
+      );
+      setData(viewportData);
+    },
+    [columns]
+  );
+
+  const handleDisconnect = useCallback(() => {
+    setError(new Error('Table disconnected'));
```
mofojed: This is likely worthy of creating a new error, `TableDisconnectError`, so that one can check if it's a table disconnect error vs. some other generic error.

packages/jsapi-utils/src/useTable.ts
```diff
@@ -0,0 +1,70 @@
...
+  const [data, setData] = useState<unknown[][]>([]);
+  const [viewport, setViewport] = useState([firstRow, lastRow]);
+  const [error, setError] = useState<Error | null>(null);
+
+  const columns = useMemo(() => {
+    try {
+      return table?.findColumns(columnNames);
+    } catch (e) {
+      log.error(`Column not found`, e, columnNames);
+      setError(new Error('Invalid columnNames argument'));
```
mofojed: Could throw custom error type `ColumnNameError` to make it more clear.

packages/jsapi-utils/src/useTable.ts
```diff
@@ -0,0 +1,70 @@
...
+  firstRow: number,
+  lastRow: number
+): {
+  columns: Column[] | undefined;
+  data: unknown[][];
+  error: Error | null;
+  setViewport: (viewport: [number, number]) => void;
+} => {
+  const [data, setData] = useState<unknown[][]>([]);
+  const [viewport, setViewport] = useState([firstRow, lastRow]);
```
mofojed: I don't think this is right... we shouldn't need to return `setViewport`, we're already passing in the viewport `firstRow, lastRow, columnNames`. Should instead just do `const viewport = { top, bottom, columnNames }` and then base the useEffects off of that.
Otherwise we're not really getting hook behaviour - we can change the values of firstRow/lastRow and it won't have an effect, we need to call `setViewport` separately. Instead just bake it into the params.

packages/jsapi-utils/src/useTable.ts
```diff
@@ -0,0 +1,70 @@
...
+import Log from '@deephaven/log';
+import useTableListener from './useTableListener';
+
+const log = Log.module('useTable');
+
+const useTable = (
+  table: Table | undefined,
+  columnNames: string[],
+  firstRow: number,
+  lastRow: number
```
mofojed: I think params should be flipped, so that `columnNames` is optional - if it's undefined, just get _all_ the columns.

packages/jsapi-utils/src/useTable.ts
```diff
@@ -0,0 +1,70 @@
...
+  error: Error | null;
+  setViewport: (viewport: [number, number]) => void;
+} => {
+  const [data, setData] = useState<unknown[][]>([]);
+  const [viewport, setViewport] = useState([firstRow, lastRow]);
+  const [error, setError] = useState<Error | null>(null);
+
+  const columns = useMemo(() => {
+    try {
+      return table?.findColumns(columnNames);
```
mofojed: If `columnNames` is `undefined`, just `return table?.columns`

packages/jsapi-utils/src/useTableListener.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+import { Evented, EventListener } from '@deephaven/jsapi-shim';
+import Log from '@deephaven/log';
+
+const log = Log.module('useTableListener');
+
+export const useTableListener = (
+  eventEmitter: Evented | undefined,
+  eventName: string,
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
mofojed: Do we need this?

mofojed (changes requested):

packages/jsapi-components/README.md
```diff
@@ -0,0 +1,17 @@
+# @deephaven/jsapi-utils
```
mofojed, line 1:
```suggestion
# @deephaven/jsapi-components
```

packages/jsapi-components/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export { default as DateInput } from './DateInput';
```
mofojed, line 1: This conflicts with your other change, I think just rebase and you don't need `DateInput` in `jsapi-components`

mofojed (changes requested):

packages/jsapi-components/src/index.ts
mofojed: No it's in `@deephaven/components` now, doesn't need the JSAPI since you made those changes

mofojed: jsapi-components/src/DateInput.jsx should also be deleted

vbabich: You're right, I'll clean it up.

mofojed (approved):

vbabich:
@mofojed One more change - removed unused dependency from the `code-studio` package

mofojed (approved):

PR #775, opened 2022-09-22
Proxy the ws socket within vite to allow tunneling

dsmmcken:
Proxy the websocket (and jsapi files) with the vite server. This is cool because it lets you open an ngrok tunnel to just port 4000 and have everything work.

### DIFF
--- a/packages/code-studio/.env.development
+++ b/packages/code-studio/.env.development
@@ -1,5 +1,4 @@
 VITE_ENABLE_LOG_PROXY=false
-VITE_CORE_API_URL=http://localhost:10000/jsapi
 VITE_FAVICON=/favicon-cc-app-dev.svg
 
 BASE_URL=/
@@ -8,3 +7,4 @@ PORT=4000
 VITE_LOG_LEVEL=4
 
 VITE_PROXY_URL=http://localhost:10000
+VITE_PROXY_WS=ws://localhost:10000
--- a/packages/code-studio/.env.development
+++ b/packages/code-studio/.env.development
@@ -7,4 +7,3 @@ PORT=4000
 VITE_LOG_LEVEL=4
 
 VITE_PROXY_URL=http://localhost:10000
-VITE_PROXY_WS=ws://localhost:10000
--- a/packages/code-studio/vite.config.ts
+++ b/packages/code-studio/vite.config.ts
@@ -40,13 +40,14 @@ export default defineConfig(({ mode }) => {
       target: `http://localhost:${port}/src/styleguide/index.html`,
       rewrite: () => '',
     },
+
     // proxy the websocket requests, allows tunneling to work with a single port
-    '^/arrow.*': {
+    '^/arrow\\.*': {
       target: env.VITE_PROXY_URL,
       changeOrigin: true,
       ws: true,
     },
-    '^/io.deephaven.*': {
+    '^/io\\.deephaven\\..*': {
       target: env.VITE_PROXY_URL,
       changeOrigin: true,
       ws: true,
--- a/packages/code-studio/vite.config.ts
+++ b/packages/code-studio/vite.config.ts
@@ -40,15 +40,34 @@ export default defineConfig(({ mode }) => {
       target: `http://localhost:${port}/src/styleguide/index.html`,
       rewrite: () => '',
     },
-  };
-
-  // Proxy all paths, including websocket
-  if (env.VITE_PROXY_URL) {
-    proxy['/'] = {
+    // proxy the websocket requests, allows tunneling to work with a single port
+    '^/arrow.*': {
       target: env.VITE_PROXY_URL,
       changeOrigin: true,
       ws: true,
-    };
+    },
+    '^/io.deephaven.*': {
+      target: env.VITE_PROXY_URL,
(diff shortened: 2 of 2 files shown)


packages/code-studio/.env.development
```diff
@@ -1,5 +1,4 @@
 VITE_ENABLE_LOG_PROXY=false
-VITE_CORE_API_URL=http://localhost:10000/jsapi
```
mattrunyon, line 2: This is used in the `index.html` and should be updated/checked there. The JS API script tags link using this env variable

packages/code-studio/vite.config.ts
```diff
@@ -40,6 +40,13 @@ export default defineConfig(({ mode }) => {
       target: `http://localhost:${port}/src/styleguide/index.html`,
       rewrite: () => '',
     },
+    // proxy the websocket, allows tunneling to work with a single port
+    '/': {
```
mattrunyon, line 44: Can't comment b/c it's not a changed line, but in the array in the block below this, you can remove `env.VITE_CORE_API_URL` if we don't have the env variable and don't need to proxy http requests to it

mattrunyon (changes requested):

mattrunyon:
Should also update in `embed-grid` and `embed-chart` packages

packages/code-studio/.env.development
dsmmcken: It's loading the jsapi successfully, so I assume that means it works.

dsmmcken:
nope that's not right, I now have just everything proxied, which isn't what we want. Will have to play a bit more.

packages/code-studio/vite.config.ts
dsmmcken: No, we still need to proxy that so that we can get the jsapi files.

dsmmcken: No, we still need to proxy that so that we can get the jsapi files.

packages/code-studio/.env.development
dsmmcken: > Should also update in `embed-grid` and `embed-chart` packages

They aren't setup with any proxy right now. I guess we could, but it's not necessary.

dsmmcken:
> Should also update in `embed-grid` and `embed-chart` packages

They aren't proxied at all right now. I guess we could, but I don't think that would ever matter from a testing perspective.

packages/code-studio/vite.config.ts
```diff
@@ -40,6 +40,17 @@ export default defineConfig(({ mode }) => {
       target: `http://localhost:${port}/src/styleguide/index.html`,
       rewrite: () => '',
     },
+    // proxy the websocket requests, allows tunneling to work with a single port
+    '^/arrow.*': {
+      target: env.VITE_PROXY_URL,
+      changeOrigin: true,
+      ws: true,
+    },
+    '^/io.deephaven.*': {
```
mofojed, line 49: This regex isn't quite right, you should be escaping the periods, e.g.:
```suggestion
    '^/io\.deephaven\..*': {
```
But aside from that - gRPC could have any path, we should be doing a negative lookup for any of the routes we reserve instead, e.g.:
```
'^(?!/ide).*'
```
```

**Review by mofojed — CHANGES_REQUESTED** [review]
We should also proxy embed-grid and embed-chart as well?

File: packages/code-studio/vite.config.ts (hunk 3 above)
Comment (dsmmcken) [reply packages/code-studio/vite.config.ts:49]: Can't do negative lookup. vite serves from a whole wack of routes. gRPC could have any path, but will it?

image

Comment (mofojed) [reply packages/code-studio/vite.config.ts:49]: Okay, seems like just `arrow.` and `io.deephaven.` are the only two we (currently) need to worry about.
However should still make the changes I suggested to have the correct regexes, `'^/arrow\..*'` and `'^/io\.deephaven\..*'`

**Review by mofojed — CHANGES_REQUESTED** [review]

**Comment by dsmmcken** [comment]
Something to note, the proxy looks to add about 3ms of latency.

**Review by mofojed — APPROVED** [review]

**Comment by mattrunyon** [comment]
I'm getting a ton of error spam in my dev server log. Mostly this error repeated a bunch

```
4:35:13 PM [vite] ws proxy error:
@deephaven/code-studio: Error: This socket has been ended by the other party
@deephaven/code-studio:     at Socket.writeAfterFIN [as write] (node:net:474:14)
@deephaven/code-studio:     at Socket.ondata (node:internal/streams/readable:754:22)
@deephaven/code-studio:     at Socket.emit (node:events:526:28)
@deephaven/code-studio:     at addChunk (node:internal/streams/readable:315:12)
@deephaven/code-studio:     at readableAddChunk (node:internal/streams/readable:289:9)
@deephaven/code-studio:     at Socket.Readable.push (node:internal/streams/readable:228:10)
@deephaven/code-studio:     at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

dsmmcken:
I don't see any log spam with jetty, this might be different now. @mattrunyon do you?

mattrunyon (approved):
No more log spam for me now. Appears to be working with ngrok

mofojed (approved):

PR #782, opened 2022-09-26
File explorer path and display initializes from focused dir/file instead of root

### LINKED ISSUE #734: File explorer create from selected path instead of always root
When creating a new folder in file explorer, the root path is always chosen by default even if I have a folder selected. Instead, it should default to the folder that is currently selected

This also applies to files, but we don't save files when you press the "New notebook" button, so I'm not sure if we would want to switch to saving if new notebook is created from file explorer, or ignore for notebooks

When selected like this and clicking new folder
image

Current behavior
image

Proposed behavior
image

jason-wang24:
Fixes #734 

## Breaking change
- `collapseAll()` has been added as a method to `FileStorageTable` in `FileStorage.ts` so all implementations of `FileStorageTable` will need to be updated.


packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
```diff
@@ -262,6 +262,7 @@ export class FileExplorerPanel extends React.Component<
             onDelete={this.handleDelete}
             onRename={this.handleRename}
             onSelect={this.handleFileSelect}
+            isNewItemModalClicked={false}
```
mofojed, line 265: This prop is always `false`, and i don't think is necessary. You should be able to remove it from all classes

packages/dashboard-core-plugins/src/panels/MockFileStorageTable.ts
```diff
@@ -25,6 +29,25 @@ export class MockFileStorageTable implements FileStorageTable {
...
+        }
+      }
+    }
+    return Promise.resolve();
+  }
+
+  collapseAll(): Promise<void> {
+    for (let i = 0; i < this.items.length; i += 1) {
+      this.setExpanded(`${this.items[i].filename}/`, false);
+    }
```
mofojed, line 50: We're iterating through all items here, and then iterating through them again in `setExpanded`, which is `O(N^2)`. To be more efficient, we should only need to loop through them once:
```suggestion
    for (let i = 0; i < this.items.length; i += 1) {
      const item = this.items[i];
      if (isDirectory(item)) {
        item.isExpanded = false;
      }
    }
```

packages/file-explorer/src/FileList.tsx
```diff
@@ -485,6 +488,18 @@ export const FileList = (props: FileListProps): JSX.Element => {
     }
   }, [draggedItems, dropTargetItem]);
 
+  const { isNewItemModalClicked, focusedPath } = props;
+  useEffect(() => {
+    if (isNewItemModalClicked && focusedPath !== undefined) {
```
mofojed, line 493: Is this needed? `isNewItemModalClicked` was always false, so I'm not sure this code ever ran.

packages/file-explorer/src/FileStorage.ts
```diff
@@ -43,6 +43,7 @@ export interface FileStorageTable extends StorageTable<FileStorageItem> {
    * @param expanded What expanded state to set
    */
   setExpanded(path: string, expanded: boolean): Promise<void>;
+  collapseAll(): Promise<void>;
```
mofojed, line 46: Add JSDocs to the method to explain how it's used.
Also you'll need to add the `breaking` label to this PR, and mention in the PR description there is a breaking change now that `collapseAll()` is added, so all implementations of `FileStorageTable` will need to be updated.

Another option would instead to not change the API at all, and instead use the existing `setExpanded` method to have collapseAll behaviour, e.g. `setExpanded('/', false)`. Though I guess existing implementations would still need to handle that special case as well, so it could introduce more subtle bugs.

I'll mention it in our 1:1

packages/file-explorer/src/WebdavFileStorageTable.ts
```diff
@@ -105,6 +110,20 @@ export class WebdavFileStorageTable implements FileStorageTable {
...
+    const { items } = directoryData;
+    for (let i = 0; i < items.length; i += 1) {
+      if (
+        isDirectory(items[i]) &&
+        !items[i].filename.includes('/', 1) &&
+        (items[i] as DirectoryStorageItem).isExpanded
+      ) {
+        this.setExpanded(items[i].filename, false);
+      }
+    }
```
mofojed, line 124: There's an easier way you can do this because of how the expanded tables are handled in `WebdavFileStorageTable`.
The way `WebdavFileStorageTable` works, is it builds a tree of `WebdavFileStorageTable`... each `WebdavFileStorageTable` is just the items in one directory, and then it has a map of `childTables` that are used for each doc. So all you need to do, is just clear out the map of `childTables`, and that removes all directories that are expanded. Something like:
```suggestion
    this.childTables.clear();
    if (this.currentViewport()) {
      await this.refreshInternal();
    }
```

mofojed (changes requested):

mofojed (changes requested):
For some reason, when I try to click Create Directory from a deeply nested directory (like `foo/bar/biz`), the file list shows up empty, e.g.:
image
image

I'll be available in the morning to help debug.

mofojed (approved):

PR #785, opened 2022-09-28
Downgrading monaco-editor to v0.31.1

### LINKED ISSUE #773: Autocomplete suggestions cannot be ignored
**Description**

When typing in the Console Input, you cannot escape the suggestions box. Subsequently, up/down arrows get stuck in the suggestions box and it is frustrating.

**Steps to reproduce**
1. Type some text in console input until suggestions come up
2. Press "Escape" to cancel out of the suggestions

**Expected results**

2. The suggestions box disappears

**Actual results**
2. The suggestion box remains, and pressing up/down moves through suggestions instead of the code. You need to press "Space" or something else to get rid of it.

https://user-images.githubusercontent.com/4505624/191830462-5f8269db-9c32-4e74-87a4-fd020bf2994e.mp4

**Versions**
 - Deephaven: `3ac2ee5568c5d5d40bf56c5b9c5888aaae708c63`
 - Web UI: 
 - OS: Ubuntu 20.04
 - Browser: Chrome
 - Docker: ...

mofojed:
- Fixes #773  being unable to escape the suggestion widget
- I don't know what change introduced the issue, as it's not reproducible in v0.34.0 playground and it's just in our code for some reason

I went down a rabbit hole debugging this...
The issue is that container is null when initializing the suggest widget contextview: https://github.com/microsoft/vscode/blob/0c22a33a9d670a84309447b36abdbd8c04ee6219/src/vs/base/browser/ui/contextview/contextview.ts#L162
Because it is null, this listener never gets registered: https://github.com/microsoft/vscode/blob/main/src/vs/base/browser/ui/contextview/contextview.ts#L183
Subsequently, this.hide() never gets called: https://github.com/microsoft/vscode/blob/0c22a33a9d670a84309447b36abdbd8c04ee6219/src/vs/base/browser/ui/contextview/contextview.ts#L352
I don't know how it's supposed to work since the container is always null for that service: https://github.com/microsoft/vscode/blob/0c22a33a9d670a84309447b36abdbd8c04ee6219/src/vs/editor/standalone/browser/standaloneLayoutService.ts#L27
I don't understand why this works in the Playground. I'm missing something, but for now I'm just going to roll back to v0.31.1 which doesn't have this issue.

### DIFF
--- a/packages/console/src/monaco/MonacoUtils.ts
+++ b/packages/console/src/monaco/MonacoUtils.ts
@@ -6,6 +6,7 @@ import { Shortcut } from '@deephaven/components';
 import { IdeSession } from '@deephaven/jsapi-shim';
 import { assertNotNull } from '@deephaven/utils';
 import * as monaco from 'monaco-editor';
+import type { Environment } from 'monaco-editor';
 // @ts-ignore
 import { KeyCodeUtils } from 'monaco-editor/esm/vs/base/common/keyCodes.js';
 // @ts-ignore
@@ -20,6 +21,18 @@ import LogLang from './lang/log';
 import { Language } from './lang/Language';
 
 const log = Log.module('MonacoUtils');
+
+declare global {
+  interface Window {
+    MonacoEnvironment: Environment;
+  }
+}
+
+window.MonacoEnvironment = {
+  getWorker() {
+    return editorWorker();
+  },
+};
 
 class MonacoUtils {
   static init(): void {
--- a/packages/console/src/monaco/MonacoUtils.ts
+++ b/packages/console/src/monaco/MonacoUtils.ts
@@ -21,12 +21,6 @@ import { Language } from './lang/Language';
 
 const log = Log.module('MonacoUtils');
 
-window.MonacoEnvironment = {
-  getWorker(workerId, label) {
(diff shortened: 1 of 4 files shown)


packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -21,12 +21,6 @@ import { Language } from './lang/Language';
 
 const log = Log.module('MonacoUtils');
 
-window.MonacoEnvironment = {
-  getWorker(workerId, label) {
-    return editorWorker();
-  },
-};
-
```
mattrunyon, line 29: I'm guessing you removed this b/c of TS. You can use this instead to keep worker support. Monaco bumped their types at some point after 0.31 I think to extend window

```ts
declare global {
  interface Window {
    MonacoEnvironment: monaco.MonacoEnvironment;
  }
}

window.MonacoEnvironment = {
  getWorker() {
    return editorWorker();
  },
};
```

mattrunyon (changes requested):

mattrunyon (approved):

mofojed:
Monaco issue is: https://github.com/microsoft/monaco-editor/issues/2947

PR #791, opened 2022-10-03
Allow up to 1% drop in code coverage per commit

mofojed:
- Previously required each change to increase test code coverage, which is unrealistic for a lot of bug fixes
- Change the treshold to 1% so bug fixes can go in without needed to increase the threshold, but bigger features should still trigger it

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,10 @@
 flag_management:
   default_rules:
     carryforward: true
+coverage:
+  status:
+    project:
+      default:
+        # Allow up to a 1% drop in threshold for codecov
+        # A lot of bug fixes don't change test coverage, and then codecov fails
+        threshold: 1%
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,6 +5,6 @@ coverage:
   status:
     project:
       default:
-        # Allow up to a 1% drop in threshold for codecov
+        # Allow up to a 0.05% drop in project threshold for codecov
         # A lot of bug fixes don't change test coverage, and then codecov fails
         threshold: 0.05%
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -1,3 +1,10 @@
 flag_management:
   default_rules:
     carryforward: true
+coverage:
+  status:
+    project:
```
mattrunyon, line 6: I think you want something targeting `patch` as well or ignoring it. `project` is for the project as a whole while `patch` is coverage of the changed lines. The `patch` check fails most often.

I think the `patch` defaults to a target of the `project` coverage. So our project sits around 36.3% right now and the `patch` check wants at least 36.3% coverage of the changed lines.

I think a 1% project level threshold would allow project coverage to drop 1% in a commit which is pretty big actually. We are at 36.3% coverage (so anything down to 35.3% is allowed).

We could just disable the patch check depending on the 

Here's more info on the config
https://docs.codecov.com/docs/commit-status#patch-status

We could also disable the annotations on GitHub if desired
https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations-via-yaml

There's also an unexpected coverage change check
https://docs.codecov.com/docs/commit-status#changes-status
https://docs.codecov.com/docs/unexpected-coverage-changes

mattrunyon (changes requested):

codecov.yml
mofojed: Okay yea project-wise I think threshold could be done to 0.05% even and we'd be fine...
For patches, I think it still should be auto (matching the project percentage), as we _should_ try and write more tests for each patch

mattrunyon (approved):

codecov.yml
```diff
@@ -1,3 +1,10 @@
...
+      default:
+        # Allow up to a 1% drop in threshold for codecov
```
mattrunyon, line 8: Fix comment percentage

mattrunyon (changes requested):

mattrunyon (approved):

PR #793, opened 2022-10-03
Change Boostrap px-2 to separate custom-px-2 class for icon only buttons

### LINKED ISSUE #717: Remove px-2 from icon only button
We currently add the class `px-2` to icon only buttons which is a bootstrap utility class. Unfortunately, bootstrap uses `!important` on the style it adds, so the only way to override it is using `!important` which is bad practice. Utility classes should be easily overridden

Checking for `kind="ghost"` should find any uses of the button with many of those being not icon only.

jason-wang24:
Fixes #717


packages/components/src/Button.tsx
```diff
@@ -63,7 +63,7 @@ function getClassName(kind: ButtonKind, iconOnly: boolean): string {
     case 'ghost':
       return classNames('btn-link', {
         'btn-link-icon': iconOnly,
-        'px-2': iconOnly,
+        'custom-px-2': iconOnly,
```
mofojed, line 66: Can we just remove the `px-2` class entirely and then add `padding-left`/`padding-right` to the css for `btn-link-icon` instead?

jason-wang24: `btn-link-icon` already has its own `padding-left`/`padding-right` and a lot of non icon-only buttons use it as a class so I don't think its possible to simply move the `px-2` padding into `btn-link-icon` as that may alter the style of other buttons.

dsmmcken: You are correct. Maybe just name this class `btn-link-icon-only` then or something.

Follow up task: clean up `<button/>` uses to component `<Button/>`?

mattrunyon: There's also this in the base stylesheet

```
.btn-link-icon:after {
  padding-left: calc(100% - 2px);
}
```

packages/components/scss/BaseStyleSheet.scss
```diff
@@ -361,6 +361,11 @@ span.btn-disabled-wrapper {
   }
 }
 
+.custom-px-2 {
```
mattrunyon, line 364: Change this classname to `btn-link-icon-only` as Don suggested

mattrunyon (changes requested):

mattrunyon (approved):

mofojed (approved):

PR #794, opened 2022-10-05
Catch and display errors in downloading in the `TableCsvExporter` component

### LINKED ISSUE #768: Catch and display errors in downloading in the `TableCsvExporter` component. Right now, the promise rejection is unhandled and can cause a bad UI state

jason-wang24:
Fixes #768

### DIFF
--- a/packages/iris-grid/src/sidebar/TableCsvExporter.test.tsx
+++ b/packages/iris-grid/src/sidebar/TableCsvExporter.test.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import TableCsvExporter from './TableCsvExporter';
+import IrisGridTestUtils from '../IrisGridTestUtils';
+
+const COLUMN_NAMES = ['A', 'B', 'C', 'D'];
+
+function makeTableCsvExporterWrapper({
+  name = 'TEST',
+  isDownloading = false,
+  tableDownloadStatus = '',
+  tableDownloadProgress = 0,
+  tableDownloadEstimatedTime = undefined,
+  onDownloadStart = () => null,
+  onDownload = (
+    fileName = null,
+    frozenTable = null,
+    tableSubscription = null,
+    snapshotRanges = null
+  ) => null,
+  onCancel = () => null,
+  selectedRanges = [],
+  model = IrisGridTestUtils.makeModel(
+    IrisGridTestUtils.makeTable(
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      COLUMN_NAMES.map(name => IrisGridTestUtils.makeColumn(name))
+    )
+  ),
+} = {}) {
+  return render(
+    <TableCsvExporter
+      name={name}
+      isDownloading={isDownloading}
+      tableDownloadStatus={tableDownloadStatus}
+      tableDownloadProgress={tableDownloadProgress}
+      tableDownloadEstimatedTime={tableDownloadEstimatedTime}
+      onDownloadStart={onDownloadStart}
+      onDownload={onDownload}
+      onCancel={onCancel}
+      selectedRanges={selectedRanges}
+      model={model}
+    />
+  );
+}
+
+it('renders without crashing', () => {
+  makeTableCsvExporterWrapper();
+});
+
+it('downloads properly with default settings', async () => {
+  makeTableCsvExporterWrapper();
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -184,12 +201,27 @@ class TableCsvExporter extends Component<
...
+                tableSubscription,
+                snapshotRanges
+              );
+            })
+            .catch(error => {
+              this.setDownloadPromiseError(error);
+            });
+        })
+        .catch(error => {
+          this.setDownloadPromiseError(error);
```
mofojed, line 223: This could probably be rewritten as an async/await function, and it'd be a bit cleaner, you'd only need one catch statement I think then.

```
try {
   const frozenTable = await model.export();
   const tableSubscription = frozenTable.setViewport(0, 0);
   const viewportData = await tableSubscription.getViewportData();
   onDownload(...)
} catch (e) {
   ...
}
```

packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -184,12 +186,24 @@ class TableCsvExporter extends Component<
...
+        await tableSubscription.getViewportData();
+        onDownload(fileName, frozenTable, tableSubscription, snapshotRanges);
+      } catch (error) {
+        log.error('CSV download failed', error);
+
+        this.setState({
+          errorMessage: (
+            <p>
+              <FontAwesomeIcon icon={vsWarning} /> CSV download failed,
+              TableCsvExporter rejected promise.
```
mofojed, line 201: This is a user facing message that we are displaying. Instead of displaying `TableCsvExporter rejected promise.`, which most users have no idea what that means, we should print the error itself, which we need to stringify to get it to appear correctly, e.g.:
```suggestion
              <FontAwesomeIcon icon={vsWarning} /> {`${error}`}
```

packages/iris-grid/src/sidebar/TableCsvExporter.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+//   tableDownloadProgress = 0,
+//   tableDownloadEstimatedTime = undefined,
+//   onDownloadStart = () => null,
+//   onDownload = (
+//     fileName = null,
+//     frozenTable = null,
+//     tableSubscription = null,
+//     snapshotRanges = null
+//   ) => null,
+//   onCancel = () => null,
```
mofojed, line 22: In general in tests, for anywhere you need a default function, just pass in a `jest.fn()`. You can pass it in anywhere and you can inspect it to see if it's been called and with what values easily.

mofojed (changes requested):
We can go over the tests in the am - we'll likely need to debug a bit why it's not downloading.

packages/iris-grid/src/sidebar/TableCsvExporter.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+it('downloads properly with default settings', async () => {
+  const onDownloadStart = jest.fn();
+  const onDownload = jest.fn();
+  const onCancel = jest.fn();
+  makeTableCsvExporterWrapper({ onDownloadStart, onDownload, onCancel });
+
+  userEvent.click(screen.getByRole('button', { name: 'Download' }));
+  expect(onDownloadStart).toHaveBeenCalledTimes(1);
+  await new Promise(resolve => setTimeout(resolve, 1));
```
mofojed, line 55: We have a utility function for flushing out pending promises already that you can use:
```suggestion
  await TestUtils.flushPromises();
```

packages/iris-grid/src/sidebar/TableCsvExporter.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  makeTableCsvExporterWrapper({
+    onDownloadStart,
+    onDownload,
+    onCancel,
+    model,
+  });
+
+  userEvent.click(screen.getByRole('button', { name: 'Download' }));
+  expect(onDownloadStart).toHaveBeenCalled();
+  await new Promise(resolve => setTimeout(resolve, 1));
```
mofojed, line 74: Use utility function

mofojed (changes requested):
Excellent work! Just use the utility function instead of `setTimeout` and should be good.

mofojed (approved):

PR #797, opened 2022-10-06
Playwright e2e tests

mofojed:
- Add some basic Playwright tests
  - Print a console message
  - Open a table
  - Open a figure
  - Create a notebook, run it, delete it

Partial #804


tests/table.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  //   await page.locator('placeholder=Enter value').click();
+  //   await page.keyboard.type('0');
+
+  //   await page.locator('button:has-text("Done")').click();
+
+  //   // Check snapshot
+  //   await expect(page.locator('.iris-grid-column')).toHaveScreenshot();
+
+  //   // Close the download panel
+  //   await page.locator('data-testid=btn-page-close').click();
```
mofojed, line 137: We're not closing the download panel here, I think just closing the sidebar. Either way, check that it's within the scope of `.iris-grid` as well.

tests/table.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    ).toHaveCount(0);
+
+    await expect(
+      page.locator('.progress .progress-bar.bg-success')
+    ).toHaveCount(1);
+
+    // Check snapshot
+    await expect(
+      page.locator('.iris-grid .navigation-page')
+    ).toHaveScreenshot();
```
mofojed, line 85: I wouldn't bother with a screenshot here - just check that the progress bar is there, that it shows success, etc. is fine.

Screenshots/snapshots are mainly used for canvas elements and things like that.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
```
mofojed, line 6: We should disable this action for now.

tests/figure.spec.ts
```diff
@@ -10,11 +10,8 @@ test('can open a simple figure', async ({ page }) => {
   const figureName = generateVarName();
 
   // Create a figure that uses the table we just created
-  await page.keyboard.type(`from deephaven import empty_table\r`);
-  await page.keyboard.type(`from deephaven.plot.figure import Figure\r`);
-  await page.keyboard.type(
-    `${figureName} = Figure().plot_xy(series_name="Test", t=empty_table(100).update(["x=i", "y=Math.sin(i)", "z=Math.cos(i)"]), x="x", y="y").show()\r`
-  );
+  const command = `from deephaven import empty_table\rfrom deephaven.plot.figure import Figure\r${figureName} = Figure().plot_xy(series_name="Test", t=empty_table(100).update(["x=i", "y=Math.sin(i)", "z=Math.cos(i)"]), x="x", y="y").show()\r`;
```
mofojed, line 13: Probably makes sense to do `\n` instead of `\r` as people are more familiar with that. In fact, it would look better to just re-write it with a multi-line string, e.g. (note that white space matters):
```
const command = `
${figureName} = Figure().plot_xy(series_name="Test", t=empty_table(100).update(["x=i", "y=Math.sin(i)", "z=Math.cos(i)"]), x="x", y="y").show()
`;
```

tests/utils.ts
```diff
@@ -14,4 +15,19 @@ export function generateVarName(prefix = 'v'): string {
...
 }
 
-export default { generateVarName };
+/**
+ * Types into monaco input and avoids autocomplete suggestions
+ * @param page The Page instance for each test
+ * @param text Text to be typed, with carriage returns
+ */
+export async function typeInMonaco(page: Page, text: string): Promise<void> {
+  const splitByLine = text.split('\r');
```
mofojed, line 24: Split on `\n` instead, then it'll work with multi-line strings.

tests/utils.ts
```diff
@@ -14,4 +15,19 @@ export function generateVarName(prefix = 'v'): string {
...
+  for (let i = 0; i < splitByLine.length; i += 1) {
+    await page.keyboard.type(splitByLine[i]);
+    await page.keyboard.press(' ');
```
mofojed, line 27: Worth adding a comment here why you're doing a space/backspace/shift+Enter

mofojed (commented):
You should also use this new function in table.spec.ts and notebook.spec.ts

mofojed (commented):
You should also use this in table.spec.ts and notebook.spec.ts

mofojed (commented):
Looking good 👍

jason-wang24 (approved):

PR #799, opened 2022-10-06
Set aria-label for icon only buttons

### LINKED ISSUE #798: Set aria-label for icon only buttons
Icon only `<Buttons/>` require a tooltip, they should probably require an aria-label as well.

Can set the aria-label to that tooltip, if a separate aria label isn't set, and the tooltip isn't a jsx element.

dsmmcken:
Resolves #798

Re-uses tooltip as aria-label for iconOnly buttons that don't have a defined aria-label. If tooltip is a jsx element require a aria-label to be set with typescript.

### DIFF
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -38,7 +38,7 @@ type ButtonWithChildren = BaseButtonProps & {
   children: React.ReactNode;
 };
 
-type IconOnlyButtoStringTooltip = BaseButtonProps & {
+type IconOnlyButtonStringTooltip = BaseButtonProps & {
   tooltip: string;
   icon: IconDefinition | JSX.Element;
   children?: undefined;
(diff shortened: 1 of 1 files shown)


packages/components/src/Button.tsx
```diff
@@ -38,12 +38,21 @@ type ButtonWithChildren = BaseButtonProps & {
   children: React.ReactNode;
 };
 
-type IconOnlyButton = BaseButtonProps & {
-  tooltip: string | JSX.Element;
+type IconOnlyButtoStringTooltip = BaseButtonProps & {
```
mofojed, line 41:
```suggestion
type IconOnlyButtonStringTooltip = BaseButtonProps & {
```

packages/components/src/Button.tsx
```diff
@@ -38,12 +38,21 @@ type ButtonWithChildren = BaseButtonProps & {
...
 };
 
+type IconOnlyButtonJsxTooltip = BaseButtonProps & {
+  tooltip: JSX.Element;
+  'aria-label': string;
+  icon: IconDefinition | JSX.Element;
+  children?: undefined;
+};
+
+type IconOnlyButton = IconOnlyButtoStringTooltip | IconOnlyButtonJsxTooltip;
```
mofojed, line 54:
```suggestion
type IconOnlyButton = IconOnlyButtonStringTooltip | IconOnlyButtonJsxTooltip;
```

mofojed (approved):

mofojed (changes requested):
Check my previous comments

mofojed (approved):

PR #808, opened 2022-10-13
Update to use new $GITHUB_OUTPUT instead of set-output

mofojed:
- set-output is deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


devinrsmith (approved):
Update merge message, you mean "set-output" not "set_state".

mofojed:
That's weird, I noticed in their blog post they had a typo and mentioned `set-state` instead of `save-state`, and then I just went and made the same mistake twice (with an `_` even).
Updated PR title/description, will update commit message as well.

PR #810, opened 2022-10-13
790 file explorer render fail

### LINKED ISSUE #790: File explorer within `NewItemModal` sometimes fails to render when a nested directory is focused
**Description**

Sometimes, when the New Folder button is clicked, the file explorer within the opened modal does not render.

**Steps to reproduce**
1. Click into a nested directory in the file explorer
2. Click the New Folder button
3. If the file explorer inside `NewItemModal` does in fact render, close the modal and repeat steps 1-3.

**Expected results**

The file explorer inside the modal always renders.

**Actual results**

Sometimes, the file explorer inside the modal does not render (the area below the breadcrumbs is empty)

Info from debugging:
- The for loop in the `fetchData` method inside `GrpcFileStorageTable` mutates the `items` array
- It does not catch a promise if the `getViewportData()` call is cancelled
- The same bug likely causes the folders to sometimes not expand/contract when they are clicked

image
image

jason-wang24:
Fixes #790 
- added Tooltip to NewItemModal home button and renamed aria-label to uppercase "Home"
- refactored GrpcFileStorageTable and added timers to its test file

## Breaking Change
The `setExpanded()` and `collapseAll()` methods in the `FileStorageTable` interface in `FileStorage.ts` have changed (they now return `void` instead of `Promise<void>`).

### DIFF
--- a/packages/code-studio/src/storage/grpc/GrpcFileStorageTable.ts
+++ b/packages/code-studio/src/storage/grpc/GrpcFileStorageTable.ts
@@ -9,13 +9,18 @@ import {
   IndexRange,
   StorageSnapshot,
 } from '@deephaven/storage';
-import { CancelablePromise, PromiseUtils } from '@deephaven/utils';
+import {
+  CancelablePromise,
+  CanceledPromiseError,
+  PromiseUtils,
+} from '@deephaven/utils';
 import {
   FileStorageItem,
   FileStorageTable,
   FileUtils,
 } from '@deephaven/file-explorer';
 import { StorageService } from '@deephaven/jsapi-shim';
+import debounce from 'lodash.debounce';
 
 const log = Log.module('GrpcFileStorageTable');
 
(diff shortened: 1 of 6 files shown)


packages/code-studio/src/storage/grpc/GrpcFileStorageTable.ts
```diff
@@ -161,28 +174,59 @@ export class GrpcFileStorageTable implements FileStorageTable {
...
+    log.debug('doRefresh', this.root);
+
+    const viewportDataPromise = this.fetchData();
+
+    viewportDataPromise.then(callback).catch(e => {
+      if (!PromiseUtils.isCanceled(e)) {
+        log.error('Error fetching viewport data', e);
+      }
+    });
+  }
```
mofojed, line 187: I know why it was complaining about the unchained version of this. When calling `viewportDataPromise.then(callback)`, that creates a new promise (each `.then` or `.catch` creates a new promise), and that new promise did not have a catch on it so was uncaught.

A better way to handle this (than silently eating the error) is to actually just return the promise, rather than resolve it in the `doRefresh` method. Then the `refresh()` method can still set the `viewportUpdatePromise` correctly, and a call to `getViewportData()` will reject correctly if the operation was cancelled.

```suggestion
  doRefresh(
    callback: (fetchPromise: Promise<ViewportData<FileStorageItem>>) => void
  ): void {
    log.debug2('doRefresh', this.root);
    callback(this.fetchData());
  }
```

mofojed (changes requested):

mofojed (approved):

PR #818, opened 2022-10-14
Extract PouchDB storage into @deephaven/pouch-storage package

vbabich:


mattrunyon:
Should probably add this to the root tsconfig as a project reference since they're all listed there. Although it probably won't affect anything since code-studio consumes it and it will get built by tsc just building code-studio

I think they were all listed there because it's one of the examples from typescript for projects in a monorepo

packages/pouch-storage/src/PouchCommandHistoryStorage.ts
```diff
@@ -1,14 +1,14 @@
 /* eslint-disable class-methods-use-this */
 import Log from '@deephaven/log';
+import {
```
mofojed, line 3: I'm a little confused about these re-ordered imports - why are they re-ordered?

mofojed (changes requested):

mofojed:
Also need to do an `npm install` to update the package-lock

packages/pouch-storage/src/PouchCommandHistoryStorage.ts
vbabich: Must be a result of cherry-picking from `0.13` to `main`.  Cleaned up.

mofojed (approved):

PR #822, opened 2022-10-17
Refactor HTML buttons in `code-studio` to Deephaven Buttons

### LINKED ISSUE #815: Refactor HTML `button`s to Deephaven `Button`s in `code-studio` package

jason-wang24:
Fixes #815 
- Refactored all HTML buttons except for the close button in the style guide's example `modal` and the buttons in the `Controls` menu (they have no matching `class`es with any `kind` of Deephaven Button)
- Added support in Button.tsx for `tabIndex` and onMouseUp/Down events


jason-wang24:
Note: There is a failed proptype warning because the `Controls` button in `AppMainContainer` (the one in the top bar with the Deephaven logo, Panels, and settings icon) does not have an `onClick()` function. The clicking functionality is handled elsewhere. What should be done about this?

packages/components/src/Button.tsx
```diff
@@ -220,6 +226,21 @@ Button.propTypes = {
...
+    }
+    return null;
+  },
+  onMouseDown(props) {
+    const { onMouseDown } = props;
+    if (onMouseDown !== undefined && typeof onMouseDown !== 'function') {
+      return new Error('onMouseDown must be a function');
+    }
+    return null;
+  },
```
mofojed, line 242: You can just write this as `PropTypes.func`. The only reason it's split up in the `onClick` is because it's only checking if the `onClick` function is there for type of `'button'`.

```suggestion
  onMouseUp: PropTypes.func,
  onMouseDown: PropTypes.func,
```

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -756,23 +756,22 @@ export class AppMainContainer extends Component<
               className="ml-1"
             />
             <div>
-              <button type="button" className="btn btn-link btn-panels-menu">
-                <FontAwesomeIcon icon={dhShapes} />
+              <Button kind="ghost" className="btn-panels-menu" icon={dhShapes}>
```
mofojed, line 759: As mentioned in your comment, this is complaining because no `onClick` is provided.
The reason no `onClick` is provided is because `AppControlsMenu`/`DropdownMenu` adds an onclick listener to it's parent: https://github.com/deephaven/web-client-ui/blob/75204eda3d72f91ad80215af6177927ae1bea1ac/packages/components/src/menu-actions/DropdownMenu.tsx#L92
This paradigm that `DropdownMenu` uses is not very React-y at all - I thought it was clever at the time, but it's stupid.
Changing all instances of `DropdownMenu` does not make sense right now, so I'd say just leave that and then just add an empty `onClick` handle to Button here so it works:
 ```suggestion
              <Button
                kind="ghost"
                className="btn-panels-menu"
                icon={dhShapes}
                onClick={() => {
                  // no-op: click is handled in `AppControlsMenu` (which uses a `DropdownMenu`)
                }}
              >
```

mofojed (changes requested):

mofojed (changes requested):

mofojed (approved):

PR #826, opened 2022-10-18
Refactor HTML buttons to Deephaven Buttons in `iris-grid` package

### LINKED ISSUE #819: Refactor HTML `button`s to Deephaven `Button`s in `iris-grid` package

jason-wang24:
Fixes #819  
- Did not refactor buttons which have CSS classes that do not correspond well with the `kind` options of Deephaven Buttons
- Did not refactor buttons in enterprise only


packages/components/src/Button.tsx
```diff
@@ -31,6 +31,8 @@ interface BaseButtonProps extends React.ComponentPropsWithRef<'button'> {
   icon?: IconDefinition | JSX.Element;
   active?: boolean;
   'data-testid'?: string;
+  'data-index'?: number;
+  'data-operator'?: string;
```
mofojed, line 35: We shouldn't add `data-index` and `data-operator` attributes (which are quite specific to the `AdvancedFilterCreator`).
Since `data-*` attributes are common (see MDN article: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), and the names can be whatever are necessary for the context.
I _think_ this is possible using Mapped Types: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html
See this [TypeScript playground here](https://www.typescriptlang.org/play?jsx=0#code/C4TwDgpgBAhgJnACgJwgMwJYA8AqB7AeQCMArCAY2AB4cAaKRKCLYCAOzgGcpPhkM2AcwB8UALxQA3gFgAUFCgBtANJQBUANYQQeNFByxuq5qw7de-IVAD8UAAYASSYgC+T5S7tQAXFDYQANwhkAF1fHBUQuRc5OQB6OP1waABVYAwAG25AMGpAeD-Y2QFWZDQYcmgCMHS8Nk4AQSk5BQU2GABbCF8LAUEAbmiCouDS8qhK6tqAIUb5ZrAYZGBOXzYAVzaiYP7ZGNk5IZKyiqqMGs4AYRrMQRnmpS0QLr4esLGTs4aAH1gEFHRsfDEMiUKjjU5TegAcgAApDhNtdnJyGdgFA2iBLmxrr4wWdMddxFIXNt4okAOp4ZAaZZydH4jCCAB0MGAfAAjITJK0Or5IdyIJDibSMVcGczWcgAEycgBE0Pmi04Mt8bKFsjpoqZLL4AGZZfKFktlVA2fR+bz+YKSbIElAAKLIZCU8LJKCQyRQBVLFbrTbIXpQFyQtTcNh4VEwTicBmtIgZaDAPBQUCQN242pfH5IVCYXCEUgUajpziTehymXCSGM4X0rUSgAssq9SpVarkQA) where a prop key is prefixed with `@`.
We essentially want to do the same thing but for keys prefixed with `data-`, and value is always a `string`.

Then we use a spread operator to apply them to the `<button>` tag.

packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -18,7 +18,7 @@ import {
   SortDirection,
   FilterItem,
 } from '@deephaven/jsapi-utils';
-import { ContextActionUtils, Tooltip } from '@deephaven/components';
+import { Button, ContextActionUtils } from '@deephaven/components';
```
mofojed, line 21: There's something weird going on when showing advanced filters. I don't see the bug, but when it is showing it appears in the wrong place first, then moves, which is incorrect (and a regression):
advanced-filter-blip

packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
```diff
@@ -336,15 +336,11 @@ class CustomColumnBuilder extends Component<
...
-      <button
-        type="button"
-        className={classNames(
-          'btn btn-apply',
-          {
-            'btn-spinner': isCustomColumnApplying,
-          },
-          isSuccessShowing ? 'btn-success' : 'btn-primary'
-        )}
+      <Button
```
mofojed, line 339: Not part of your change, but I noticed the padding is incorrect on the title "Custom Column Name and Formula":
image

If you can fix this as part of this change, that would be helpful. Check with @dsmmcken if we want to just add the correct padding there (so it lines up with the inputs below), or if we can just remove that subtitle (it seems redundant)

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -883,9 +883,8 @@ class VisibilityOrderingBuilder extends Component<
         }}
       >
         <div className="top-menu">
-          <button
-            type="button"
-            className="btn btn-link"
+          <Button
```
mofojed, line 886: Something looks messed up in this top menu, the Toggle All button is wrapping for some reason:
image
Should look more like:
image

mofojed (changes requested):
For some reason the Ctrl+F and Ctrl+G keyboard shortcuts in IrisGrid are no longer working as well. Please ensure those work before submitting.

EDIT: It looks like they're broken in `main`, so not caused by this change. We should submit a fix for that ASAP.

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
jason-wang24: Looks like kind="ghost" adds the class `btn-link-icon-only` which doubles the padding on the 6 icon only buttons in the div.  This causes the rightmost one to just barely overflow and Toggle All matches the new height as a result.

mofojed: Might need to add some css to reduce the padding on these buttons then.

packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
jason-wang24: The subtitle will be removed.

packages/components/src/Button.tsx
mattrunyon: Data attributes are kind of anti-React I think. You can create any attribute you want as a prop in React, so we should be leaning on those instead unless there's a specific reason it needs to be a data attribute

Also, I saw there's an aria attribute added the the prop types, but all valid aria props are already part of the type since it extends the react component type.

packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
mattrunyon: That padding is always broken for me, but only in dev mode. Works fine in a built version. Not sure why it's broken in dev though

packages/components/src/Button.tsx
```diff
@@ -31,6 +36,7 @@ interface BaseButtonProps extends React.ComponentPropsWithRef<'button'> {
   icon?: IconDefinition | JSX.Element;
   active?: boolean;
   'data-testid'?: string;
+  dataAttributes?: DataAttribute;
```
mofojed, line 39: As mentioned in my other comment, I don't think we should have a `dataAttributes` prop on button. We should instead look at refactoring the `AdvancedFilterCreator` data attribute usage so that we don't need to use data attributes there, then we don't need to put data attributes on the button itself.

mofojed (changes requested):

mofojed (approved):

PR #827, opened 2022-10-19
`run` , `send to console`, `send to notebook` buttons not working in safari

### LINKED ISSUE #805: User reports run button in notebook does not work with Safari
Community slack user reports Safari run button was not executing code in safari. They were running latest/0.17.0

Investigate to confirm reproducibility. 

Screen Shot 2022-10-10 at 5 00 55 PM

Zhou-Ziheng:
Fixes #805

### DIFF
--- a/packages/dashboard/src/PanelManager.ts
+++ b/packages/dashboard/src/PanelManager.ts
@@ -187,13 +187,20 @@ class PanelManager {
     C extends ComponentType<P> = ComponentType<P>
   >(types: PanelComponentType<P, C>[]): PanelComponent<P> | undefined {
     return this.getLastUsedPanel(panel =>
-      types.some(
-        type =>
-          panel instanceof type ||
+      types.some(type => {
+        let instanceofType;
+        try {
+          instanceofType = panel instanceof type;
+        } catch (e) {
+          instanceofType = false;
+        }
+        return (
+          instanceofType ||
           (isWrappedComponent(type) &&
             type.WrappedComponent &&
             panel instanceof type.WrappedComponent)
-      )
+        );
+      })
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/PanelManager.ts
```diff
@@ -187,13 +187,20 @@ class PanelManager {
...
-      types.some(
-        type =>
-          panel instanceof type ||
+      types.some(type => {
+        let instanceofType;
+        try {
+          instanceofType = panel instanceof type;
+        } catch (e) {
+          instanceofType = false;
+        }
```
mattrunyon, line 196: Add a comment mentioning that the catch block is needed due to safari behavior

mattrunyon (dismissed):

packages/dashboard/src/PanelManager.ts
mattrunyon: Oops didn't notice this was in draft state. Still add a comment for the final though

mofojed: It's interesting - in this case, `type` that is passed in is a typeof `"object"`, so I don't think `instanceof` would _ever_ be true, and it's kind of surprising that Chrome/Firefox don't throw an error here. We probably shouldn't be checking `instanceof` on just `type` if it's a wrapped component.
So I think to be more correct, we could just do something like:
```
return (isWrappedComponent(type) && panel instanceof type.WrappedComponent) || (!isWrappedComponent(type) && panel instanceof type);
```

mofojed (changes requested):

mofojed (approved):

PR #831, opened 2022-10-19
Fix keyboard shortcuts not working in Grid

### LINKED ISSUE #829: Keyboard Shortcuts not working in Grids
**Description**

Keyboard shortcuts don't seem to be working in Grid

**Steps to reproduce**
1. Open up a table
2. Focus in the table and press Ctrl+F

**Expected results**
2. Quick filter bar appears

**Actual results**
2. Browser Find bar triggers.

**Versions**
 - OS: Ubuntu
 - Browser: Chrome

jason-wang24:
Fixes #829 
- Added keyHandlers to support shortcuts but could be overly complex because it does not use the fact that these actions are `contextActions`
- Does not have support for `discard` or `commit` yet

### DIFF
--- a/packages/components/src/context-actions/ContextActionUtils.ts
+++ b/packages/components/src/context-actions/ContextActionUtils.ts
@@ -104,7 +104,7 @@ class ContextActionUtils {
   }
 
   /**
-   * Returns true if the modifier key for the current platform is  down for the event (Ctrl for windows/linux, Command (meta) for mac)
+   * Returns true if the modifier key for the current platform is down for the event (Ctrl for windows/linux, Command (meta) for mac)
    * @param event The event to get the meta key status from
    */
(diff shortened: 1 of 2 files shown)


mofojed (changes requested):
This will add the key handler to the grid canvas only which is a problem when other DOM elements (like the settings button, or quick filter input). You can see the problem fairly easily by just pressing Ctrl+F to show the quick filter bar, then pressing Ctrl+F again to try and hide the quick filter bar - instead it will show find.

I think we should figure out exactly which commit the issue was introduced. We can use [git bisect](https://git-scm.com/docs/git-bisect) to narrow it down.

mofojed (approved):

mofojed:
Excellent work!

PR #838, opened 2022-10-21
Add scientific notation number format option to table context menu

### LINKED ISSUE #836: Add "Scientific Notation" option to number formatting context menu
image (16)

Maybe in the menu after "Millions"?

see: https://www.gwtproject.org/javadoc/latest/com/google/gwt/i18n/client/NumberFormat.html

jason-wang24:
Fixes #836 
- Rounds to 4 decimal places (same as default format)


mofojed (changes requested):
I think we should add to `IntegerFormatContextMenu` as well, as scientific notation also applies to large numbers.

mofojed (approved):

PR #840, opened 2022-10-24
Disable line numbers in Log Panel

### LINKED ISSUE #839: Extra Number on Bottom of Log Panel

jason-wang24:
Fixes #839


packages/console/src/log/LogView.tsx
```diff
@@ -211,6 +211,7 @@ class LogView extends PureComponent<LogViewProps, LogViewState> {
       fixedOverflowWidgets: true,
       folding: false,
       fontFamily: 'Fira Mono',
+      lineNumbers: 'off',
```
dsmmcken, line 214: Can probably delete this comment now:

```
      // I commented this out since '' is not a valid parameter for line Numbers
      // lineNumbers: '',
 ```
 
 Looks like this was switched off during TS migration, rather than looked up.

dsmmcken:
Minor but you should probably sort this option adjacent to the other "line" related options, as all the options appear sorted alphabettically.

dsmmcken (approved):

PR #844, opened 2022-10-26
resolve missing onclick proptype validation

dsmmcken:


packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -1050,6 +1050,7 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
                   disabled={toolbarDisabled}
                   icon={vsKebabVertical}
                   tooltip="More Actions..."
+                  onClick={() => undefined}
```
jason-wang24, line 1053:
```suggestion
                  onClick={() => {
                    // no-op: click is handled in `DropdownMenu`
                  }}
```
To keep it consistent with other Buttons that use DropdownMenus

jason-wang24 (changes requested):

jason-wang24 (approved):

PR #848, opened 2022-10-26
Fix incorrect table sizes in panel tab tooltip

### LINKED ISSUE #820: Web console reports incorrect table sizes in panel tab tooltip when table is changing sizes
Running 'latest' at 7dd6468139ffc0234c72236107f4b51354c4f581

My query is
```
test1 = time_table("00:00:00.000001").update_view(["Int64Value = ii"]).tail(3000000)
```

Expected value for 'Number of Rows': 3,000,000. Displayed value: 1.

image

Using up-arrow to run the same query but store it in `test2`
```
test2 = time_table("00:00:00.000001").update_view(["Int64Value = ii"]).tail(3000000)
```

Now `test1` has the correct count but `test2` has the incorrect count:
image
image

If I do the same thing with `test3`, then `test2` will have the correct count and `test3` will be wrong. And so on.

jason-wang24:
Fixes #820  
- Added 'Updated' event listener to IrisGridPanel to listen for dynamically changing row counts
- Added `rowCount` as a state in IrisGridPanel to cause it to re-render when row count changes


packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -1212,10 +1226,7 @@ export class IrisGridPanel extends PureComponent<
       this.getPluginContent(Plugin, model, user, workspace, pluginState);
     const { permissions } = user;
     const { canCopy, canDownloadCsv } = permissions;
-    const formattedRowCount = model?.displayString(
-      model?.rowCount ?? 0,
-      'long'
-    );
+    const formattedRowCount = model?.displayString(rowCount, 'long');
```
mofojed, line 1229: rowCount could be undefined here, use `?? 0` or set a default value when pulling it from state at top of render function.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -807,6 +818,7 @@ export class IrisGridPanel extends PureComponent<
       IrisGridModel.EVENT.TABLE_CHANGED,
       this.handleTableChanged
     );
+    model.addEventListener(IrisGridModel.EVENT.UPDATED, this.handleUpdated);
```
mofojed, line 821: I was incorrect about which event to listen to - this will only trigger when there's an update within the current viewport you're viewing, and isn't triggering when the table grows larger than the current viewport. We want to listen to the JS API `EVENT_SIZECHANGED`, which we'll need to wire up through `IrisGridModel`. 
Add a `SIZE_CHANGED` event in `IrisGridModel.EVENT`, then in `IrisGridTableModelTemplate` add an event listener on the table to listen for `EVENT_SIZECHANGED` and then emit the `SIZE_CHANGED` event, listen to that from here.

mofojed (approved):

PR #855, opened 2022-10-27
Fix incorrect table sizes in panel tab tooltip

### LINKED ISSUE #820: Web console reports incorrect table sizes in panel tab tooltip when table is changing sizes

jason-wang24:
Fixes #820


packages/dashboard-core-plugins/src/panels/IrisGridPanelTooltip.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { ReactElement, useEffect, useState } from 'react';
```
mofojed, line 1: `useEffect`, `useState` unused
```suggestion
import React, { ReactElement } from 'react';
```

mofojed (changes requested):

mofojed (approved):

PR #860, opened 2022-10-28
Replaces old Deephaven logo spinner with more minimal animation

dsmmcken:
The old logo that was used as a spinner while the table loaded has served us well, but it is time to say goodbye. It is a little too overt and can look bad when the table window is really small. It is also too in your face when you are watching a dozen of them spin at once on the initial load of a complex dashboard. It is also no longer even our current logo design.

old_spinner
👎 

Change to a more minimal indeterminate loading bar integrated with the table loading status bar. Also moved the cancel button to be more seamless with the table status bar. I think being more subtle with it can actually make loading seem faster. The same previous logic still applies that the loading animation/cancel button a staggered in after a few ms has already elapsed so that on quick loads, you don't even see an animation. 

loading
👍


packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -301,6 +301,19 @@ class IrisGridRenderer extends GridRenderer {
       (scrimBlurSize * 2 + (height - gridY)) * loadingScrimProgress
     );
 
+    // add an overlay of the scrim color to darken things a bit
+    context.globalCompositeOperation = 'overlay';
+    if (scrimColor != null) {
+      context.fillStyle = scrimColor;
+    }
```
mofojed, line 308: I don't think this is needed again, as it's already set above

mofojed (dismissed):

packages/iris-grid/src/IrisGridRenderer.ts
dsmmcken: right.

mofojed (approved):

PR #861, opened 2022-10-28
Fixes command history not scrolled to bottom on load

### LINKED ISSUE #859: Command history not scrolled to bottom on load
1. When command history is loaded in as the panel in the background, it loads scrolled to top rather than bottom as expected. If the command history tab is on front when loaded, it loads correctly scrolled to bottom.
  command_history_bug_1

2. If the command history tab is picked up and dropped in place, it's contents disappear until scrolled again. 
  command_history_bug

dsmmcken:
Resolves #859

- Use initialScrollOffset prop to set initial position instead of scrolling to bottom in did mount (which doesn't work when in a backgrounded tab)
- forceUpdate the virtualized list when it is picked up and dropped in the same place

### DIFF
--- a/packages/components/src/ItemList.tsx
+++ b/packages/components/src/ItemList.tsx
@@ -181,18 +181,11 @@ export class ItemList<T> extends PureComponent<
     };
   }
 
-  componentDidMount(): void {
-    const { isStickyBottom } = this.props;
-    if (isStickyBottom && !this.isListAtBottom()) {
-      this.scrollToBottom();
-    }
-  }
-
   componentDidUpdate(
     prevProps: ItemListProps<T>,
     prevState: ItemListState
   ): void {
-    const { selectedRanges: propSelectedRanges } = this.props;
+    const { selectedRanges: propSelectedRanges, itemCount } = this.props;
     const {
       focusIndex,
       isStuckToBottom,
@@ -200,7 +193,7 @@ export class ItemList<T> extends PureComponent<
       height,
       selectedRanges,
     } = this.state;
-    if (isStuckToBottom && !this.isListAtBottom()) {
+    if (isStuckToBottom && !this.isListAtBottom() && itemCount > 0) {
       this.scrollToBottom();
     }
 
@@ -322,6 +315,10 @@ export class ItemList<T> extends PureComponent<
 
   focus(): void {
     this.listContainer.current?.focus();
+  }
+
+  update(): void {
+    this.list.current?.forceUpdate();
   }
 
   getElement(itemIndex: number): Element | null {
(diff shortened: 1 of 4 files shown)


packages/components/src/ItemList.tsx
```diff
@@ -181,26 +181,19 @@ export class ItemList<T> extends PureComponent<
...
+    const { selectedRanges: propSelectedRanges, itemCount } = this.props;
     const {
       focusIndex,
       isStuckToBottom,
       scrollOffset,
       height,
       selectedRanges,
     } = this.state;
-    if (isStuckToBottom && !this.isListAtBottom()) {
+    if (isStuckToBottom && !this.isListAtBottom() && itemCount > 0) {
```
mofojed, line 196: I think we still need the `componentDidMount` - if you have a virtual list that's stuckToBottom and has items already loaded, it won't scroll to bottom.

packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx
```diff
@@ -105,6 +106,11 @@ class CommandHistoryPanel extends Component<
     }
   }
 
+  handleShow(): void {
+    // virtual list requires a force re-render when picked up and dropped in same place
+    this.container.current?.update();
```
mofojed, line 111: Does show always get called in this case? What about listening to `onResize` instead?

mofojed (changes requested):

dsmmcken:
Todo: doesn't always reset correctly on pick up and place back if you scroll first

packages/components/src/ItemList.tsx
mofojed: I could be wrong, I just tried this by modifying the styleguide and doesn't seem to be an issue. Must be something causing a re-render on mount

dsmmcken: I don't think it's needed, that's what the initialScrollOffset is for. when it mounts?

packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx
dsmmcken: show is enough

packages/console/src/command-history/CommandHistory.tsx
```diff
@@ -351,6 +354,10 @@ class CommandHistory extends Component<
     }
   }
 
+  update(): void {
```
mattrunyon, line 357: I'd prefer this to be named something else. It's not an update, it's on show or scroll to bottom

packages/components/src/ItemList.tsx
```diff
@@ -324,6 +317,15 @@ export class ItemList<T> extends PureComponent<
     this.listContainer.current?.focus();
   }
 
+  update(): void {
```
mattrunyon, line 320: Same about naming this `scrollToBottom` instead of `update`

mattrunyon (changes requested):

mofojed (approved):

mattrunyon:
@dsmmcken Did you fix the position when dropping it in the same position?

dsmmcken:
> @dsmmcken Did you fix the position when dropping it in the same position?

Yes, that's why I added the restoreScrollPosition function

mattrunyon (approved):

PR #867, opened 2022-11-02
Add word wrap and minimap to notebook monaco editor

### LINKED ISSUE #845: Add an option for word wrap in code editing panels
In the notebook panel overflow menu add an option for toggling word wrap of the monaco editor (off by default). Setting is per panel.

image

We don't have any global configuration of monaco settings at this time.

Request by a community user in slack.

jason-wang24:
Fixes #845 
Word wrap is set per editor while the minimap is a global setting for all editors


mofojed (changes requested):

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -40,9 +44,13 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     isLoaded: false,
     focusOnMount: true,
     onChange: (): void => undefined,
+    onEditorInitialized: (): void => undefined,
+    onEditorWillDestroy: (): void => undefined,
+    updateWorkspaceData: (): void => undefined,
```
mofojed, line 49: Unused, shouldn't be here
```suggestion
```

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -40,9 +44,13 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
...
     session: null,
     sessionLanguage: null,
     settings: null,
+    workspace: null,
```
mofojed, line 53: Unused, shouldn't be here.
```suggestion
```

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -40,9 +44,13 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     isLoaded: false,
     focusOnMount: true,
     onChange: (): void => undefined,
+    onEditorInitialized: (): void => undefined,
+    onEditorWillDestroy: (): void => undefined,
+    updateWorkspaceData: (): void => undefined,
     session: null,
     sessionLanguage: null,
     settings: null,
```
mofojed, line 52: Side note, I don't think we need to specific `null` for these params, since we can just leave them as `undefined` (they are both optional).

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -21,11 +21,15 @@ interface ScriptEditorProps {
   focusOnMount?: boolean;
   onChange: (e: editor.IModelContentChangedEvent) => void;
   onRunCommand: (command: string) => void;
+  onEditorInitialized: (editor: editor.IStandaloneCodeEditor) => void;
+  onEditorWillDestroy: () => void;
   session: IdeSession;
   sessionLanguage?: string;
   settings?: {
```
mofojed, line 28: Instead of defining our own type for settings, we should use the monaco editor type, `editor.IStandaloneEditorConstructionOptions`.

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -21,11 +21,15 @@ interface ScriptEditorProps {
   focusOnMount?: boolean;
   onChange: (e: editor.IModelContentChangedEvent) => void;
   onRunCommand: (command: string) => void;
+  onEditorInitialized: (editor: editor.IStandaloneCodeEditor) => void;
+  onEditorWillDestroy: () => void;
```
mofojed, line 25: `Editor` passes through the editor as a param, we should here as well:
```suggestion
  onEditorWillDestroy: (editor: monaco.editor.IStandaloneCodeEditor) => void;
```

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -168,10 +177,14 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     if (focusOnMount != null && focusOnMount) {
       innerEditor.focus();
     }
+
+    onEditorInitialized(this.editor);
   }
 
   handleEditorWillDestroy(): void {
```
mofojed, line 184: Take in the editor param here and pass it through

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -578,6 +646,48 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     }
   }
 
+  handleMinimap() {
```
mofojed, line 649: In terms of naming of these functions, I'd make the following changes:
`handleMinimap` => `updateEditorMinimap`: Makes it clear that the function is updating the child editor minimap settings
`handleMinimapWrapper` => `handleMinimapChange`: Makes it clear that the function is handling the option being changed
`handleWordWrap` => `handleWordWrapChange`: Same as above. Also, instead of calling `this.editor.updateOptions` right in here, I'd have an `updateEditorWordWrap` similar to `updateEditorMinimap`, and call that from `componentDidUpdate` when the minimap settings change in the state. I think that will clean it up a bit.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -578,6 +646,48 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+        wordWrap,
+      });
+      this.setState(prevState => ({
+        settings: {
+          ...prevState.settings,
+          wordWrap,
+        },
+      }));
+    }
+  }
```
mofojed, line 689: As mentioned in another comment, this should be broken into two functions:
`updateEditorWordWrap`: Should take the word wrap from settings, and call `editor.updateOptions`. Call it from `componentDidUpdate`
`handleWordWrapChange`: Should just toggle the option in the state settings, doesn't need to set on the actual editor.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -927,10 +1038,17 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
     const isMarkdown = itemName.endsWith('.md');
     const isExistingItem = fileMetadata?.id != null;
-    const overflowActions = this.getOverflowActions();
+    const minimap = {
+      minimap: { enabled: defaultNotebookSettings.isMinimapEnabled },
+    };
     const settings = {
       ...initialSettings,
+      ...minimap,
     };
```
mofojed, line 1047: Doesn't look like it's from your change, but we should be memoizing this result (I'm not sure why we were even copying it before??), similar to how we're assigning `overflowActions`
Right now we're creating a new `settings` object on each re-render, which may give the wrong impression (something changing when it hasn't actually).

mofojed (changes requested):

packages/console/src/notebook/ScriptEditor.tsx
mofojed: This works fine, I would have passed the value through `handleEditorWillDestroy` and used that though (then no need for `assertNotNull`)

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -170,9 +164,10 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     this.handleEditorWillDestroy = this.handleEditorWillDestroy.bind(this);
     this.handleEditorChange = this.handleEditorChange.bind(this);
     this.handleFind = this.handleFind.bind(this);
-    this.handleMinimapWrapper = this.handleMinimapWrapper.bind(this);
-    this.handleMinimap = this.handleMinimap.bind(this);
-    this.handleWordWrap = this.handleWordWrap.bind(this);
+    this.handleMinimapChange = this.handleMinimapChange.bind(this);
+    this.updateEditorMinimap = this.updateEditorMinimap.bind(this);
+    this.handleWordWrapChange = this.handleWordWrapChange.bind(this);
+    this.updateEditorWordWrap = this.updateEditorWordWrap.bind(this);
```
mofojed, line 170: `updateEditorMinimap` and `updateEditorWordWrap` need not be bound, as they are not passed in as a prop or event handler.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -666,7 +669,17 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+      const { wordWrap } = settings;
+      this.editor.updateOptions({
+        wordWrap,
+      });
+    }
+  }
+
+  handleWordWrapChange() {
     if (this.editor) {
       const { settings } = this.state;
```
mofojed, line 684: You could just rewrite this whol block within the `setState`, e.g.
```
this.setState(prevState => {
  const { settings } = prevState;
  const wordWrap = settings.wordWrap === 'on' ? 'off' : 'on';
  return {
    settings: {
      ...settings,
      wordWrap,
    },
  }
});
```

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -511,6 +503,16 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+  getSettings = memoize(
+    (
+      initialSettings: editor.IStandaloneEditorConstructionOptions,
+      minimap: editor.IStandaloneEditorConstructionOptions
+    ) => ({
+      ...initialSettings,
+      ...minimap,
+    })
+  );
+
```
mofojed, line 515: Should just be `isMinimapEnabled` here, take a boolean. Taking a second `IStandaloneEditorConstructionOptions` doesn't make sense. Should also specify the return type as `IStandaloneEditorConstructionOptions`:
```suggestion
  getSettings = memoize(
    (
      initialSettings: editor.IStandaloneEditorConstructionOptions,
      isMinimapEnabled: boolean
    ): editor.IStandaloneEditorConstructionOptions => ({
      ...initialSettings,
      minimap: { enabled: isMinimapEnabled },
    })
  );
```

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -1038,13 +1048,9 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     const itemName = fileMetadata?.itemName ?? NotebookPanel.DEFAULT_NAME;
     const isMarkdown = itemName.endsWith('.md');
     const isExistingItem = fileMetadata?.id != null;
-    const minimap = {
+    const settings = this.getSettings(initialSettings, {
```
mofojed, line 1051: With the `getSettings` change this should be simplified
```
    const settings = this.getSettings(
      initialSettings,
      defaultNotebookSettings.isMinimapEnabled
    );
```

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -1196,7 +1202,9 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
                 sessionLanguage={sessionLanguage}
                 onEditorInitialized={this.handleEditorInitialized}
                 onEditorWillDestroy={this.handleEditorWillDestroy}
-                settings={settings}
+                settings={
+                  settings as editor.IStandaloneEditorConstructionOptions
+                }
```
mofojed, line 1207: Don't need to cast here if you define the return type on `getSettings` correctly.

mofojed (changes requested):

mofojed (approved):
Excellent work!

PR #872, opened 2022-11-03
Fix Tab/Shift+Tab not cycling through quick filter input columns

### LINKED ISSUE #841: Tab/Shift+Tab not cycling through quick filter inputs
Tab and Shift+Tab not cycling through all quick filter columns.

Steps to reproduce:
1. Make a table with several columns (enough to go off screen).
2. Open quick filters (ctrl+f) on the table and select the first column
3. Press tab key several times/shift+tab several times

Expected:
tab key cycles through all the columns, and is focus trapped

Actual:
Tab key leaves the quick filters to a external button when the last visible column is reached.

jason-wang24:
Fixes #841


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2017,7 +2017,11 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     } else if (rightVisible < column) {
       const metricState = this.grid?.getMetricState();
       assertNotNull(metricState);
-      const newLeft = metricCalculator.getLastLeft(metricState, column, gridX);
+      const newLeft = metricCalculator.getLastLeft(
+        metricState,
+        column - 1,
```
mofojed, line 2022: I don't think you should need the `- 1` here. You want to get the left most column for that right column to be visible

mofojed (changes requested):
This is closer, but there's still something else funky going on. If I create a table with a bunch of columns:
```
t = empty_table(100).update(["a=i", "b=i", "c=i", "d=i", "e=i", "f=i", "g=i", "h=i", "j=i"])
```

Then resize a bunch of the columns to be wider, and put the table in a narrow viewport, show the filters, then press tab a view times to go to a filter offscreen, the view gets messed up:
image

I'm not sure why it would do that, as it looks like the canvas width is off? See if you can figure that out before merging

mofojed:
While debugging, it looks like the `iris-grid-column` element is being shifted to the left for some reason??
image

mofojed (approved):
Great work!

PR #879, opened 2022-11-07
Fix sass relative imports that rely on node_modules structure

### LINKED ISSUE #878: Cleanup sass imports
**Description**

This undoes #60 for the most part, but we no longer use the sass CLI watcher.

We currently import our app-wide variables as relative paths like `../../../components/scss/custom.scss`. Instead, we can switch to `@import '@deephaven/components/scss/custom.scss'` to clean it up. Locally, the components package is already symlinked to the root node_modules. Then, we can change our imports of bootstrap from `'../../../bootstrap/scss/_functions.scss'` to just ``bootstrap/scss/_functions.scss'`

We had previously removed these imports because our dev mode included starting watchers for sass in all the packages. Sass has an issue with load-path that watches the entire directory, so we were starting thousands of watchers on node_modules files that were unnecessary. With the switch to Vite, we no longer need these watch scripts since running dev mode just runs the Vite CLI in code-studio.

Cleaning up these helps a bit w/ DHE as well. Because of the relative imports, Vite in DHE currently prints a bunch of warnings about missing source maps. That could be fixed by either adding the `--include-source` flag for Sass (makes source maps bigger, but more portable). Or removing the relative imports to components and bootstrap. Vite resolves to node_modules with no special characters (webpack uses `~bootstrap` to indicate `node_modules/bootstrap`).

The `sass:watch` scripts can be removed from packages. AFAIK, we don't develop at the package level ever (i.e. no code-studio running), so we don't need watch scripts at the package level.

mattrunyon:
Fixes #878

We don't use the watch scripts any more, so we shouldn't have to worry about the sass watch CLI watching thousands of node_modules files and blowing up

Also noticed the stylelint checks weren't applying to `components/scss` so I added those and fixed a few complaints it had

Vite currently ignores the embedded sources in sourcemaps, but once that is fixed, this should not spam warnings in our DHE start script about missing sourcemap files

https://github.com/vitejs/vite/issues/10870


packages/dashboard/src/layout/GoldenLayout.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../../components/scss/custom.scss';
+@import '@deephaven/components/scss/custom.scss';
 @import '../../../golden-layout/src/scss/goldenlayout-base.scss';
 @import '../../../golden-layout/src/scss/goldenlayout-dark-theme.scss';
```
mofojed, line 3: Any reason we don't have the `golden-layout` package embedding it's scss as well? Or even `dashboard`? It looks like this works, just curious where the line was drawn/why.

mattrunyon: Fixed these imports

I tried to make it similar to the eslint rule that prevents you from importing outside the source of the package. So any relative path that would have passed through the `packages` dir should be avoided

mofojed (approved):

PR #881, opened 2022-11-11
Fix EditableItemList selection, change to multiselect

vbabich:
EditableItemList was using wrong event for tracking selection.
- Changed `onSelect` to `onSelectionChange`
- Also changed the list to multiselect, as suggested by QA

Breaking change: `onDelete` is now triggered with array of values instead of a single value

### DIFF
--- a/packages/components/src/EditableItemList.test.tsx
+++ b/packages/components/src/EditableItemList.test.tsx
@@ -16,12 +16,7 @@ function makeWrapper({
   onAdd = jest.fn(),
   onDelete = jest.fn(),
   validate = undefined,
-}: {
-  items?: string[];
-  onAdd?: (item: string) => void;
-  onDelete?: (items: string[]) => void;
-  validate?: (item: string) => Error | null;
-} = {}) {
+}: Partial<EditableItemListProps> = {}) {
   return render(
     <EditableItemList
       items={items}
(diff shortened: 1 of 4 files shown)


packages/components/src/EditableItemList.test.tsx
```diff
@@ -18,9 +18,9 @@ function makeWrapper({
   validate = undefined,
 }: {
   items?: string[];
-  onAdd?: (string) => void;
-  onDelete?: (string) => void;
-  validate?: (string) => Error | null;
+  onAdd?: (item: string) => void;
+  onDelete?: (items: string[]) => void;
+  validate?: (item: string) => Error | null;
```
mofojed, line 23: Instead of redefining the types here, you could just do a `Partial<EditableItemListProps>`

packages/components/src/EditableItemList.tsx
```diff
@@ -29,25 +29,27 @@ const EditableItemList = (props: EditableItemListProps): React.ReactElement => {
...
+    const itemsToDelete: string[] = selectedRanges.reduce(
+      (acc: string[], range: Range) => {
+        const result = [...acc];
+        for (let i = range[0]; i <= range[1]; i += 1) {
+          result.push(items[i]);
+        }
+        return result;
+      },
+      []
+    );
```
mofojed, line 49: We could probably put this in `RangeUtils` in the utils package. Seems like this could be common functionality. It's kind of like a flatten.

mofojed (approved):
Changes look good, a couple of possible optimizations/cleanups.

packages/components/src/EditableItemList.tsx
vbabich: Would `getItemsInRanges` be a good name for the util function?

mofojed (approved):

PR #882, opened 2022-11-11
Cleanup unused package scripts

### LINKED ISSUE #880: Cleanup unused scripts
We can remove the watch and dev scripts for most of the packages. We only need to run dev scripts in the apps (except for icons), so no need for `watch`, `watch:babel`, `watch:sass`, `build-dev`

Could also change `babel` to `build:babel`  and `sass` to `build:sass` to remove 2 more. No need after removing watch to have `build` call `build:babel` which just calls `babel`

mattrunyon:
Fixes #880. Removes several unused package scripts now that we've moved to Vite.

This is a nice step before lerna v6 which uses nx since nx config includes any script that can be cached. Just helps keep things clean.


mofojed (dismissed):

mofojed (commented):
> ● Validation Error:
> Directory /home/runner/work/web-client-ui/web-client-ui/packages/*/src in the roots[0] option was not found.

mattrunyon:
That error seems to be from me adding `roots` to the unit test jest config. I added that to try to quiet jest's initial spam about duplicate mocks, but I removed it since it's not a big deal

mofojed (approved):

PR #892, opened 2022-11-18
Fix null in links on grouped columns

### LINKED ISSUE #875: DH-13599 Don't send nulls to the target when the link source is a grouping column (in a rollup table) -- jackson hotfix

vbabich:
- Pass `isGrouped` and `isExpandable` flags in selected data in addition to column type, text, and value
- Clear quick filters when applying links on columns with `undefined` value
- Fixes #875

### DIFF
--- a/packages/dashboard-core-plugins/src/linker/LinkerUtils.tsx
+++ b/packages/dashboard-core-plugins/src/linker/LinkerUtils.tsx
@@ -38,6 +38,8 @@ export type LinkDataMapValue = {
   type: string;
   text: string;
   value: string;
+  isExpandable: boolean;
+  isGrouped: boolean;
 };
 
 export type LinkDataMap = Record<string, LinkDataMapValue>;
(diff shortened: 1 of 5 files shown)


packages/dashboard-core-plugins/src/linker/LinkerUtils.tsx
```diff
@@ -38,6 +38,8 @@ export type LinkDataMapValue = {
   type: string;
   text: string;
   value: string;
```
mofojed: I think we need to update this - the value isn't always a `string`, this should be `unknown`. Then you don't need that ugly casting in `Linker.tsx`.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2429,19 +2443,27 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
     const dataMap: Record<
       string,
-      { value: unknown; text: string | null; type: string }
+      {
+        value: unknown;
+        text: string | null;
+        type: string;
+        isExpandable: boolean;
+        isGrouped: boolean;
+      }
```
mofojed, line 2455: We should define this type above, use it in `onDataSelected` and here.
Also, `text` should just be `string` - we don't allow `null` there.

We can't import the Linker type directly since IrisGrid doesn't depend on linker - either way there should be some abstraction between the two.

mofojed (changes requested):

mofojed (approved):

PR #893, opened 2022-11-18
Added engine version to settings menu

### LINKED ISSUE #770: Add engine version to settings, read other config settings
In the settings panel, we currently show the UI version but not the engine version. These versions can be different. Will need some API support to get the engine version

Zhou-Ziheng:
Fixes #770 
- Added engine version to settings menu
- On hover, Java version and barrage version are also displayed
- All versions are exported when one exports logs.


packages/code-studio/src/redux/selectors.ts
```diff
@@ -10,3 +10,13 @@ import LayoutStorage from '../storage/LayoutStorage';
...
 export const getLayoutStorage = (store: RootState): LayoutStorage =>
   store.layoutStorage as LayoutStorage;
+
+/**
+ * Get the configuration values of the server
+ * @param store The redux store
+ * @returns The layout storage instance
+ */
+
+// eslint-disable-next-line import/prefer-default-export
```
mofojed, line 20: You can delete the empty line and the disabling of the warning, including for the export above as well. It's only necessary when the file only has one export, and we want it named.
```suggestion
```

packages/code-studio/src/settings/SettingsMenu.scss
```diff
@@ -180,3 +180,17 @@ $settings-menu-z-index: $zindex-modal;
...
+
+.detailed-server-config-info {
+  padding: 10px;
+}
+
+.detailed-server-config-row {
+  display: flex;
+  justify-content: space-between;
+  gap: $spacer-3;
+}
```
mofojed, line 196: I would nest these under `.app-settings-menu` above just so they are in a more specific scope.

packages/code-studio/src/settings/SettingsMenu.tsx
```diff
@@ -177,7 +195,25 @@ export class SettingsMenu extends Component<
...
+                  <div className="detailed-server-config">
+                    <FontAwesomeIcon icon={vsInfo} />
+                    <Tooltip>
+                      <div className="detailed-server-config-info">
+                        {getRow('Deephaven Version     ', deephavenVersion)}
+                        {getRow('Java Version', javaVersion)}
+                        {getRow('Barrage Version', barrageVersion)}
+                        {/* <div>Deephaven Version {deephavenVersion}</div>
+                        <div>Java Version {javaVersion}</div>
+                        <div>Barrage Version {barrageVersion}</div> */}
```
mofojed, line 212: Delete the commented out code, and remove the unnecessary whitespace after Deephaven Version.
```suggestion
                        {getRow('Deephaven Version', deephavenVersion)}
                        {getRow('Java Version', javaVersion)}
                        {getRow('Barrage Version', barrageVersion)}
```

packages/code-studio/src/settings/SettingsMenu.tsx
```diff
@@ -177,7 +195,25 @@ export class SettingsMenu extends Component<
...
               <div className="app-settings-footer-item">
-                <div className="text-muted">Version {version}</div>
+                <div className="text-muted">UI Version {version}</div>
+              </div>
+              <div className="app-settings-footer-item">
+                <div className="text-muted">
+                  Engine Version {deephavenVersion}{' '}
+                  <div className="detailed-server-config">
+                    <FontAwesomeIcon icon={vsInfo} />
+                    <Tooltip>
```
mofojed, line 205: Interesting, the tooltip is hiding immediately after appearing in many cases... I think because it's just the right distance from the edge of the screen, where the tooltip arrow appears overtop of the item you are hovering over, so it triggers `mouseleave`.
I think _ideally_ we'd fix that on the Popper side - it should reposition the tooltip to be above because there isn't quite enough room on the right. But tracking that down might be tricky, and it's an upstream change.

Instead, just flag this tooltip as `interactive`. That way the tooltip will stay up even if they hover over the tooltip, and also has the added benefit (for this case) that you can select the text and copy it.

```suggestion
                    <Tooltip interactive>
```

packages/redux/src/reducers/serverConfigValues.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * The plugins that have been loaded
```
mofojed, line 2: Update comment - something about the server config values instead of this, which doesn't apply.

packages/code-studio/src/storage/LocalWorkspaceStorage.ts
```diff
@@ -72,8 +72,11 @@ export class LocalWorkspaceStorage implements WorkspaceStorage {
 
   private layoutStorage: LayoutStorage;
 
-  constructor(layoutStorage: LayoutStorage) {
+  private serverVersion?: string;
+
+  constructor(layoutStorage: LayoutStorage, serverVersion?: string) {
     this.layoutStorage = layoutStorage;
+    this.serverVersion = serverVersion;
```
mofojed, line 79: I don't think `LocalWorkspaceStorage` needs the `serverVersion`, revert the changes in this file.

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -243,7 +247,24 @@ const AppInit = (props: AppInitProps) => {
...
+        const [config, version] = configs[i];
+
+        if (config === 'java.version') {
+          serverConfig.javaVersion = version;
+        } else if (config === 'deephaven.version') {
+          serverConfig.deephavenVersion = version;
+        } else if (config === 'barrage.version') {
+          serverConfig.barrageVersion = version;
+        }
+      }
```
mofojed, line 264: I think instead of building the server config values into our own object, we should just create a Map, which would eliminate most of this code, e.g. `const serverConfig = new Map(configs);`
Then have a few constants for the settings that we know about. This has a couple of advantages:
1) We don't need to change this part for each new settings
2) We could pass the server config to plugins, and they can read a server config value that we may not know/care about.

mofojed (changes requested):
Great start, some clean up necessary.

mofojed (approved):

packages/code-studio/src/settings/SettingsMenu.tsx
```diff
@@ -177,7 +193,20 @@ export class SettingsMenu extends Component<
...
+                    <Tooltip interactive>
+                      <div className="detailed-server-config-info">
+                        {getRow('UI Version', uiVersion)}
+                        {getRow('Deephaven Version', deephavenVersion)}
+                        {getRow('Java Version', javaVersion)}
+                        {getRow('Barrage Version', barrageVersion)}
+                      </div>
+                    </Tooltip>
+                  </div>
+                </div>
```
dsmmcken, line 209:
```suggestion
                <div className="font-weight-bold">Version</div>
                <span className="d-inline-block text-muted">
                  {deephavenVersion} <FontAwesomeIcon icon={vsInfo} />
                  <Tooltip interactive>
                    <div className="detailed-server-config">
                      {getRow('Engine Version', deephavenVersion)}
                      {getRow('Web UI Version', uiVersion)}
                      {getRow('Java Version', javaVersion)}
                      {getRow('Barrage Version', barrageVersion)}
                    </div>
                  </Tooltip>
                </span>
```

Re-styled a bit, and adjusted the order/labels.

packages/code-studio/src/settings/SettingsMenu.tsx
```diff
@@ -84,11 +83,28 @@ export class SettingsMenu extends Component<
...
+    const { serverConfigValues } = this.props;
+    const barrageVersion = serverConfigValues.get('barrage.version');
+    const javaVersion = serverConfigValues.get('java.version');
+    const deephavenVersion = serverConfigValues.get('deephaven.version');
+
+    const getRow = (text: string, ver?: string) => (
+      <div className="detailed-server-config-row">
+        <div>{text}</div>
+        <div>{ver}</div>
+      </div>
```
dsmmcken, line 105:
```suggestion
      <>
        <div>{text}</div>
        <div>{ver}</div>
      </>
```

If you use a css grid, you don't need this extra wrapper.

packages/code-studio/src/settings/SettingsMenu.scss
```diff
@@ -177,6 +177,20 @@ $settings-menu-z-index: $zindex-modal;
...
```
dsmmcken, line 196:
```suggestion
.detailed-server-config {
  display: grid;
  grid-template-columns: auto auto;
  gap: $spacer-1 $spacer-3;
  padding: $spacer-2;
  text-align: left;
}
```

CSS grid is good for any 2 directional layout, helps things stay nicely aligned. Also that 10px padding should have also been a spacer.

packages/code-studio/src/settings/SettingsMenu.scss
```diff
@@ -177,6 +177,20 @@ $settings-menu-z-index: $zindex-modal;
     border-top: 1px solid $gray-800;
     .app-settings-footer-item:not(:last-child) {
       padding-bottom: $settings-menu-padding;
+
+      .detailed-server-config {
+        display: inline-block;
+      }
```
dsmmcken, line 183:
```suggestion
```

Can be removed in css grid version

dsmmcken (changes requested):
Added a variety of style changes.

dsmmcken (approved):

mofojed (approved):

PR #894, opened 2022-11-18
Update browserslist to latest defaults

### LINKED ISSUE #658: Audit browserslist configs
Is our browserslist configs setup correctly since moving to lerna? I only see it in code-studio and embed-grid?

```
"browserslist": [
    ">0.2%",
    "not dead",
    "not ie >= 0",
    "not op_mini all"
  ],
  ```

Also the new default config is https://github.com/browserslist/browserslist#browserslist-
```
  "browserslist": [
    "defaults",
    "not IE 11",
    "maintained node versions"
  ]
  ```

mofojed:
- Just use the latest defaults specified with browserslist instead of our own list
- Use a root `.browserlistrc` file instead of specifying it in package.json
- Fixes #658

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,3 @@
+# Just use the defaults value for browsers we support
+
+defaults
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,3 @@
-# Just use the defaults value for browsers we support
+# Use the default value for browsers we support https://browsersl.ist/#q=defaults
 
 defaults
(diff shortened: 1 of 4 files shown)


.browserslistrc
```diff
@@ -0,0 +1,3 @@
+# Just use the defaults value for browsers we support
```
dsmmcken, line 1:
```suggestion
# Use the default value for browsers we support https://browsersl.ist/#q=defaults
```

dsmmcken (dismissed):

.browserslistrc
```diff
@@ -0,0 +1,3 @@
+# Use the default value for browsers we support https://browsersl.ist/#q=defaults
```
dsmmcken, line 1:
```suggestion
# Use the default value for browsers we support https://browsersl.ist/#q=defaults
# Used by post-css only, JS browser targeting is controlled by Vite https://vitejs.dev/config/build-options.html#build-target
```

dsmmcken (approved):

PR #895, opened 2022-11-19
Add types override for memoizee

mattrunyon:
This is the only way I could get a types override for `memoizee` and `memoize-one`. The way the @types package is declared for it, I was unable to override its default. These packages are simple enough that I would be shocked if the API signature ever changed.

React `useCallback` is another potential problem point where we can omit types and not get any warning about the implicit any. That will likely need to be overridden differently, but should be doable since it is not the default export we would override from the package. The types have been changed in definitely typed for React 18. I found [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/52873#issuecomment-845806435) useful for getting an implicit any warning at the definition instead of using never and getting warnings on usage.

`useMemo` should not cause any issues since the function passed can not contain parameters


mattrunyon:
It's fine to review/merge now. React 18 will just show a few more potential issues with implicit any inside `useCallback`

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -80,15 +78,17 @@ class IrisGridTableModel extends IrisGridTableModelTemplate<Table, UIRow> {
   }
 
   getMemoizedFrontColumns = memoize(
-    layoutHintsFrontColumns => layoutHintsFrontColumns ?? EMPTY_ARRAY
+    (layoutHintsFrontColumns: ColumnName[] | undefined) =>
+      layoutHintsFrontColumns ?? []
```
mofojed, line 82: This was intentionally using `EMPTY_ARRAY` as a const, so that at a higher level it was memoized properly... though upon further reflection, I guess in this case it doesn't matter because it'll only return the `[]` when `layoutHintsFrontColumns` is `undefined`, and the memoize here will keep that object cached whenever `undefined` is passed; so disregard my first sentence and well done :)

@types/memoize-one/index.d.ts
```diff
@@ -0,0 +1,12 @@
+export declare type EqualityFn = (
```
mofojed, line 1: Instead of adding our own type overrides, can we just update memoize-one to `6.0.0`? Looks like they added types and tightened it up for TypeScript usage: https://github.com/alexreardon/memoize-one/releases
https://github.com/alexreardon/memoize-one/blob/master/src/memoize-one.ts

@types/memoizee/index.d.ts
```diff
@@ -0,0 +1,28 @@
+declare namespace memoizee {
```
mofojed, line 1: Looks like a new version of `@types/memoizee` exists `v0.4.8`, can we just use that and maybe we don't need to specify our own types at all?

mofojed (changes requested):
Changes look good, I think we can just update memoize packages and get the typings for free (`memoize-one` at least).

If there are still changes to `memoizee/index.d.ts`, see if you can open a PR for it in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/memoizee/index.d.ts (though I don't see any differences).

@types/memoizee/index.d.ts
mattrunyon: Nope. The most up to date for `memoizee` are the loose types still which just allow you to ignore the param types and then get an uncaught implicit any. Types [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/memoizee/index.d.ts)

@types/memoize-one/index.d.ts
mattrunyon: I'll have to check. At first glance, their types still appear to let through untyped function params without warning which is the whole point of my type overrides here

mofojed:
If that's the case, can you open a PR against the memoizee/memoize-one types with these additions?

mofojed (approved):

PR #896, opened 2022-11-19
Add floating action buttons to linker and ability to change the comparison operator

### LINKED ISSUE #887: Add ability to change comparison operator for linkers with floating action buttons on select

jason-wang24:
Fixes #887 
- Added selected state to linker which displays two floating action buttons when selected: delete and comparison selection panel
- Linkers can be deleted with alt + click
- New styling for linkers based on selection state, and danger when alt is held


packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -52,20 +70,55 @@ export class LinkerOverlayContent extends Component<
...
+    this.handleKeyUp = this.handleKeyUp.bind(this);
     this.handleEscapePressed = this.handleEscapePressed.bind(this);
+    this.handleComparisonOperatorChanged = this.handleComparisonOperatorChanged.bind(
+      this
+    );
 
     this.state = {
       mouseX: undefined,
       mouseY: undefined,
+      isAltPressed: false,
```
mofojed, line 114: Instead of having an `isAltPressed` state variable, make the state variable either something like `isDeleting: boolean`, or even `mode: 'select' | 'delete'` or something to that effect. We're using Alt to trigger delete mode here, but it wouldn't be unusual to trigger it via some other method.

packages/filters/src/Utils.ts
```diff
@@ -74,3 +74,43 @@ export function getLabelForBooleanFilter(filterType: TypeValue): string {
       throw new Error(`Unrecognized boolean filter type ${filterType}`);
   }
 }
+
+export function getSymbolForTextFilter(filterType: TypeValue): string {
```
mofojed, line 78: You should be able to just use `TableUtils.getFilterOperatorString` instead of defining a new function here. May need to add a couple cases (e.g. `eqIgnoreCase`).

mofojed (changes requested):

packages/filters/src/Utils.ts
jason-wang24: According to the `TableUtils` tests, any case not currently in the switch statement is supposed to throw. Luckily, for the purposes of the linker, I don't think we need to add any new cases except for `startsWith` and `endsWith` but that can just be handled via conditionals inside `Linker.tsx`.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -448,25 +442,28 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
+        if (operator !== undefined && operator !== 'eq') {
+          if (operator === 'startsWith' || operator === 'endsWith') {
+            symbol = '*';
+          } else {
+            symbol = TableUtils.getFilterOperatorString(operator);
+          }
+        }
+        let text = `${symbol}${value}`;
+        if (operator === 'startsWith') {
+          text = `${value}${symbol}`;
```
mofojed, line 463: Don't do all this operator stuff here. Instead, just pass the `operator` through in the `filterMap` (along with `columnType`, `text`, `value`) and then in `IrisGridPanel` or `IrisGrid` build the text there (as you'll have the array of filters as well).

mofojed (changes requested):

PR #905, opened 2022-11-29
Fix broken ctrl+e to clear filters shortcut (#664)

### LINKED ISSUE #663: "Clear All Tables Filters" Keyboard Shortcut not working
In DHC versions 0.13.0 and 0.14.1, the keyboard shortcut for clearing filters on all tables (cmd+E by default) doesn't seem to work (it just doesn't do anything).

Clearing filters for all tables (cmd+shift+E) still works.

The shortcuts panel does show cmd+E as set:

image

Changing it to a different key combo doesn't seem to work either.

vbabich:
* Resolves #663 by copying enterprise fix DH-12258. Removes dead code.
* Fix shortcut broken by #203 Separating dashboard into own package
* Cleanup hjkl shortcuts to not conflict with ctrl+L for linker
* add shift key support

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.jsx
+++ b/packages/code-studio/src/main/AppMainContainer.jsx
@@ -13,6 +13,7 @@ import {
   GLOBAL_SHORTCUTS,
   Popper,
 } from '@deephaven/components';
+import { SHORTCUTS as IRIS_GRID_SHORTCUTS } from '@deephaven/iris-grid';
 import {
   Dashboard,
   DashboardUtils,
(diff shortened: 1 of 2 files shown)


mofojed:
Something's not right here - this _used_ to work. Why are the global context actions not triggering before the grid key handler anymore?

dsmmcken:
https://github.com/deephaven/web-client-ui/pull/664 has context

mofojed (approved):

mofojed:
I've filed another ticket #913 , as I think Linker should have first crack at keyboard events rather than the underlying panels.

PR #914, opened 2022-11-30
Clean up non-semantic uses of defined colours

dsmmcken:
Colors defined with `interface` prefix was intended for use within bootstrap overrides only.

Everywhere else should use either `$foreground`, `$background` or `$content-bg`. If the usage is not semantically relevant, then an explicit `$white` or `$black` is also acceptable.

This might be considered a breaking change as I also removed them from ThemeExport.

### DIFF
--- a/packages/chart/src/Chart.scss
+++ b/packages/chart/src/Chart.scss
@@ -1,6 +1,6 @@
 @import '@deephaven/components/scss/custom.scss';
 
-$chart-bg: $interfacegray;
+$chart-bg: $content-bg;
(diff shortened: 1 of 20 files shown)


mofojed (approved):
Would also be nice to add a comment to `bootstrap_overrides.scss` saying we should avoid `$interfacewhite`/`$interfaceblack` etc. and instead use the semantic names.

dsmmcken:
I intend to re-factor them out of existence entirely.

PR #916, opened 2022-11-30
Subtract the pending row count when displaying row count

### LINKED ISSUE #889: Wrong row numbers for InputTable in the WebUI
image

This was discovered during the debugging of a problem with the new InputTable Python wrapper.
I have already talked to @niloc132 about it and experimented a bit more, the preliminary diagnosis is that UI might confuses with the reserved rows for editing and the actual table size.

mofojed:
- We fudge the `rowCount` when using InputTables to make the infinite scrolling input table effect. Displaying this `rowCount` was confusing to the user, as that's not the actual size of the table.
- Don't include the pending row count in tooltips when displaying row count. 
- Fixes #889

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanelTooltip.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanelTooltip.tsx
@@ -17,7 +17,8 @@ const IrisGridPanelTooltip = (
 ): ReactElement => {
   const { model, widgetName, glContainer, description } = props;
 
-  const formattedRowCount = model?.displayString(model?.rowCount ?? 0, 'long');
+  const rowCount = (model?.rowCount ?? 0) - (model?.pendingRowCount ?? 0);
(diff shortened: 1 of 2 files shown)


vbabich (commented):
Is this all of the `rowCount` uses in the codebase? There might be cases like ```const { columnCount, rowCount } = model;``` in `SelectionKeyHandlers.ts`. Do we need to fix those as well?

mofojed:
> Is this all of the `rowCount` uses in the codebase? There might be cases like `const { columnCount, rowCount } = model;` in `SelectionKeyHandlers.ts`. Do we need to fix those as well?

In most other cases, we do want to use the `rowCount` of the displayed rows so that you can move around the grid correctly, keyboard navigation, scrolling, rendering the cells, etc. It's only in the display of "Number of Rows" where we want to display the _actual_ table size. I think these are the only places where we want that specific value.

vbabich (approved):

PR #920, opened 2022-11-30
Copy unformatted cell for dates should copy full date string

### LINKED ISSUE #919: "Copy Cell Unformatted" on a datetime column should provide a complete datetime
Pulled from #614.

"Copy Cell Unformatted" on a datetime column should provide a complete datetime (e.g. `2022-06-14T15:08:13.000000000 <zone>`), instead of the time offset in nanoseconds. The time offset in nanoseconds is fairly useless.

mofojed:
- Instead of copying the numeric value in nanoseconds of the date, copy it as a full date string
- Tested with a date column, changing time zones, and copying values. The date was correct and time zone was included correctly. Also tested non-date columns still copied as intended.
- Fixes #919

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -184,7 +184,7 @@ const DEFAULT_AGGREGATION_SETTINGS = Object.freeze({
   showOnTop: false,
 });
 
-const UNFORMATTED_DATE_PATTERN = 'yyyy-MM-dd HH:mm:ss.SSSSSSSSS z';
+const UNFORMATTED_DATE_PATTERN = `yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS z`;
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -183,6 +184,8 @@ const DEFAULT_AGGREGATION_SETTINGS = Object.freeze({
   showOnTop: false,
 });
 
+const UNFORMATTED_DATE_PATTERN = 'yyyy-MM-dd HH:mm:ss.SSSSSSSSS z';
```
dsmmcken, line 187: I think the intent is the value would then be immediately usable as part of a query, in which case you'll want a T separator to be included.

dsmmcken (changes requested):

dsmmcken (approved):

PR #921, opened 2022-11-30
Fixes Grid isStuckToBottom prop doesn't work correctly on mount if rowCount doesn't increase

### LINKED ISSUE #581: Grid isStuckToBottom prop doesn't work correctly on mount if rowCount doesn't increase
**Description**

**Steps to reproduce**
1. In the Grids page in Styleguide, make the following change to make the top grid `isStuckToBottom`:
```
diff --git a/packages/code-studio/src/styleguide/Grids.jsx b/packages/code-studio/src/styleguide/Grids.jsx
index d818d31d39..c7666475a1 100644
--- a/packages/code-studio/src/styleguide/Grids.jsx
+++ b/packages/code-studio/src/styleguide/Grids.jsx
@@ -32,7 +32,7 @@ class Grids extends PureComponent {
         <ThemeContext.Provider value={contextTheme}>
           <h2 className="ui-title">Grid</h2>
           <div style={{ height: 500 }}>
-            <Grid model={model} theme={theme} />
+            <Grid model={model} theme={theme} isStuckToBottom />
           </div>
           <h2 className="ui-title">Static Data</h2>
           <div style={{ height: 200 }}>
```
2. Open the style guide and look at the top grid

**Expected results**
2. The grid should be scrolled to the bottom

**Actual results**
2. The grid is at the top

dsmmcken:
Fixes #581 

@mofojed why does lastTop take into account floating rows but lastLeft doesn't? Noticed, because lastTop was not taking into account floating top rows.

### DIFF
--- a/packages/grid/src/GridMetricCalculator.ts
+++ b/packages/grid/src/GridMetricCalculator.ts
@@ -796,9 +796,9 @@ export class GridMetricCalculator {
     visibleWidth: number = this.getVisibleWidth(state)
   ): VisibleIndex {
     const { model } = state;
-    const { columnCount } = model;
+    const { columnCount, floatingRightColumnCount } = model;
 
-    let lastLeft = columnCount - 1;
+    let lastLeft = Math.max(0, columnCount - floatingRightColumnCount - 1);
     if (right != null) {
       lastLeft = right;
     }
@@ -820,15 +820,10 @@ export class GridMetricCalculator {
   /**
    * The last row that can be the top row (e.g. scrolled to the bottom)
    * If no bottom row is provided, then the last row that is not floating is used
-   */
-
-  /**
-   * The last row that can be the top row (e.g. scrolled to the bottom)
-   * If no bottom row is provided, then the last row that is not floating is used
    * @param state The current grid state
    * @param bottom The bottom-most row to be visible, or null to default to last cell
    * @param visibleHeight The height of the "visible" area (excluding floating items)
-   * @returns The index of the last left visible column
+   * @returns The index of the last top visible row
    */
   getLastTop(
     state: GridMetricState,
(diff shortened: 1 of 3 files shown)


packages/grid/src/GridMetricCalculator.ts
```diff
@@ -836,9 +836,12 @@ export class GridMetricCalculator {
     visibleHeight: number = this.getVisibleHeight(state)
   ): VisibleIndex {
     const { model } = state;
-    const { rowCount, floatingBottomRowCount } = model;
+    const { rowCount, floatingTopRowCount, floatingBottomRowCount } = model;
```
mofojed, line 839: It purposely does not take into account the `floatingTopRowCount` here - this is because the `lastTop` is actually scrolled to underneath the floating rows. This is incorrect, you can check this by going to the Iris Grid example in style guide and scrolling to the bottom, you'll be missing ~4 rows.
`getLastLeft` should probably take the `floatingRightColumnCount` into account, and we just haven't noticed that issue because we never use floating right columns.

Can you also fix the comment up for this method? There's a duplicated column block right before this one.

mofojed (changes requested):

packages/grid/src/GridMetricCalculator.ts
dsmmcken: image

Hmm, but if I add isStuckToBottom to that example without accounting for top floating rows it initializes like this though.? If I scroll up then down it's right again though. Not sure why though

dsmmcken: I don't really get why it wasn't working, and how I fixed it.... lastTop was different between the results of `getMetricState` and the value of `lastTop` in `this.metrics`, switched to just use the value of lastTop in metrics, and it seems to be correct now?

dsmmcken: Okay, the root cause was I was calling lastTop with only two params, instead of three. The default one in metrics is called with 3, with a different value for the last param then the default which accounts for the difference.

To make this less likely of an error to occur in future, I have chosen to make them all required params.

same with getLastLeft. Please check what I did to focusFilterBar call to make sure that makes sense.

mofojed (approved):

PR #922, opened 2022-12-01
Add optional parameter choosing whether or not to consolidate ranges in IrisGridTableModel snapshot

### LINKED ISSUE #597: Grid with moved columns copies in incorrect order
**Description**

The copy selection handler copies the original order of the table, not the visible order

**Steps to reproduce**
1. Use this Python code
```python
from deephaven.column import int_col

base = new_table([
   int_col("A", [1]),
   int_col("B", [2]),
   int_col("C", [3]),
   int_col("D", [4]),
   int_col("E", [5]),
   int_col("F", [6])
])
```

2. Drag column B to the right to switch with column C.
3. Select and copy the row
4. Paste somewhere else (like a notebook or the console)

**Expected results**

Pastes 1 3 2 4 5 6

**Actual results**

Pastes 1 2 3 4 5 6

**Versions**
 - Deephaven: 0.12

jason-wang24:
Partial #597

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModelTemplate.ts
+++ b/packages/iris-grid/src/IrisGridTableModelTemplate.ts
@@ -1309,13 +1309,16 @@ class IrisGridTableModelTemplate<
   async snapshot(
     ranges: GridRange[],
     includeHeaders = false,
-    formatValue: (value: unknown, column: Column) => unknown = value => value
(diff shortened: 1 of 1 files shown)


mofojed (approved):
@jason-wang24 this fixes the Copy issue were seeing, but Download CSV is still not taking moving columns (or hidden columns for that matter) into account.

You can still merge this and then split the ticket up, or file a new ticket for CSV.

PR #924, opened 2022-12-02
Stringify the error details before outputting to console

mofojed:
- While looking through logs reported from a user, the error printed out was just `[AppInit] Connection failure Object`. Stringify it so we may have at least a bit more detail.

### DIFF
--- a/packages/code-studio/src/main/AppInit.tsx
+++ b/packages/code-studio/src/main/AppInit.tsx
@@ -167,7 +167,7 @@ const AppInit = (props: AppInitProps) => {
         dh.IdeConnection.HACK_CONNECTION_FAILURE,
         event => {
           const { detail } = event;
-          log.error('Connection failure', `${detail}`);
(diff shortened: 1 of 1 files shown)


packages/code-studio/src/main/AppInit.tsx
```diff
@@ -167,7 +167,7 @@ const AppInit = (props: AppInitProps) => {
         dh.IdeConnection.HACK_CONNECTION_FAILURE,
         event => {
           const { detail } = event;
-          log.error('Connection failure', detail);
+          log.error('Connection failure', `${detail}`);
```
vbabich, line 170: Should probably do `JSON.stringify` if `detail` is an object, otherwise we'll just get `'[object Object]'`
```suggestion
          log.error('Connection failure', `${JSON.stringify(detail)}`);
```

vbabich (changes requested):

vbabich (approved):

Issue #843, opened 2022-10-25
EPIC: Linker improvements


dsmmcken:
We have a several improvements to linker that we plan to make:
  
## Linker comparison operators:

The purpose of these improvements is to provide users with more flexibility with creating links. Specifically, the combination of setting a comparison operator per link, with conjunctively combining links from the same source table linked to the same target column, will allow the users to create new powerful linker driven filters, such as their own "range" filter.

Ex. User has Table A with columns: `id`, `start_time`, `end_time`, and Table B with columns: `id` and `timestamp`. By linking `id` to `id`, and `start_time` <= to `timestamp` and  `timestamp` <= `end_time` the user would be able to filter Table B based on a range provided from Table A.

- [x] #887
  - [x] Add new default styling $blue, mix($blue, $content-bg, 25%), hover styling ($primary, $primary-hover), and selected styling is previous default styling as per mockup:
  ![image](https://user-images.githubusercontent.com/1576283/197881574-4389d10b-a070-4353-b0e0-c5115c2371dc.png)
  - [x] Add support for setting filter comparison operator on links
  ![image](https://user-images.githubusercontent.com/1576283/197880900-743bdf83-3bcf-4dca-83e1-ec704b8d513b.png)
  Note: linker colors are incorrect in this mockup, see previous point for updated style
  - [x] Click to delete no longer applies, must now be selected first then deleted. Delete being a two-step operation I don't like as much, so open to suggestions. Maybe at least an alt-click (with delete mode cursor during alt press?)
  - [x] Links originating from multiple source columns in the same table, going to the same target column in target table should be combined. Conjunctively with && and unless one of the source columns is set to an equals operator, in which case use disjunctive || or filter.
  - [x] Comparison operators and their combination should still show correctly in the quick filter bar.

## Other minor ux improvements: 
- [x] #875
- [x] #1035
- [x] When hovering a link, it should be brought to front, so the arrow and ball on each of the link is on top of the hover stack
- [x] #964
  ![image](https://user-images.githubusercontent.com/1576283/207144544-68241871-2f5d-4c47-b7e3-3bd06664a1bc.png)
- [x] #906
- [x] #901
  ![image](https://user-images.githubusercontent.com/1576283/197880089-8310cc91-97a2-449e-ae5e-0a5a92db106a.png)
  Note: allow left edge to act as grab handle, and be moveable anywhere in bounds.

devinrsmith:
[...]

PR #926, opened 2022-12-05
Add buttons to linker, ability to change the comparison operator, and link combination

### LINKED ISSUE #887: Add ability to change comparison operator for linkers with floating action buttons on select

jason-wang24:
Fixes #887 
Supercedes PR [https://github.com/deephaven/web-client-ui/pull/896](url)
See the issue for more information: [https://github.com/deephaven/web-client-ui/issues/843](url)

- Linkers can be deleted with alt + click
- New styling for linkers based on selection state, and danger when alt is held
- removed outer LinkerOverlay svg
- moved operator building from linker to irisgrid
- null values are sent and combined appropriately
- sort filters by column left to right when combining them in the quick filter
- the left to right order adapts to when columns are moved
- changed irisgrid test to match new changes
- set other links to not selected state when building a new one (only one link is selected at one time)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1484,20 +1494,67 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       }
       const columnIndex = model.getColumnIndexByName(column.name);
       assertNotNull(columnIndex);
-      if (value === undefined) {
-        this.removeQuickFilter(columnIndex);
-      } else if (value === null) {
-        this.setQuickFilter(columnIndex, column.filter().isNull(), '=null');
+      filterList.sort((a, b) => {
```
mofojed, line 1497: This whole function could be extracted into a utils function, add some unit testing around it.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2467,10 +2524,18 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
         isExpandableGridModel(model) && model.isRowExpandable(rowIndex);
       const isGrouped = groupedColumns.find(c => c.name === name) != null;
-      dataMap[name] = { value, text, type, isGrouped, isExpandable };
+      dataMap[name] = {
+        value,
+        text,
+        type,
+        isGrouped,
+        isExpandable,
+        visibleIndex,
```
mofojed, line 2537: I'm not crazy about `visibleIndex` being part of this, but I guess it works.

packages/jsapi-utils/src/TableUtils.ts
```diff
@@ -52,6 +52,7 @@ export type RowDataMapValue = {
   value: unknown;
   isExpandable: boolean;
   isGrouped: boolean;
+  visibleIndex: number;
 };
 
 export type RowDataMap = Record<string, RowDataMapValue>;
```
mofojed, line 58: We could be clearer about `RowDataMap` by using the type `ColumnName` (imported from Formatter) instead of just `string`.

```suggestion
export type RowDataMap = Record<ColumnName, RowDataMapValue>;
```

mofojed (changes requested):

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -109,13 +124,37 @@ export class LinkerOverlayContent extends Component<
...
+    const { links, onLinksUpdated } = this.props;
+    const newLinks: Link[] = links.map(link =>
+      link.id === linkId ? { ...link, operator: type } : link
+    );
+    onLinksUpdated(newLinks);
+  }
+
   handleMouseMove(event: MouseEvent): void {
     this.setState({
       mouseX: event.clientX,
```
mofojed, line 137: I think you should set the `mode` here as well based on if the `altKey` is pressed... Or listen for when the window goes out of focus and change the mode back.
Reason being when I alt-tab out of this window, the window never gets the AltUp, so when I come back, it still looks like it's going to delete it.

packages/dashboard-core-plugins/src/linker/LinkerLink.scss
```diff
@@ -12,89 +12,144 @@ $dash-delete-color-1: $danger;
...
     }
-    path {
-      stroke-width: $link-width;
-      stroke: $invalid-dash-color-2;
-      &.link-select {
-        stroke-width: $selectable-width;
-        stroke: transparent;
+
+    &.link-is-selected {
+      .link-dot,
```
mofojed, line 62: You'll want to add a `z-index: 10` here or something so that the selected link appears overtop of the other links.

packages/dashboard-core-plugins/src/linker/LinkerLink.scss
```diff
@@ -12,89 +12,144 @@ $dash-delete-color-1: $danger;
 $dash-delete-color-2: mix($danger, $content-bg, 25%);
 $dash-size: 10;
 
-.linker-overlay svg {
-  .link-dot,
-  .link-triangle {
-    fill: $dash-color-1;
+.linker-overlay {
+  .btn-fab {
+    pointer-events: all;
```
mofojed, line 17: You'll want a `z-index: 20` here as well so buttons appear above any other links

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1495,6 +1495,69 @@ class IrisGridUtils {
...
+
+    let combinedText = '';
+    for (let i = 0; i < filterList.length; i += 1) {
+      let { operator } = filterList[i];
+      const { text, value } = filterList[i];
+      if (value !== undefined) {
+        let symbol = '';
+        if (operator !== undefined) {
+          if (value == null && operator !== 'notEq') {
+            symbol = '=';
```
mofojed, line 1531: I don't think this is right. If I have the `>` or `<` operator selected, and then double click a null value, I'm getting a filter `=null` which doesn't seem right? Or I guess that's intentional and what @dsmmcken asked for?
If we're doing that for a reason, should put a comment here. And with the test case that tests that behaviour as well, it doesn't seem intuitive to me (I would've expected `>null` which would be invalid of course, so then just dropping that part of the filter...)

mofojed (changes requested):

packages/iris-grid/src/IrisGridUtils.ts
dsmmcken: Confirming behaviour should be as bender said, put the invalid link through so it visually fails to the user.

packages/dashboard-core-plugins/src/linker/LinkerLink.scss
```diff
@@ -12,89 +12,151 @@ $dash-delete-color-1: $danger;
...
+    position: absolute;
+    min-width: 1rem;
+    min-height: 1rem;
+    padding: 0.25rem 0.45rem;
+    z-index: 30;
   }
```
dsmmcken, line 23:
```suggestion
  .btn-fab {
    position: absolute;
    display: flex;
    justify-content: center;
    align-items: center;
    pointer-events: all;
    min-width: $input-height;
    min-height: $input-height;
    height: $input-height;
    width: $input-height;
    line-height: $font-size-lg; // vertical centering for text inside fa-layer
    padding: 0;
    margin: 0;
    z-index: 30;
    box-shadow: $box-shadow;
    .fa-layers {
      display: inline-flex;
      justify-content: center;
      b {
        // comparison operator
        font-size: $font-size-lg;
      }
    }
  }
```

- Button insides weren't centered
- Try to use existing variables instead of hardcode values
- Increased text size
- Added box-shadow

packages/dashboard-core-plugins/src/linker/LinkerLink.tsx
```diff
@@ -106,22 +250,117 @@ export class LinkerLink extends PureComponent<LinkerLinkProps> {
...
+              }}
+              onClick={() => {
+                // no-op: click is handled in `DropdownMenu'
+              }}
+              icon={
+                <div className="fa-md fa-layers">
+                  <b>{symbol}</b>
+                  <FontAwesomeIcon
+                    icon={vsTriangleDown}
+                    transform="right-8 down-10 shrink-6"
```
dsmmcken, line 341:
```suggestion
                    transform="right-8 down-9 shrink-5"
```

Minor visual tweak.

packages/dashboard-core-plugins/src/linker/LinkerLink.tsx
```diff
@@ -106,22 +250,117 @@ export class LinkerLink extends PureComponent<LinkerLinkProps> {
...
+                popperOptions={{ placement: 'bottom-start' }}
+              />
+            </Button>
+
+            <Button
+              kind="primary"
+              className="btn-fab btn-delete"
+              style={{ top: midY + deleteOffsetY, left: midX + deleteOffsetX }}
+              onClick={this.handleDelete}
+              icon={<FontAwesomeIcon icon={vsTrash} transform=" down-1" />}
```
dsmmcken, line 358:
```suggestion
              icon={vsTrash} />}
```

Shouldn't need any positional offset with the CSS centering change I shared above.

dsmmcken (changes requested):
1. Holding alt seems super inconsistent with the visual behaviour. Cursor and hover state flickers. Is the hover being done with JS instead of CSS only? Or is oscillating between states on move or something? Some times it doesn't even show at all.
flicker

2, When a link is going down does it seem weird that the comparison operator is now on the "right"? I wonder if it should always be the left most position. It seems like it switches places in this case?
image

packages/dashboard-core-plugins/src/linker/LinkerLink.tsx
```diff
@@ -14,6 +14,7 @@ import { TableUtils } from '@deephaven/jsapi-utils';
 import memoize from 'memoize-one';
 
 import './LinkerLink.scss';
+import classNames from 'classnames';
```
mofojed, line 17: I thought we had a rule that enforced import order... but in any case, put this order at the top after the `react` import.
We import from packages first, then from local files (in this case, `LinkerLink.scss` is a local file, where `classNames` is from `node_modules` and should be at the top)

packages/dashboard-core-plugins/src/linker/LinkerLink.scss
```diff
@@ -119,6 +122,10 @@ $dash-size: 10;
       pointer-events: none;
     }
 
+    &.hovering {
```
mofojed, line 125: I think you can just use `&:hover` and skip all the JS stuff adding this class

mofojed (changes requested):
Shouldn't the "Delete" key delete the link as well when the link is selected? That doesn't seem to be working right now.

packages/dashboard-core-plugins/src/linker/LinkerLink.tsx
mofojed: Still need to move this import up

mofojed (changes requested):
Move that classnames import up, then I'm good to merge this.
Can have another PR for making the "Delete" key delete the selected link.

mofojed (approved):

dsmmcken (approved):
:shipit:

PR #933, opened 2022-12-07
Fix tsconfig to type check code studio

mattrunyon:
Code studio was not being type checked on builds (errors would still show up in VSCode if you had the file open). This should fix the config so code-studio is type checked on build. Also fixes a few TS errors from code-studio

### DIFF
--- a/packages/code-studio/src/main/AppControlsMenu.tsx
+++ b/packages/code-studio/src/main/AppControlsMenu.tsx
@@ -151,7 +151,7 @@ const AppControlsMenu = (props: AppControlsMenuProps): ReactElement => {
         title: 'Input Filter',
         icon: dhInput,
         menuElement: <DragSourceMenuItem />,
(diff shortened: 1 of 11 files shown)


dsmmcken:
Is this also the case with other packages? I just came across a typescript error in layoututils in packages/dashboard when I opened it.

mofojed (dismissed):

mattrunyon:
> Is this also the case with other packages? I just came across a typescript error in layoututils in packages/dashboard when I opened it.

I'm not sure why that shows up in vscode but not in the type check. I changed a random `return;` to `return false;` in that file and it flagged it, so the file is being checked. Not sure why vscode has different results when it's set to the workspace version of typescript

dsmmcken:
I was talking about the dragEvent in that file, it's the wrong dragEvent type, should be based on the golden-layout jquery one

mofojed (approved):

PR #935, opened 2022-12-07
Adds copy button to version tooltip

dsmmcken:
Lol, I just wanted to added a copy version info button to the version tooltip...

- Moved ContextActionsUtils.CopyToClipboard to CopyClipboardUtils
- Created useCopyToClipboard hook
- Create CopyButton component using useCopyToClipboard
- Replaced instances of existing copy buttons with CopyButton
- Additonal refactoring as required
- Finally, add copy button to settings version tooltip


packages/components/src/CopyButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * props:
+ * @param {string} copy the value to copy when clicked
+ * @param {string} ButtonKind the kind of button
+ * @param {string} tooltip optional tooltip label ex. 'Copy column name'
+ * @param {string} className optional extra classname
+ * @param {string} data-testid optional extra testid
+ *
+ *
+ */
+
```
mofojed, line 13: Why'd you put the comments at the top of the file? This won't appear in Intellisense.

Instead, what you should do is put the comments above the component and the props themselves, then it'll appear. Something like:
```
type CopyButtonProps = {
  /** The value to copy when clicked */
  copy: string;
  ...
}

/**
 * Copy button that has an icon change when clicked, and copies the string passed in.
 */
const Copy Button = ({...}: CopyButtonProps): ...
```

packages/utils/src/CopyClipboardUtils.ts
```diff
@@ -0,0 +1,43 @@
+class CopyClipboardUtils {
```
mofojed, line 1: I would just rename this file `ClipboardUtils`, no need to specify Copy only.

I know this is how we've done Utils before, but I've kind of changed my ways as I've learned. Instead of using a class and doing static methods, do something like we do in `SessionUtils` where we export each function. So in this case, it would be something like:

```
export function copyToClipboard(...) { ... }

export function copyToClipboardExecCommand(...) { ... }
```
You don't even need an export Default; if another file wants to import all the functions, they can do: `import * as ClipboardUtils from './ClipboardUtils'`. But most of the time it'll just be `import { copyToClipboard } from './ClipboardUtils'` or something.

Some discussions on this topic:
https://www.typescriptlang.org/docs/handbook/modules.html#export-as-close-to-top-level-as-possible
https://stackoverflow.com/questions/29893591/es6-modules-export-single-class-of-static-methods-or-multiple-individual-method/29895235#29895235
https://olegvaraksin.medium.com/what-is-the-best-way-to-write-utilities-in-typescript-e3cae916fe30

packages/iris-grid/src/ColumnStatistics.tsx
```diff
@@ -211,16 +194,10 @@ class ColumnStatistics extends Component<
...
           <span className="column-statistics-type">&nbsp;({columnType})</span>
-          <Button
-            kind="ghost"
+          <CopyButton
             className="column-statistics-copy"
-            icon={copied ? vsPassFilled : vsCopy}
-            onClick={() => {
-              ContextActionUtils.copyToClipboard(column.name)
-                .then(this.handleCopyHeader)
-                .catch(e => log.error('Unable to column name', e));
```
mofojed, line 221: What a dumb error handler this was.

packages/utils/src/CopyClipboardUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+  static copyToClipboard(text: string): Promise<void> {
+    const { clipboard } = navigator;
+    if (clipboard === undefined) {
+      CopyClipboardUtils.copyToClipboardExecCommand(text);
+      return Promise.resolve();
+    }
+    return navigator.clipboard.writeText(text).catch(() => {
+      CopyClipboardUtils.copyToClipboardExecCommand(text);
+    });
+  }
```
mofojed, line 16: This could be written as an `async` function - right now `copyToClipbardExecCommand` could throw before a promise is returned and that would screw up a consumer. Something like 
```
  static async copyToClipboard(text: string): Promise<void> {
    const { clipboard } = navigator;
    if (clipboard !== undefined) {
      try {
        return navigator.clipboard.writeText(text);
      } catch (e) {
        // fallback to copyToClipboardExecCommand
      }
    }
    copyToClipboardExecCommand(text);
  }
```

packages/components/src/CopyButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useCopyToClipboard } from '@deephaven/react-hooks';
+import Button, { ButtonKind } from './Button';
+
+type CopyButtonProps = {
+  copy: string;
+  kind?: ButtonKind;
+  tooltip?: string;
+  className?: string;
+  'data-testid'?: string;
+  children?: React.ReactNode;
```
mofojed, line 25: Should have an optional error handler for if the copy fails. It's possible for execCommand to fail if it's called from a timer, for example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#using_execcommand
There may be other cases with a permission locked browser, not sure.

packages/components/src/CopyButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const [copied, copyToClipboard] = useCopyToClipboard();
+  return (
+    <Button
+      kind={kind}
+      className={className}
+      data-testid={dataTestId}
+      icon={copied ? vsPassFilled : vsCopy}
+      tooltip={copied ? 'Copied' : tooltip}
+      onClick={() => {
+        copyToClipboard(copy);
```
mofojed, line 45: Need to catch error here.

packages/react-hooks/src/useCopyToClipboard.ts
```diff
@@ -0,0 +1,34 @@
...
+const log = Log.module('useCopyToClipboard');
+
+export default function useCopyToClipboard(): [
+  boolean,
+  (value: string) => void
+] {
+  const [copied, setCopied] = useState(false);
+  const copiedTimeout = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  function copyToClipboard(value: string): void {
```
mofojed, line 15: `useCallback` probably

packages/react-hooks/src/useCopyToClipboard.ts
```diff
@@ -0,0 +1,34 @@
...
+    CopyClipboardUtils.copyToClipboard(value)
+      .then(() => {
+        setCopied(true);
+      })
+      .catch(e => log.error(`Unable to copy ${value}`, e));
```
mofojed, line 20: Ahh we're just eating the error here? Hmmm okay...

packages/react-hooks/src/useCopyToClipboard.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  useEffect(() => {
+    if (copiedTimeout.current) clearTimeout(copiedTimeout.current);
+    copiedTimeout.current = setTimeout(() => {
+      setCopied(false);
+    }, TIMEOUT_DELAY);
```
mofojed, line 27: Should only need to set this `if (copied)`, if we transition to `!copied` doesn't need to set a timer again.

packages/components/src/CopyButton.tsx
dsmmcken: no, this returns void. Error is eaten elsewhere.

packages/react-hooks/src/useCopyToClipboard.ts
dsmmcken: Yep. User feedback is that setCopied is never true in that case.

packages/components/src/CopyButton.tsx
dsmmcken: Copy button should handle it internally, not users of the component. And I don't think I have anything I want to do here other than the checkmark never appears, which is what it does.

mofojed (changes requested):
You also need to fix the unit tests

mofojed (approved):

PR #937, opened 2022-12-08
Allow embed-grid to open pandas dataframes

mofojed:
- Pass in an optional `type` parameter that specifies the variable type
- No need for reload button in this case/marking it like a pandas dataframe like we do in panels, just use the browser refresh
- Preparation for https://github.com/deephaven/deephaven-ipywidgets/issues/4 (which I'll get Joseph to do)

Tested by executing the following code:
```
import pandas as pd

data_frame = pd.DataFrame(
    {'A': [1, 2, 3],
     'B': ['X', 'Y', 'Z']}
)

t = empty_table(100).update("x=i")
```

Then navigated to the following two addresses, made sure they both appeared correctly:
http://localhost:4010/?name=data_frame
http://localhost:4010/?name=t

### DIFF
--- a/packages/embed-grid/src/App.tsx
+++ b/packages/embed-grid/src/App.tsx
@@ -6,13 +6,8 @@ import {
   IrisGridModel,
   IrisGridModelFactory,
 } from '@deephaven/iris-grid'; // iris-grid is used to display Deephaven tables
-import dh, {
-  IdeConnection,
-  Sort,
-  Table,
-  VariableTypeUnion,
-} from '@deephaven/jsapi-shim'; // Import the shim to use the JS API
-import { TableUtils } from '@deephaven/jsapi-utils';
+import dh, { IdeConnection, Sort, Table } from '@deephaven/jsapi-shim'; // Import the shim to use the JS API
+import { fetchVariableDefinition, TableUtils } from '@deephaven/jsapi-utils';
 import Log from '@deephaven/log';
 import './App.scss'; // Styles for in this app
 
@@ -42,12 +37,11 @@ export type SortCommandType = {
  */
 async function loadTable(
   connection: IdeConnection,
-  name: string,
-  type: VariableTypeUnion = dh.VariableType.TABLE
+  name: string
 ): Promise<Table> {
   log.info(`Fetching table ${name}...`);
 
-  const definition = { name, type };
+  const definition = await fetchVariableDefinition(connection, name);
   return (await connection.getObject(definition)) as Table;
 }
 
@@ -71,8 +65,6 @@ function App(): JSX.Element {
   );
   const canCopy = searchParams.get('canCopy') != null;
   const canDownloadCsv = searchParams.get('canDownloadCsv') != null;
-  const type = (searchParams.get('type') ??
-    dh.VariableType.TABLE) as VariableTypeUnion;
 
   useEffect(
     function initializeApp() {
@@ -99,7 +91,7 @@ function App(): JSX.Element {
           log.debug('Loading table', name, '...');
 
           // Load the table up.
-          const table = await loadTable(connection, name, type);
+          const table = await loadTable(connection, name);
 
           // Create the `IrisGridModel` for use with the `IrisGrid` component
           log.debug(`Creating model...`);
@@ -117,7 +109,7 @@ function App(): JSX.Element {
       }
       initApp();
     },
-    [searchParams, type]
+    [searchParams]
   );
 
   useEffect(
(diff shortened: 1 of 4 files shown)


packages/embed-grid/src/App.tsx
```diff
@@ -61,6 +71,8 @@ function App(): JSX.Element {
   );
   const canCopy = searchParams.get('canCopy') != null;
   const canDownloadCsv = searchParams.get('canDownloadCsv') != null;
+  const type = (searchParams.get('type') ??
+    dh.VariableType.TABLE) as VariableTypeUnion;
 
   useEffect(
     function initializeApp() {
```
mattrunyon, line 78: The majority of this hook could probably be pulled into a custom hook like `const connection = useConnection({ url, credentials, onSuccess, onError })` or something. But that's for another day

mattrunyon (approved):
Kind of annoying the type has to be in the URL, but I guess that's a limitation of the API. Not sure why when you shouldn't be able to have multiple objects with the same name but different types? 1 variable => 2 values isn't possible AFAIK

mofojed:
> Kind of annoying the type has to be in the URL, but I guess that's a limitation of the API. Not sure why when you shouldn't be able to have multiple objects with the same name but different types? 1 variable => 2 values isn't possible AFAIK

Actually yea - I could do `subscribeToFieldUpdates`, and get the variable type from there, but that's also a cumbersome API (would have to add a timeout for when we don't get the updates or something). Maybe I should do that instead?

mattrunyon:
If you can make it not require the datatype in the URL, I think that's preferred from a user standpoint. I would say depends how much effort that will be vs how we think this will be used. Are people embedding grids w/ URLs themselves, or is it mostly for future stuff like Jupyter where we'll be creating the URL for them and they never see it?

packages/jsapi-utils/src/ConnectionUtils.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+beforeEach(() => {
+  mockRemoveListener.mockClear();
+  mockSubscribeToFieldUpdates.mockClear();
+});
+
+it('finds the right definition if variable exists', async () => {
+  const fetchPromise = fetchVariableDefinition(
+    connection,
+    testDefinition1.title ?? ''
```
mattrunyon, line 39: Why `?? ''`?

packages/jsapi-utils/src/ConnectionUtils.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  const result = await fetchPromise;
+
+  expect(result).toBe(testDefinition1);
+  expect(mockRemoveListener).toHaveBeenCalled();
+});
+
+it('finds the definition in the second update if not after the first update', async () => {
+  const fetchPromise = fetchVariableDefinition(
+    connection,
+    testDefinition2.title ?? ''
```
mattrunyon, line 62: Why nullish coalesce again

packages/jsapi-utils/src/ConnectionUtils.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  const result = await fetchPromise;
+
+  expect(result).toBe(testDefinition2);
+  expect(mockRemoveListener).toHaveBeenCalled();
+});
+
+it('throws a TimeoutError if variable not found', async () => {
+  const fetchPromise = fetchVariableDefinition(
+    connection,
+    testDefinition2.title ?? ''
```
mattrunyon, line 93: `?? ''`

mattrunyon (changes requested):
Need to test locally and my only comments are minor

I'm guessing if you were to pass it the name of an existing Figure, it would just show an error?

packages/jsapi-utils/src/ConnectionUtils.test.ts
mofojed: This would be a case where the `satisfies` operator would be handy.
`testDefinition1` is defined as a `VariableDefintion`, where `title` is optional; even though I very clearly define it prior to this. I'll rewrite this some other way to be cleaner.

mattrunyon: You just don't need to define the type on `testDefinition1` and `testDefinition2`. TS will figure it out anywhere that it needs a `VariableDefinition`. If in the future `VariableDefinition` changes, anywhere that requires it within the test will complain since the variables won't match the updated type

mofojed: Yea I just don't need to define the type. I think I did originally for autocomplete. D'oh

mofojed:
> I'm guessing if you were to pass it the name of an existing Figure, it would just show an error?

Schrodinger's guess... yes, it does show an error (now) ;)

mattrunyon (approved):

PR #950, opened 2022-12-15
Integrate Playwright into GitHub Actions

### LINKED ISSUE #825: GitHub action integration for running end-to-end/Playwright tests on each PR

mofojed:
- Update tests to start up production version of app for testing if not already running. If you're already running `npm start` locally, it won't try and start it up and will just use that.
- Increase some timeouts that were triggering (some commands take a long time to run).
- Update `baseUrl` so goto links can be relative.
- Cache node modules based on package-lock.json. Don't run `npm ci` if it's cached.
- Pull the specified version of core and run it. Only pull if it has changed.
- Update version of playwright.
- Install playwright browsers and run playwright tests.
- Add some delays for typing in Monaco fix some flakey tests.
- Update README instructions for updating snapshots using a local docker container.
- Update snapshots using the instructions provided.
- Only retain video on failure.
- Tested a bunch of runs on https://github.com/mofojed/web-client-ui/actions/workflows/node.js.yml
- Fixes #825


playwright.config.ts
```diff
@@ -13,32 +13,36 @@ import { devices } from '@playwright/test';
...
-  /* Run tests in files in parallel */
-  fullyParallel: true,
+  /* We don't want to run tests in parallel because we have the same backend, tests may conflict with eachother */
+  fullyParallel: false,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Global setup file for initializing before all other tests */
   globalSetup: require.resolve('./tests/globalSetup.ts'),
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
```
mattrunyon, line 31: What is this retrying? Could this prevent heisenbugs from being detected in CI unless they fail 3 times in a row?

.github/workflows/node.js.yml
```diff
@@ -8,15 +8,31 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-
+env:
+  DOCKER_TAG: 0.19.1
```
mattrunyon, line 12: This seems brittle. Why do we need to specify this? Can we instead have an override + default value to latest?

.github/workflows/node.js.yml
```diff
@@ -53,24 +57,33 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-lintcache-
 
+      - name: Cache node modules
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: |
+            node_modules
+            packages/*/node_modules
```
mattrunyon, line 66: How big is this cache? I think there's a cache explorer now for github actions. We might want to ignore something like `node_modules/.cache` if it's too big. Although I don't think it should be that big since we're no longer using webpack and terser which create huge (1GB+) caches

This is 1 area that yarn (with PnP mode) or pnpm have some advantages. They don't extract the package contents so there's significantly less file overhead. Only a few hundred files instead of (often) tens of thousands. But that would be a completely different ticket and maybe not worth changing package managers if they cause a headache

.github/workflows/node.js.yml
```diff
@@ -53,24 +57,33 @@ jobs:
...
+          key: node-modules-${{ hashFiles('package-lock.json')}}
+
       - name: Install dependencies
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci --no-audit
 
       - name: Build
-        run: npm run build
-
+        run: VITE_CORE_API_URL=http://localhost:10000/jsapi npm run build
```
mattrunyon, line 74: If running `npm run preview:app` for e2e, this might not be necessary. The Vite preview server runs w/ the proxies which default to `localhost:10000`

.github/workflows/node.js.yml
```diff
@@ -53,24 +57,33 @@ jobs:
...
       # Run unit tests on all files for accurate code coverage
       # Lint only changed files so linting is faster
       # Caching with the absolute path b/c Jest will make a folder in each project
       # Then there's caches in all package folders
-      - name: Test (Pull Request)
+      - name: Unit Test (Pull Request)
```
mattrunyon, line 79: Might be worth making unit tests a separate action that can run in parallel since it doesn't require build

.github/workflows/node.js.yml
```diff
@@ -83,3 +96,15 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           directory: ./coverage
           flags: unit
+
+      - name: Install Playwright dependencies
+        run: PLAYWRIGHT_BROWSERS_PATH=0 npx playwright install --with-deps
+      - name: Run Playwright tests
+        run: PLAYWRIGHT_BROWSERS_PATH=0 npx playwright test
+      - name: Upload Playwright report
```
mattrunyon, line 104: Can you put a blank line above each of the `name` blocks so it visually breaks the steps up a bit?

.github/workflows/node.js.yml
```diff
@@ -83,3 +96,15 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 14
```
mattrunyon, line 110: Any reason to not just leave this as the default maximum of 90 days?

README.md
```diff
@@ -115,3 +115,48 @@ When adding new features, it can be useful to analyze the final package size to
 npm run build
 npx source-map-explorer 'packages/code-studio/build/static/js/*.js'
 ```
+
+## Updating Snapshots
+
+Snapshots are used by end-to-end tests to visually verify the output. Sometimes changes are made requiring snapshots to be updated. Since snapshots are platform dependent, you may need to use a docker image to [update snapshots for CI](https://playwright.dev/docs/test-snapshots). You mount the current directory into a docker image and re-run the tests from there.
+
+First start with a clean build. `node_modules` and some other build output is platform dependent.
```
mattrunyon, line 123:
```suggestion
First start with a clean repo. `node_modules` and some other build output is platform dependent.
```

README.md
```diff
@@ -115,3 +115,48 @@ When adding new features, it can be useful to analyze the final package size to
...
+Next, use nvm to install and use the correct version of node:
+
+```
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.2/install.sh | bash
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+nvm install
+```
+
+Install npm dependencies and build the production app, pointing to an API running on your machine:
```
mattrunyon, line 151: Do you not have to copy the repo files into docker?

Would it be possible to combine several of these steps into a dockerfile instead?

mattrunyon (changes requested):

package.json
```diff
@@ -10,7 +10,6 @@
     "npm": ">=8"
   },
   "scripts": {
-    "postinstall": "run-s build:necessary",
```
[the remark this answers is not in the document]
mofojed: I tried with a fresh clone locally, and it's fine. Tried each of these cases:
1) Fresh clone, `npm install`, `npm test`. All tests passed
2) Fresh clone, `npm install`, `npm start`. It all starts up fine.
3) Fresh clone, `npm install`, `npm run build`. It builds fine.

There are already `prestart`, `pretest` steps that do this, and the `build:necessary` is built into the `build` step. I think this `postinstall` script is just redundant and no longer necessary. When do you think it is necessary?

package.json
```diff
@@ -32,8 +31,11 @@
     "build:necessary": "lerna run build --scope=@deephaven/icons",
     "analyze": "lerna run analyze",
     "preview": "lerna run --scope=@deephaven/{code-studio,embed-chart,embed-grid} preview --stream",
+    "preview:app": "lerna run --scope=@deephaven/code-studio preview --stream",
     "prestart": "npm run build:necessary",
     "start": "run-p watch:types start:*",
+    "start:e2e": "run-p start:docker start:app",
+    "start:docker": "docker run --detach --publish 10000:10000 ghcr.io/deephaven/server:latest",
```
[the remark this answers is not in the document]
mofojed: I added this into the GitHub action. I removed these from the npm startup scripts, since they're not actually used (and they conflicted with normal `npm start` functionality).

playwright.config.ts
mofojed: Yes, it retries on CI to prevent heisenbugs/flakey tests. The Playwright Report that is still output includes information on flakey tests. We could reduce it to 1 retry only... I don't think any of the tests should be flakey anymore right now (after this change).

.github/workflows/node.js.yml
mofojed: Yea I debated between `latest` and `0.19.1`. `latest` works fine, I just selected `0.19.1` so that our tests wouldn't block all our PRs if core put in a breaking change... Really we should have some tests running against `edge` in `core` so that they break if something bad is checked in.
I'll change this to `latest` for now (since it should be latest stable version).

.github/workflows/node.js.yml
mofojed: This cache is ~420MB:
```
node-modules-a852c260fe0187e618cdb16aaed716abc473d2fa7a857c0f7695f7b489a0e629
420 MB cached 10 hours ago
```
It saves ~1 minute to use a cache instead of doing an `npm ci` each time.

.github/workflows/node.js.yml
mofojed: This wasn't working for me without specifying the core api URL when I was testing. I'll try it one more time to see if it breaks.

.github/workflows/node.js.yml
mofojed: Good call, these can definitely run in parallel.

.github/workflows/node.js.yml
mofojed: 90 seems excessive, the Playwright yml defaulted to 30 and I knocked it down to 14. I guess there's no reason not to though, so I've bumped it up to 90.

README.md
mofojed: It adds `$(pwd)` as a volume, so don't need to copy any files into docker.
These steps could definitely be put into a Dockerfile instead, because these steps are brutal. I leave that as a follow-up exercise.

.github/workflows/node.js.yml
mofojed: With me splitting these into two different tests (e2e and unit), I'll need to update `infra` as well to make them both required status checks (instead of the one just called `build` now).

.github/workflows/node.js.yml
mofojed: Pretty sure removing this causes the tests to just stall: https://github.com/deephaven/web-client-ui/actions/runs/3704468319/jobs/6277124073

I'll add it back in a bit.

package.json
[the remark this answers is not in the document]
mattrunyon: It is necessary for

1. Fresh clone, `npm install`, browse in VSCode without running any other scripts
2. `npm run clean`, `npm install`, browse code without running other scripts

In these cases, icons aren't built so VSCode will error. But if you run start/test then should be fine and it should be rare those cases happen

.github/workflows/node.js.yml
mattrunyon: You can try with `VITE_PROXY_URL=http://localhost:10000` when you run preview, but this should be fine too.

Actually without the proxy, things like notebooks might be broken since the notebook URL is proxied to port 10000 only if the proxy url is specified and not running w/ Docker. `preview` runs in production mode, so there's no proxy url by default

README.md
mattrunyon: Can you make a ticket for this

.github/workflows/node.js.yml
mofojed: The e2e tests check notebook saving/opening functionality, so notebooks seem to be working

README.md
mofojed: https://github.com/deephaven/web-client-ui/issues/960

.github/workflows/node.js.yml
mofojed: I'll name one `build`, and update the description.

package.json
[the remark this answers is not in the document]
mofojed: I'll just add it back since it only takes a few seconds to build.

mattrunyon (approved):

PR #951, opened 2022-12-15
Update eslint to v8

### LINKED ISSUE #382: Update @typescript-eslint to v5+
Relies on #381 because the @typescript-eslint v5 packages require node `^14.17.0 || >=16.0.0`

TypeScript 4.5 is not in the support list for @typescript-eslint v4 and it prints a warning when the eslint tests run. Doesn't seem to affect the linter right now, but probably better to do this sooner rather than later.

There is 1 breaking change in v5 [CHANGELOG](https://github.com/typescript-eslint/typescript-eslint/blob/main/CHANGELOG.md#500-2021-10-11)

Might also help to set `parserOptions.project` to a glob to find all `tsconfig.json` files and then removing paths. Need to test if eslint can resolve modules that aren't built w/ this change though.

mattrunyon:
Fixes #382 

Upgrades eslint to v8 and corresponding plugins/configs to latest versions

These 4 rules were disabled (likely added by one of the config upgrades) and may or may not be worth re-enabling in the future

```
react/no-unused-class-component-methods // iris-grid heavily violates this to make imperative calls
react/default-props-match-prop-types // This seems to conflict w/ TS since it requires you mark default props optional
react/require-default-props // Requires default props for all optional props. Kind of messes with TS
react/jsx-no-bind // Can be annoying with basic keyboard handlers where memoization is more overhead than gain
```


package.json
```diff
@@ -72,7 +72,7 @@
     "@types/bootstrap": "^4.4.1",
     "@types/color-convert": "^2.0.0",
     "@types/deep-equal": "^1.0.1",
-    "@types/eslint": "^7.22.0",
+    "@types/eslint": "^8.0.0",
```
mofojed, line 75: Any reason not to specify the latest version 8.4.10?

packages/components/src/BasicModal.tsx
```diff
@@ -142,7 +142,7 @@ const BasicModal: React.FC<BasicModalProps> = props => {
       </ModalFooter>
     </Modal>
   );
-};
+}
```
mofojed, line 145: Need to fix this up for failing test

packages/components/src/DraggableItemList.tsx
```diff
@@ -275,10 +277,11 @@ class DraggableItemList<T> extends PureComponent<
             'draggable-item-list-dragging-item-container',
             draggingItemClassName
           )}
-          // eslint-disable-next-line react/jsx-props-no-spreading
+          // eslint-disable-next-line react/jsx-props-no-spreading, react/destructuring-assignment
           {...provided.draggableProps}
-          // eslint-disable-next-line react/jsx-props-no-spreading
+          // eslint-disable-next-line react/jsx-props-no-spreading, react/destructuring-assignment
           {...provided.dragHandleProps}
+          // eslint-disable-next-line react/destructuring-assignment
```
mofojed, line 284: I don't think destructuring is needed on the next line?

packages/components/src/ValidateLabelInput.tsx
```diff
@@ -65,6 +65,7 @@ export function ValidateLabelInput(
       </label>
       {React.Children.toArray(children).map(child => {
         if (!React.isValidElement(child)) {
+          // eslint-disable-next-line react/jsx-key
```
mofojed, line 68: We probably should give this a key here and in the `.cloneElement`... though the only key I'd give would just be the index of the array.

packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -340,7 +340,7 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     const errorMessage = error ? `Unable to open document. ${error}` : null;
     const editorLanguage = settings ? settings.language ?? null : null;
     const completionProviderEnabled =
-      model && session && editorLanguage === sessionLanguage;
+      model && session != null && editorLanguage === sessionLanguage;
```
mofojed, line 343: Shouldn't it be checking `model != null` as well? How is strict boolean not tripping for this?

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from '@deephaven/filters';
 import TableUtils from './TableUtils';
 import DateUtils from './DateUtils';
+// eslint-disable-next-line import/no-relative-packages
 import IrisGridTestUtils from '../../iris-grid/src/IrisGridTestUtils';
```
mofojed, line 15: IrisGridTestUtils could likely be relocated. Doesn't need to be this change though.

mofojed (changes requested):
Need to fix the failing tests.

packages/console/src/notebook/ScriptEditor.tsx
mattrunyon: Oh actually this is because strict boolean is ok with `object | null` omitting the check since it's safe. It's not ok w/ boolean, number, string omitting since they are unsafe. Boolean is mostly unsafe if you invert the logic of the variable name and it's mostly to remind you to set a default in that case (at least to me)

`session` is always defined here. Strict boolean expression I guess is smarter now and knows it's always defined, so it's useless in the condition. I added the `!= null` mostly for safety in case our types were wrong and the session is undefined somehow

package.json
mattrunyon: Guess not

packages/components/src/DraggableItemList.tsx
mofojed: Ah duh. Well why not do what it's asking and just destructure all of these above, e.g.:
```
      const { innerRef, dragHandleProps, draggableProps } = provided;
      return (
        <div
          className={classNames(
            'draggable-item-list-dragging-item-container',
            draggingItemClassName
          )}
          {...draggableProps}
          {...dragHandleProps}
          ref={innerRef}
        >
```

package.json
```diff
@@ -55,7 +55,6 @@
     "@babel/core": "7.17.7",
     "@deephaven/components": "file:../components",
     "@deephaven/console": "file:../console",
-    "@deephaven/eslint-config": "file:../eslint-config",
```
mofojed, line 58: I'm a little confused...
- `.eslintrc` references `@deephaven/eslint-config`, does this not need to be a dev dep?
- Why do the `devDependencies` all have `../` instead of `package/` in there value?

mofojed (approved):

package.json
mattrunyon: Oops. I removed this while trying to fix package-lock stuff. I checked out the main package-lock and when running `npm install` after updating the versions, I was getting a bunch of peer dep issues. Turned out I need to `npm install eslint@version -w @deephaven/eslint-config` to get npm to play nice

mofojed (approved):

mofojed (approved):

PR #952, opened 2022-12-15
Make linker dialog draggable

### LINKED ISSUE #901: Make linker dialog moveable
Allow left edge to act as grab handle and be moveable anywhere in bounds.

jason-wang24:
Fixes #901


dsmmcken (changes requested):
1. Since you are clamping to the window with JS instead of CSS, you'll needs to re-check windows bounds on window resize events. Otherwise you can move to bottom left, and shrink the window and it will be offscreen.
2. Tooltip on button should be dismissed on drag start, might want to just use a div and a tooltip for more control over the tooltip.
3. You shouldn't need to clamp both the mouse and the dialog, I would clamp only the dialog.
4. Not sure why you have isLoading? If you need the state of the transition you can get that from the parent. I am guessing for the right/bottom setting. Rather than setting top/left you could instead set bottom/right and accomplish the same thing.

I'll let bender do a more through code review first. That btn-drag handle should use flex box for centering though.

packages/dashboard-core-plugins/package.json
```diff
@@ -47,6 +47,7 @@
     "memoize-one": "^5.1.1",
     "memoizee": "^0.4.15",
     "prop-types": "^15.7.2",
+    "react-beautiful-dnd": "^13.1.0",
```
mofojed, line 50: Need to run `npm install` again and commit changes from `package-lock.json`

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -48,6 +50,11 @@ export type LinkerOverlayContentProps = {
 export type LinkerOverlayContentState = {
   mouseX?: number;
   mouseY?: number;
+  dialogX?: number;
+  dialogY?: number;
+  offsetX: number;
+  offsetY: number;
```
mofojed, line 56: Since these are all in pairs, one could define them all as points or something, like:
```
type Point = { x: number, y: number }

export type LinkerOverlayContentState = {
  mouse?: Point;
  dialog?: Point;
  offset?: Point;
};
```
Might make it a little bit cleaner. Of course, not a big deal.

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -64,23 +71,41 @@ export class LinkerOverlayContent extends Component<
...
   componentDidMount(): void {
     window.addEventListener('blur', this.handleBlur, true);
     window.addEventListener('mousemove', this.handleMouseMove, true);
+    window.addEventListener('mouseup', this.handleMouseUp, true);
     window.addEventListener('keydown', this.handleKeyDown, true);
     window.addEventListener('keyup', this.handleKeyUp, true);
+    window.addEventListener('resize', this.handleResize, true);
+    this.setState({
+      dialogX:
+        window.innerWidth - (this.dialogRef.current?.offsetWidth ?? 0) - 16,
```
mofojed, line 105: Where'd the `16` come from? Extract it into a constant at the top of the file and name is something descriptive, e.g.
```
const DIALOG_MARGIN = 16;
```

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -139,11 +168,68 @@ export class LinkerOverlayContent extends Component<
     this.setState({ mode: 'select' });
   }
 
+  handleResize(): void {
+    const { dialogX, dialogY } = this.state;
+    const boundedDialogX = Math.min(
+      Math.max(0, dialogX ?? 0),
+      window.innerWidth - (this.dialogRef.current?.offsetWidth ?? 0)
```
mofojed, line 175: If you drag the dialog to the very bottom right, then resize, the dialog becomes all bunched up. May need some more constraints in CSS to ensure it maintains it's width/height.

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.scss
```diff
@@ -14,17 +14,38 @@
...
+      border: none;
+      position: absolute;
+      background: $gray-300;
+      color: $gray-600;
+      top: 50%;
+      left: 0;
+      height: 100%;
+      min-width: 1.5rem;
+      max-width: 2rem;
+      transform: translateY(-50%);
```
mofojed, line 36: As @dsmmcken mentioned, use flexbox here to position/center the button - you shouldn't need to use this transform (or many of these other properties).

`min-width` and `max-width` should be the same as well, and should be referencing a variable. Could probably just use `$spacer-4`.

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.scss
```diff
@@ -14,17 +14,38 @@
...
+        background-color: $gray-400;
+        color: $gray-700;
+      }
+    }
+
     .toast-body {
       padding: $toast-inner-padding;
+      position: relative;
+      margin-right: 10px;
+      left: 20px;
```
mofojed, line 48: Should use `margin-left` here instead of just `left`. Right now the div is actually bleeding out of the dialog on the right if you inspect element (can't tell because of padding/margin though).

Also should use `$spacer-2` and `$spacer-4` instead of 10px/20px.

mofojed (changes requested):

mofojed (approved):

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -140,9 +157,47 @@ export class LinkerOverlayContent extends Component<
   }
 
   handleMouseMove(event: MouseEvent): void {
```
dsmmcken, line 159: My bad, you still actually need to handle a window resize.

Positioning the dialog near the top edge and resizing shorter, now makes it go off screen at the top, rather than bottom :(

bottom right positioning is still better, as now it keeps position better though.

dsmmcken (changes requested):

dsmmcken (approved):
:shipit: 

Great work.

mofojed (approved):

PR #958, opened 2022-12-16
Change onclick to separate mousedown and mouseup events in Modal.tsx

### LINKED ISSUE #934: Save As Modal dismisses on mouse up outside of dialog, without corresponding mouse down
**Description**

The Save As modal dialog dismisses when you click inside and then drag outside of the modal

**Steps to reproduce**
1. Create a new Notebook
2. Hit Save
3. In the Save As modal dialog, click inside the "name" field and drag to select, dragging past the outside of the modal then releasing the mouse

**Expected results**
3. The text should select, modal should stay open

**Actual results**
3. The text selects, but the modal dismisses on mouse up

Engine Version
0.19.0
Web UI Version
0.22.2
Java Version
11.0.15
Barrage Version
0.5.0

jason-wang24:
Fixes #934

### DIFF
--- a/packages/components/src/modal/Modal.tsx
+++ b/packages/components/src/modal/Modal.tsx
@@ -38,7 +38,8 @@ function Modal({
   'data-testid': dataTestId,
 }: ModalProps): React.ReactElement | null {
   const element = useRef<HTMLElement>();
-  const [contentClicked, setContentClicked] = useState(false);
+  const background = useRef<HTMLDivElement>(null);
+  const [backgroundClicked, setBackgroundClicked] = useState(false);
 
   const handleKeyDown = useCallback(
     (event: KeyboardEvent): void => {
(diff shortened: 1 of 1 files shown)


packages/components/src/modal/Modal.tsx
```diff
@@ -139,7 +151,11 @@ function Modal({
               >
                 <div
                   className="modal-content"
-                  onClick={e => e.stopPropagation()}
+                  onMouseDown={e => {
+                    e.stopPropagation();
+                    setContentClicked(true);
```
mofojed, line 156: Instead of doing `setContentClicked`, can you just do `setBackgroundClicked` when mousedown in the background? Then you don't need to listen to click, down, or up in the content

mofojed (changes requested):

mofojed (approved):

PR #959, opened 2022-12-16
FileExplorer tests

vbabich:
- `FileListItemEditor` - 100% coverage
- `FileUtils` - 88% coverage

### DIFF
--- a/packages/file-explorer/src/FileListItemEditor.test.tsx
+++ b/packages/file-explorer/src/FileListItemEditor.test.tsx
@@ -1,0 +1,147 @@
+import React from 'react';
+import { act, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import {
+  FileListItemEditor,
+  FileListItemEditorProps,
+} from './FileListItemEditor';
+import { FileStorageItem } from './FileStorage';
+
+function makeItem(name = 'DEFAULT_NAME'): FileStorageItem {
+  return {
+    id: `id-${name}`,
+    type: 'file',
+    filename: name,
+    basename: name,
+  };
+}
+
+function makeWrapper({
+  item = makeItem(),
+  onCancel = jest.fn(),
+  onSubmit = jest.fn(),
+  validate,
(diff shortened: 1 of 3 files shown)


packages/file-explorer/src/FileListItemEditor.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+      onCancel,
+      onSubmit,
+      validate,
+    });
+    const input: HTMLInputElement = screen.getByRole('textbox');
+    userEvent.type(input, '123{enter}');
+    // Wait to resolve the validation promise
+    await act(async () => {
+      await Promise.resolve();
+    });
```
mofojed, line 77: Could put this in a convenience function or in `TestUtils.flushPromises()`

mofojed (dismissed):

packages/file-explorer/src/FileListItemEditor.test.tsx
vbabich: Updated

mofojed (changes requested):
Now it's getting the `act` errors logged:
```
    console.error
      Warning: An update to FileListItemEditor inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at FileListItemEditor (/home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/file-explorer/src/FileListItemEditor.tsx:28:3)

```

Might want to just wrap the flush in `TestUtils.flushPromises` with `act`

vbabich:
Looks like we have a few files that use `TestUtils.flushPromises` in a non-RTL context. Adding `act` to `flushPromises` would add RLT as a new dependency in `TestUtils`, would it be better to define a local function with `act` in the file that needs it?

mofojed (approved):

PR #963, opened 2022-12-19
Column grouping sidebar

### LINKED ISSUE #962: Side panel group management

mattrunyon:
Fixes #962

I strayed from the spec on creating groups and instead opted for a button that is only active when a user has selected items. The selected items will be moved into a new group instead. This removes a bunch of places that would otherwise need a special case just for handling the new group item.

Items left todo

- [x] CSS cleanup
- [x] Fix TS (mostly some templated types that need to be piped all the way through)
- [x] Change selections to be name based instead of index so groups can be selected
- [x] Update create group to allow creation with selected groups
- [x] Update create group to move non-adjacent items next to each other
- [x] Finish multi-select handling
> - [x] Drag and drop
> - [x] Move up/down/top/bottom
- [x] Write a comprehensive list of behaviors to test (there's a lot in the unit tests)
- [x] Should probably write unit tests


dsmmcken:
> I strayed from the spec on creating groups and instead opted for a button that is only active when a user has selected items. The selected items will be moved into a new group instead. 

Acknowledging approval of this change. Not ideal, but acceptable trade-off in terms of development effort.

mattrunyon:
I do think there's some benefit to the change from usability for the expediency of adding a group. Spec as written would require click add group, name group, select columns you want in group, drag into group. This at least just requires select columns, click create group, name group. Skips the need for the drag on initial creation if you know the columns you want

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+  renderItem(
+    props: Required<Omit<SortableTreeItemProps, 'renderItem'>> & {
+      item: TreeItem<T>;
+    }
+  ): React.ReactNode;
+}
+
+export function SortableTree<T = undefined>({
+  collapsible = false,
+  defaultItems = [] as TreeItem<T>[],
```
mofojed, line 152: Define this array outside as a constant.
Like this as a default arg, it could cause an infinite loop since the `[]` will not equal twice through two renders (specifically that `useEffect` using `defaultItems` in it's dependency array below).

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+  );
+  const sensors = useSensors(
+    useSensor(PointerWithInteraction, {
+      activationConstraint: {
+        distance: 5,
+      },
+    })
+    // useSensor(KeyboardSensor, {
+    //   coordinateGetter,
+    // })
```
mofojed, line 207: Unused
```suggestion
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+                indentationWidth={indentationWidth}
+                indicator={indicator}
+                collapsed={Boolean(collapsed && children.length)}
+                onCollapse={
+                  collapsible && children.length
+                    ? () => handleCollapse(id)
+                    : undefined
+                }
+                onRemove={
+                  removable && onRemove ? () => handleRemove(id) : undefined
```
mofojed, line 253: `handleRemove` undefined - I think you mean `onRemove(item)`.

Could also set `onRemove` to a default function (defined outside as a const) and then this could just be `removable ? onRemove(item) : undefined`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+            {activeId && activeItem ? (
+              <SortableTreeItem
+                id={activeId}
+                depth={activeItem.depth}
+                clone
+                childCount={getChildCount(items, activeId) + 1}
+                value={activeId.toString()}
+                indentationWidth={indentationWidth}
+                renderItem={renderItem}
+                item={activeItem}
```
mofojed, line 275: Type error

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+        )}
+      </SortableContext>
+    </DndContext>
+  );
+
+  function handleDragStart({ active: { id: newActiveId } }: DragStartEvent) {
+    setActiveId(newActiveId);
+    setOverId(newActiveId);
+
+    document.body.style.setProperty('cursor', 'grabbing');
```
mofojed, line 289: Not sure if there's any differences between this and doing it on `document.documentElement` like `Grid` does. In any case, this works.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+    setOverId(null);
+    setActiveId(null);
+    setOffsetLeft(0);
+
+    document.body.style.setProperty('cursor', '');
+  }
+
+  function handleCollapse(id: UniqueIdentifier) {
+    setItems(oldItems =>
+      setProperty(oldItems, id, 'collapsed', value => !value)
```
mofojed, line 340: Need to clean up typing on `setProperty`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+    );
+  }
+}
+
+const adjustTranslate: Modifier = ({ transform }) => {
```
mofojed, line 345: I'd prefer `function adjustTranslate`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
```diff
@@ -0,0 +1,350 @@
...
+
+  function handleDragMove({ delta }: DragMoveEvent) {
+    setOffsetLeft(delta.x);
+  }
+
+  function handleDragOver({ over }: DragOverEvent) {
+    setOverId(over?.id ?? null);
+  }
+
+  function handleDragEnd({ active, over }: DragEndEvent) {
```
mofojed, line 300: Not sure I like all these functions declared after the return; should also be using `useCallback` with a proper dependencies array.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTreeItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { AnimateLayoutChanges, useSortable } from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
+import { TreeItem, Props as TreeItemProps } from './TreeItem';
+import { iOS } from './utilities';
+
+export interface Props extends TreeItemProps {
+  id: UniqueIdentifier;
+}
+
+const animateLayoutChanges: AnimateLayoutChanges = ({
```
mofojed, line 13: Declare as `function`?

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTreeItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+    transform,
+    transition,
+  } = useSortable({
+    id,
+    animateLayoutChanges,
+  });
+  const style: CSSProperties = {
+    transform: CSS.Translate.toString(transform),
+    transition,
+  };
```
mofojed, line 35: I'd probably memoize this just to ensure no unnecessary re-rendering of the child.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+
+  &:first-child {
+    padding-top: $spacer-1;
+  }
+
+  &.clone {
+    // display: inline-block;
```
mofojed, line 18:
```suggestion
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+          position: absolute;
+          left: -8px;
+          top: -4px;
+          display: block;
+          content: '';
+          width: 12px;
+          height: 12px;
+          border-radius: 50%;
+          border: 1px solid #2389ff;
+          // background-color: #ffffff;
```
mofojed, line 54:
```suggestion
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+      opacity: 1;
+      position: relative;
+      z-index: 1;
+      margin-bottom: -1px;
+
+      .TreeItem {
+        position: relative;
+        padding: 0;
+        height: 8px;
+        border-color: #2389ff;
```
mofojed, line 41: Should be declared as a var at the top of the file... also, should probably use an already defined variable and lighten or darken it? @dsmmcken

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  flex-wrap: nowrap;
+  justify-content: flex-start;
+  align-items: center;
+  margin-right: $spacer-1;
+  margin-left: $spacer-1;
+  flex-grow: 1;
+
+  border-radius: $border-radius;
+
+  // margin-bottom: 2px;
```
mofojed, line 90:
```suggestion
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+    clone: boolean;
+    childCount: number;
+    value: string;
+    item: IrisGridTreeItem;
+  }): React.ReactNode;
+}
+
+export const TreeItem = forwardRef<HTMLDivElement, Props>((props, ref) => {
+  const {
+    childCount = 0,
```
mofojed, line 34: Unused prop... are you trying to assign default values to the props here? I think would need to do in the `renderItem` function defined in `VisibilityOrderingBuild` instead.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+    style,
+    value,
+    wrapperRef,
+    renderItem,
+    item,
+    ...rest
+  } = props;
+
+  const DepthMarker = <span className={styles.DepthLine} />;
+  const depthMarkers = Array(depth).fill(DepthMarker);
```
mofojed, line 55: `useMemo`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/keyboardCoordinates.ts
```diff
@@ -0,0 +1,158 @@
+/* eslint-disable import/prefer-default-export */
+import {
+  closestCorners,
+  getFirstCollision,
+  KeyboardCode,
+  KeyboardCoordinateGetter,
+  DroppableContainer,
+} from '@dnd-kit/core';
+
```
mofojed, line 9:
```suggestion
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/keyboardCoordinates.ts
```diff
@@ -0,0 +1,158 @@
...
+import type { SensorContext } from './types';
+import { getProjection } from './utilities';
+
+const directions: string[] = [
```
mofojed, line 13: `SCREAMING_SNAKE_CASE`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/keyboardCoordinates.ts
```diff
@@ -0,0 +1,158 @@
...
+  KeyboardCode.Down,
+  KeyboardCode.Right,
+  KeyboardCode.Up,
+  KeyboardCode.Left,
+];
+
+const horizontal: string[] = [KeyboardCode.Left, KeyboardCode.Right];
+
+export const sortableTreeKeyboardCoordinates: (
```
mofojed, line 22: Why not declare as `function`? Also probably should have some tests for this function.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/utilities.ts
```diff
@@ -0,0 +1,345 @@
+import type { UniqueIdentifier } from '@dnd-kit/core';
```
mofojed, line 1: I'd add some tests for some of the functions in this file.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/utilities.ts
```diff
@@ -0,0 +1,345 @@
...
+  GridUtils,
+  ModelSizeMap,
+  MoveOperation,
+  ModelIndex,
+} from '@deephaven/grid';
+import type IrisGridModel from '../../../IrisGridModel';
+import type ColumnHeaderGroup from '../../../ColumnHeaderGroup';
+import type { FlattenedItem, TreeItem, TreeItems } from './types';
+
+export const iOS = /iPad|iPhone|iPod/.test(navigator.platform);
```
mofojed, line 13: Should be named `isIOS`

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -675,77 +617,333 @@ class VisibilityOrderingBuilder extends Component<
...
+    onColumnHeaderGroupChanged(newGroups);
+  }
+
+  handleGroupCreate(childIndexes: number[]): void {
+    const { model, onColumnHeaderGroupChanged } = this.props;
+    const columnGroupMap = model.getColumnHeaderGroupMap();
+    const newGroups = [...columnGroupMap.values()];
+
+    // TODO: Change to name based so you can seed with groups
+    // TODO: Remove children from their parents
```
mofojed, line 749: Some TODOs. Noticed some errors:
- Creating a group when a group named `NewGroup` already exists
- Creating a group with child elements selected
- Creating a group with some top level and child elements selected

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/utilities.ts
```diff
@@ -0,0 +1,345 @@
...
+      visibleIndex += item.data.visibleIndex[1] - item.data.visibleIndex[0] + 1;
+    } else {
+      visibleIndex += 1;
+    }
+  }
+
+  return items;
+}
+
+export function getProjection(
```
mofojed, line 135: I'd add some docs on these utility functions

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -96,17 +107,18 @@ class VisibilityOrderingBuilder extends Component<
 
     this.handleSearchInputChange = this.handleSearchInputChange.bind(this);
     this.searchColumns = this.searchColumns.bind(this);
-    this.handleOrderDraggingStart = this.handleOrderDraggingStart.bind(this);
-    this.handleOrderDraggingEnd = this.handleOrderDraggingEnd.bind(this);
+    this.handleGroupMove = this.handleGroupMove.bind(this);
+    this.handleGroupDelete = this.handleGroupDelete.bind(this);
+    this.handleGroupNameChange = this.handleGroupNameChange.bind(this);
+    this.validateGroupName = this.validateGroupName.bind(this);
+    // this.handleGroupColorChange = this.handleGroupColorChange.bind(this);
```
mofojed, line 114:
```suggestion
```

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -142,15 +155,12 @@ class VisibilityOrderingBuilder extends Component<
       lastSelectedColumns: [],
       searchFilter: '',
     });
-    onMovedColumnsChanged([]);
+    onMovedColumnsChanged(model.movedColumns);
+    model.setColumnHeaderGroups(model.layoutHints?.columnGroups);
```
mofojed, line 159: As mentioned elsewhere shouldn't need to call `setColumnHeaderGroups` from anywhere if you make it a `get`/`set` property and set it through `TableViewportUpdater`

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -675,77 +617,333 @@ class VisibilityOrderingBuilder extends Component<
...
+          </div>
+        </>
+      );
+    }
+  );
+
+  makeVisibilityOrderingList = memoize(
+    (
+      movedColumns: MoveOperation[],
+      columnHeaderGroups: ColumnHeaderGroup[],
```
mofojed, line 860: This is unused in the memoized function

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -675,77 +617,333 @@ class VisibilityOrderingBuilder extends Component<
...
+      userColumnWidths: ModelSizeMap,
+      selectedColumns: ModelIndex[]
+    ) => {
+      const { model } = this.props;
+      const { columns } = model;
```
mofojed, line 865: `columns` should be passed in as a param to the memoized function.

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
```diff
@@ -901,133 +1011,125 @@ class VisibilityOrderingBuilder extends Component<
...
+          <span className="vertical-divider" />
+          <Button
+            kind="ghost"
+            tooltip={
+              selectedColumns.length === 0
+                ? 'Select items to create group'
+                : 'Create group'
+            }
+            disabled={selectedColumns.length === 0}
+            onClick={() => this.handleGroupCreate(selectedColumns)}
```
mofojed, line 1041: Not sure why you're doing this instead of just passing `onClick={this.handleGroupCreate}` and get `selectedColumns` from state in there.

mofojed (changes requested):
Also an `Uncaught Error: Column group child MyByte specified in multiple groups` if you click on a header that's in a group already, then click the `Group` button.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
+@import '@deephaven/components/scss/custom.scss';
+
+.Wrapper {
+  list-style: none;
+  box-sizing: border-box;
+  padding-left: var(--spacing);
```
dsmmcken, line 6: Should be a $spacer-N?

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+        &:before {
+          position: absolute;
+          left: -8px;
+          top: -4px;
+          display: block;
+          content: '';
+          width: 12px;
+          height: 12px;
+          border-radius: 50%;
+          border: 1px solid #2389ff;
```
dsmmcken, line 53: 😞 why all these hardcoded colours?

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  padding-left: 0.5rem;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
+
+.Count {
+  padding: 0 $spacer-2;
+  border-radius: 1rem;
+  background-color: rgba($white, 0.25);
```
dsmmcken, line 121: $foreground?

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  &.collapsed svg {
+    transform: rotate(-90deg);
+  }
+}
+
+.MarchingAnts {
+  @include ants-base($foreground, $background);
+  animation-name: march;
```
dsmmcken, line 150: shouldn't need to include animation name

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+}
+
+$depth-line-color: $gray-600;
+$depth-margin: 18px;
+
+.DepthLine {
+  margin-left: $depth-margin;
+  padding-bottom: 2px;
+  height: 28px;
```
dsmmcken, line 159: seems like a magic number, should probably be extracted to variable at the top, or calculated if it can be.

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
...
+  .Text,
+  .Count {
+    user-select: none;
+    -webkit-user-select: none;
+  }
+}
+
+.Collapse {
+  svg {
+    transition: transform 250ms ease;
```
dsmmcken, line 140: $transition-mid for time

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
```diff
@@ -0,0 +1,162 @@
+@import '@deephaven/components/scss/custom.scss';
+
+.Wrapper {
```
dsmmcken, line 3: class names are normally kebab-case by convention. Did you do it this way because you are using it as a module?

dsmmcken (changes requested):

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
mattrunyon: I did it this way because I copied it from dnd-kit. I'll clean up to our conventions

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
mattrunyon: These aren't used. They're from the example which has an option for a smaller indicator of where you're dropping instead of a ghost item

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.scss
```diff
@@ -11,9 +11,25 @@
...
     justify-content: space-between;
+    align-items: center;
     background-color: $content-bg;
-    padding: $spacer-1;
+    padding: $spacer-1 $spacer-2;
     z-index: 2;
+
+    .btn-link-icon {
+      padding-left: $spacer-1 !important;
+      padding-right: $spacer-1 !important;
```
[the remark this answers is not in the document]
mattrunyon: Nope. Removing

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.scss
```diff
@@ -143,7 +159,7 @@
     border-left: 1px solid $black;
     padding-right: $spacer-1;
     margin-left: $spacer-1;
-    height: 1rem;
+    height: 70%;
```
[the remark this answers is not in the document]
mattrunyon: Only reason I can remember changing this is `1rem` is really difficult to see the divider for me. It's also low contrast which doesn't help

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -170,7 +169,9 @@ class IrisGridTableModelTemplate<
 
   private _columnHeaderMaxDepth: number | null = null;
 
-  private _columnHeaderGroups: ColumnGroup[] = [];
+  private _columnHeaderGroups: ColumnHeaderGroup[] = [];
```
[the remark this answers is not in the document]
mattrunyon: I'll remove it

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: I think it should just be required. This was code from dnd-kit examples

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: I'm removing the removable and collapsible code from this. It was part of the example and we aren't using it. Should be easy enough to add back later if desired

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: Not needed any more since I'm removing collapse. Not using that functionality

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: Ya no idea why the example did it that way. I'll clean up

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/utilities.ts
mattrunyon: Not needed since we don't do any mobile stuff

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
mattrunyon: I added a comment. It isn't used, but it affects the list when it changes. I think I need to adjust the `getTreeItems` util to take this in instead of grabbing it from the model

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingGroup.tsx
```diff
@@ -0,0 +1,171 @@
...
+            onClick={handleCancel}
+          />
+        </div>
+        {!isValid && (
+          <p className="mb-0 validate-label-error text-danger">
+            Invalid or duplicate name
+          </p>
+        )}
+      </div>
+    );
```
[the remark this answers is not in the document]
mattrunyon: Fair point. I don't think `useCallback` is necessary for the handle methods here though since they don't get passed down any further and will change on every update anyway. This only renders base elements so they'll all re-render if anything triggers a render. And controlled inputs will change the callback every time a key is pressed so memoizing them doesn't prevent re-renders

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
mattrunyon: Removed the collapse stuff since we aren't using that functionality from the example

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/TreeItem.module.scss
mattrunyon: Not used any more. Was initially the way the example set the indent for each item. Instead I added depth lines so this is now removed

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -338,8 +349,16 @@ class IrisGridUtils {
       invertSearchColumns = true,
       pendingDataMap = [],
       frozenColumns,
+      columnHeaderGroups,
     } = irisGridState;
     const { columns, formatter } = model;
+
+    let hydratedColumnHeaderGroups: ColumnHeaderGroup[] | null = null;
+
+    if (columnHeaderGroups) {
```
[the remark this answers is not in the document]
mattrunyon: Nope. If the user has deleted all `columnHeaderGroups`, it will be set to an empty array. The setter method handles empty arrays fine (as well as undefined for clearing). `null` (or undefined) is used for the "no overrides" value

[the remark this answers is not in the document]
mofojed: Ah okay well then just `if (columnHeaderGroups != null) {`

I was more confused why the `strict-boolean-expressions` isn't tripping here, I thought it would require `!= null`

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: Probably not unless you render content outside the body

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTree.tsx
mattrunyon: It's typed this way because `Modifier` is a function type. Changing would have to be `function adjustTranslate({ transform }: Parameters<Modifier>[0]): ReturnType<Modifier> { ... }` which is much more fragile

packages/iris-grid/src/sidebar/visibility-ordering-builder/sortable-tree/SortableTreeItem.tsx
mattrunyon: Same as before. Not as clean in TS to declare as `function` since `AnimateLayoutChanges` is a function type

packages/iris-grid/src/sidebar/VisibilityOrderingBuilder.tsx
mattrunyon: These should all be fixed now

dsmmcken (changes requested):
Styling is looking pretty good. Getting closer.

Here's the things I noticed:

- [x] Bug: Toggle All doesn't toggle all to hidden when all our showing, only all to shown when one is hidden, that's a regression. When a selection is made the Hide Selected button doesn't work either
- [x] Bug: internal group name appears while it is first being created in table, before a name is even given :newgroup..
- [x] Create new group should be immediately cancelable triggering delete on first creation, otherwise you have to name it just to delete it
- [x] Rename section to "Hide, Group, and Order Columns"
- [x] Bug: Yes, animation is wonky with arrow moves, looks like it is always animating from [0,0]?
- [x] Bug: Create group with group as first item, select two other items further down the list and move them to top with arrow, then move down one with arrow. The first item entirely disappears from the list.
- [x] Bug: Moving a group down with the down arrow causes the group to be deleted (
- [x] Bug? Exiting with a group name in progress
- [x] Bug? Creating a new group, never naming it then close sidebar keeps it around in the initial edit state? Should it remove it?

mofojed:
There's one weird case where if you create a group, then don't type anything, then select a few others and create a group, you get two input fields, and in the grid they look like they're in the same (empty) group:
image

I think if the input field blurs without any changes, should just cancel the grouping.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3885,8 +3899,10 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
               model={model}
               movedColumns={movedColumns}
               userColumnWidths={userColumnWidths}
+              columnHeaderGroups={model.getColumnHeaderGroups()}
```
mofojed, line 3902: Hmmm why are you using the `model.getColumnHeaderGroups()` instead of the `columnHeaderGroups` from state?

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -487,6 +489,25 @@ abstract class IrisGridModel<
...
+  setColumnHeaderGroups(groups: ColumnGroup[] | undefined): void {
+    // no-op
+  }
+
+  getColumnHeaderParentGroup(
+    modelIndex: ModelIndex,
+    depth: number
+  ): ColumnHeaderGroup | undefined {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
mofojed: This comment hasn't been addressed.

mofojed (changes requested):
Still need to address my IrisGridModel comments (`setColumnHeaderGroups` => `set columnHeaderGroups`, etc)

mattrunyon:
> There's one weird case where if you create a group, then don't type anything, then select a few others and create a group, you get two input fields, and in the grid they look like they're in the same (empty) group: image
> 
> I think if the input field blurs without any changes, should just cancel the grouping.

My concern with this is it makes it impossible to do something like create a group and then copy the group name from another source. Maybe the user is trying to clone a spreadsheet they have and want to copy the name.

I guess I could make it to where if another group is created with a new group that hasn't had its name changed, then the old new group would be deleted

packages/components/src/popper/Popper.tsx
```diff
@@ -150,6 +150,7 @@ class Popper extends Component<PopperProps, PopperState> {
     let { options } = this.props;
     options = {
       placement: 'auto',
+      modifiers: { preventOverflow: { boundariesElement: 'viewport' } },
```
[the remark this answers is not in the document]
mattrunyon: It's part of that ticket. Binds to viewport. Didn't change the placement from auto though. Don wanted to check that it didn't cause overflow issues by changing default to bottom instead of auto

dsmmcken:
- [ ] Dragging a group to the top breaks uhoh
- [ ] Creating two unamed groups, then reset. It only removes one group and mylong, myfloat, mydouble are entirely missing from the list untile the second reset.
uhoh2

mattrunyon:
I changed to only allow 1 new group at a time which fixes the reset issue.

Fixed the issue of dragging a group to top. This only occurred if dragging from the top to the top.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -133,7 +136,7 @@ abstract class IrisGridModel<
   }
 
   /** List of column movements defined by the model. Used as initial movements for IrisGrid */
-  get movedColumns(): MoveOperation[] {
+  get initialMovedColumns(): MoveOperation[] {
     return [];
   }
```
mofojed, line 142: `get movedRows()` should be `get initialMovedRows()` now to match new naming

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3885,8 +3900,10 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
               model={model}
               movedColumns={movedColumns}
               userColumnWidths={userColumnWidths}
+              columnHeaderGroups={model.columnHeaderGroups}
```
mofojed, line 3903: Can just be `columnHeaderGroups`

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -142,6 +145,113 @@ abstract class IrisGridModel<
...
 
+  /**
+   * Parses the column header groups provided.
+   * If undefined, should provide default groups such as from layoutHints
+   *
+   * @returns Object containing groups array, max depth, map of name to parent group, and map of name to group
+   */
+  parseColumnHeaderGroups = memoize(
```
mofojed, line 154: Pull this into `IrisGridUtils` as a utility function. Memoization shouldn't be needed in `IrisGridUtils` or `IrisGridTableModelTemplate`; `VisibilityOrderingBuilder` should be calling something like `get initialColumnHeaderGroups` instead.

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -56,6 +56,8 @@ const log = Log.module('IrisGridTableModel');
 const SET_VIEWPORT_THROTTLE = 150;
 const APPLY_VIEWPORT_THROTTLE = 0;
 
+const EMPTY_ARRAY: unknown[] = [];
```
mofojed, line 59: Just making a note that I'll replace this in my readonly change: https://github.com/deephaven/web-client-ui/pull/986

mofojed (dismissed):
Side note, any reason we don't allow changing the colour on columns individually and limit it to just groups? Seems like that would be nice functionality to have @dsmmcken

dsmmcken:
> any reason we don't allow changing the colour on columns individually and limit it to just groups

Yeah, we could consider that for a future revision. And if we are colouring columns, we could also maybe consider having columns inherit their parents colour?

dsmmcken (dismissed):

mofojed (approved):

dsmmcken (approved):

PR #967, opened 2022-12-20
Fix linker spamming errors, change linker dialog text, add delete key to delete selected links

jason-wang24:
- Fix linker spamming errors
    - Add listener for `PanelEvent.CLOSE` because we originally only listened to `PanelEvent.CLOSED` and it does not emit every time we need it to
- Change linker dialog text to reflect the ability to change comparison operator and new ways to delete links. Feel free to make it better
- `Delete` key now deletes the selected link, if any. Currently, this is event.key or event.code `Delete` and not `Backspace`. I know on mac keyboards the key labelled delete is actually backspace, which currently shouldn't work. Open to discussion

### DIFF
--- a/packages/dashboard-core-plugins/src/linker/Linker.tsx
+++ b/packages/dashboard-core-plugins/src/linker/Linker.tsx
@@ -153,6 +153,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       this.handleFilterColumnSelect
     );
     eventHub.on(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
+    eventHub.on(PanelEvent.CLOSE, this.handlePanelClosed);
     eventHub.on(PanelEvent.CLOSED, this.handlePanelClosed);
   }
 
@@ -170,6 +171,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       this.handleFilterColumnSelect
     );
     eventHub.off(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
+    eventHub.off(PanelEvent.CLOSE, this.handlePanelClosed);
     eventHub.off(PanelEvent.CLOSED, this.handlePanelClosed);
   }
 
@@ -509,7 +511,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
   }
 
   handlePanelClosed(panelId: string): void {
-    // Delete links on PanelEvent.CLOSED instead of UNMOUNT
+    // Delete links on PanelEvent.CLOSE and PanelEvent.CLOSED instead of UNMOUNT
     // because the panels can get unmounted on errors and we want to keep the links if that happens
     log.debug(`Panel ${panelId} closed, deleting links.`);
     this.deleteLinksForPanelId(panelId);
@@ -663,8 +665,8 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     const disabled = linkInProgress != null && linkInProgress.start != null;
     const linkerOverlayMessage =
       isolatedLinkerPanelId === undefined
-        ? 'Click a column source, then click a column target to create a filter link. Remove a filter link by clicking again to erase. Click done when finished.'
-        : 'Create a link between the source column button and a table column by clicking on one, then the other. Remove the link by clicking it directly. Click done when finished.';
+        ? 'Click a column source, then click a column target to create a filter link. Change its comparison operator by clicking on the link and then the dropdown menu button. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.'
+        : 'Create a link between the source column button and a table column by clicking on one, then the other. Change its comparison operator by clicking on the link and then the dropdown menu button. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.';
 
     return (
       <CSSTransition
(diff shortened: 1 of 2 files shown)


packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
```diff
@@ -222,6 +222,10 @@ export class LinkerOverlayContent extends Component<
       this.setState({
         mode: 'delete',
       });
+    } else if (event.key === 'Delete') {
```
dsmmcken, line 225: should be backspace or delete I think

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -153,6 +153,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       this.handleFilterColumnSelect
     );
     eventHub.on(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
+    eventHub.on(PanelEvent.CLOSE, this.handlePanelClosed);
```
dsmmcken, line 156: remove

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -170,6 +171,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       this.handleFilterColumnSelect
     );
     eventHub.off(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
+    eventHub.off(PanelEvent.CLOSE, this.handlePanelClosed);
```
dsmmcken, line 174: remove

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -509,7 +511,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
   }
 
   handlePanelClosed(panelId: string): void {
-    // Delete links on PanelEvent.CLOSED instead of UNMOUNT
+    // Delete links on PanelEvent.CLOSE and PanelEvent.CLOSED instead of UNMOUNT
```
dsmmcken, line 514: revert

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -663,8 +665,8 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     const disabled = linkInProgress != null && linkInProgress.start != null;
     const linkerOverlayMessage =
       isolatedLinkerPanelId === undefined
-        ? 'Click a column source, then click a column target to create a filter link. Remove a filter link by clicking again to erase. Click done when finished.'
-        : 'Create a link between the source column button and a table column by clicking on one, then the other. Remove the link by clicking it directly. Click done when finished.';
+        ? 'Click a column source, then click a column target to create a filter link. Change its comparison operator by clicking on the link and then the dropdown menu button. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.'
+        : 'Create a link between the source column button and a table column by clicking on one, then the other. Change its comparison operator by clicking on the link and then the dropdown menu button. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.';
```
dsmmcken, line 669:
```suggestion
        ? 'Click a column source, then click a column target to create a filter link. The filter comparison operator used by a selected link can be changed. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.'
        : 'Create a link between the source column button and a table column by clicking on one, then the other. Delete a filter link by clicking the delete button or with alt+click. Click done when finished.';
```

For a follow up ticket, needs checking, if working correctly:
1. isolated linker mode shouldn't have a comparison operator
2. same for plot one clicks I think

dsmmcken (changes requested):

packages/dashboard-core-plugins/src/linker/Linker.tsx
mofojed: Uhhh this change is needed. Don't revert.

@dsmmcken The `PanelEvent.CLOSED` doesn't get fired in this case, as the `PanelEvent.CLOSE` event gets fired before the `PanelErrorBoundary` gets mounted, so it can't emit the `PanelEvent.CLOSED`

packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx
mofojed: Yes add `Backspace` for Mac

mofojed (changes requested):
Just add `Backspace` for Mac

packages/dashboard-core-plugins/src/linker/Linker.tsx
mofojed: Correct. `ConsolePanel` fires it to close these panels. They never emit a `.CLOSED` event because those panels are never actually mounted. So we're listening for the `.CLOSE` here as well to pick up when a panel would get closed before it's mounted.
Golden layout mounts the panels in order, so when it mounts the `ConsolePanel`, it emits all the `.CLOSE` events to close the panels from the previous session, those panels that haven't been initialized yet won't emit the `.CLOSED` event.
We could have layout initialization check if there's any panels that need to be closed and remove them from the config beforehand, then Linker would need to check on startup too if all panels still exist; we could re-emit the `.CLOSED` event in cases when `.CLOSE` occurs on a panel that isn't mounted yet, but that's a little complicated because `.CLOSED` expects a `glContainer` which may not still exist; or we can just listen for `.CLOSE` and `.CLOSED` in linker and fix that case.

mofojed (approved):

packages/dashboard-core-plugins/src/linker/Linker.tsx
dsmmcken: oh... that panelevent.close. I am an idiot yes.

dsmmcken (approved):

PR #968, opened 2022-12-20
Fix weird UI when dragging a panel with linker overlay open

### LINKED ISSUE #964: Dragging a panel while linker is opened results in weird UI
Fix is to simply close the linker overlay when a panel with at least one link is being dragged.

image

jason-wang24:
Fixes #964 
- Added `PanelEvent.DRAGGING` and `PanelEvent.DROPPED` events and `isDraggingPanel` state to `Linker.tsx`. isDraggingPanel is set to false if the panel being dragged has a link starting or ending in that panel and set back to true when the panel is dropped.
- Perhaps we can also reopen the linker overlay when the panel is dropped (this would require adding a new event but is otherwise trivial)

### DIFF
--- a/packages/dashboard-core-plugins/src/linker/Linker.tsx
+++ b/packages/dashboard-core-plugins/src/linker/Linker.tsx
@@ -86,6 +86,7 @@ export type LinkerProps = OwnProps &
 export type LinkerState = {
   linkInProgress?: Link;
   selectedIds: Set<string>;
+  showOverlay: boolean;
 };
 
 export class Linker extends Component<LinkerProps, LinkerState> {
@@ -109,9 +110,14 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     this.handleExited = this.handleExited.bind(this);
     this.handleLinkSelected = this.handleLinkSelected.bind(this);
     this.handlePanelDragging = this.handlePanelDragging.bind(this);
+    this.handlePanelDropped = this.handlePanelDropped.bind(this);
     this.isColumnSelectionValid = this.isColumnSelectionValid.bind(this);
 
-    this.state = { linkInProgress: undefined, selectedIds: new Set<string>() };
+    this.state = {
+      linkInProgress: undefined,
+      selectedIds: new Set<string>(),
+      showOverlay: true,
+    };
   }
 
   componentDidMount(): void {
@@ -156,6 +162,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     eventHub.on(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
     eventHub.on(PanelEvent.CLOSED, this.handlePanelClosed);
     eventHub.on(PanelEvent.DRAGGING, this.handlePanelDragging);
+    eventHub.on(PanelEvent.DROPPED, this.handlePanelDropped);
   }
 
   stopListening(layout: GoldenLayout): void {
@@ -174,6 +181,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     eventHub.off(InputFilterEvent.COLUMNS_CHANGED, this.handleColumnsChanged);
     eventHub.off(PanelEvent.CLOSED, this.handlePanelClosed);
     eventHub.off(PanelEvent.DRAGGING, this.handlePanelDragging);
+    eventHub.off(PanelEvent.DROPPED, this.handlePanelDropped);
   }
 
   handleCancel(): void {
@@ -517,7 +525,19 @@ export class Linker extends Component<LinkerProps, LinkerState> {
       const link = links[i];
       const { start, end } = link;
       if (start.panelId === componentId || end?.panelId === componentId) {
-        this.handleDone();
+        this.setState({ showOverlay: false });
+        return;
+      }
+    }
+  }
+
+  handlePanelDropped(componentId: string): void {
+    const { links } = this.props;
+    for (let i = 0; i < links.length; i += 1) {
+      const link = links[i];
+      const { start, end } = link;
+      if (start.panelId === componentId || end?.panelId === componentId) {
+        this.setState({ showOverlay: true });
         return;
       }
     }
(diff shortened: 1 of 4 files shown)


dsmmcken (dismissed):
After trying it, I agree. You are right we should hide it then re-show it instead.

Could make it unmount the children, or display none on a parent, but it should be with zero-transition as picking up the component is instant. Looks still a bit weird to have it fade out in the wrong spot as they fade out.

jason-wang24:
I added a `showOverlay` state to `Linker.tsx` which gets toggled by the DRAGGING and DROPPED events. I'm slightly concerned this is not defensive enough, as if the panel is somehow not dropped, `showOverlay` may be stuck being false. Not sure if this is a legitimate concern or not but either way I'm not sure how to make it more defensive (set it to true elsewhere?) if need be.

dsmmcken:
react.development.js:1139 Uncaught Error: React.Children.only expected to receive a single React element child.
    at Object.onlyChild [as only] (react.development.js:1139:13)
    at Transition2.render (Transition.js:408:108)
    at finishClassComponent (react-dom.development.js:17485:31)
    at updateClassComponent (react-dom.development.js:17435:24)
  
  Seeing this warning

dsmmcken:
(fat fingered the close with comment button, ignore)

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -666,7 +702,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
         ? 'Click a column source, then click a column target to create a filter link. Remove a filter link by clicking again to erase. Click done when finished.'
         : 'Create a link between the source column button and a table column by clicking on one, then the other. Remove the link by clicking it directly. Click done when finished.';
 
-    return (
+    return isLinkOverlayShown ? (
```
mofojed, line 705: Shouldn't need a ternary here, just keep it in the `CSSTransition` `in` attribute - then there will be an appropriate fade animation when you start dragging, and another fade animation when you drop.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -108,9 +109,15 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
     this.handleLinkSelected = this.handleLinkSelected.bind(this);
+    this.handlePanelDragging = this.handlePanelDragging.bind(this);
+    this.handlePanelDropped = this.handlePanelDropped.bind(this);
     this.isColumnSelectionValid = this.isColumnSelectionValid.bind(this);
 
-    this.state = { linkInProgress: undefined, selectedIds: new Set<string>() };
+    this.state = {
+      linkInProgress: undefined,
+      selectedIds: new Set<string>(),
+      showOverlay: true,
```
mofojed, line 119: I'd name this `isDraggingLinkedPanel` or even just `isDraggingPanel` instead, as it's kind of confusing naming it `showOverlay` when the overlay is shown based on the active tool in `isOverlayShown()` method

mofojed (changes requested):

packages/dashboard-core-plugins/src/linker/Linker.tsx
jason-wang24: > After trying it, I agree. You are right we should hide it then re-show it instead.
> 
> Could make it unmount the children, or display none on a parent, but it should be with zero-transition as picking up the component is instant. Looks still a bit weird to have it fade out in the wrong spot as they fade out.

I think we are trying to avoid the fade animation. Is there a better way to get the overlay to disappear as quickly as possible?

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -108,9 +109,15 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
     this.handleLinkSelected = this.handleLinkSelected.bind(this);
+    this.handlePanelDragging = this.handlePanelDragging.bind(this);
+    this.handlePanelDropped = this.handlePanelDropped.bind(this);
     this.isColumnSelectionValid = this.isColumnSelectionValid.bind(this);
 
-    this.state = { linkInProgress: undefined, selectedIds: new Set<string>() };
+    this.state = {
+      linkInProgress: undefined,
+      selectedIds: new Set<string>(),
+      isDraggingPanel: true,
```
mofojed, line 119: I think you've got the `boolean` backwards for the way this is named. `isDraggingPanel` should be `true` while there's a panel that's currently being dragged.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -580,7 +615,8 @@ export class Linker extends Component<LinkerProps, LinkerState> {
 
   isOverlayShown(): boolean {
     const { activeTool } = this.props;
-    return activeTool === ToolType.LINKER;
+    const { isDraggingPanel } = this.state;
```
mofojed, line 618: I didn't realize we don't want the fade transition for this case, so it should be a separate boolean from `isOverlayShown()`.
Just wrap the whole CSSTransition in `!isDraggingPanel ? <CSSTransition in={isOverlayShown} ... > : null`, then you'll have a fade when activating/deactivating the linker by changing the tool, but it'll still be instant when dragging the panel.

Even with these changes, it looks like Linker gets a couple renders in before receiving the `.DRAGGING` event and still logs the error a couple time.  I think logging an error when trying to draw the link may be a bit aggressive. Changing that to `log.warn` instead may be useful: https://github.com/deephaven/web-client-ui/blob/f5794d797403eb3ab2c553d237de11171ac7c9e6/packages/dashboard-core-plugins/src/linker/LinkerOverlayContent.tsx#L291

mofojed (changes requested):

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -510,6 +521,22 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
+  handlePanelDragging(componentId: string): void {
+    const { links } = this.props;
+    for (let i = 0; i < links.length; i += 1) {
+      const link = links[i];
+      const { start, end } = link;
+      if (start.panelId === componentId || end?.panelId === componentId) {
+        this.setState({ isDraggingPanel: true });
+        return;
+      }
+    }
```
mofojed, line 533: We probably don't even really need a check here, we could just always hide the links when dragging... but whatever, no biggie.

mofojed (approved):

PR #974, opened 2022-12-21
Added goto value feature

### LINKED ISSUE #652: Go to value
Continuation of #616, adds go to value functionality. Requires api support for `SeekRow`.

image

Zhou-Ziheng:
Fixes #652 
Manually tested to ensure the goto feature works on integer columns and string columns


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3152,6 +3169,47 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
   }
 
+  async seekRow(inputString: string, isBackwards = false): Promise<void> {
+    const {
+      gotoValueSelectedColumn: selectedColumn,
+      gotoValueSelectedFilter,
+      lastSeekedRow,
```
mofojed, line 3176: I think instead of `lastSeekedRow`, we should use the users current selection for seeking. If it's forward, go from the last row in their selection, if it's backward, go from before the first row in their selection.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3152,6 +3169,47 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    } = this.state;
+    const { model } = this.props;
+    if (isIrisGridProxyModel(model)) {
+      const { table } = model;
+      if (table !== undefined && !TableUtils.isTreeTable(table)) {
+        if (selectedColumn.type === 'java.lang.String') {
```
mofojed, line 3182: Use `TableUtils.isStringType(selectedColumn.type)` instead

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3152,6 +3169,47 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
 
+  async seekRow(inputString: string, isBackwards = false): Promise<void> {
+    const {
+      gotoValueSelectedColumn: selectedColumn,
+      gotoValueSelectedFilter,
+      lastSeekedRow,
+    } = this.state;
+    const { model } = this.props;
+    if (isIrisGridProxyModel(model)) {
+      const { table } = model;
```
mofojed, line 3180: We should define `seekRow` as a function in `IrisGridModel` so this functionality can be abstracted out. We generally shouldn't call `table` functions directly from anything except within `IrisGridModel`.

mofojed (changes requested):
Also `BigDecimal` is giving me an error, will need to check that.
Will need to update the snapshots as well. I have instructions in the `README.md` (though they're a little cumbersome, it should work).

packages/iris-grid/src/IrisGrid.tsx
mofojed: To be clear, add an `abstract seekRow` method and `get isSeekRowAvailable()` method to `IrisGridModel` (similar to `abstract export` and `get isExportAvailable`), implement `seekRow` in `get isSeekRowAvailable` in `IrisGridTableModel` (where `isSeekRowAvailable` returns `this.table.seekRow != null`).

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3152,6 +3169,47 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      if (table !== undefined && !TableUtils.isTreeTable(table)) {
+        if (selectedColumn.type === 'java.lang.String') {
+          // is string column,
+          const isContains = gotoValueSelectedFilter === 1;
+          const rowIndex = await table.seekRow(
+            isBackwards ? lastSeekedRow - 1 : lastSeekedRow + 1,
+            selectedColumn,
+            'String',
```
mofojed, line 3188: I was incorrect about this value always being `'String'`. You'll want it to match the column type, so check if column `isStringType`, set as `String`, check `TableUtils.isNumberType`, set to `Number`. Otherwise it'll compare the value as a string, and you may get `95` matching `9500` for a number instead of just `95`.

mofojed: You'll also want to try/catch around this, as the `seekRow` function could throw an exception if there's an error. Ideally pass the error down to the Goto value and display it.

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -28,6 +37,13 @@ interface GotoRowProps {
   onEntered: () => void;
   onExiting: () => void;
   onExited: () => void;
+
+  gotoValueSelectedColumn: Column;
+  gotoValueSelectedFilter: number;
```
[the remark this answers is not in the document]
Zhou-Ziheng: Would we also want a `containsIgnoreCase`? The API has case sensitivity and equals/contains separated. This is more of a question for Don, but would we be better off by including a separate toggle option that let's the user select if they want case sensitivity?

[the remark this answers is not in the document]
mofojed: Interestingly, the existing `FilterType.contains` is actually insensitive: https://github.com/deephaven/web-client-ui/blob/main/packages/jsapi-utils/src/TableUtils.ts#L1401. Removing the `(?i)` from that regex would make it case-sensitive.
We should probably also add `containsIgnoreCase` that uses regex `(?s)(?i).*\\Q${value}\\E.*`, then change `contains` to just use `(?s).*\\Q${value}\\E.*`. Not sure why it was omitted originally, will want to check with @dsmmcken when he's back.
It wouldn't require many changes, though it would change the behaviour of the `~` operator in quick filters: https://github.com/deephaven/web-client-ui/blob/main/packages/jsapi-utils/src/TableUtils.ts#L1335

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -42,19 +58,59 @@ function GotoRow({
...
+      gotoRowInputRef.current?.select();
+    } else if (
+      !isGotoRowActive &&
+      document.activeElement !== gotoValueInputRef.current
+    ) {
+      gotoValueInputRef.current?.select();
     }
-  }, [gotoRow]);
+  };
+  useEffect(selectInput, [isGotoRowActive, gotoRow, gotoValue]);
```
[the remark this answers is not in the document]
Zhou-Ziheng: see Don's comment inside `GotoRow.tsx`: `// when row changes without focus (i.e. via context menu), re-select input`

packages/iris-grid/src/GotoRow.tsx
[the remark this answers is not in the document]
dsmmcken: UI inputted values we always treat as case-insensitive.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3171,42 +3176,88 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+        let rowIndex;
+        if (columnDataType === 'DateTime') {
+          const { formatter } = model;
+          const [startDate] = DateUtils.parseDateRange(
             inputString,
-            false,
-            isContains,
+            formatter.timeZone
+          );
+          rowIndex = await table.seekRow(
```
mofojed, line 3236: This still needs to be abstracted through `IrisGridModel`, we should not be calling anything on the `Table` object directly.

mofojed (changes requested):

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -182,89 +276,43 @@ function GotoRow({
...
+                  disabled={gotoValue === ''}
+                  onClick={() => {
+                    onGotoValueChanged(gotoValue, false);
+                  }}
+                >
+                  <FontAwesomeIcon icon={vsArrowDown} />
+                </Button>
+              </div>
+              {gotoValueError && (
+                <div className="goto-row-error text-danger">
```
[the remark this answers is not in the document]
Zhou-Ziheng: Turns out `goto-value-error` isn't even defined in `GotoRow.scss`. I removed the class for now and changed e2e tests to look for `'.goto-row-wrapper .text-danger'`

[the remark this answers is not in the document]
Zhou-Ziheng: Will see if this breaks e2e once it's ready for merging

mofojed (commented):
Looks good - just need to update the snapshots now, instructions are in the README.md. Let me know if you have issues.

packages/iris-grid/src/GotoRow.scss
```diff
@@ -5,13 +5,17 @@
 }
 
 .goto-row {
-  display: flex;
-  justify-content: space-between;
-  align-items: flex-start;
   background-color: $content-bg;
-  padding: $spacer-2;
   border-top: $border-width solid black;
```
dsmmcken, line 9:
```suggestion
  border-top: $border-width solid $black;
```

dsmmcken (changes requested):
1. per spec: If accessed via the context menu, populate the input with the row number AND the go to value input with the cell value that the menu was opened from (be sure to set the column correctly too)
2. per spec: Focus go to value input on open
3. go to value doesn't reset on close and re-open (row input does)
4. Weird behavior: open menu with ctrl+g, press tab to cycle to the value input, start typing a1, the a goes into that input and the 1 goes into the row input (focus event of any input should also trigger selecting of that row probably)

dsmmcken (changes requested):
Found several more things while playing with it:

- [ ] go to value `null` doesn't work for any column type
- [x] Boolean columns accept anything not true as false, goto string "dasfadf" is false. Could consider select box for booleans, of true,false,null instead?
- [x] Char columns don't work.
- [ ] Go to value postive/negative.inf works for floats/doubles but the text field is blank
- [x] Open go to, scroll table to bottom, close box, grid now has excess space at bottom. Should we render the grid with the correct lastTop again? Thoughts on this.

dsmmcken (dismissed):
- [x] scroll to bottom of a ticking table (aka stick to bottom), enter a value that is in one of the first few rows. Hit enter, but the next table tick will return you to stuck to bottom. Should probably unset the stuck to bottom setting. Also hitting enter repeatedly occasionally does get it to that row (probably has to do with getting it during the right time during an update) and it sticks there and you can see that updates (number of rows) "freeze" until you mouse over the table.

uhoh3

dsmmcken:
- [ ] If go to is initiated via shortcut or the sidebar menu, and there is a table row cursor (a selected row) populate the initial values for the row index with that row number
- [ ] If go to has a row index, then populate the go to value with the value of the selected column matching that row index, if the column selected dropdown is changed, pre-fill the new matching value for that row index.

dsmmcken:
I am good with splitting the reminder of the work out into a separate ticket so this can ship.

> for the issue with excess space, I think resetting lastTop should really be taken care of in a IrisGridBottomBar. Thoughts?
 Yes probably.

packages/iris-grid/src/IrisGrid.tsx
mofojed: Will also need to check for `BigDecimal` / `BigInteger` types, as those _will_ need to pass the value as string (a `Number` could get truncated).

Zhou-Ziheng:
`seekRow` function in IrisGrid was restructured into a switch for better style

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3173,6 +3202,126 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      const columnDataType = TableUtils.getNormalizedType(selectedColumn.type);
+
+      let rowIndex;
+
+      switch (columnDataType) {
+        case TableUtils.dataType.STRING: {
+          rowIndex = await model.seekRow(
+            isBackwards === true ? searchFromRow - 1 : searchFromRow + 1,
+            selectedColumn,
+            'String',
```
mofojed, line 3251: Instead of passing a raw string here, pass in the `dh.ValueType` constants: https://github.com/deephaven/deephaven-core/blob/89cbd6aafb2b10f349546b91220245fabac74c53/web/client-api/src/main/java/io/deephaven/web/client/api/JsTable.java#L96. You'll need to add to `dh.types`.
Basically anywhere you're calling `model.seekRow`, you should pass a `dh.ValueType` constant in.

See how `VariableType` is defined in dh.types.ts and used in `IrisGridPanel` as an example.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -488,6 +488,20 @@ abstract class IrisGridModel<
    */
   abstract delete(ranges: GridRange[]): Promise<void>;
 
+  abstract seekRow(
+    startRow: number,
+    column: Column,
+    valueType: string,
```
mofojed, line 494: Should restrict this to `VariableTypeUnion` (see my other comment about defining a `VariableType` type)

mofojed (changes requested):
Looking great! Just the update to the types for `ValueType` and should be good to go.

mofojed (approved):

PR #975, opened 2022-12-22
Fix closing linker using shortcut not resetting state

### LINKED ISSUE #971: Closing linker using cmd/ctrl+L shortcut should cancel in-progress links
**Steps to reproduce**
1, open linker dialog (CMD-L on mac)
2. select a column to start the linker from
3. close the linker dialog (CMD-L on mac)

**Expected results**

The in-progress link is discarded when linker mode is exited

**Actual results**

The in-progress link re-appears the next time linker is opened

Behaviour has always been this way, but first reported by @rachelmbrubaker 

Engine Version: 0.20.0
Web UI Version: 0.23.0
Java Version: 11.0.16
Barrage Version: 0.5.0

jason-wang24:
Fixes #971 
- Set `linkInProgress` to undefined and `selectedIDs` to an empty set when `activeTool` is set to default 
- Remove unnecessary setStates based on above change

### DIFF
--- a/packages/dashboard-core-plugins/src/linker/Linker.tsx
+++ b/packages/dashboard-core-plugins/src/linker/Linker.tsx
@@ -109,7 +109,6 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     this.handleExited = this.handleExited.bind(this);
     this.handleLinkSelected = this.handleLinkSelected.bind(this);
     this.isColumnSelectionValid = this.isColumnSelectionValid.bind(this);
-    this.reset = this.reset.bind(this);
 
     this.state = { linkInProgress: undefined, selectedIds: new Set<string>() };
   }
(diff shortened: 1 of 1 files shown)


packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -127,6 +127,12 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     }
     if (activeTool !== prevProps.activeTool) {
       this.updateSelectionValidators();
+      if (activeTool === ToolType.DEFAULT) {
+        this.setState({
+          linkInProgress: undefined,
+          selectedIds: new Set<string>(),
+        });
```
mofojed, line 134: You could probably put this in a `reset()` method. I just noticed in `handleAllLinksDeleted` we're setting `linkInProgress:undefined` but we're not resetting `selectedIds`, and we probably should.

mofojed (dismissed):

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -109,6 +109,7 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     this.handleExited = this.handleExited.bind(this);
     this.handleLinkSelected = this.handleLinkSelected.bind(this);
     this.isColumnSelectionValid = this.isColumnSelectionValid.bind(this);
+    this.reset = this.reset.bind(this);
```
mofojed, line 112: You don't need to do this, since you're not passing the function as a reference anywhere.

mofojed (approved):

mofojed (approved):

PR #978, opened 2022-12-28
DH-13988 Fix issue with old buttons not getting disabled in console history

### LINKED ISSUE #325: Command History gets really slow when >1000 items
**Description**

When there are lots of items in the command history (more than 1000) it gets really slow when scrolling

**Steps to reproduce**
1. Add >1000 items to command history
2. Scroll up and down quickly

**Expected results**
2. Able to see command history quickly

**Actual results**
2. Shows a blank screen when scrolling, takes a few seconds for the information to appear

### LINKED ISSUE #928: Deleted references to tables are not being marked as disabled in console history
image

Deleted references to tables are not being marked as disabled.

Steps to reproduce:
1. create a table
2. delete the table

Expected:
Table in console history table button is greyed out/disabled 

Actual:
Table it still clickable, and it fails to open

Engine Version
0.19.0
Web UI Version
0.22.2
Java Version
11.0.16
Barrage Version
0.5.0

mofojed:
- We were mutating objects within the state, violating the immutability principle.
- Clone any object that we were mutating.
- Cleaned up some of the code.
- All tests pass
- Tested by running the steps in the description of #928, also tested importing a CSV and giving it the same name as a prior object. Old objects were updated correctly.
- Updated e2e tests:
  - Refactored creating a table command into it's own function
  - Fixed up old print commands tests - wasn't verifying the result correctly
  - Added test to check that an object button was created when creating a table, and that the old button was disabled when re-running the same command twice
- Fixes #928 
- Fixes [DH-13988](https://deephaven.atlassian.net/browse/DH-13988)

### DIFF
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -331,21 +331,34 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
           changes: VariableChanges;
         }
       | undefined,
-    historyItemParam: ConsoleHistoryActionItem,
+    historyItem: ConsoleHistoryActionItem,
     workspaceItemPromise: Promise<CommandHistoryStorageItem>
   ): void {
-    const historyItem = historyItemParam;
-    historyItem.wrappedResult = undefined;
-    historyItem.cancelResult = undefined;
+    const newHistoryItem = {
+      ...historyItem,
+      wrappedResult: undefined,
+      cancelResult: undefined,
+      result: result ?? historyItem.result,
+    };
+
+    this.setState(({ consoleHistory }) => {
+      const itemIndex = consoleHistory.lastIndexOf(historyItem);
+      if (itemIndex < 0) {
+        log.error(`historyItem not found in consoleHistory`);
+        return { consoleHistory };
+      }
+      const newHistory = consoleHistory.concat();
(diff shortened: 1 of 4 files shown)


packages/console/src/Console.tsx
```diff
@@ -331,21 +331,34 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
...
+      wrappedResult: undefined,
+      cancelResult: undefined,
+      result: result ?? historyItem.result,
+    };
+
+    this.setState(({ consoleHistory }) => {
+      const itemIndex = consoleHistory.lastIndexOf(historyItem);
+      if (itemIndex < 0) {
+        log.error(`historyItem not found in consoleHistory`);
+        return { consoleHistory };
```
vbabich, line 348: Do we want a re-render in this case? Might just return a `null` if we don't.

```suggestion
        return null;
```

vbabich (approved):

packages/console/src/Console.tsx
mofojed: Sweet, I was trying to return `undefined` before and getting a TypeScript error - should've just returned `null` instead, I'm a bit of a moron. Good suggestion.

PR #981, opened 2022-12-28
Add support for plotting series point shapes and point size

mofojed:
- Add the series shape to markers
- Add shape size
- Fix when the shape is an empty string
- Tested with my deephaven-core branch from PR https://github.com/deephaven/deephaven-core/pull/3246
- Add e2e tests for checking point shapes and sizes (will need to update docker image used once my above PR is merged and container is published)
- Tested with following code in python:
```
from deephaven.plot import Figure
import math

math_funcs = ["sin", "cos"]
shapes = ["SQUARE", "CIRCLE", "UP_TRIANGLE", "DOWN_TRIANGLE", "RIGHT_TRIANGLE", "LEFT_TRIANGLE", "DIAMOND"]
# Unsupported shapes: ["ELLIPSE", "HORIZONTAL_RECTANGLE", "VERTICAL_RECTANGLE"]

# Create a generic table that has enough columns to display all the shapes
# Re-uses some of the funcs
t = empty_table(50).update(["x=i*0.1"])
for i in range(len(shapes)):
    t = t.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])

# Generate the table and figure based on the shapes created
p = Figure()
for i in range(len(shapes)):
    p = p.plot_xy(series_name=shapes[i], t=t, x="x", y=f"y{i}").point(shape=shapes[i], size=len(shapes) - i, visible=True)

p = p.show();
```


packages/chart/src/ChartUtils.ts
```diff
@@ -623,13 +653,27 @@ class ChartUtils {
     seriesVisibility: boolean | 'legendonly',
     theme = ChartTheme
   ): Partial<PlotData> {
-    const { name, plotStyle, lineColor, shapeColor, sources } = series;
+    const {
+      name,
+      isLinesVisible,
+      isShapesVisible,
```
mattrunyon, line 659: I think our types in the shim are slightly off for these. We have them as booleans when they should be optional booleans or `boolean | null`?

packages/chart/src/ChartUtils.ts
```diff
@@ -757,13 +805,60 @@ class ChartUtils {
...
   }
 
+  /**
+   * Get the marker symbol for the provided Deephaven shape
+   * Deephaven shapes: https://deephaven.io/enterprise/docs/plotting/visual-formatting/#point-formatting
+   * Plotly shapes: https://plotly.com/javascript/reference/scattergl/#scattergl-marker-symbol
+   * Table of plotly shapes: https://plotly.com/python/marker-style/
+   * @param shape Shape to get the marker symbol for
+   */
+  static getMarkerSymbol(shape: string): MarkerSymbol {
```
mattrunyon, line 834: Can you rename the param to `deephavenShape` just so it's a bit more clear which one is the param

packages/chart/src/ChartUtils.ts
```diff
@@ -757,13 +805,60 @@ class ChartUtils {
...
+
     // Skipping pie charts
     // Pie slice visibility is configured in chart layout instead of series data
     if (seriesVisibility != null && plotStyle !== dh.plot.SeriesPlotStyle.PIE) {
       seriesData.visible = seriesVisibility;
     }
   }
 
+  /**
+   * Get the marker symbol for the provided Deephaven shape
```
mattrunyon, line 828:
```suggestion
   * Get the Plotly marker symbol for the provided Deephaven shape
```

packages/chart/src/ChartUtils.ts
```diff
@@ -757,13 +805,60 @@ class ChartUtils {
...
+      case 'DOWN_TRIANGLE':
+        return 'triangle-down';
+      case 'RIGHT_TRIANGLE':
+        return 'triangle-right';
+      case 'LEFT_TRIANGLE':
+        return 'triangle-left';
+      // There don't seem to be any plotly equivalents for ellipse or rectangles
+      // Rectangles could be `line-ew`, `line-ns`, or `hourglass` and `bowtie` instead?
+      // Ellipse could be `asterisk` or `diamond-wide` instead?
+      // Just throw an error for now, we've already got a bunch of types.
```
mattrunyon, line 853: Should open a ticket for these?

tests/figure.spec.ts
```diff
@@ -30,3 +39,44 @@ ${figureName} = Figure().plot_xy(series_name="Test", t=empty_table(100).update([
...
+    ${tableName} = ${tableName}.update([f"y{i}=Math.{math_funcs[i % len(math_funcs)]}(x+{math.floor(i / len(math_funcs))})"])
+
+# Generate the table and figure based on the shapes created
+${figureName} = Figure()
+for i in range(len(shapes)):
+    ${figureName} = ${figureName}.plot_xy(series_name=shapes[i], t=${tableName}, x="x", y=f"y{i}").point(shape=shapes[i], size=len(shapes) - i, visible=True)
+
+${figureName} = ${figureName}.show();`;
+
+  await typeInMonaco(page, command);
```
mattrunyon, line 67: Why `typeInMonaco` here instead of just pasting the command? I feel like we should only type in monaco when testing things directly related to typing (autocomplete, etc.). This feels fine to just paste which should also speed up the test a bit

mattrunyon (changes requested):

packages/chart/src/ChartUtils.ts
```diff
@@ -210,18 +217,41 @@ class ChartUtils {
   /**
    * Converts the Iris plot style into a plotly chart mode
    * @param plotStyle The plotStyle to use, see dh.plot.SeriesPlotStyle.*
+   * @param shape The shape of the markers to use
    */
   static getPlotlyChartMode(
-    plotStyle: SeriesPlotStyle
+    plotStyle: SeriesPlotStyle,
+    isLinesVisible: boolean | null,
+    isShapesVisible: boolean | null
```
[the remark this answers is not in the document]
mofojed: Yea I was mainly matching the API but all true and I'll clean this up instead.

packages/chart/src/ChartUtils.ts
mofojed: Yup

packages/chart/src/ChartUtils.ts
mofojed: https://github.com/deephaven/deephaven-core/issues/3253

tests/figure.spec.ts
mofojed: Yea, I couldn't get paste working right... let me try taking another crack at it.

mattrunyon: It might be the clipboard context isn't allowed in unsecure environments or something

You could try `page.howeverYouGetMonaco().fill(command)`

mofojed: Interestingly, `locator('.console-input textarea').fill(text)` does work in that it fills in the text immediately; _however_ is screws up the indenting in monaco editor:
image

I can simulate a paste event in Chromium and Safari though, and fall back to typing in Firefox (paste doesn't work there for some reason).

mofojed: For some reason on the latest run, the download table is flaking and had to run twice, with the first run taking ~2m. Not sure why it flaked out, but pasting here for posterity:
image

As I wasn't changing this test, and it seems to be intermittent, I won't worry about it right now...

mattrunyon (approved):

tests/table.spec.ts
```diff
@@ -1,16 +1,16 @@
 import { test, expect, Page } from '@playwright/test';
-import { generateVarName, pasteInMonaco } from './utils';
+import { makeTableCommand, pasteInMonaco } from './utils';
 
 // Run tests serially since they all use the same table
 test.describe.configure({ mode: 'serial' });
 
-async function openSimpleTable(page: Page) {
+async function openSimpleTable(browserName: string, page: Page) {
```
mattrunyon, line 7: I think it's possible to get this info from the `Page`, but this is fine too

`page.context().browser()?.browserType().name()` gets the name

mattrunyon (approved):

tests/table.spec.ts
mofojed: I was trying to figure that out, sweet. I'm going to clean that up.

tests/console.spec.ts
```diff
@@ -1,12 +1,16 @@
...
 
 let page: Page;
+let consoleInput: Locator;
 
 test.beforeAll(async ({ browser }) => {
   page = await browser.newPage();
   await page.goto('');
+
+  consoleInput = page.locator('.console-input');
+  await consoleInput.click();
```
mattrunyon, line 13: Are these clicks fallback for when "pasting" in Firefox? If so, I would put it in the `pasteInMonaco` function instead.

Doesn't look like the `typeInMonaco` function requires the click, but that seems to be a method that is really just "type in whatever is active"

Can be a follow up ticket to fix these utils though. If you open another ticket just re-request and I'll approve

mattrunyon (changes requested):

mattrunyon (approved):

PR #984, opened 2022-12-29
Use showLegend specified by JS API

### LINKED ISSUE #983: Pie chart legend visbility is ignored

mofojed:
- Read the `showLegend` property
- Apply it to the series instead of the layout
- Layout `showLegend` checks the series when no value is set: https://plotly.com/javascript/reference/layout/#layout-showlegend
- Fixes #983

Tested with the snippet in the issue. Also tested with `visible=False` to make sure it did not appear, and that not specifying the legend defaults to showing (as specified by the API).


packages/jsapi-shim/src/dh.types.ts
```diff
@@ -400,7 +400,7 @@ export interface Chart extends Evented {
   readonly title: string;
   readonly titleFont: string;
   readonly titleColor: string;
-  readonly isShowLegend: boolean;
+  readonly showLegend: boolean;
   readonly legendFont: string;
   readonly legendColor: string;
```
mattrunyon, line 405: Should we be passing these through too? Or are they passed through already

mattrunyon (changes requested):

packages/jsapi-shim/src/dh.types.ts
mofojed: They are not passed through; we're also not passing through `titleFont`. I've made a ticket for that: https://github.com/deephaven/web-client-ui/issues/985

mofojed:
Added an issue on deephaven-core so we can possibly differentiate between `showLegend` being explicitly set and not: https://github.com/deephaven/deephaven-core/issues/3254

With this change we end up always showing a legend unless the user explicitly disables it, which doesn't match our previous behaviour of not showing the legend if there's just 1 series on a line plot.

packages/chart/src/FigureChartModel.ts
```diff
@@ -180,6 +180,14 @@ class FigureChartModel extends ChartModel {
       this.layout.hiddenlabels = ChartUtils.getHiddenLabels(this.settings);
     }
 
+    // We only want to force hide the legend if there is only one series that is not a PIE
+    // Right now this means that if the user only has one series, they cannot explicitly show the legend until deephaven-core#3254 is implemented
+    // TODO: deephaven-core#3254, once done, this can be removed.
+    this.layout.showlegend =
+      this.data.length > 1 || series.plotStyle === dh.plot.SeriesPlotStyle.PIE
+        ? undefined
+        : false;
```
mattrunyon, line 189: Should this be something like this instead?
`this.layout.showlegend = this.data.length === 1 ? false : showLegend`

Since `showLegend` defaults to true, just use it unless only 1 series. That way a user could disable on pie still if they wanted to?

mattrunyon (changes requested):

packages/chart/src/FigureChartModel.ts
mofojed: There are two different `showlegend` properties - this one on the entire layout, and then there's one on the data trace itself.
In the JS API it is set on the `JsChart`, which I'm using to set on the all the series below that. In our API, a `Figure` can have multiple `Chart`s, which can have multiple `Series`s. Mapping that to plotly, the `Figure` essentially maps to `layout`, and `Series` maps to trace, whereas `Chart` maps to the axes/properties within a data trace.

And I guess this is a convoluted was of saying yea I think you're right. But we should still remove it once https://github.com/deephaven/deephaven-core/issues/3254 is implemented.

mattrunyon (approved):

PR #989, opened 2023-01-03
Add pre-dist-tag when publishing an alpha package

mofojed:
- It was always just using the tag `canary` instead of a pre-dist-tag specified
- Docs: https://github.com/lerna/lerna/tree/main/commands/publish#--preid


mofojed:
Yea, you can run `npm dist-tag rm @deephaven/code-studio markdownnotebooks` for example to remove the `markdownnotebooks` dist-tag.
But you make a good point, no need in making so much noise. I thought doing `pre-dist-tag` would not publish it in that list, but I was wrong. I'll just close this, and all alphas can publish under `canary` rather than poluting it further

PR #992, opened 2023-01-05
Change tooltip default placement to bottom instead of auto

### LINKED ISSUE #966: Change tooltips to be bound by viewport and show at bottom by default
We currently have tooltips default to auto positioning and bounded by the scrollParent (or whatever the default value in popper is). This can cause tooltips to not display in some odd cases. It also makes them inconsistent which could be annoying with several icon buttons next to each other if the tooltips decided to go left/right and cover the other buttons

Will need to add something like `modifiers: { preventOverflow: { boundariesElement: 'viewport' } },` to the Tooltip default options

dsmmcken:
Fixes #966

- Sets tooltip placement to bottom by default. Tooltips belong below by convention and not just the side with the most space. 
- Sets boundaries element to viewport by default on popper, instead of scroll parent. We want to allow poppers to escape their panel/scroll parent in almost all cases.

Testing: tried all the tooltips I could find in community, some have new positioning, but it's all fine.

### DIFF
--- a/packages/components/scss/BaseStyleSheet.scss
+++ b/packages/components/scss/BaseStyleSheet.scss
@@ -109,6 +109,7 @@ button:focus {
 }
 
 span.btn-disabled-wrapper {
+  display: inline-block;
   .btn.disabled,
   .btn:disabled {
     pointer-events: none;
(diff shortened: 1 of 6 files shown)


mofojed (changes requested):
Just clean up the PR description a bit - fix the typos/grammar, then add some details in the description of the PR about testing performed and how to verify this change.

mofojed (approved):

PR #940, opened 2022-12-12
Fix dragging empty tables

mattrunyon:
Fixes #932

PR #993, opened 2023-01-06
Fix scrollbars drawing without handles when grid is the perfect size

### LINKED ISSUE #561: Vertical scroll bar can be drawn without handle
**Description**

If the table is sized just right, the vertical scroll track is drawn, but not the handle. This occurs if the last row of the table goes under the horizontal scroll bar.

image

**Steps to reproduce**
1. Create a table with only a few rows such as this Python
```python
from deephaven.column import string_col, char_col, int_col

table = new_table([
    string_col("Long", ["A"*50 , "B"*500, "C"*5000, "D"*49900]),
    char_col("Char", "ABCD"),
    int_col("Int", [123456789, 234567890, 42, 8675309]),
])
```
2. Shrink to where the horizontal scroll bar appears.
3. Shrink vertically until the last row is just under the horizontal scroll bar.

**Expected results**

Both scroll bars fully appear

**Actual results**

Vertical scroll track appears, but not the handle

**Versions**
 - Deephaven: 0.11.9

mattrunyon:
Fixes #561

### DIFF
--- a/packages/grid/src/GridMetricCalculator.ts
+++ b/packages/grid/src/GridMetricCalculator.ts
@@ -824,7 +824,7 @@ export class GridMetricCalculator {
       x += columnWidth;
 
       if (x >= visibleWidth) {
-        return Math.min(lastLeft + 1, right ?? columnCount - 1);
+        return clamp(lastLeft + 1, 0, right ?? columnCount - 1);
       }
 
       lastLeft -= 1;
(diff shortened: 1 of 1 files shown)


packages/grid/src/GridMetricCalculator.ts
```diff
@@ -803,7 +819,7 @@ export class GridMetricCalculator {
       lastLeft = right;
     }
     let x = 0;
-    while (lastLeft > 0) {
+    while (lastLeft >= 0) {
```
mofojed, line 822: This is reverting your fix from https://github.com/deephaven/web-client-ui/pull/940, and that problem is now back. Need some way to handle that case.

mofojed (changes requested):

mofojed (approved):

PR #998, opened 2023-01-07
Fix cannot scroll left after marking columns as hidden

### LINKED ISSUE #936: Cannot scroll left after marking columns as hidden
**Description**

If you have a table scrolled all the way to the right, then open the Column Visibility & Ordering menu, then hide a bunch of columns, you get stuck showing only the far right column.

**Steps to reproduce**
1. Create a table with a bunch of columns, e.g.
```

size = 100
scale = 1000

random_values_with_null_NaN = empty_table(size).update([
    "MyString=(i%11==0? null : `a`+(int)(scale*(Math.random()*2-1)))",
    "MyInt=(i%12==0 ? null : (int)(scale*(Math.random()*2-1)))",
    "MyLong=(i%13==0 ? null : (long)(scale*(Math.random()*2-1)))",
    "MyFloat=(float)(i%14==0 ? null : i%10==0 ? 1.0F/0.0F: i%5==0 ? -1.0F/0.0F : (float) scale*(Math.random()*2-1))",
    "MyDouble=(double)(i%16==0 ? null : i%10==0 ? 1.0D/0.0D: i%5==0 ? -1.0D/0.0D : (double) scale*(Math.random()*2-1))",
    "MyBoolean = (i%17==0 ? null : (int)(10*Math.random())%2==0)",
    "MyChar = (i%18==0 ? null : new Character((char) (((26*Math.random())%26)+97)) )",
    "MyShort=(short)(i%19==0 ? null : (int)(scale*(Math.random()*2-1)))",
    "MyByte=(Byte)(i%19==0 ? null : new Byte( Integer.toString((int)(Byte.MAX_VALUE*(Math.random()*2-1)))))",
    "MyBigDecimal=(i%21==0 ? null : new java.math.BigDecimal(scale*(Math.random()*2-1)))",
    "MyBigInteger=(i%22==0 ? null : new java.math.BigInteger(Integer.toString((int)(scale*(Math.random()*2-1)))))"
])
```
2. Open the Column Visibility & Ordering menu
3. Scroll to the far right of the table
4. Hide a bunch of the columns

**Expected results**
4. Table to scroll back to the left and show all columns when no more scroll bar is necessary

**Actual results**
4. Table stays scrolled to the right, the scroll bar disappears and you cannot scroll to the left anymore.

Engine Version 0.19.0
Web UI Version 0.22.2
Java Version 11.0.15 
Barrage Version 0.5.0

Dec-07-2022 16-42-04

mattrunyon:
Fixes #936 

Also removed `left` and `top` from the render state passed to the render function. They aren't used and should be retrieved from metrics instead.

I have a version of this which instead updates `left` and `top` in the metric calculator instead of updating the grid state after metrics if that is preferred (relies on #997). In that case we would probably still want to update the state in Grid since a few handlers in Grid use that state.

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -810,25 +810,25 @@ class Grid extends PureComponent<GridProps, GridState> {
     canvasContext.scale(scale, scale);
   }
 
+  updateScrollBounds(): void {
+    if (!this.metrics) throw new Error('metrics not set');
+    const { left, top } = this.state;
+    const { lastLeft, lastTop } = this.metrics;
+    if (left > lastLeft) {
+      this.setState({ left: lastLeft, leftOffset: 0 });
+    }
+    if (top > lastTop) {
+      this.setState({ top: lastTop, topOffset: 0 });
+    }
+  }
+
   updateMetrics(state = this.state): GridMetrics {
     this.prevMetrics = this.metrics;
 
     const { metricCalculator } = this;
     const metricState = this.getMetricState(state);
     this.metrics = metricCalculator.getMetrics(metricState);
-
-    // Fix for https://github.com/deephaven/web-client-ui/issues/936
-    // If metrics update from something like visibility panel, the lastLeft
-    // Can change due to hidden columns. Left should always be <= lastLeft
-    // Same for top
-    const { left, top } = state;
-    const { lastLeft, lastTop } = this.metrics;
-    if (left > lastLeft) {
-      this.setState({ left: lastLeft });
-    }
-    if (top > lastTop) {
-      this.setState({ top: lastTop });
-    }
+    this.updateScrollBounds();
 
     return this.metrics;
   }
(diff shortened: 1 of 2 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -817,6 +817,19 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+    // Can change due to hidden columns. Left should always be <= lastLeft
+    // Same for top
+    const { left, top } = state;
+    const { lastLeft, lastTop } = this.metrics;
+    if (left > lastLeft) {
+      this.setState({ left: lastLeft });
+    }
+    if (top > lastTop) {
+      this.setState({ top: lastTop });
+    }
```
mofojed, line 831: You also need to set `leftOffset` and `topOffset` when resetting the scroll position.
```suggestion
    const { left, top } = state;
    const { lastLeft, lastTop } = this.metrics;
    if (left > lastLeft) {
      this.setState({ left: lastLeft, leftOffset: 0 });
    }
    if (top > lastTop) {
      this.setState({ top: lastTop, topOffset: 0 });
    }
```

We do this in `handleResize` as well, consider refactoring it into it's own `checkScrollBounds` method or something instead.

mofojed (changes requested):

PR #1000, opened 2023-01-09
Fix default button focusing on BasicModal re-render

### LINKED ISSUE #792: Close All locks up when there are multiple unsaved tabs
The web UI (v0.19.0, and earlier I believe) locks up when the "Close All" action tries to close multiple tabs with unsaved changes. In my case, Firefox eventually tells me that the website is locked up and gives me a "stop" button.

image

vbabich:
- Replace inline function in the `Modal` `onOpened` prop with a `useCallback` to prevent re-focusing default button on each `BasicModal` render
- Unit tests
- Fixes #792

### DIFF
--- a/packages/components/src/BasicModal.test.tsx
+++ b/packages/components/src/BasicModal.test.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { render } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


mofojed (approved):
Amazing. Great find!

...another reason to almost always use `useCallback` - to avoid bugs like this!

PR #1001, opened 2023-01-10
Update Jest to v29 and user-event to v14

### LINKED ISSUE #987: Update testing libraries
We're falling behind on Jest, RTL, user-event at least. Being behind on RTL and user-event causes some annoyance in finding the old docs. The API around user-event has changed fairly significantly between 12 and 14.

We're 3 major versions behind in Jest right now. Just good to try to keep up to date on that so it doesn't get harder and harder down the road.

mattrunyon:
Fixes #987 

Removed `flushPromises` util since we can instead use `waitFor` or `findBy` queries in RTL to await events instead.

Updated our `user-event` syntax to the latest recommended [here](https://testing-library.com/docs/user-event/setup#direct-apis). I'm guessing the old API of calling directly on `userEvent` import will go away at some point

Can't update `@testing-library/react` to v13 because it is for React 18+ only


packages/components/src/ItemList.test.tsx
```diff
@@ -45,44 +45,66 @@ it('mounts and unmounts properly', () => {
...
 describe('mouse', () => {
-  function clickItem(itemIndex: number, options = {}): void {
+  async function clickItem(
+    user: ReturnType<typeof userEvent.setup>,
+    itemIndex: number,
+    options: {
+      ctrlKey?: boolean;
+      shiftKey?: boolean;
+      dblClick?: boolean;
+    } = {}
```
mofojed, line 55: These options get used in a few places, may want to declare as a type in `TestUtils` to refactor it a bit.

packages/iris-grid/src/IrisGridModel.test.ts
```diff
@@ -154,11 +159,13 @@ it('closes the table correctly when the model is closed', () => {
 });
 
 describe('totals table tests', () => {
-  const TOTALS_CONFIG = { operationMap: { 1: ['Max', 'Min'] } };
+  const TOTALS_CONFIG = ({
+    operationMap: { 1: ['Max', 'Min'] },
+  } as unknown) as UITotalsTableConfig;
```
mofojed, line 164: Could rewrite to remove the cast:
```suggestion
  const TOTALS_CONFIG: UITotalsTableConfig = {
    operationMap: { 1: ['Max', 'Min'] },
    operationOrder: [],
    showOnTop: true,
  };
```

mofojed (changes requested):
Just need to update `@types/jest`

mofojed (approved):

PR #1007, opened 2023-01-11
Fix silent error when saving notebook with duplicate name

### LINKED ISSUE #956: Saving a new notebook with an existing name silently errors
**Description**

Saving a new notebook with a name that exists already silently errors in the browser console log.

**Steps to reproduce**

1. Create and save a notebook.
2. Create and save another notebook, but change the name to the name of the 1st saved notebook in the modal.

**Expected results**

3. Error message about file already existing (or ask to overwrite maybe)

**Actual results**

3. Modal stays open, no error is shown, browser console has an error message.

Browser error

```
[NewItemModal] Error: Filename Untitled-1.py does not start with expected root /notebooks
    at FileUtils.removeRoot (FileUtils.ts:235:13)
    at _GrpcFileStorage.removeRoot (GrpcFileStorage.ts:38:22)
    at _GrpcFileStorage.info (GrpcFileStorage.ts:114:22)
    at async _NewItemModal.validateName (NewItemModal.tsx:208:32)
```

**Versions**

 - Engine Version: 0.19.0
 - Web UI Version: 0.22.2

emilyhuxng:
Fixes #956 

When saving a notebook with a duplicate name, it will ask you if you want to overwrite.

### DIFF
--- a/packages/code-studio/src/storage/grpc/GrpcFileStorage.ts
+++ b/packages/code-studio/src/storage/grpc/GrpcFileStorage.ts
@@ -111,7 +111,7 @@ export class GrpcFileStorage implements FileStorage {
     const itemDetails = allItems[0];
     return {
       filename: this.removeRoot(itemDetails.filename),
-      basename: this.removeRoot(itemDetails.basename),
+      basename: itemDetails.basename,
       id: this.removeRoot(itemDetails.filename),
       type: itemDetails.type,
     };


mofojed (approved):
Great work! This can merge.

While testing this, I think I did find another error (which you may have mentioned in a chat). If you already have the file open, and then rename another file to match that name, that file stays open but with the old contents (which have now been overwritten). If you could file a new ticket for that, we can have a separate PR for that fix.

emilyhuxng:
> Great work! This can merge.
> 
> While testing this, I think I did find another error (which you may have mentioned in a chat). If you already have the file open, and then rename another file to match that name, that file stays open but with the old contents (which have now been overwritten). If you could file a new ticket for that, we can have a separate PR for that fix.

What should the expected result be?

mofojed:
@emilyhuxng good question. I _think_ the ideal expected result would be to transition the previously open file to an unsaved state, and if you want to save it again you need to pick a new name; that way we don't ever lose user data. Another acceptable solution would be to simply close other panels that were overwritten, though it's possible to lose data that way (the user did explicitly say "overwrite" though).
@dsmmcken may have some input.

dsmmcken:
image

vscode just prevents you from doing that.

PR #1011, opened 2023-01-12
Upgrade to lerna v6

### LINKED ISSUE #864: Update to Lerna v6
Lerna v6 is out: https://blog.nrwl.io/lerna-reborn-whats-new-in-v6-10aec6e9091c

Update to use it.

mattrunyon:
Fixes #864 

Lerna v6 comes w/ nx by default. I've setup the caching for build tasks so builds should be faster/cached locally when possible. `test` could be cached (see [here](https://nx.dev/recipes/other/root-level-scripts#configuring-a-root-level-target)), but we run it in watch mode usually so doesn't make as much sense.

Running `npm run build` twice in a row should result in the 2nd time taking 1 second or so to cache-hit everything. You can run `npm run clean:build` and then `npm run build` again and it will only take a few seconds to restore the output from the cache. `npm run clean` also cleans the nx cache.


nx.json
```diff
@@ -0,0 +1,71 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": [
+          "build",
+          "build:babel",
```
mofojed, line 8: Interesting - so the cacheable actions are the package actions, not the top level actions?

mofojed (changes requested):
It looks like `npm start` doesn't work anymore - something happen with the vite proxy?

package.json
```diff
@@ -86,7 +87,7 @@
     "@types/lodash.throttle": "^4.1.1",
     "@types/memoizee": "^0.4.5",
     "@types/node": "^16.11.7",
-    "@types/papaparse": "^5.3.2",
+    "@types/papaparse": "5.3.2",
```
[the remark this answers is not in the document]
mattrunyon: If it updates to the latest `^5.3.2`, there's TS errors. I had to delete and recreate `package-lock.json` at one point and then got papaparse type errors

nx.json
mattrunyon: Yep. This was generated from their CLI tool which generates a list of operations and asks which are cacheable

mattrunyon:
For now, I've disabled nx until https://github.com/nrwl/nx/issues/14010 is fixed

An alternative would be to change our `.env` files a bit and use [Nx's convention](https://nx.dev/recipes/environment-variables/define-environment-variables), but name the task `development` so we can at least get `.env`, `.env.local`, and `.env.development`. Then we have just 1 `.env.local` file, but it would apply to any task run since Nx doesn't load `.env.development.local` and Vite won't override the env variable if it's already set when Vite starts

mofojed (dismissed):
I guess we're still checking in the `nx.json` file even if we're not using it yet so we don't have to rebuild later?

mofojed:
Changed type to `build`: 
> build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)

It could be `chore` but I feel like `build` is more specific:
> chore: Other changes that don't modify src or test files

mofojed (approved):

PR #1013, opened 2023-01-12
Validate custom column names

### LINKED ISSUE #939: Creating a custom column with invalid name errors panel
**Description**

If you create a custom column on a table with an invalid name, an error is thrown and the panel errors out

**Steps to reproduce**
1. Create a table. E.g. in Python

```python

table = new_table([
])
```

2. Create a custom column from the table sidebar with an invalid name. E.g. "test name" which has a space and is not allowed

**Expected results**

3. Error message in the input

**Actual results**

3. Panel catches error and displays error state

Should also check the column name does not exist in addition to being valid

**Versions**
 - Deephaven: 0.22

mattrunyon:
Fixes #939 

Does not validate that the names aren't duplicates of existing column names, only of existing custom columns. The API seems to allow custom columns that overwrite an existing column name. For example, table with coiumns `A, B, C`. Create a custom column `A=C` and `B=A`. All of the columns will display column `C`. If you swap the order of the custom columns, then `B` will show the original `A` and `A` will show `C`.

The API does not allow invalid custom column names (table column validation) or duplicate custom column names.

### DIFF
--- a/packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
+++ b/packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
@@ -9,6 +9,7 @@ import { dhNewCircleLargeFilled, vsWarning, vsPass } from '@deephaven/icons';
 import CustomColumnInput from './CustomColumnInput';
 import './CustomColumnBuilder.scss';
 import IrisGridModel from '../IrisGridModel';
+import { DbNameValidator } from '@deephaven/utils';
 
 export type CustomColumnKey = 'eventKey' | 'name' | 'formula';
 
@@ -334,6 +335,9 @@ class CustomColumnBuilder extends Component<
   renderSaveButton(): ReactElement {
     const { inputs, isCustomColumnApplying, isSuccessShowing } = this.state;
     const saveText = inputs.length > 1 ? 'Save Columns' : 'Save Column';
+    const areNamesValid = inputs.every(({ name }) =>
+      DbNameValidator.isValidColumnName(name)
+    );
 
     return (
       <Button
@@ -341,7 +345,7 @@ class CustomColumnBuilder extends Component<
         className={classNames('btn-apply', {
           'btn-spinner': isCustomColumnApplying,
         })}
-        disabled={isSuccessShowing || isCustomColumnApplying}
+        disabled={isSuccessShowing || isCustomColumnApplying || !areNamesValid}
         onClick={this.handleSaveClick}
       >
         {isCustomColumnApplying && (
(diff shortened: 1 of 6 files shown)


mofojed (approved):
Would be nice to add some `CustomColumnBuilder` tests as well.

mattrunyon:
> Would be nice to add some `CustomColumnBuilder` tests as well.

Ask and you shall receive

The tests found a bug in Tab handling within the monaco editor (monaco was eating the tab). I have the fix for that behavior in this PR. I can split it out if you prefer

packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
```diff
@@ -334,21 +336,30 @@ class CustomColumnBuilder extends Component<
...
+      new Set(inputs.map(({ name }) => name)).size === inputs.length;
 
     return (
       <Button
         kind={isSuccessShowing ? 'success' : 'primary'}
         className={classNames('btn-apply', {
           'btn-spinner': isCustomColumnApplying,
         })}
-        disabled={isSuccessShowing || isCustomColumnApplying}
+        disabled={
```
mofojed, line 351: This is disabling the button when there are no custom columns; if you add a custom column, then want to remove it, there's no long a way to do that. May want to consider just dropping any that are both blank name and formula.

Also weird behaviour if you add a name, but then don't add a formula - you can still hit Save, but the column with the name doesn't appear. May want to check there are no blank formulas.

mofojed (changes requested):

packages/iris-grid/src/sidebar/CustomColumnBuilder.tsx
mattrunyon: The previous behavior was to drop anything that had either part blank.

But I also just discovered if both are blank and you save, it gets stuck in an infinite loader. Checked on Jackson too and same behavior. So need to check/filter that if allowing save w/ all blank

mattrunyon: I made the button visible if there's no custom columns. I also fixed the loader issue in the case where you had no custom columns and clicked save with no custom columns. That case triggered the loading state, but the grid panel listens for a table change event to stop the loader. That event never happens. The column builder panel also listens for the same event to dismiss the loader

dsmmcken: Adding a custom column with a name and formula, then removing the formula and hitting save still succeeds and leaves the previous formula on the table

mattrunyon: Looks like that bug exists in the published version too. I'll fix it

mattrunyon: I tried adding a test for this w/ RTL, but it seems if I load the panel w/ an existing formula, I can't delete the text in monaco. Oddly, I can add w/ keyboard events (adds to the front of the existing string), but delete key doesn't delete the existing text, select all then delete doesn't either

mattrunyon: Managed to get a test working by creating a new custom column, saving, then deleting the text and saving again. Guess it's something to do with monaco starting w/ the text vs adding it via RTL keyboard events.

mofojed (approved):

PR #1014, opened 2023-01-12
Adjust margin in column header grouping list

dsmmcken:
The negative margin was looking weird with the transparency overlapping creating a darker border, give the items a 1px bottom margin to create a small gap between items, then remove that gap on the depth lines with negative margin so they are still continuous.


dsmmcken:
Before:
image

After:
image

mattrunyon (approved):

PR #1015, opened 2023-01-12
Clean up the release action a little bit

mofojed:
- I verified that it was getting the changelog and bumping the version correctly by testing on my forked repo. You can take a look at the actions run: https://github.com/mofojed/web-client-ui/actions
- I think the action was already working, but was failing due to branch protection. New branch protection rules should allow pushing from GitHub actions and was recently fixed: https://github.com/integrations/terraform-provider-github/pull/1453/files
- Should probably move to conventional commits, and simply remove the use of `lerna-changelog`. That would be a team-wide change, but ultimately would be a cleaner process, as we could just run `lerna version` and it would automatically detect whether to do a major, minor, or patch version bump instead of having to manually enter the version

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -12,20 +12,21 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: '0' # This action defaults to only getting the latest commit. Setting to 0 makes it retrieve the full git commit history
+          # This action defaults to only getting the latest commit. Setting to 0 makes it retrieve the full git commit history
+          fetch-depth: '0'
       - run: |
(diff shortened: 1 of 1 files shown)


mattrunyon (changes requested):
Can't comment on the line directly since it wasn't changed, but we should switch to a different release action. The github provided action is unmaintained and warning about Node 12. There's a list of alternatives listed in the repo https://github.com/actions/create-release

This just creates the changelog and adds a release on GitHub, right? Which should then trigger the release action to actually publish to npm? If so, can you add a comment at the top of this workflow that briefly explains its purpose in the overall release flow?

mofojed:
Superceded by: https://github.com/deephaven/web-client-ui/pull/1015

PR #1016, opened 2023-01-12
Cancelled promises reject immediately

### LINKED ISSUE #949: Cancelled promises do not reject until original promise resolves
**Description**

When you call `cancellablePromise.cancel()`, the promise does not reject immediately. Instead, we wait for the original promise that was wrapped to resolve/reject and only then do we reject the promise. If calling cancel, we shouldn't care about the original promise any more, so we should just discard it and reject our cancellable promise

Also, you cannot do `PromiseUtils.makeCancelable(something).then().catch()` and still call cancel. You have to create the promise in 1 statement and add then/catch in another. Might be a nice thing to cleanup as well

**Steps to reproduce**
```ts
import { PromiseUtils } from '@deephaven/utils';

const a = new Promise((resolve, reject) => {
  setTimeout(() => resolve(), 120 * 1000); // 2 minutes until promise resolves
});

/* This syntax doesn't actually work, but it should. Cancel gets removed
const b = PromiseUtils.makeCancelable(a)
  .then(() => console.log('resolved'))
  .catch(() => console.log('rejected'));

b.cancel();
*/

const b = PromiseUtils.makeCancelable(a);
b.then(() => console.log('resolved')).catch(() => console.log('rejected'));

b.cancel();
```

**Expected results**

Logs rejected immediately

**Actual results**

Logs rejected after 2 minutes

**Versions**
 - Web UI Version: 0.22.2

emilyhuxng:
Fixes #949 

When the cancel function is called on a cancellablePromise, the wrapper promise immediately rejects.

### DIFF
--- a/packages/utils/src/PromiseUtils.ts
+++ b/packages/utils/src/PromiseUtils.ts
@@ -16,30 +16,22 @@ export class PromiseUtils {
     promise: Promise<T> | T,
     cleanup?: (val: T) => void | null
   ): CancelablePromise<T> {
-    let hasCanceled = false;
     let resolved: T | undefined;
+    let rejectFn: (val?: any) => void;
 
     const wrappedPromise = new Promise((resolve, reject) => {
-      Promise.resolve(promise).then(
-        val => {
-          if (hasCanceled) {
-            if (cleanup) {
-              cleanup(val);
-            }
-            reject(new CanceledPromiseError());
-          } else {
-            resolved = val;
-            resolve(val);
-          }
-        },
-        error =>
-          hasCanceled ? reject(new CanceledPromiseError()) : reject(error)
-      );
+      rejectFn = reject;
+      Promise.resolve(promise)
+        .then(val => {
+          resolved = val;
+          resolve(val);
+        })
+        .catch(error => reject(error));
     });
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (wrappedPromise as any).cancel = () => {
-      hasCanceled = true;
+      rejectFn(new CanceledPromiseError());
 
       if (resolved != null && cleanup) {
         cleanup(resolved);
--- a/packages/utils/src/PromiseUtils.ts
+++ b/packages/utils/src/PromiseUtils.ts
@@ -17,6 +17,7 @@ export class PromiseUtils {
     cleanup?: (val: T) => void | null
   ): CancelablePromise<T> {
     let resolved: T | undefined;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let rejectFn: (val?: any) => void;
 
     const wrappedPromise = new Promise((resolve, reject) => {
--- a/packages/utils/src/PromiseUtils.ts
+++ b/packages/utils/src/PromiseUtils.ts
@@ -18,8 +18,7 @@ export class PromiseUtils {
   ): CancelablePromise<T> {
     let hasCanceled = false;
     let resolved: T | undefined;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    let rejectFn: (val?: any) => void;
+    let rejectFn: (val?: unknown) => void;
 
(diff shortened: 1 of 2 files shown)


packages/utils/src/PromiseUtils.ts
```diff
@@ -16,30 +16,22 @@ export class PromiseUtils {
...
-            resolve(val);
-          }
-        },
-        error =>
-          hasCanceled ? reject(new CanceledPromiseError()) : reject(error)
-      );
+      rejectFn = reject;
+      Promise.resolve(promise)
+        .then(val => {
+          resolved = val;
```
mofojed, line 27: We should still check if `hasCanceled` is true here, and run the cleanup if it has canceled.
This behaviour is used in a few places, where you can just call `.cancel()` on the promise and it will run the cleanup when it can.

An example is in IrisGridPanel: https://github.com/deephaven/web-client-ui/blob/main/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx#L517

mofojed (changes requested):

packages/utils/src/PromiseUtils.ts
emilyhuxng: I think I'm a bit confused about what is happening. I thought that this if statement covers it since if it's pending then nothing needs to be done. If it is fulfilled and rejected, then no cleanup necessary. If it is fulfilled and resolved, then resolved shouldn't be null so cleanup will run.
```
if (resolved != null && cleanup) {
  cleanup(resolved);
}
```

mofojed: This `.then` is after the underlying promise is fulfilled. The below block (`if (resolved != null && cleanup)`) is after the cancel is called.
I think there are four scenarios involving cancelation:
1) Original promise resolved, then cancel is called
2) Original promise rejected, then cancel is called
3) Cancel is called, then original promise is resolved
4) Cancel is called, then original promise is rejected

In both cases 1 and 3, when the original promise is resolved, the cleanup function should be called when cancel is called. Without checking `hasCanceled` within our `.then` here, that cleanup function wasn't getting called in scenario 3.

packages/utils/src/PromiseUtils.ts
```diff
@@ -18,10 +18,13 @@ export class PromiseUtils {
...
     let resolved: T | undefined;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let rejectFn: (val?: any) => void;
 
     const wrappedPromise = new Promise((resolve, reject) => {
-      Promise.resolve(promise).then(
-        val => {
+      rejectFn = reject;
+      Promise.resolve(promise)
+        .then(val => {
```
mofojed, line 27: With these changes, `reject` is called right after `hasCanceled` is set to true. If `hasCanceled` is true, there's no reason to call `reject` in those scenarios. It will just get swallowed up since the promise is already fulfilled, but it is unnecessary and could be removed (lines 32, and the true case on line 39).

packages/utils/src/PromiseUtils.ts
```diff
@@ -18,10 +18,13 @@ export class PromiseUtils {
   ): CancelablePromise<T> {
     let hasCanceled = false;
     let resolved: T | undefined;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let rejectFn: (val?: any) => void;
```
mofojed, line 22: You can use `unknown` here instead of `any` and get rid of the disabling eslint.

packages/utils/src/PromiseUtils.ts
emilyhuxng: Okay, I understand now. I mistakenly thought that reject and resolve statements act as return statements so they would stop the execution code once reject was called.

mofojed (commented):
Works, some minor cleanup suggestions.

mofojed (approved):

PR #1021, opened 2023-01-13
Convert `serverConfigValues` from a `Map` to a JSON Object

Zhou-Ziheng:
On Enterprise, `serverConfigValues` are read from the server and saved as a JSON object in redux. However, in community, `serverConfigValues` are stored as a map. This creates incompatibility between redux states and all redux related functions between enterprise and community. 

Converted community serverConfigValues to a JSON object to fix the incompatibility.

### DIFF
--- a/packages/code-studio/src/main/AppInit.tsx
+++ b/packages/code-studio/src/main/AppInit.tsx
@@ -255,9 +255,14 @@ function AppInit(props: AppInitProps) {
           ),
         },
       };
+      const serverConfigValues = {
+        barrageVersion: serverConfig.get('barrage.version'),
+        javaVersion: serverConfig.get('java.version'),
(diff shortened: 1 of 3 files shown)


mofojed (changes requested):
I think I would prefer changing the Enterprise API to match Community instead, and continue using a Map for these properties.

See this MDN page for Map vs. Object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps

In particular, iterating on items and seeing the size of the map can be useful.

Also, in this change you are stripping all values other than `barrageVersion`, `javaVersion`, and `deephavenVersion` from the server config values - we shouldn't be doing that, we should keep _all_ the values the server sends in the serverConfigValues. It's possible we will pass this value into plugins, or some other component and we need another value from the server.

Issue #554, opened 2022-05-05
Collapsed quick filter bar display issues


mattrunyon:
**Description**

There are at least 2 issues w/ collapsed quick filter bar displays

1. When a table has frozen columns and horizontally scrolls, the quick filters do not draw properly for the frozen columns
![image](https://user-images.githubusercontent.com/8619789/167028001-2e31e4e7-5320-4471-825d-2af806637bb9.png)

2. When a table has frozen columns w/ a collapsed quick filter bar, the bar does not draw on frozen columns when horizontally scrolled more than 1 column to the right
![image](https://user-images.githubusercontent.com/8619789/167028101-fbf956b6-0130-4439-bbc6-5f5592628c80.png)

3. When a table has a collapsed quick filter bar, it does not extend to the right edge. This is easiest to see the inconsistency by adding a sort to the table as well (ignore the missing burger menu in this screenshot)
![image](https://user-images.githubusercontent.com/8619789/167028368-3bfbb304-e777-4a1d-8718-824ae29adeac.png)

**Versions**
 - Deephaven: 0.11.8

vbabich:
Don is saying the frozen column filter and the header should get a double border, too.

PR #1025, opened 2023-01-14
Fix vertical scrollbar appearing over column header group names

### LINKED ISSUE #1022: Vertical scrollbar does not start below column header groups
**Description**

When creating a table that scrolls with column header groups, the vertical scrollbar is not below the header groups.

**Steps to reproduce**

Create a table w/ column header groups

Example Python w/ examples data installed
```python
from deephaven.parquet import read

source = read("/data/examples/Taxi/parquet/taxi.parquet")

groups = [
    {'name': 'DirectCosts', 'children': ['fare_amount', 'mta_tax', 'tolls_amount'], 'color': 'PALE_GREEN'},
    {'name': 'IndirectCosts', 'children': ['extra', 'improvement_surcharge'], 'color': 'TOMATO'},
    {'name': 'Costs', 'children': ['IndirectCosts', 'DirectCosts', 'tip_amount', 'total_amount'], 'color': 'SLATEBLUE'},
    {'name': 'Times', 'children': ['tpep_pickup_datetime', 'tpep_dropoff_datetime']}
]

source_with_groups = source.layout_hints(column_groups=groups)
```

**Expected results**

Scrollbar starts below all headers

**Actual results**

image

**Versions**

 - Web UI Version: 0.25.0

mattrunyon:
Fixes #1022 

I think we should do a patch release to put this in the next core release since column grouping will be released w/ that.


mofojed (changes requested):
Looks like the snapshots need updating - should check with @dsmmcken if this is desired behaviour (the scrollbar being below the quick filter bar now).
Also - this would be a good case to have an e2e test for as well (column grouping in general).

dsmmcken:
Quick filter bar being below the scroll bar fixes this old [enterprise issue](https://deephaven.atlassian.net/browse/DH-9181) where you can't grab the scrollbar if it is overlapping. So it's probably better to be below it for multiple reasons.

image

However this looks dumb... so I am not sure this is quite what we want:
2f24ce551ffb3d28be4190af96e8a7b0211ca932

If we think about it in context of https://github.com/deephaven/web-client-ui/issues/554
> When a table has a collapsed quick filter bar, it does not extend to the right edge. This is easiest to see the inconsistency by adding a sort to the table as well (ignore the missing burger menu in this screenshot).

Maybe the correct solution would be to fill that dead area with that dark blue, or even a blue that is even darker so it's not a distractingly long blue bar:

image

Maybe like this? And do the same for collapsed.

dsmmcken (changes requested):
Add dark blue bar per previous comment

dsmmcken (approved):

mofojed (dismissed):

dsmmcken (approved):
nice, adding more e2e for grid is definitely useful

PR #1028, opened 2023-01-16
Clicking save on a new notebook automatically selects input field

### LINKED ISSUE #1020: Save as on a notebook should focus the name field
When I click the save button on a notebook, the name input should be selected when the modal opens because a user is likely to rename the file.

`input.select()` should focus and select all text. Focus would also be acceptable, but probably opt for select.

emilyhuxng:
Fixes #1020 

After creating a new file, saving opens the modal but the focus is on a div rather than the input field, so I removed the tabIndex field.

### DIFF
--- a/packages/file-explorer/src/NewItemModal.tsx
+++ b/packages/file-explorer/src/NewItemModal.tsx
@@ -329,7 +329,8 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
       const { value } = input;
       const selectionEnd =
         type === 'directory' ? value.length : value.lastIndexOf('.');
-      setTimeout(() => input.focus(), 100);
+      // setTimeout(() => input.focus(), 100);
+      input.focus();
       input.setSelectionRange(
         0,
         selectionEnd > 0 ? selectionEnd : value.length
--- a/packages/file-explorer/src/NewItemModal.tsx
+++ b/packages/file-explorer/src/NewItemModal.tsx
@@ -329,7 +329,6 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
       const { value } = input;
       const selectionEnd =
         type === 'directory' ? value.length : value.lastIndexOf('.');
-      // setTimeout(() => input.focus(), 100);
       input.focus();
       input.setSelectionRange(
         0,
--- a/packages/file-explorer/src/NewItemModal.tsx
+++ b/packages/file-explorer/src/NewItemModal.tsx
@@ -329,7 +329,7 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
       const { value } = input;
       const selectionEnd =
         type === 'directory' ? value.length : value.lastIndexOf('.');
-      input.focus();
+      setTimeout(() => input.focus(), 100);
       input.setSelectionRange(
         0,
         selectionEnd > 0 ? selectionEnd : value.length
--- a/packages/golden-layout/src/container/ItemContainer.ts
+++ b/packages/golden-layout/src/container/ItemContainer.ts
@@ -31,7 +31,7 @@ export default class ItemContainer<
   _element = $(
     [
       '<div class="lm_item_container">',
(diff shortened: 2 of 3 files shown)


dsmmcken:
I don't think that should be removed, it's needed for another reason:
https://github.com/illumon-public/golden-layout/commit/0140f78373e63b00967b0f068606031df2f9b231

Pretty sure it relates to focusing inside the container on click.

@mofojed we should archive that repo.

emilyhuxng:
Okay, I'll look into other solutions.

dsmmcken:
See matt's original comment on the the ticket. I think on open of the save as modal, you just need to have a ref to the input then call `input.select()`

emilyhuxng:
The weird thing is this feature was already implemented. When I open some tabs and reload the page, then try to save those tabs, the select works. However, when I make a new file and immediately try to save it, the input will not focus.

emilyhuxng:
I suspect that even though it's returning an input element, focus can not be called on it at the time so focus is called on the div instead, but I'm unsure. Adding a timeout seems to work and the div is no longer ever focused but I read that using a timeout can be bad practice. Is it acceptable in this case?

mofojed:
@emilyhuxng This blog post is a good read that details the React event system, including capture/bubbling: https://betterprogramming.pub/whats-the-difference-between-synthetic-react-events-and-javascript-events-ba7dbc742294
One thing it doesn't detail is how using something like jquery to add events listeners works when using React as well. That could probably be it's own blog post. 

@dsmmcken thinks the issue is that the Tab `click` handler is being triggered _after_ this new file click handler, and thus focus is getting set on the panel after it's set on this input. This is another case where we can listen to the event in the `capture` phase to set focus within the panel, as that will ensure it doesn't interfere with event listeners within the panel handling the event.

As a side note, `golden-layout` is a package that was originally written in JS/jquery which we forked because it [dropped React support](https://github.com/pbklink/golden-layout/commit/6c3935262ec98d370825d0b89e9d146ecfaa7011). There was a [discussion](https://github.com/golden-layout/golden-layout/issues/606) about taking that direction with golden-layout. 
Ideally, we'd have a layout system with a similar feature as golden-layout that had better React support; however rewriting the layout engine would be quite a lot of work (including re-writing dashboards, and the event hub system panels use to communicate with each other).

packages/golden-layout/src/container/ItemContainer.ts
```diff
@@ -31,7 +31,7 @@ export default class ItemContainer<
   _element = $(
     [
       '<div class="lm_item_container">',
-      '<div class="lm_content" tabindex="-1"></div>',
+      '<div class="lm_content" tabIndex="-1"></div>',
```
mofojed, line 34: This is `jQuery`, not react, so you still want it to be just [`tabindex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex). 
```suggestion
      '<div class="lm_content" tabindex="-1"></div>',
```

packages/file-explorer/src/NewItemModal.tsx
```diff
@@ -329,6 +329,7 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
       const { value } = input;
       const selectionEnd =
         type === 'directory' ? value.length : value.lastIndexOf('.');
+      // setTimeout(() => input.focus(), 100);
```
mofojed, line 332: Remove commented out code.
```suggestion
```

packages/golden-layout/src/controls/Tab.ts
```diff
@@ -81,8 +81,13 @@ export default class Tab {
     if (isComponent(this.contentItem)) {
       // add focus class to tab when content
       this.contentItem.container._contentElement
-        .on('focusin click', this._onTabContentFocusIn)
+        .on('focusin', this._onTabContentFocusIn)
         .on('focusout', this._onTabContentFocusOut);
+      this.contentItem.container._contentElement[0].addEventListener(
```
mofojed, line 86: You should also `removeEventListener` in `_$destroy` for completeness.

PR #1029, opened 2023-01-17
Fix cell overflow viewer with browser zoom

### LINKED ISSUE #834: Cell overflow viewer with browser zoom is scaled incorrectly
1. Add 150% browser zoom
2. Click the view contents button on a cell with many lines of overflow

Expected: overflow modal is full height, shows all the contents
Actual: overflow modal is incorrectly sized, really short. Toggling the line number triggers it re-draw and show correctly.

image

emilyhuxng:
Fixes #834 

Set the height when the editor is initialized.

### DIFF
--- a/packages/iris-grid/src/IrisGridCellOverflowModal.tsx
+++ b/packages/iris-grid/src/IrisGridCellOverflowModal.tsx
@@ -130,6 +130,7 @@ export default function IrisGridCellOverflowModal({
         setHeight(0);
       }}
       className="theme-bg-dark cell-overflow-modal"
+      size="xl"
     >
       <ModalHeader toggle={onClose}>
         <h5 className="overflow-modal-title">Cell Contents</h5>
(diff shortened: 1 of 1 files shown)


mattrunyon:
Looks good. Can you add a `size='xl'` to the overflow modal? Looks like it was removed by mistake in #611. This should give us more width when needed up to a certain amount. It does introduce a little weirdness when zooming in since you can cross the browser size thresholds in bootstrap and actually get a smaller (by percent) modal, but I wouldn't worry about that for now

mattrunyon (changes requested):
Looks good. Can you add a `size='xl'` to the overflow modal? Looks like it was removed by mistake in #611. This should give us more width when needed up to a certain amount. It does introduce a little weirdness when zooming in since you can cross the browser size thresholds in bootstrap and actually get a smaller (by percent) modal, but I wouldn't worry about that for now

mattrunyon (approved):

mofojed (approved):

PR #1031, opened 2023-01-18
Existing rows in append-only tables cannot be edited

### LINKED ISSUE #990: Append-only input tables don't properly add data if you attempt to modify a pre-existing row
**Description**

Attempting to modify a cell in an append-only input table results in the new data being written twice.

**Steps to reproduce**

Create an append-only input table.

```python
from deephaven import InputTable

t = empty_table(10).update(["X = i"])
t2 = InputTable(init_table=t)
```

Then, attempt to modify a pre-existing cell. Hit enter (or the commit button), and the new data will be written twice to the bottom of the column.

**Expected results**

I expect one of two things to happen:

- To not be able to type in the cell
- My data gets written to the end of the table once.

**Actual results**

The data is entered into the table twice. There was an instance where I found that `table.size` reported one less row than what the UI shows. I have yet to repro that.

**Versions**

 - Engine Version: 0.19 (latest)
 - Web UI Version: latest
 - Java Version: 11
 - Barrage Version: unsure
 - OS: OS X Catalina
 - Browser: Chrome
 - Docker: 20.10.13

emilyhuxng:
Fixes #990 

Allow only pendingRow changes if there are no key columns.

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModelTemplate.ts
+++ b/packages/iris-grid/src/IrisGridTableModelTemplate.ts
@@ -1542,22 +1542,40 @@ class IrisGridTableModelTemplate<
   }
 
   isEditableRange(range: GridRange): boolean {
-    return (
-      this.inputTable != null &&
-      GridRange.isBounded(range) &&
-      ((this.isPendingRow(range.startRow) && this.isPendingRow(range.endRow)) ||
-        (range.startColumn >= this.inputTable.keyColumns.length &&
-          range.endColumn >= this.inputTable.keyColumns.length)) &&
-      range.startRow >= this.floatingTopRowCount &&
-      range.startRow <
-        this.floatingTopRowCount + this.table.size + this.pendingRowCount &&
-      range.endRow <
-        this.floatingTopRowCount + this.table.size + this.pendingRowCount &&
-      (this.inputTable.keyColumns.length !== 0 ||
-        (this.inputTable.keyColumns.length === 0 &&
-          this.isPendingRow(range.startRow) &&
-          this.isPendingRow(range.endRow)))
-    );
+    // Make sure we have an input table and a valid range
+    if (this.inputTable == null || !GridRange.isBounded(range)) {
+      return false;
+    }
+
+    // Check that the edit is in the editable range
+    // If an input table has keyed columns, the non-key columns are always editable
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1552,7 +1552,11 @@ class IrisGridTableModelTemplate<
       range.startRow <
         this.floatingTopRowCount + this.table.size + this.pendingRowCount &&
       range.endRow <
-        this.floatingTopRowCount + this.table.size + this.pendingRowCount
+        this.floatingTopRowCount + this.table.size + this.pendingRowCount &&
+      (this.inputTable.keyColumns.length !== 0 ||
+        (this.inputTable.keyColumns.length === 0 &&
+          this.isPendingRow(range.startRow) &&
+          this.isPendingRow(range.endRow)))
```
mofojed, line 1559: Nice, this fixes it, but I think we could re-write this a little bit so it's a little more readable.
On line `1548` there's already the `this.isPendingRow(range.startRow) && this.isPendingRow(range.endRow)` check. I would write this logic into that part into there, e.g.:
```
      ((this.isPendingRow(range.startRow) && this.isPendingRow(range.endRow)) ||
        (this.inputTable.keyColumns.length > 0 && range.startColumn >= this.inputTable.keyColumns.length &&
range.endColumn >= this.inputTable.keyColumns.length)) &&
```

Side note, because of how large this conditional has become, it would almost make more sense to just expand it out so it's not so hard on the brain, something like:
```
isEditableRange(range: GridRange): boolean {
  // Make sure we have an input table and a valid range
  if (this.inputTable == null || !GridRange.isBounded(range)) {
    return false
  }
  
  // Check that the edit is in the editable range
  // If an input table has keyed columns, the non-key columns are always editable
  // If an input table does not have key columns, it is append only and existing rows cannot be editable
  // Pending rows are always editable
  if (...) {
    return false;
  }
  
  // Editing the aggregations/totals which are floating above/below is not allowed
  if (...) {
    return false;
  }
  
  return true;
}
```

mofojed (changes requested):

mofojed (approved):
Great work!

PR #1034, opened 2023-01-18
Fix up re-hydration of hierarchical tables

### LINKED ISSUE #1033: Hierarchical tables fail to load on page refresh
**Description**

When a hierarchical/tree table is in the layout, it fails to load the table on page refresh

**Steps to reproduce**

1. Create a hierarchical table (Python example below)
   ```python
   from deephaven.constants import NULL_INT

   t1 = empty_table(10_000).update_view(["ID = i", "Parent = i == 0 ? NULL_INT : (int)(i / 10)"])
   t2 = empty_table(10_000).update_view(["I = i % 10_000"])

   source = t1.join(t2, "ID = I")

   result = source.tree(id_col = "ID", parent_col = "Parent")
   ```

2. Refresh the page

**Expected results**

3. Hierarchical table loads on page refresh

**Actual results**

3. Hierarchical table fails to open for the 10 second timeout

The table can be reopened from the panels tab

**Versions**

Engine Version: 0.20.0
Web UI Version: 0.26.0
Java Version: 17.0.5
Barrage Version: 0.5.0

mofojed:
- Add the widget type to the panel metadata
- Use that type when re-hydrating
- Fixes #1033

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -61,7 +61,12 @@ import {
   PromiseUtils,
 } from '@deephaven/utils';
 import { ContextAction, ContextMenuRoot } from '@deephaven/components';
-import { Column, FilterCondition, Sort } from '@deephaven/jsapi-shim';
+import {
+  Column,
+  FilterCondition,
+  Sort,
+  VariableTypeUnion,
+} from '@deephaven/jsapi-shim';
 import {
   GridRangeIndex,
   GridState,
@@ -93,6 +98,7 @@ type ModelQueue = ModelQueueFunction[];
 
 interface Metadata {
   table: string;
+  type?: VariableTypeUnion;
   query?: string;
   querySerial?: string;
 }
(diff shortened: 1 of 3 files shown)


packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -93,6 +98,7 @@ type ModelQueue = ModelQueueFunction[];
 
 interface Metadata {
   table: string;
+  type?: VariableTypeUnion;
```
mattrunyon, line 101: This should maybe be required. Although I don't think the spot where it's passed in (`GridPlugin`) is strictly checked, but if it were in the future, we would want to make sure this is required

mattrunyon (approved):

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
mofojed: Yea would need to worry about migration from older data then. On Enterprise this isn't really an issue since we know all the variable types on a query, so we just check what the variable type is for that variable name. Theoretically we could do that here as well, but it would require a round trip (calling the `subscribeToFieldUpdates` API, listening for "changes", scanning for the variable, ugh).

Issue #965, opened 2022-12-19
Remove uses of JS API column index


mattrunyon:
`Column.index` is deprecated in JS API and Colin would like to remove it since web is the only place still using it

Within the TS files, I only see it in 2 places. You can remove the field in `dh.types` and then run the type checker to find the uses.

```
TableUtils.ts
IrisGridUtils.ts
```

Alternatively, can add an eslint plugin like https://www.npmjs.com/package/eslint-plugin-deprecation

bmingles:
@mattrunyon Are there any specific features I should regression test to make sure I don't break anything?

mattrunyon:
Nothing off the top of my head, but I see there's no tests for `TableUtils.getSortIndex`, so I would look into how that's used and write some tests first maybe. The only other place I see it used is `IrisGridUtils.dehydrateSort`, so it looks like the related features would just be applying sorts and reloading the page to ensure the sorts saved and restored correctly

PR #1036, opened 2023-01-19
Don't allow starting or ending links from non-filterable columns

### LINKED ISSUE #1035: Don't allow starting or ending links from non-filterable columns (array columns etc.)

jason-wang24:
Fixes #1035 
- Cursor type is set to 'not-allowed' when hovering over a non-filterable column (and clicking it does not start or complete a link)
- Link type shows 'invalid' when a link in progress hovers over a non-filterable column
- Currently only works for groupedColumns in TreeTables, as other `isFilterable` methods are incomplete (waiting for JS API??)


packages/iris-grid/src/mousehandlers/IrisGridColumnSelectMouseHandler.ts
```diff
@@ -26,7 +26,8 @@ class IrisGridColumnSelectMouseHandler extends GridMouseHandler {
   }
 
   private isValidColumn(tableColumn: Column | null): boolean {
-    if (tableColumn == null) {
+    const { model } = this.irisGrid.props;
+    if (tableColumn == null || !model.isFilterable(tableColumn.index)) {
```
mofojed, line 30: So `column.index` JS API is actually deprecated and we should remove all uses of it: https://github.com/deephaven/web-client-ui/issues/965
Instead we should be storing/tracking by name, and then looking up the column by name (`model.getColumnIndexByName`, so this would be `model.isFilterable(model.getColumnIndexByName(tableColumn.name))`).
Side note, maybe we could allow `ModelIndex | ColumnName` to be passed in to `model.isFilterable` for convenience, and then if a `ColumnName` is passed in then convert that to model index automatically.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -267,7 +268,16 @@ export class Linker extends Component<LinkerProps, LinkerState> {
     this.deleteLinks(linksToDelete);
   }
 
-  handleGridColumnSelect(panel: PanelComponent, column: LinkColumn): void {
+  handleGridColumnSelect(
+    panel: PanelComponent,
+    model: IrisGridModel,
+    column: LinkColumn
+  ): void {
+    const { index } = column;
```
mofojed, line 276: Same comment as below, `column.index` is deprecated and we shouldn't use it.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -651,7 +661,8 @@ export class Linker extends Component<LinkerProps, LinkerState> {
 
   isColumnSelectionValid(
     panel: PanelComponent,
-    tableColumn?: LinkColumn
+    tableColumn?: LinkColumn,
+    model?: IrisGridModel
```
mofojed, line 665: This method is used by `ChartPanel` as well, which is going to have a different type of model... you should be checking within `IrisGridPanel` if the selection is valid before even passing it to this linker method, so you shouldn't need the `model` here. Put this check in `IrisGridPanel.isColumnSelectionValid`

mofojed (changes requested):
I double checked behaviour in Swing, and indeed you cannot start links from non-filterable columns. I'm not sure why that restriction exists, but yes let's continue with that behaviour.

dsmmcken:
> I double checked behaviour in Swing, and indeed you cannot start links from non-filterable columns. I'm not sure why that restriction exists, but yes let's continue with that behaviour.

By non-filterable, I meant columns like array columns, not normally filterable column types that aren't filterable due to being part of a rollup. Is there a better subset that we can use to restrict?

mofojed:
@dsmmcken we could check `TableUtils.getNormalizedType` and if it's `UNKNOWN` then don't allow it to be used? I think that makes sense, since if it isn't one of those normalized types, we wouldn't know how to handle it anyway.

packages/dashboard-core-plugins/src/linker/Linker.tsx
jason-wang24: Right. This does unfortunately make it harder to update the link in progress type to `invalid` since I would return before calling `Linker.isColumnSelectionValid`. What do you think is the best way to approach this? Would we have to add another event, or perhaps an optional parameter to `columnSelectionValidator()` indicating if the column is already invalid?

mofojed: Let's double check with @dsmmcken and see my last comment about using `TableUtils.getNormalizedType` - I think that would make more sense than checking `model.isFilterable`.
So for array type columns (or anything of type `UNKNOWN` returns from `getNormalizedType`), we don't want to allow starting or ending a link there, because we wouldn't know how to apply that filter. So in that case we should change the cursor to `invalid` and not allow it as an option. In which case `model` is left entirely out of this, and we're just looking at the column type.

dsmmcken: Sounds reasonable.

mofojed:
So because of how columns with preview types currently work, those columns we don't want to be linkable just appear to the client as `String` columns. That's because they are converted to such as a preview type: https://github.com/deephaven/deephaven-core/blob/main/engine/table/src/main/java/io/deephaven/engine/table/impl/preview/ColumnPreviewManager.java

I've created a ticket to include that information on the Column object, so that we know when it has been converted and we can check the original type: https://github.com/deephaven/deephaven-core/issues/3358

@jason-wang24 : For this ticket - we can add a hack to check if the `column.description?.startsWith("Preview of type: ")`, and in those cases assume it's a preview column and that it cannot be linked. Add that check into `Linker.isColumnSelectionValid`, with a TODO comment referencing the above deephaven-core ticket to use the new "preview" or "original" type property once it is implemented.

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -656,17 +663,31 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
     }
 
-    if (tableColumn == null) {
-      // Link started, end point is not a valid target
-      this.updateLinkInProgressType(linkInProgress);
+    // TODO: Use preview/original type property when core/#3358 is completed
+    if (
+      tableColumn.description !== undefined &&
+      tableColumn.description?.startsWith('Preview of type: ')
+    ) {
```
mofojed, line 678: I'd extract this logic into a utility function (put in `LinkerUtils`), e.g.:
```
function isLinkableColumn(column: LinkColumn): boolean {
  // TODO: core/#3358 Use preview/original type property instead of checking description
  return !column.description?.startsWith('Preview of type: ');
}
```

mofojed (approved):

PR #1037, opened 2023-01-19
Grid selection tick mark goes away after deselection

### LINKED ISSUE #1024: Grid selection tick mark does not go away after deselection
**Description**

Once a selection is made in a grid, the selection tick mark will always stay active, even if the selection is removed.

**Steps to reproduce**

1. Create a table that scrolls vertically
2. Select a row
3. Click the row again to deselect

**Expected results**

4. No selection tick mark in the scroll bar

**Actual results**

4. Selection tick mark remains in the scroll bar

**Versions**

Engine Version: 0.19.0
Web UI Version: 0.25.0
Java Version: 17.0.5
Barrage Version: 0.5.0

emilyhuxng:
Fixes #1024 

Whenever the selectedRange is empty, set cursorRow and cursorColumn to null.


mofojed (approved):
Looks good!

Side note - previously was maintaining the `cursorRow`/`cursorColumn` so that if you deselected, then used the keyboard it would re-appear from where it had last been. With this change it now just appears in the top left... I think that is fine.

PR #1040, opened 2023-01-20
Overwriting file keeps old tab open with old contents

### LINKED ISSUE #1017: Overwriting a file keeps old tab open with old contents
**Description**

After creating a file with a duplicate name and overwriting the file, the old tab will stay open and still contain the old contents.

**Steps to reproduce**

1. Go to file explorer and create a file.
2. Write some content into the file and save it.
3. Create another file and write some content.
4. Save it as the same name as the first file and overwrite it.

**Expected results**

Either change the old tab to an unsaved state and if you try to save it, force the user to change the name or close the old tab.

**Actual results**

Old tab stays open containing the old content. 

https://user-images.githubusercontent.com/61912640/212344680-972bda8d-05e7-4ed2-afd2-51558a48b81d.mov

**Versions**

Engine Version: 0.19.1
Web UI Version: 0.25.0
Java Version: 17.0.5
Barrage Version: 0.5.0

emilyhuxng:
Fixes #1017 

Overwritten tabs are now closed upon confirming the overwrite modal.

### DIFF
--- a/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
+++ b/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
@@ -249,7 +249,7 @@ export function ConsolePlugin(
         if (panelId === existingPanelId) {
           log.debug(`Update tab title for file ${fileId}`);
           const { itemName } = fileMetadata;
-          renameFilePanel(fileId, FileUtils.getBaseName(itemName));
+          renameFilePanel(fileId, itemName);
         } else {
           log.error(
             `File ${fileId} already associated with a different tab ${existingPanelId}`
@@ -298,7 +298,7 @@ export function ConsolePlugin(
   );
 
   const closeFilePanel = useCallback(
-    fileMetadata => {
+    (fileMetadata, isOverwrite = false) => {
       log.debug('closeFilePanel', fileMetadata);
       const { id: fileId } = fileMetadata;
       let panelId = null;
@@ -313,7 +313,7 @@ export function ConsolePlugin(
         return;
       }
       unregisterFilePanel(fileMetadata, isPreview);
-      LayoutUtils.closeComponent(layout.root, { id: panelId });
+      LayoutUtils.closeComponent(layout.root, { id: panelId }, isOverwrite);
     },
     [layout.root, openFileMap, previewFileMap, unregisterFilePanel]
   );
--- a/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
+++ b/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
@@ -68,14 +68,50 @@ export function ConsolePlugin(
   } = props;
   const notebookIndex = useRef(0);
   // Map from file ID to panel ID
-  const [openFileMap] = useState(new Map<string, string>());
-  const [previewFileMap] = useState(new Map<string, string>());
+  const [openFileMap, setOpenFileMap] = useState(new Map<string, string>());
+  const [previewFileMap, setPreviewFileMap] = useState(
+    new Map<string, string>()
+  );
 
   const dispatch = useDispatch();
 
   const getConsolePanel = useCallback(
     () => panelManager.getLastUsedPanelOfType(ConsolePanel),
     [panelManager]
+  );
+
+  const addOpenFileMapEntry = useCallback(
+    (key: string, value: string) => {
+      setOpenFileMap(map => new Map(map.set(key, value)));
+    },
+    [setOpenFileMap]
+  );
+
+  const deleteOpenFileMapEntry = useCallback(
+    (key: string) => {
+      setOpenFileMap(map => {
+        map.delete(key);
+        return new Map(map);
+      });
+    },
+    [setOpenFileMap]
+  );
+
+  const addPreviewFileMapEntry = useCallback(
+    (key: string, value: string) => {
+      setPreviewFileMap(map => new Map(map.set(key, value)));
+    },
+    [setPreviewFileMap]
+  );
+
+  const deletePreviewFileMapEntry = useCallback(
+    (key: string) => {
+      setPreviewFileMap(map => {
+        map.delete(key);
+        return new Map(map);
+      });
+    },
+    [setPreviewFileMap]
   );
 
   const handleSendCommand = useCallback(
@@ -149,16 +185,16 @@ export function ConsolePlugin(
       let panelId;
       if (openFileMap.has(oldName)) {
         panelId = openFileMap.get(oldName);
-        openFileMap.delete(oldName);
+        deleteOpenFileMapEntry(oldName);
         if (panelId != null) {
-          openFileMap.set(newName, panelId);
+          addOpenFileMapEntry(newName, panelId);
         }
       }
       if (previewFileMap.has(oldName)) {
         panelId = previewFileMap.get(oldName);
-        previewFileMap.delete(oldName);
+        deletePreviewFileMapEntry(oldName);
         if (panelId != null) {
-          previewFileMap.set(newName, panelId);
+          addPreviewFileMapEntry(newName, panelId);
         }
       }
 
@@ -169,7 +205,15 @@ export function ConsolePlugin(
 
       renamePanel(panelId, FileUtils.getBaseName(newName));
     },
-    [openFileMap, previewFileMap, renamePanel]
+    [
+      openFileMap,
+      previewFileMap,
+      renamePanel,
+      addOpenFileMapEntry,
+      addPreviewFileMapEntry,
+      deleteOpenFileMapEntry,
+      deletePreviewFileMapEntry,
+    ]
   );
 
   /**
@@ -197,7 +241,7 @@ export function ConsolePlugin(
       }
       const { id: fileId } = fileMetadata;
       if (isPreview) {
-        previewFileMap.set(fileId, panelId);
+        addPreviewFileMapEntry(fileId, panelId);
         return;
       }
       if (openFileMap.has(fileId)) {
@@ -214,14 +258,21 @@ export function ConsolePlugin(
         return;
       }
 
-      openFileMap.set(fileId, panelId);
+      addOpenFileMapEntry(fileId, panelId);
 
       // De-register preview tab
       if (previewFileMap.has(fileId)) {
-        previewFileMap.delete(fileId);
+        deletePreviewFileMapEntry(fileId);
       }
     },
-    [openFileMap, previewFileMap, renameFilePanel]
+    [
+      openFileMap,
+      previewFileMap,
+      renameFilePanel,
+      addOpenFileMapEntry,
+      addPreviewFileMapEntry,
+      deletePreviewFileMapEntry,
+    ]
   );
 
   const unregisterFilePanel = useCallback(
@@ -238,12 +289,12 @@ export function ConsolePlugin(
       }
       const { id: fileId } = fileMetadata;
       if (isPreview) {
-        previewFileMap.delete(fileId);
+        deletePreviewFileMapEntry(fileId);
         return;
       }
-      openFileMap.delete(fileId);
+      deleteOpenFileMapEntry(fileId);
     },
-    [openFileMap, previewFileMap]
+    [deleteOpenFileMapEntry, deletePreviewFileMapEntry]
   );
 
   const closeFilePanel = useCallback(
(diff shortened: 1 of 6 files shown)


packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -600,6 +607,41 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     this.setState({ showCloseModal: false });
   }
 
+  /**
+   * Close overwritten tabs
+   * @params {string} fileName = The name of the file to be overwritten
```
mofojed, line 612: You shouldn't need the `{string}` part here, since it's already specified in the TypeScript. May need to check your VS Code settings if that was automatically added for you when adding the JS Doc snippet.
```suggestion
   * @params fileName The name of the file to be overwritten
```

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -600,6 +607,41 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+      for (let i = 0; i < tabsToBeRemoved?.length; i += 1) {
+        if (isComponent(tabsToBeRemoved[i].contentItem)) {
+          (tabsToBeRemoved[i].contentItem as glComponent).container.close();
+        } else {
+          tabsToBeRemoved[i].header.parent.removeChild(
+            tabsToBeRemoved[i].contentItem
+          );
+        }
+      }
+    }
```
mofojed, line 634: This is the wrong way to go about it:
- It's only looking in the same tab stack as the current tab
- It ends up closing the tab that you just renamed as well.

I think instead what you should be doing is emitting the `NotebookEvent.RENAME_FILE` event, and then `NotebookPanel` can listen for that, and if it picks up that the new file name is matching itself and it's not the panel we renamed from, then it can close itself.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -600,6 +601,18 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
 
+  /**
+   * Closes overwritten tabs
+   * @param fileName The name of the file to be overwritten
+   */
+  handleOverwrite(fileName: string): void {
+    const { glEventHub } = this.props;
+    const { panelState } = this.state;
+
+    glEventHub.emit(NotebookEvent.RENAME_FILE, fileName, fileName, panelState);
```
mofojed, line 612: Not sure why this works - we're emitting the same `fileName` for both the old name and new name, which isn't how this event should be used. It should be emitting `oldName`, `newName`, shouldn't need to pass the `panelState` as well...
Thinking about this a bit further - I guess it's not always a rename, since you could be saving a file that hasn't been saved yet (i.e. just writing a new file).
So emitting a `NotebookEvent.CLOSE_FILE` with the `newName` would be more appropriate - really we just want to close any files that are already open with the new name anyway.
Try that out. One edge case we need to be concerned about is if the old file that's already open has been modified, since it may prompt to save/discard changes if we try and close it (which we've agreed we should probably just ignore and close the file).

packages/file-explorer/src/NewItemModal.tsx
```diff
@@ -294,9 +295,13 @@ class NewItemModal extends PureComponent<NewItemModalProps, NewItemModalState> {
...
     this.setState({ showOverwriteModal: false });
-    const { onSubmit } = this.props;
+    const { onSubmit, onOverwrite } = this.props;
     const { path, value } = this.state;
     log.debug('handleOverwriteConfirm', path, value);
+
+    if (onOverwrite) {
+      onOverwrite(value);
+    }
     onSubmit(`${path}${value}`);
```
mofojed, line 305: Instead of adding an `onOverwrite` method, add a parameter `isOverwrite = false` (defaults to false) to `onSubmit`.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
emilyhuxng: This works with files that have already been saved but I'm not sure how to go about this with unsaved files. For example, if you make a file called "foo.py" and make another file and try to save it as "foo.py", since "foo.py" is not saved, it doesn't have an id so the handler for CLOSE_FILE doesn't have any record of it.

packages/dashboard-core-plugins/src/ConsolePlugin.tsx
```diff
@@ -78,6 +80,40 @@ export function ConsolePlugin(
     [panelManager]
   );
 
+  const addOpenFileMapEntry = useCallback(
```
mofojed, line 83: Yoooo great work! Good catch, didn't even realize we were mutating the `openFileMap`/`previewFileMap` before.

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -597,8 +601,8 @@ class LayoutUtils {
         // container property exists on a contentItem if contentItem is a component,
         // however, this is not included in the types for a contentitem, thus the casting
         ((oldContentItem as unknown) as {
-          container: { close: () => void };
-        }).container.close();
+          container: { close: (isOverwrite?: boolean) => void };
```
mofojed, line 604: A cleaner way to write this whole block without the casting/comment is to use the `isComponent` type predicate that is defined in `AbstractContentItem`.
Combined with my other comment about defining `CloseOptions`, we shouldn't need any casting and should just be able to do:
```
if (isComponent(oldContentItem)) {
  oldContentItem.container.close(closeOptions);
}
```

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -574,7 +574,11 @@ class LayoutUtils {
    * @param root The GoldenLayout root to search and close the component in
    * @param config The GoldenLayout component config definition to close, eg. { component: 'IrisGridPanel', id: 'table-t' }
    */
-  static closeComponent(root: ContentItem, config: LayoutConfig): void {
+  static closeComponent(
+    root: ContentItem,
+    config: LayoutConfig,
+    isOverwrite = false
```
mofojed, line 580: You'll need to update the JS Docs to include this parameter... but before you do that, I suggest we change this to be an options object instead.
`isOverwrite` doesn't really make sense from the context of `LayoutUtils` - it doesn't (and shouldn't) know that the panel it is closing is a notebook panel type, or whatever. It doesn't care.

In fact, the way we're overriding the `close` behaviour has a bunch of hacks:
- Below where we're casting to unknown and then redefining close instead of just using the `ItemContainer` type
- We have to re-assign the `glContainer.close` property in `NoteboolPanel`, which is pretty hacky
  - That method returns `true` for seemingly no reason at all

We may as well clean this up while we're in here messing with it. My suggestions:
- Add a type `CloseOptions = { force?: boolean }` type to `ItemContainer`
- Add a `beforeClose` method that takes a callback or `null` to register a before close handler, e.g. `beforeClose(callback: ((options?: CloseOptions) => boolean) | null)`
- In the `close` method, redefine to take `CloseOptions` as an optional parameter, then check if there is a `beforeClose` callback registered, and if so then await the result of that, and only close if that resolves to success
- `NotebookPanel` then calls `beforeClose` instead of re-assigning `glContainer.close`

Let me know if you have questions. We could also make the `beforeClose` callback return a `Promise` instead for async, but I don't think we need that right now.

mofojed (changes requested):

packages/dashboard-core-plugins/src/ConsolePlugin.tsx
emilyhuxng: Yes, it was doing something weird where the console was logging it with length 0 but it had entries.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -343,24 +348,21 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
 
-  // override glContainer.close() with a custom closure that checks if needs saving
   initTabCloseOverride() {
     const { glContainer } = this.props;
-    const close = glContainer.close.bind(glContainer);
-    glContainer.close = (isOverwrite = false) => {
-      if (isOverwrite) {
-        close();
+    glContainer.beforeClose((options?: CloseOptions) => {
+      if (options !== undefined && options.force === true) {
```
mofojed, line 354: You can use optional chaining to shorten this, e.g.:
```
if (options?.force) {
  return true;
}
```

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -597,12 +602,8 @@ class LayoutUtils {
...
     const maintainFocusElement = document.activeElement; // attempt to retain focus after dom manipulation, which can break focus
     if (oldContentItem) {
-      if (oldContentItem.isComponent) {
-        // container property exists on a contentItem if contentItem is a component,
-        // however, this is not included in the types for a contentitem, thus the casting
-        ((oldContentItem as unknown) as {
-          container: { close: (isOverwrite?: boolean) => void };
-        }).container.close(isOverwrite);
+      if (isComponent(oldContentItem)) {
+        oldContentItem.container.close(closeOptions ?? undefined);
```
mofojed, line 606: The `?? undefined` is unnecessary - `closeOptions` is already `CloseOptions | undefined`, so this null coalescing just falls back from `undefined` to `undefined`.
Should just be `oldContentItem.container.close(closeOptions);`

packages/golden-layout/src/container/ItemContainer.ts
```diff
@@ -9,6 +9,7 @@ import type { AbstractContentItem, Component } from '../items';
 import type LayoutManager from '../LayoutManager';
 import EventEmitter from '../utils/EventEmitter';
 
+export type CloseOptions = { force?: boolean };
```
mofojed, line 12: Generally I'd keep an empty line between class/type definitions

packages/golden-layout/src/container/ItemContainer.ts
```diff
@@ -146,14 +151,32 @@ export default class ItemContainer<
...
       this.emit('close');
       this.parent.close();
     }
   }
 
+  /**
+   * Sets the beforeCloseHandler to callback
+   * @param callback Callback function to call before closing
+   */
+  beforeClose(callback: (options?: CloseOptions) => boolean) {
```
mofojed, line 176: Need to allow `null` to be passed in here as well, so before handler can be removed if necessary (by passing `beforeClose(null)`), e.g.:
```
beforeClose(callback: ((options?: CloseOptions) => boolean) | null) {
```

packages/golden-layout/src/container/ItemContainer.ts
```diff
@@ -146,14 +151,32 @@ export default class ItemContainer<
    * Closes the container if it is closable. Can be called by
    * both the component within at as well as the contentItem containing
    * it. Emits a close event before the container itself is closed.
+   * @param options Options to pass into the beforeClose handler
    */
-  close() {
-    if (this._config.isClosable) {
+  close(options?: CloseOptions) {
```
mofojed, line 156: This method could be rewritten to be clearer; most notably, we don't need to do anything if `isClosable` is false (no need to check beforeCloseHandler). Could write something like

```
if (!this._config.isClosable) return;

if (this.beforeCloseHandler?.(options) ?? true) {
  this.emit('close');
  this.parent.close();
}
```

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
mofojed: I think the behaviour you have is correct :+1:

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
emilyhuxng: Getting this ESlint error: Unexpected any value in conditional. An explicit comparison or type cast is required. Is this okay?
```
if (options?.force === true) {
  return true;
}
```

mofojed: Yup that's good

mofojed (approved):

PR #1047, opened 2023-01-26
Fix freeze column from context menu

### LINKED ISSUE #1003: Freeze/Unfreeze columns broken
freeze bug

Steps to reproduce:
1. Right click a column and select "Freeze column"
2. Right click the frozen column

Expected:
Option to unfreeze column

Actual:
Only see option to freeze column. The column with the original index has an option to "unfreeze", but that actually just moves that column to the start. See screenshot.

mattrunyon:
Fixes #1003 

Also disables freeze column entry if the column is in a group because that breaks current behavior (groups can't contain frozen columns and must be contiguous)

### DIFF
--- a/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
+++ b/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
@@ -22,7 +22,6 @@ import {
   GridPoint,
   isEditableGridModel,
   ModelIndex,
-  VisibleIndex,
 } from '@deephaven/grid';
 import dh, {
   Column,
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -212,32 +214,36 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
-        if (isColumnFrozen) {
-          this.irisGrid.unFreezeColumnByColumnName(column.name);
-        } else {
-          this.irisGrid.freezeColumnByColumnName(column.name);
-        }
+        this.irisGrid.hideColumnByVisibleIndex(visibleIndex);
       },
-      order: 10,
     });
+    if (isColumnFreezable) {
```
mofojed, line 233: I think instead of not showing this menu item at all, we'd want to disable it in this situation.

@dsmmcken do you have a pref?

mofojed (dismissed):
Approved, but I'd double check if we want to hide the menu item or just disable it.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
dsmmcken: disable.

mofojed (approved):

PR #1048, opened 2023-01-26
feat: Expand all UI support for hierarchical tables

### LINKED ISSUE #1030: Expand all UI support for hierarchical tables
The API will be supporting expand all in the future, and it would be nice to be able to expand all from the UI. Suggested methods:

- [ ] ctrl/cmd+click of the regular expand arrow/cell will expand all nested depths below the selected node
- [ ] (after filter section) _for rows that are expandable_ with two to four new menu options:
  - "Expand "Cell Value"" / "Collapse "Cell Value'"  - Expands or contracts the selected row, only show on expandable rows.
    Show the appropriate expand/collapse text based on if the row is collapsed or not. For "Cell value" insert the value of the cell 
    depth that would be expanded.
  - "Expand All in “Cell value”" - Expand all children of the selected row, only show option on expandable rows.
    Doesn't change to collapse if current row is expanded, as it may still have children that aren't expanded.
  - "Expand Entire Table" - expands all from root (if table is expandable, option should appear on all rows)
  - "Collapse Entire Table" - collapses all from root (if table is expandable, option should appear on all rows)

I think it would be useful for the context menu options on the whole row, not just the cell with the arrow.

Mockup: Note the truncating cell values, please set a max-width on them of 30ch
image

emilyhuxng:
Closes #1030 

Ctrl/cmd+click expands all nested rows of the row that is clicked on. Added two context menu options that appear only on expandable rows and two more that appear on any row of a table with expandable rows.

Shortcut will instead be implemented in another ticket as a tooltip.

### DIFF
--- a/packages/components/src/shortcuts/Shortcut.ts
+++ b/packages/components/src/shortcuts/Shortcut.ts
@@ -92,7 +92,6 @@ export enum KEY {
   F10 = 'F10',
   F11 = 'F11',
   F12 = 'F12',
-  LEFT_CLICK = 'Left Click',
 }
 
 const ALLOWED_SINGLE_KEY_SET: Set<KEY> = new Set([
--- a/packages/components/src/shortcuts/Shortcut.ts
+++ b/packages/components/src/shortcuts/Shortcut.ts
@@ -92,6 +92,7 @@ export enum KEY {
   F10 = 'F10',
   F11 = 'F11',
   F12 = 'F12',
+  LEFT_CLICK = 'Left Click',
 }
 
 const ALLOWED_SINGLE_KEY_SET: Set<KEY> = new Set([
--- a/packages/grid/src/ExpandableGridModel.ts
+++ b/packages/grid/src/ExpandableGridModel.ts
@@ -4,7 +4,10 @@ import { ModelIndex } from './GridMetrics';
 export function isExpandableGridModel(
   model: GridModel
 ): model is ExpandableGridModel {
-  return (model as ExpandableGridModel)?.hasExpandableRows !== undefined;
+  return (
+    (model as ExpandableGridModel)?.hasExpandableRows !== undefined &&
+    (model as ExpandableGridModel)?.hasExpandableRows
+  );
 }
 
 /**
--- a/packages/grid/src/ExpandableGridModel.ts
+++ b/packages/grid/src/ExpandableGridModel.ts
@@ -30,8 +30,13 @@ export interface ExpandableGridModel extends GridModel {
    * Change the expanded status of an expandable row
    * @param row Row to expand
    * @param isExpanded True to expand the row, false to collapse
+   * @param expandAll True to expand nested rows, false otherwise
    */
-  setRowExpanded(row: ModelIndex, isExpanded: boolean): void;
+  setRowExpanded(
+    row: ModelIndex,
+    isExpanded: boolean,
+    expandAll?: boolean
+  ): void;
 
   /**
    * Get the depth of a row (ie. How indented the row should be)
--- a/packages/grid/src/ExpandableGridModel.ts
+++ b/packages/grid/src/ExpandableGridModel.ts
@@ -30,13 +30,23 @@ export interface ExpandableGridModel extends GridModel {
    * Change the expanded status of an expandable row
    * @param row Row to expand
    * @param isExpanded True to expand the row, false to collapse
-   * @param expandAll True to expand nested rows, false otherwise
+   * @param expandDescendants True to expand nested rows, false otherwise
    */
   setRowExpanded(
     row: ModelIndex,
     isExpanded: boolean,
-    expandAll?: boolean
+    expandDescendants?: boolean
   ): void;
+
+  /**
+   * Expand all rows
+   */
+  setExpandAll(): void;
+
+  /**
+   * Collapse all rows
+   */
+  setCollapseAll(): void;
 
   /**
    * Get the depth of a row (ie. How indented the row should be)
--- a/packages/grid/src/ExpandableGridModel.ts
+++ b/packages/grid/src/ExpandableGridModel.ts
@@ -17,6 +17,9 @@ export interface ExpandableGridModel extends GridModel {
   /** Whether the grid has rows that can be expanded */
   hasExpandableRows: boolean;
 
+  /** Whether the grid can expand all */
+  isExpandAllAvailable?: boolean;
+
   /**
    * @param row Row to check
    * @returns True if the row is expandable
@@ -44,12 +47,12 @@ export interface ExpandableGridModel extends GridModel {
   /**
    * Expand all rows
    */
-  setExpandAll(): void;
+  expandAll(): void;
 
   /**
    * Collapse all rows
    */
-  setCollapseAll(): void;
+  collapseAll(): void;
 
   /**
    * Get the depth of a row (ie. How indented the row should be)
--- a/packages/grid/src/ExpandableGridModel.ts
+++ b/packages/grid/src/ExpandableGridModel.ts
@@ -4,10 +4,7 @@ import { ModelIndex } from './GridMetrics';
 export function isExpandableGridModel(
   model: GridModel
 ): model is ExpandableGridModel {
-  return (
-    (model as ExpandableGridModel)?.hasExpandableRows !== undefined &&
-    (model as ExpandableGridModel)?.hasExpandableRows
-  );
+  return (model as ExpandableGridModel)?.hasExpandableRows !== undefined;
 }
 
 /**
@@ -18,7 +15,7 @@ export interface ExpandableGridModel extends GridModel {
   hasExpandableRows: boolean;
 
   /** Whether the grid can expand all */
-  isExpandAllAvailable?: boolean;
+  isExpandAllAvailable: boolean;
 
   /**
    * @param row Row to check
--- a/packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
+++ b/packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
@@ -3,7 +3,7 @@ import { EventHandlerResult } from '../EventHandlerResult';
 import Grid from '../Grid';
 import { getOrThrow } from '../GridMetricCalculator';
 import GridMouseHandler from '../GridMouseHandler';
-import { GridPoint } from '../GridUtils';
+import GridUtils, { GridPoint } from '../GridUtils';
 
 /**
  * Detect when the tree expand/collapse button is clicked
@@ -60,7 +60,7 @@ class GridRowTreeMouseHandler extends GridMouseHandler {
     if (GridRowTreeMouseHandler.isInTreeBox(gridPoint, grid)) {
       const { row } = gridPoint;
       if (row !== null) {
-        grid.toggleRowExpanded(row, event.ctrlKey || event.metaKey);
+        grid.toggleRowExpanded(row, GridUtils.isModifierKeyDown(event));
         return true;
       }
     }
--- a/packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
+++ b/packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
@@ -52,11 +52,19 @@ class GridRowTreeMouseHandler extends GridMouseHandler {
     return GridRowTreeMouseHandler.isInTreeBox(gridPoint, grid);
   }
 
-  onClick(gridPoint: GridPoint, grid: Grid): EventHandlerResult {
+  onClick(
+    gridPoint: GridPoint,
+    grid: Grid,
+    event: React.MouseEvent
+  ): EventHandlerResult {
     if (GridRowTreeMouseHandler.isInTreeBox(gridPoint, grid)) {
       const { row } = gridPoint;
       if (row !== null) {
-        grid.toggleRowExpanded(row);
+        if (event.ctrlKey || event.metaKey) {
+          grid.toggleRowExpanded(row, true);
+        } else {
+          grid.toggleRowExpanded(row, false);
+        }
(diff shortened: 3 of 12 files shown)


mofojed:
@emilyhuxng for the context menu - it should show either one of "Expand Row" or "Collapse Row", depending on what row is selected.

packages/components/src/shortcuts/Shortcut.ts
```diff
@@ -92,6 +92,7 @@ export enum KEY {
   F10 = 'F10',
   F11 = 'F11',
   F12 = 'F12',
+  LEFT_CLICK = 'Left Click',
```
mofojed, line 95: Left click isn't really a key from the keyboard, so I don't think this should be here.
I don't think we show anything showing mouse click shortcuts in the context menu - @dsmmcken should we bother showing a shortcut for this action?

packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
```diff
@@ -52,11 +52,19 @@ class GridRowTreeMouseHandler extends GridMouseHandler {
...
+  ): EventHandlerResult {
     if (GridRowTreeMouseHandler.isInTreeBox(gridPoint, grid)) {
       const { row } = gridPoint;
       if (row !== null) {
-        grid.toggleRowExpanded(row);
+        if (event.ctrlKey || event.metaKey) {
+          grid.toggleRowExpanded(row, true);
+        } else {
+          grid.toggleRowExpanded(row, false);
+        }
```
mofojed, line 67: You can just pass the condition straight into the parameter to shorten this block 


```suggestion
        grid.toggleRowExpanded(row, event.ctrlKey || event.metaKey);
```

packages/iris-grid/src/IrisGridShortcuts.ts
```diff
@@ -37,6 +37,12 @@ const TABLE = {
     shortcut: [MODIFIER.CTRL, KEY.G],
     macShortcut: [MODIFIER.CMD, KEY.G],
   }),
+  EXPAND_ROWS_BELOW: ShortcutRegistry.createAndAdd({
+    id: 'TABLE.EXPAND_ROWS_BELOW',
+    name: 'Expand Rows Below',
+    shortcut: [MODIFIER.CTRL, KEY.LEFT_CLICK],
```
mofojed, line 43: As mentioned above probably don't need to add shortcuts here.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -66,7 +67,9 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
 
   static GROUP_FILTER = ContextActions.groups.high + 50;
 
-  static GROUP_GOTO = ContextActions.groups.high + 51;
+  static GROUP_EXPAND_COLLAPSE = ContextActions.groups.high + 51;
+
+  static GROUP_GOTO = ContextActions.groups.high + 52;
```
mofojed, line 72: Could probably space these priorities out a bit so we don't need to adjust if we insert another. 


```suggestion
  static GROUP_EXPAND_COLLAPSE = ContextActions.groups.high + 55;

  static GROUP_GOTO = ContextActions.groups.high + 60;
```

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -546,6 +552,47 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+        group: IrisGridContextMenuHandler.GROUP_EXPAND_COLLAPSE,
+        order: 20,
+        action: () => {
+          model.setRowExpanded(modelRow, true, true);
+        },
+      });
+
+      let expandAll = false;
+      // If there is a row that isn't expanded, expand all from root
+      for (let i = 0; i < rowCount; i += 1) {
```
mofojed, line 578: Tables could have billions of rows - we shouldn't iterate through all the rows. 
I think the API could give us the current size of the table and the maximum size of the table when all rows are expanded, then we could just compare the two. But iterating every row we should avoid at all costs. There's got to be a better way to check if there are any expanded rows.

mattrunyon: Maybe just expose both entries as options. Then you don't need to try to detect if there's expanded rows or not, just have 2 buttons

It might even be preferable since it seems the current way would mean the only way to collapse all is to first expand all. Maybe the user expanded many groups manually and wants to collapse them all again

mattrunyon: Collapsing the root is not equivalent to collapse all. There may not be a single root node. Also, for tree tables at least the child expanded states are maintained so if you expand the root again, the children are expanded/collapsed as before

dsmmcken: Is the new expand all behaviour on both rollups and trees, or just rollups?

mattrunyon: I don't think there's a distinction between the 2 in the JS API. Both are a tree table. Rollup just adds some aggregation columns, but the UI just sees them both as tree tables

dsmmcken: @emilyhuxng We might need to re-work both the server-side API to support expand multiple roots, and re-consider what options are shown based on the limitations. You can chill on this a bit until we figured it out.

packages/grid/src/mouse-handlers/GridRowTreeMouseHandler.ts
```diff
@@ -52,11 +52,19 @@ class GridRowTreeMouseHandler extends GridMouseHandler {
...
+  onClick(
+    gridPoint: GridPoint,
+    grid: Grid,
+    event: React.MouseEvent
+  ): EventHandlerResult {
     if (GridRowTreeMouseHandler.isInTreeBox(gridPoint, grid)) {
       const { row } = gridPoint;
       if (row !== null) {
-        grid.toggleRowExpanded(row);
+        if (event.ctrlKey || event.metaKey) {
```
mofojed, line 63: Also use `GridUtils.isModifierKeyDown(event)` instead so it is correct depending on the platform.

mofojed (changes requested):

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
emilyhuxng: As Matt mentioned, if a couple rows are expanded, should the option be Expand All or Collapse All? @dsmmcken

mattrunyon: If the API is tricky to detect if all are expanded, could just make it based on the target row. So if the row is expanded show "Collapse All". If collapsed show "Expand all"

dsmmcken: I've updated the parent ticket with requested wording changes.

packages/grid/src/ExpandableGridModel.ts
```diff
@@ -30,8 +33,23 @@ export interface ExpandableGridModel extends GridModel {
...
+
+  /**
+   * Expand all rows
+   */
+  setExpandAll(): void;
+
+  /**
+   * Collapse all rows
+   */
+  setCollapseAll(): void;
```
mofojed, line 52: I just realized - we're going to need to check if `expandAll` functionality is even available, otherwise this change is going to break tree tables on Enterprise.
Specifically, we'll need to add an optional `isExpandAllAvailable?: boolean` flag on `ExpandableGridModel`, then in `IrisGridTreeTableModel`, can implement it like 
```
get isExpandAllAvailable(): boolean {
  return this.table.expandAll !== undefined;
}
```

Then in the menu items, only display Expand/Collapse all for models that support it; in `IrisGridTreeTableModel` don't provide a third param to `table.setExpanded` if `expandAll` isn't available (since that param won't be valid on Enterprise).

I'll need to test this out with Enterprise to make sure it doesn't break existing tree table functionality.

mofojed (changes requested):

packages/grid/src/ExpandableGridModel.ts
mofojed, line 52: Also these methods should just be called `expandAll` and `collapseAll`, no need to call them `setExpandAll` or `setCollapseAll`.

emilyhuxng: When I do ```model.isExpandAllAvailable```, it doesn't run the function but instead accesses the field in ExpandableGridModel which is always undefined. How can I access this function instead?

mofojed: Did you wire it through `IrisGridProxyModel` as well?

mofojed (changes requested):

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -546,6 +549,78 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+      // If there are grouped columns, then it is a rollup
+      // For rollups, the column number will be the depth minus one
+      let cellValue =
+        model.groupedColumns.length > 0
+          ? model.textForCell(model.depthForRow(modelRow) - 1, modelRow)
+          : model.textForCell(0, modelRow);
+
+      if (cellValue === '') {
+        cellValue = 'null';
+      }
```
mofojed, line 563: We're going to need better handling for `null` vs. `empty` string. Consider this snippet:
```
from deephaven.constants import NULL_INT

source = empty_table(100).update_view(["ID = i == 0 ? `` :  `` + (i + 2)", "Parent = i == 0 ? null : i % 2 == 0 ? `` : `` + (i % 9)"])

result_no_orphans = source.tree(id_col = "ID", parent_col = "Parent")
result_w_orphans = source.tree(id_col = "ID", parent_col = "Parent", promote_orphans=True)
```
Right now it's showing "Expand row for 'null'" when the string value is empty:
image

What you want to do is get `valueForCell`, which will return `null` if the cell is actually null, and then the empty string `''` if it's just an empty string (or the value for any other situation).
Then, if the value is `null`, just display `null`, otherwise display the value of `textForCell`.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -546,6 +549,78 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+        command: string,
+        cellVal: string,
+        len: number
+      ) => {
+        let newCellVal = cellVal;
+        if (command.length + cellVal.length + 3 > len) {
+          newCellVal = `${cellVal.substring(0, len - command.length - 6)}...`;
+        }
+        return `${command} "${newCellVal}"`;
+      };
```
mofojed, line 575: This function should be a static function, it doesn't need to be declared within this method as all the methods it uses are passed into it.
`len` could also be an optional parameter, defaulting to `30` (since that seems to be the value that's passed in every time).
Instead of using three dots (`...`), use the ellipses character (`…`): https://www.compart.com/en/unicode/U+2026
That will just take up one char instead.
There is already a `GridRenderer.truncate(str, len)` method already (which uses the ellipses instead of three dots), so you could write this method something like:
```
static function getRowOptionFormatted(command: string, cellValue: string, len = 30) {
  return `${command} "${GridRenderer.truncate(cellValue, len - command.length - 3)}"`;
}
```

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -546,6 +561,82 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+        if (valForRollUp === null || valForRollUp === undefined) {
+          cellValue = 'null';
+        } else {
+          cellValue = model.textForCell(colForRollUp, modelRow);
+        }
+      } else if (valForTreeTable === null || valForTreeTable === undefined) {
+        cellValue = 'null';
+      } else {
+        cellValue = model.textForCell(0, modelRow);
+      }
```
mofojed, line 582: Here we're getting the value for both the rollup and treetable, when we only need the one we're interested in. Also, we have the same logic once we do get that value (if it's null, use `'null'`, otherwise use the `textForCell`). So you could simplify this quite a bit, to something like this:
```
      const expandingColumn = model.groupedColumns.length > 0 ? model.depthForRow(modelRow) - 1 : 0;
      const cellValue = model.valueForCell(expandingColumn, modelRow)
      const cellText =  cellValue == null ? 'null' : model.textForCell(expandingColumn, modelRow);
```

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -546,6 +561,82 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+      actions.push({
+        title: model.isRowExpanded(modelRow)
+          ? IrisGridContextMenuHandler.getRowOptionFormatted(
+              'Collapse',
+              cellValue
+            )
+          : IrisGridContextMenuHandler.getRowOptionFormatted(
+              'Expand',
+              cellValue
+            ),
```
mofojed, line 593: Similar here, can invert this to be simpler:
```
IrisGridContextMenuHAndler.getRowOptionFormatted(model.isRowExpanded(modelRow) ? 'Collapse' : 'Expand');
```

mofojed (approved):

PR #1051, opened 2023-01-26
Add optional worker authentication

vbabich:
- Check query params for the `authRequired` flag
- if flag present, send auth request to the parent window, wait for response with the auth token
- authenticate using the received token

### DIFF
--- a/packages/code-studio/src/main/SessionUtils.ts
+++ b/packages/code-studio/src/main/SessionUtils.ts
@@ -1,10 +1,19 @@
 import { SessionWrapper } from '@deephaven/dashboard-core-plugins';
-import dh, { CoreClient, IdeConnection } from '@deephaven/jsapi-shim';
+import dh, {
+  CoreClient,
+  IdeConnection,
+  LoginOptions,
+} from '@deephaven/jsapi-shim';
 import Log from '@deephaven/log';
 import shortid from 'shortid';
 import NoConsolesError from './NoConsolesError';
 
 const log = Log.module('SessionUtils');
+
+export enum AUTH_TYPE {
+  ANONYMOUS = 'anonymous',
+  PARENT = 'parent',
+}
 
 export function getBaseUrl(): URL {
   return new URL(import.meta.env.VITE_CORE_API_URL ?? '', `${window.location}`);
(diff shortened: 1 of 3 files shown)


packages/code-studio/src/main/SessionUtils.ts
```diff
@@ -15,6 +15,11 @@ export function getWebsocketUrl(): string {
   return `${baseUrl.protocol}//${baseUrl.host}`;
 }
 
+export function isAuthRequired(): boolean {
+  const searchParams = new URLSearchParams(window.location.search);
+  return searchParams.get('authRequired') != null;
+}
```
mofojed, line 21: I think we should be more explicit about the auth mechanism here; we're just saying `authRequired`, and then assuming that means we ask the parent window for an authentication token, which will be kind of confusing when we add other authentication methods.

Param like `authProvider=window.opener` or just `authProvider=opener` would make it more explicit, and then instead of calling it `requestAuthToken`, it'd be `requestAuthTokenFromOpener`. (Or `parent` instead of `opener`, your call).

mofojed (changes requested):

packages/code-studio/src/main/SessionUtils.ts
vbabich: Agree. Changed the query param to `authProvider=parent`, request to `requestAuthTokenFromParent`, and response to
```
{ message: 'loginOptions', payload: { type, token }:LoginOptions }
```

mofojed (approved):

packages/code-studio/src/main/SessionUtils.ts
```diff
@@ -64,4 +83,42 @@ export function createCoreClient(): CoreClient {
...
+    const listener = (
+      event: MessageEvent<{
+        message: string;
+        payload: LoginOptions;
+      }>
+    ) => {
+      const { data } = event;
+      log.info('Received message', data);
+      if (data?.message !== 'loginOptions') {
+        log.info('Ignore received message', data);
```
mofojed, line 100: We should change these to `debug` level, as it may have login credential information which we don't want in the logs.

PR #1056, opened 2023-01-28
refactor: TypeScript Type Improvements

### LINKED ISSUE #1122: Improve redux types
We should probably be using generics more in redux so that DHE types get properly inferred when they differ from DHC.

One example is `getSettings` which the store from DHE has extra settings. Most selectors should probably be something in the form of `getSettings<S extends CommunityStore>(store: S): S['settings']` or something like that. Then as long as the DHE store extends the DHC store, we should get the proper DHE types for DHE specific selectors

Or even something like `getSettings<S>(store: { settings: S }): S => store.settings` since that's really the minimum required for the selector. This might play nicer w/ a few mismatches between DHC and DHE (like `ServerConfigValues` currently)

Zhou-Ziheng:
There are certain types that needs to be improved as part of the Enterprise TypeScript conversion. 
This PR aims to contain all changes that needs to be made.

Fixes #1122
BREAKING CHANGE: Selector Type removed from redux


mofojed:
If you want to make sure this doesn't get merged, you can click "Convert to Draft" under the reviewers section to mark it as a draft.

Zhou-Ziheng:
Fixes #1122

mofojed:
@Zhou-Ziheng can you rebase this off the latest and mark this as "Ready for review" if it is ready for review?

packages/code-studio/src/settings/SettingsUtils.tsx
```diff
@@ -16,7 +16,7 @@ export type FormatOption = {
 export type FormatterItem = {
   columnType: string;
   columnName: string;
-  format: Partial<TableColumnFormat>;
+  format: TableColumnFormat | Record<string, never>;
```
mofojed, line 19: I think would be cleaner to do `Partial<TableColumnFormat>`

packages/components/src/navigation/MenuItem.tsx
```diff
@@ -22,7 +22,7 @@ function isSwitchMenuItemType(item: MenuItemDef): item is SwitchMenuItemDef {
 }
 
 export type MenuItemProps = {
-  item: MenuItemDef;
+  item: MenuItemDef | SwitchMenuItemDef;
```
mofojed, line 25: Shouldn't need this - `SwitchMenuItemDef` extends `MenuItemDef` already.

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -75,6 +75,28 @@ type HydratedIrisGridState = Pick<
...
+    number,
+    {
+      options: AdvancedFilterOptions;
+    }
+  ][];
+  inputFilters?: readonly InputFilter[];
+  sorts?: readonly {
+    column: ModelIndex;
+    isAbs: boolean;
+    direction: SortDirection;
```
mofojed, line 95: Define this after the types below and then use them instead of redefining the types (`DehydratedQuickFilter`, `DehydratedAdvancedFilter`, `DehydratedSort`)

mofojed (changes requested):

mofojed:
@Zhou-Ziheng also update the PR description to put "Fixes #1122" in the description, and change the Breaking notice to [match the format we're expecting](https://github.com/deephaven/web-client-ui/blob/main/CONTRIBUTING.md#creating-a-pull-request) (e.g. "BREAKING CHANGE: Selector type removed from redux"

mofojed (changes requested):

mofojed (approved):

PR #1057, opened 2023-01-29
feat: Use Conventional Commits for release management/PRs

mofojed:
- Update web-client-ui repository to use Conventional Commits instead of the labelling system we have been using for release management.
- Remove the label check GitHub action/script; we no longer require labels to be added
- Remove `Create Release` action - the version bump/release creation process will be kicked off locally instead of by a GitHub action
- Add GitHub action to enforce conventional commit PR titles
- Update CONTRIBUTING guidelines to detail rules for creating a PR
- Update README to detail how to update release
- Pin the version of ubuntu in all GitHub actions


README.md
```diff
@@ -75,19 +75,15 @@ If you encounter an issue specific to a browser, check that your browser is up t
...
-5. Create the release: `gh release create <version> --notes-file /tmp/changelog_<version>.md --title <version>`
+We use [lerna](https://github.com/lerna/lerna) and [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to automatically handle incrementing the version, generate the changelog, and create the release.
+
+1. Generate a [GitHub Personal access token](https://github.com/settings/tokens):
+
+- Under `Repository Access`, select `Only select repositories` and add `deephaven/web-client-ui`.
+- Under `Repository Permissions`, set `Access: Read and write` for `Contents`. This will be necessary to push your version bump and create the release.
+- Copy the token created and replace `<token>` with it in the next step.
+
+2. Bump the version, update the changelog, and create a release: `GH_TOKEN=<token> npm run version-bump`
```
mattrunyon, line 86: We should make this a github action. Idk if we can then enforce permissions so only certain people on the web team can publish. We could work something out using `github.actor` or `github.user` on the action context to limit things if we wanted to.

Benefit to this being an action is I could see an accidental run of this locally without checking out and pulling latest main and I'm not sure what would happen in that case. Duplicates in changelog? Incorrect versioning?

CONTRIBUTING.md
```diff
@@ -37,21 +37,22 @@ Over time, forks will get out of sync with the upstream repository. To stay up t
...
-   - **Breaking Changes:** if your change breaks an existing API in such a way that users of the package affected will need to make some changes to migrate to the newer version, add the `breaking` label to the PR and include a section in the description saying `## Breaking Changes` and any migration instructions required.
+   - PR titles must follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
+     - The `type` provided must be one of [commitizen conventional commit types](https://github.com/commitizen/conventional-commit-types).
+     - The `scope` is not required.
+   - **BREAKING CHANGE:** if your change breaks an existing API in such a way that users of the package affected will need to make some changes to migrate to the newer version, add the `BREAKING CHANGE:` footer to the PR description, detailing the breakage and any migration instructions necessary, e.g.:
+   ```
+   BREAKING CHANGE: The API now takes a new parameter that must be provided.
+   ```
+   - **NOTE:** Do _not_ use the `!` notation for marking a breaking change - you must use the `BREAKING CHANGE:` footer and include details of the breakage/migration.
```
mattrunyon, line 53: Will this break if we allow both? Or you just want to encourage the footer so breaking always has a corresponding message?

mattrunyon (changes requested):

README.md
mofojed: You won't able to push it if you don't have the latest main checked out.

GitHub action doesn't give you the opportunity to confirm the version numbers are correct before it goes through. This is still an improvement on our process, and I wanted to make sure it worked a couple of times manually to iron out any kinks before we let the robots take over.

CONTRIBUTING.md
mofojed: I want to encourage the footer so there is always a corresponding breaking message. Also, `lerna` doesn't seem to process the exclamation mark format as breaking.

mattrunyon (changes requested):

.github/workflows/conventional-pr-check.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
```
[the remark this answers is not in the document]
mofojed: True, updated the name. Yea I can disable the status check briefly just to merge this change if we need to.

CONTRIBUTING.md
```diff
@@ -37,21 +37,22 @@ Over time, forks will get out of sync with the upstream repository. To stay up t
...
    - If you know people who should be reviewers, add them as a reviewer
    - Add yourself as the Assignee
-   - Select one label depending on the change type:
-      - `bug`: Fixing a bug in an existing feature
-      - `enhancement`: Adding a new feature
-      - `internal`: Changes to the project config or build
-      - `documentation`: Documentation change only
-   - **Breaking Changes:** if your change breaks an existing API in such a way that users of the package affected will need to make some changes to migrate to the newer version, add the `breaking` label to the PR and include a section in the description saying `## Breaking Changes` and any migration instructions required.
+   - PR titles must follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
+     - The `type` provided must be one of [commitizen conventional commit types](https://github.com/commitizen/conventional-commit-types).
```
dsmmcken, line 47: Can you list them here for easier reference?

.github/workflows/conventional-pr-check.yml
```diff
@@ -8,7 +8,7 @@ on:
       - synchronize
 
 jobs:
-  main:
+  doc-labels:
```
mattrunyon, line 11: Looks like github didn't run it which makes sense since it's a new workflow. I think it only runs workflows in main (but uses the workflow file from the PR)

As long as we can still merge this PR without the check passing/failing then we should be good.

We should cut a release (using the old style) just before or after merging this

mattrunyon (approved):

mattrunyon (changes requested):
Hit approve instead of request when I mentioned Don's comment

.github/workflows/conventional-pr-check.yml
mofojed: Yes, what I'll do is cut a release, update the required checks, then merge this. All open PRs will need to match.

dsmmcken (approved):
approved for readme change

mattrunyon (approved):

PR #1059, opened 2023-01-30
fix: Pasting with smart quotes causes double paste in monaco

### LINKED ISSUE #1050: Pasted items with smart-quotes are being duplicated
Pasted text containing a smart quote appears twice.

**Steps to reproduce**
Paste the following into the console input:  ` example” `  <- Right Double Quotation Mark
 
**Expected results**
Text appears once, as `example"`  <- Regular Quotation Mark (replaced in onDidPaste)

**Actual results**
`example"example"`

Text appears twice for some reason. Maybe something to do with the auto-completer?

**Versions**
Engine Version: 0.21.0
Web UI Version: 0.26.1
Java Version: 11.0.16
Barrage Version: 0.5.0

mattrunyon:
Fixes #1050 

Don't think I can add a test easily since paste events were giving trouble in e2e tests and monaco doesn't play nicely w/ unit tests

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -583,13 +583,12 @@ export class AppMainContainer extends Component<
 
   // eslint-disable-next-line class-methods-use-this
   handlePaste(event: ClipboardEvent<HTMLDivElement>): void {
-    let element = event.currentTarget.parentElement;
-
-    while (element != null) {
-      if (element.classList.contains('monaco-editor')) {
-        return;
-      }
-      element = element.parentElement;
+    if (
+      event.target instanceof HTMLElement &&
+      event.target.closest('.monaco-editor')
+    ) {
+      // Skip if this is inside a monaco editor
+      return;
     }
 
     const { clipboardData } = event;


mofojed:
For context, the original MR that added this functionality: https://gitlab.deephaven.io/illumon/iris/-/merge_requests/3052/diffs
Two questions I have before clicking merge:
- What changed causing this to paste twice? That wasn't happening before...
- Do we even need that `handlePaste` method? Where are we pasting other than a monaco editor that might after smart quotes?

mattrunyon:
Broke in #716 which changed `event.target` to `event.relatedTarget`. Shame React's types aren't smart enough to know on certain types (like clipbardEvent) that there isn't a related target. Or maybe there is in non-standard implementations.

You could be pasting into an input filter or any other inputs we have that aren't in monaco. I'm not really sure any of those would expect smart quotes.

In a perfect world we'd have a separate `Input` or `InputBase` component that handles the paste and use that everywhere. Then don't need the global listener and the monaco ignore

mofojed (approved):

dsmmcken:
Nice work debugging this one.

PR #1060, opened 2023-01-30
style: adds a custom linker-not-allowed cursor

dsmmcken:
Browser not-allowed didn't pair well with the linker cursor. Adds a cursor that combines a not allowed cursor with a cross hair of the same style of our linker cursor. Icon is now also BW only, as the danger color was a stronger warning than needed.


mattrunyon:
The `style:` prefix is for code style like whitespace. This would be a `feat:`

https://github.com/commitizen/conventional-commit-types/blob/master/index.json

dsmmcken:
> The `style:` prefix is for code style like whitespace. This would be a `feat:`
> 
> https://github.com/commitizen/conventional-commit-types/blob/master/index.json

My bad, didn't read the descriptions. I guess we don't have the equivalent of an "enhancement" any more, it's all just features? @mofojed

dsmmcken:
My mental model has been wrong this whole time. I thought "enhancement" as like enhancing an existing feature.

mofojed (approved):

mofojed:
Everything's a feature if you look at it the right way!

PR #1071, opened 2023-02-06
build: Add Dockerfile for easily updating snapshots

### LINKED ISSUE #960: Create a Dockerfile to update e2e snapshots
Right now the instructions to update Playwright snapshots for use in CI are cumbersome. Create a Dockerfile that encapsulates all the steps and makes it easy to do.

mofojed:
- Updated README to include new instructions on how to update snapshots
- Add snapshots for mac that weren't previously included
- Tested on M1 MacBook Pro
- Fixes #960


README.md
```diff
@@ -114,45 +114,4 @@ npx source-map-explorer 'packages/code-studio/build/static/js/*.js'
...
-npm ci
-VITE_CORE_API_URL=http://host.docker.internal:10000/jsapi npm run build
-```
-
-Next, run the tests and update the snapshots:
-
-```
-npm run e2e:update-snapshots
-```
+Snapshots are used by end-to-end tests to visually verify the output. Sometimes changes are made requiring snapshots to be updated. Since snapshots are platform dependent, you may need to use a docker image to [update snapshots for CI](https://playwright.dev/docs/test-snapshots). Run `npm run e2e:update-ci-snapshots` to mount the current directory into a docker image and re-run the tests from there. **Note:** You must have [Docker installed](https://docs.docker.com/get-docker/), and `deephaven-core` must already be running on port 10000 on your local machine.
```
mattrunyon, line 117: Update this to mention that you should run local snapshots while creating your tests and only run the docker version once before committing since it will be slow w/ the npm install.

Also after you run the docker version, you need to run `npm install` or `npm ci` locally again to fix your node_modules. See if you can find a way to not require that. Maybe copy the src files into docker, run `npm ci` within docker, and then mount the local `/test` directory to run the tests in Docker (so snapshots go to the right place)

mattrunyon (changes requested):

README.md
```diff
@@ -114,4 +114,4 @@ npx source-map-explorer 'packages/code-studio/build/static/js/*.js'
...
+Snapshots are used by end-to-end tests to visually verify the output. Sometimes changes are made requiring snapshots to be updated. Run snapshots locally to update first, by running `npm run e2e:update-snapshots`. Once you are satisfied with the snapshots and everything is passing locally, you may need to use a docker image to [update snapshots for CI](https://playwright.dev/docs/test-snapshots) (since snapshots are platform dependent). Run `npm run e2e:update-ci-snapshots` to mount the current directory into a docker image and re-run the tests from there. **Note:** You must have [Docker installed](https://docs.docker.com/get-docker/), and `deephaven-core` must already be running on port 10000 on your local machine.
```
mattrunyon, line 117: Maybe break this into 2 paragraphs. 2nd starting w/ the "Once you are satisfied".

I would change the wording to just say "you need to run npm run e2e:update-ci-snapshots" and then the short explanation why. But instead of "may need" just "need" since it's a relatively high chance (at least since most of the devs have Macs) and probably just good practice.

package.json
```diff
@@ -50,7 +50,7 @@
     "e2e:headed": "playwright test --project=chromium --headed --debug",
     "e2e:update-snapshots": "playwright test --update-snapshots",
     "version-bump": "lerna version --conventional-commits --create-release github",
-    "e2e:update-ci-snapshots": "docker build -t web-client-ui-ci-snapshots ./tests/update-ci-snapshots/ && docker run --network host -v $(pwd):/work/ web-client-ui-ci-snapshots /work/tests/update-ci-snapshots/update-snapshots.sh"
+    "e2e:update-ci-snapshots": "docker build -t web-client-ui-ci-snapshots -f ./tests/update-ci-snapshots/Dockerfile . && docker run --network host -v $(pwd)/tests:/tests/ web-client-ui-ci-snapshots /work/tests/update-ci-snapshots/update-snapshots.sh"
```
mattrunyon, line 53:
```suggestion
    "e2e:update-ci-snapshots": "docker build --tag web-client-ui-ci-snapshots --file ./tests/update-ci-snapshots/Dockerfile . && docker run --network host --volume $(pwd)/tests:/tests/ web-client-ui-ci-snapshots /work/tests/update-ci-snapshots/update-snapshots.sh"
```

tests/update-ci-snapshots/Dockerfile
```diff
@@ -4,6 +4,11 @@
 FROM mcr.microsoft.com/playwright:v1.30.0-focal
 WORKDIR /work/
 
+# Copy the web-client-ui src folder to the docker image
+# This requires the Dockerfile to be built in the context of the root of the web-client-ui repository
+# https://stackoverflow.com/a/34300129
+ADD . /work
```
mattrunyon, line 10: I think we should be using `COPY` here https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

Also, we may want to copy just a few select files initially after the nvm install (like .nvmrc, package.json files, package-lock.json), then `RUN npm ci` and then copy the remaining files (ideally ignoring dist/build related files). This should improve docker caching these steps

I can see why some repos build to a `/build` folder in the root instead of within the individual packages. Makes copy/cache easier in docker since the `packages` directory should only contain src files at any state locally

Although cacheability shouldn't matter as much if the plan is to tell people to run locally and only run this before they push since it will only take a few extra minutes.

In theory though, I think w/ those steps that this should be similar in speed to just running locally. Would need to rebuild if the src files changed, but that's it.

mattrunyon (changes requested):

tests/update-ci-snapshots/Dockerfile
mofojed: Yeah I don't know how to copy while ignoring the dist/build files. No clean way to do that (that I can tell) unless we change the build output folder to not be in package directories at all.

mattrunyon: [.dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file) will do just that actually. A bit annoying that it doesn't just read gitignore, but we could just throw it in a directory w/ the dockerfile inside test or something

mattrunyon:
I'm fine with this in its current state as working if you would rather have a followup ticket for encapsulating all of the e2e process in docker. This at least lets me update the e2e snapshots without messing with my local files so I can still just run tests/build/dev without needing to `npm install` after running the CI e2e snapshots

The followup would include adding better caching to the dockerfile as explained in the other comments so that users don't need to run anything locally. I have no idea how this might affect trying to run in headed mode though for debugging tests, so that would be an important thing to check

mattrunyon (approved):
See my other comment about opening a follow up ticket to at least investigate running e2e completely within docker w/ nicely cached layers

mattrunyon (changes requested):
Actually can you add a name to the docker container. It creates a randomly named containers right now for each run and isn't removed. Or figure out if there's a flag to remove the container after the run

mofojed:
I added the `--rm` flag, that just removes the container, the image stays cached.
Also updated the `Dockerfile` to copy files in a different order so that it caches `nvm install` and `npm ci` steps if you haven't changed the `.nvmrc` file or the `package.json` files. It's annoying that we have to specify each individual `package.json` - copying with a glob just puts all the files in the root directory (so only last package.json stays), I don't think there's another way to copy the package.json with these patterns and retain the directory structure.

.dockerignore
```diff
@@ -0,0 +1,47 @@
...
+src/**/*.css
+
+tsconfig.tsbuildinfo
+packages/*/package-lock.json
+/test-results/
+/playwright-report/
+/playwright/.cache/
+
+# Ignore Dockerfile as well, since we don't need to copy that into the snapshot container
+Dockerfile
```
mattrunyon, line 47: Should also add `/tests` to the docker ignore. Those files are currently copied over, but if you only change a test file, the build should be fully cached

package.json
```diff
@@ -50,7 +50,7 @@
     "e2e:headed": "playwright test --project=chromium --headed --debug",
     "e2e:update-snapshots": "playwright test --update-snapshots",
     "version-bump": "lerna version --conventional-commits --create-release github",
-    "e2e:update-ci-snapshots": "docker build --tag web-client-ui-ci-snapshots --file ./tests/update-ci-snapshots/Dockerfile . && docker run --network host --volume $(pwd)/tests:/tests/ web-client-ui-ci-snapshots /work/tests/update-ci-snapshots/update-snapshots.sh"
+    "e2e:update-ci-snapshots": "docker build --tag web-client-ui-ci-snapshots --file ./tests/update-ci-snapshots/Dockerfile . && docker run --rm --network host --volume $(pwd)/tests:/tests/ web-client-ui-ci-snapshots npm run e2e:update-snapshots -- --config=playwright-ci.config.ts"
```
mattrunyon, line 53: Either the workdir in the dockerfile should be `/` or the mount should be `$(pwd}/tests:/work/tests/`

This goes w/ the other comment about adding `/tests` to dockerignore

mattrunyon (changes requested):

skip.js
```diff
@@ -0,0 +1,6 @@
+// Script will skip the postinstall script if `SKIP_POSTINSTALL` env variable is set
+if (process.env.SKIP_POSTINSTALL) {
+  process.exit(0);
+} else {
+  process.exit(1);
+}
```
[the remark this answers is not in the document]
mofojed: https://github.com/deephaven/web-client-ui/issues/1083

.dockerignore
mofojed: Good call

mattrunyon (approved):

PR #1072, opened 2023-02-07
feat: Goto Value Improvements

### LINKED ISSUE #1027: Go to row improvements
- [x] If go to is initiated via shortcut or the sidebar menu, and there is a table row cursor (a selected row) populate the initial values for the row index with that row number
- [x] If go to has a row index, then populate the go to value with the value of the selected column matching that row index, if the column selected dropdown is changed, pre-fill the new matching value for that row index.
- [x] Go to value postive/negative.inf works for floats/doubles but the text field is blank
- [x] Accept "null" as input for all columns (need to check on api support)
- [x] Open go to, scroll table to bottom, close box, grid now has excess space at bottom. Should we render the grid with the correct lastTop again?

- for the issue with +-inf, currently we are enforcing the user to only enter numbers for number columns. If we allow user to enter `infinity`, we have to drop the number only requirement for the input unless I build a custom number only input field. FYI, for quick filters on numerical columns, we allow the user to enter letters. Another option would be to add a +-inf button.
- for the issue with excess space, I think resetting lastTop should really be taken care of in a IrisGridBottomBar. Thoughts?

_Originally posted by @Zhou-Ziheng in https://github.com/deephaven/web-client-ui/issues/974#issuecomment-1384349979_

Zhou-Ziheng:
Fixes #1027 
- `null` values are not supported by the API, so it is not implemented

### DIFF
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -15,7 +15,6 @@ import {
 } from '@deephaven/filters';
 import { Button, DateTimeInput } from '@deephaven/components';
 import { TableUtils } from '@deephaven/jsapi-utils';
-import { assertNotNull } from '@deephaven/utils';
 import classNames from 'classnames';
 import './GotoRow.scss';
 import IrisGridModel from './IrisGridModel';
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -102,8 +102,7 @@ function GotoRow({
 
   const index = model.getColumnIndexByName(gotoValueSelectedColumnName);
 
-  assertNotNull(index);
-  const selectedColumn = columns[index];
+  const selectedColumn = columns[index ?? 0];
 
   const columnType = selectedColumn?.type;
 
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -91,6 +91,12 @@ function GotoRow({
       e.stopPropagation();
       e.preventDefault();
       onGotoValueSubmit();
+    } else if (
+      e.key === 'Backspace' &&
+      (gotoValue === `${Number.POSITIVE_INFINITY}` ||
+        gotoValue === `${Number.NEGATIVE_INFINITY}`)
+    ) {
+      onGotoValueInputChanged('');
     }
   };
 
@@ -126,13 +132,20 @@ function GotoRow({
           <div className="goto-row-input">
             <input
               ref={gotoValueInputRef}
-              type="number"
               className={classNames('form-control', {
                 'is-invalid': gotoValueError !== '',
               })}
               onKeyDown={handleGotoValueKeyDown}
               placeholder="value"
-              onChange={e => onGotoValueInputChanged(e.target.value)}
+              onChange={e => {
+                if (/^-?[0-9]*\.?[0-9]*e?[0-9]*$/.test(e.target.value)) {
+                  onGotoValueInputChanged(e.target.value);
+                } else if (e.target.value.toLowerCase() === '-i') {
+                  onGotoValueInputChanged(`${Number.NEGATIVE_INFINITY}`);
+                } else if (e.target.value.toLowerCase() === 'i') {
+                  onGotoValueInputChanged(`${Number.POSITIVE_INFINITY}`);
+                }
+              }}
               value={gotoValue}
             />
           </div>
@@ -202,6 +215,7 @@ function GotoRow({
               }}
               value={gotoValue}
             >
+              <option aria-label="null value" key="null" value="" />
               <option key="true" value="true">
                 true
               </option>
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -85,7 +85,8 @@ function GotoRow({
   const res = 'Row number';
 
   const { rowCount } = model;
-  const handleGotoValueKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+
+  const handleGotoValueNumberKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter') {
       e.stopPropagation();
       e.preventDefault();
@@ -96,6 +97,14 @@ function GotoRow({
         gotoValue === `${Number.NEGATIVE_INFINITY}`)
     ) {
       onGotoValueInputChanged('');
+    }
+  };
+
+  const handleGotoValueKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
+      e.stopPropagation();
+      e.preventDefault();
+      onGotoValueSubmit();
     }
   };
 
@@ -133,10 +142,11 @@ function GotoRow({
               className={classNames('form-control', {
                 'is-invalid': gotoValueError !== '',
               })}
-              onKeyDown={handleGotoValueKeyDown}
+              onKeyDown={handleGotoValueNumberKeyDown}
               placeholder="value"
               onChange={e => {
                 const value = e.target.value.toLowerCase();
+                // regex tests for
                 if (/^-?[0-9]*\.?[0-9]*$/.test(e.target.value)) {
                   onGotoValueInputChanged(e.target.value);
                 } else if (value === '-i' || value === '-infinity') {
@@ -160,6 +170,7 @@ function GotoRow({
                   'is-invalid': gotoValueError !== '',
                 }
               )}
+              defaultValue={gotoValue}
               onChange={onGotoValueInputChanged}
             />
           </div>
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -92,7 +92,7 @@ function GotoRow({
       e.preventDefault();
       onGotoValueSubmit();
     } else if (
-      e.key === 'Backspace' &&
+      (e.key === 'Backspace' || e.key === 'Delete') &&
       (gotoValue === `${Number.POSITIVE_INFINITY}` ||
         gotoValue === `${Number.NEGATIVE_INFINITY}`)
     ) {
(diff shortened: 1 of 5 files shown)


packages/iris-grid/src/GotoRow.tsx
```diff
@@ -91,6 +91,12 @@ function GotoRow({
       e.stopPropagation();
       e.preventDefault();
       onGotoValueSubmit();
+    } else if (
+      e.key === 'Backspace' &&
```
mofojed, line 95: I'd probably check `Delete` here as well.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3635,6 +3670,23 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   }
 
   handleGotoValueSelectedColumnNameChanged(columnName: ColumnName): void {
+    const { model } = this.props;
+    let cursorRow;
+    if (this.grid) {
+      ({ cursorRow } = this.grid.state);
+    }
```
mofojed, line 3677: You could just write this like:
```suggestion
    const cursorRow = this.grid?.state.cursorRow;
```

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1671,6 +1671,20 @@ class IrisGridUtils {
 
     return { groups: [...groupMap.values()], maxDepth, groupMap, parentMap };
   }
+
+  static convertValueToText(value: unknown, columnName?: string): unknown {
```
mofojed, line 1675: Add unit tests and JSDocs for new utils function.
Also param name should be `columnType`, not `columnName`. And it should return a `string`.

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1671,6 +1671,20 @@ class IrisGridUtils {
...
+    if (
+      columnName != null &&
+      TableUtils.isCharType(columnName) &&
+      value != null
+    ) {
+      return String.fromCharCode(parseInt(value as string, 10));
```
mofojed, line 1681: This isn't right - you only want to do this conversion if a number is provided, and in that case doing `value as string` doesn't make sense. Should be something like:
```suggestion
    if (
      columnName != null &&
      TableUtils.isCharType(columnName) &&
      typeof value === 'number'
    ) {
      return String.fromCharCode(value);
```

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1671,6 +1671,20 @@ class IrisGridUtils {
...
+    }
+    if (value == null) {
+      return '';
+    }
+    return value;
```
mofojed, line 1686: Should return the stringified version here.
```suggestion
    return `${value}`;
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridUtils.ts
Zhou-Ziheng: value is being passed in as a string in this case, so we'll have to handle that explicitly then by checking converting value to number does not return NaN

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -91,13 +90,18 @@ function GotoRow({
...
+    ) {
+      onGotoValueInputChanged('');
     }
   };
 
   const index = model.getColumnIndexByName(gotoValueSelectedColumnName);
 
-  assertNotNull(index);
-  const selectedColumn = columns[index];
+  const selectedColumn = columns[index ?? 0];
```
mattrunyon, line 104: This seems incorrect. Why default to a column if the name isn't found? How did the user get to this state where the name isn't a valid column name?

Unless it's for handling a case where there is no selected column and the user brings up the dialog. In which case, add a comment that the dialog defaults to 1st column if there is no user selection

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2453,6 +2453,29 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       this.focusRowInGrid(row);
       return;
     }
+
+    let cursorRow;
+    let cursorColumn;
+    if (this.grid) {
+      ({ cursorRow, cursorColumn } = this.grid.state);
+    }
```
mattrunyon, line 2461: This should probably be a check of its own or an assertion maybe. If there's no grid, how are we here? Pretty sure it would mean rendering hasn't occurred yet

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2453,6 +2453,29 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    if (cursorRow != null && cursorColumn != null) {
+      // if a row is selected
```
mattrunyon, line 2463: Flip the logic here to do an early return if nothing is selected. That way the majority of the logic is at the lowest possible indent level

```typescript
if (cursorRow == null || cursorColumn == null) {
  this.setState({
    // All the empty strings from the end of this function
  });
}

// Do the value fetch and set state
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3219,18 +3242,19 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
       return;
     }
 
-    const selectedColumn = model.columns[columnIndex];
-
     let searchFromRow;
 
     if (this.grid) {
-      ({ selectionEndRow: searchFromRow } = this.grid.state);
+      ({ cursorRow: searchFromRow } = this.grid.state);
```
mattrunyon, line 3257: Again if there's no grid, we shouldn't really be doing anything at all since it's not rendered yet

I also saw Bender commented on this and said you can change it to use `this.grid?.state.cursorRow` instead. Why not use that syntax here?

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3283,7 +3308,13 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
             !TableUtils.isBigDecimalType(selectedColumn.type) &&
             !TableUtils.isBigIntegerType(selectedColumn.type)
           ) {
-            const inputValue = parseInt(inputString, 10);
+            let inputValue = parseInt(inputString, 10);
+            if (inputString === '-Infinity') {
```
mattrunyon, line 3312: In another spot you allow for `-i` and `-infinity` as valid input strings. Why not here? Or is whatever the user types in those cases replaced w/ `-Infinity`

mattrunyon (changes requested):

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1671,6 +1671,26 @@ class IrisGridUtils {
...
+      value != null &&
+      !Number.isNaN(parseInt(value as string, 10))
+    ) {
+      return String.fromCharCode(parseInt(value as string, 10));
+    }
+    if (value == null) {
+      return '';
+    }
+    return `${value}`;
+  }
```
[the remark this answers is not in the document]
Zhou-Ziheng: The problem with textForCell is that it truncates.

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -1671,6 +1671,26 @@ class IrisGridUtils {
...
+   * @returns The value of the cell converted to text
+   */
+  static convertValueToText(value: unknown, columnType?: string): string {
+    if (
+      columnType != null &&
+      TableUtils.isCharType(columnType) &&
+      value != null &&
+      !Number.isNaN(parseInt(value as string, 10))
+    ) {
+      return String.fromCharCode(parseInt(value as string, 10));
```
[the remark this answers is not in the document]
Zhou-Ziheng: Interesting, last time I tested this, `typeof value` on char column returns a string, so I was forcing a conversion from string to int, then doing a isNaN check. `typeof value` seems to be returning a number now, so I'll change it to `typeof value === 'number'`

packages/iris-grid/src/GotoRow.tsx
Zhou-Ziheng: See #1100 , go to always defaults the the first column, so when the first column disappears somehow, the `gotoValueSelectedColumn` is not updated automatically, so the assertion throws an error. Solution is to again default the selectedColumn to index 0.

packages/iris-grid/src/IrisGrid.tsx
Zhou-Ziheng: At this point in the code, we are only concerned with whether cursorRow/cursorColumn exists. So if the grid is not rendered, then shouldn't we just default everything to the empty string?

Zhou-Ziheng: Are we suggesting we never render the gotoRow UI at all if grid is not rendered?

packages/iris-grid/src/IrisGrid.tsx
Zhou-Ziheng: Assuming we already know the column and data, and we are just waiting on grid rendering, then I think it's fine if we call seekrow in advance, and render grid with the correct row selected. Thoughts?

packages/iris-grid/src/IrisGrid.tsx
Zhou-Ziheng: in the other spot, -i is automatically changed to -infinity (kind of like auto-completing). So I'm assuming the only possible sequence of letters in this input is either `Infinity` or `-Infinity` since I'm not allowing any other strings

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: `this.grid` is a ref to the Grid component rendered. The only reason it's ever not set is on the initial render. So realistically this is always set, but TS doesn't know that

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: `grid` should basically always exist except on initial render of the component

packages/iris-grid/src/IrisGrid.tsx
Zhou-Ziheng: right, assertNotNull would be a better choice then

mofojed (approved):

PR #1075, opened 2023-02-08
feat: jsapi core module

mofojed:
- Just showing that we can import the JS API as a module dynamically
- Needs https://github.com/deephaven/deephaven-core/pull/2733


mofojed:
Opening up https://github.com/deephaven/web-client-ui/pull/1084 instead

PR #1076, opened 2023-02-08
test: Increase code coverage in package file explorer

### LINKED ISSUE #1049: Increase code coverage in package file-explorer
Increase code coverage in package file-explorer.

emilyhuxng:
Closes #1049 

Add tests for files in file-explorer. Note: Drag and drop is not tested.


packages/file-explorer/src/FileExplorer.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+  let user: ReturnType<typeof userEvent.setup>;
+  let dirs: DirectoryStorageItem[] = [];
+  let files: FileStorageItem[] = [];
+  let items: FileStorageItem[] = [];
+  const onRename = jest.fn();
+  const onDelete = jest.fn();
+
+  const clickContextMenuButton = async (
+    itemName: RegExp,
+    buttonName: RegExp
```
mofojed, line 76: You could have these both accept `RegExp | string` so you could just pass a string in here instead of always having to pass in regexes.

packages/file-explorer/src/FileExplorer.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  it('should delete the item upon confirmation', async () => {
+    await clickContextMenuButton(/testfile0/i, /delete/i);
+
+    const modalDeleteButton = await screen.findByRole('button', {
+      name: /delete/i,
+    });
+    await user.click(modalDeleteButton);
+    await waitForElementToBeRemoved(modalDeleteButton);
+    expect(onDelete).toBeCalledTimes(1);
```
mofojed, line 153: Should check that it's called with the correct value for delete.

packages/file-explorer/src/FileExplorer.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+    // await user.pointer({ keys: '[MouseLeft>]', target: file });
+    fireEvent.mouseDown(file);
+    fireEvent.mouseMove(directory);
+    fireEvent.mouseUp(directory);
+    // await user.pointer({ keys: '[/MouseLeft]' });
+
+    // expect(file).not.toBeInTheDocument();
+  });
+});
+ */
```
mofojed, line 193: Remove commented out code. We can make a ticket for following up on creating drag 'n' drop tests in FileExplorer, and include this snippet in the ticket. And just have a comment here `TODO: #XXX` with the ticket number.

Ultimately the drag and drop tests may end up being done in Playwright instead of jest though.

packages/file-explorer/src/FileList.test.tsx
```diff
@@ -35,29 +44,107 @@ describe('getMoveOperation', () => {
...
+        id: '/testdir0',
+        itemName: 'testdir0',
+        type: 'directory',
+        isExpanded: false,
+      }),
+      expect.anything()
+    );
+  });
+
+  // Need to implement drag and drop tests
```
mofojed, line 149: Reference TODO ticket for drag and drop

packages/file-explorer/src/FileListContainer.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+   * @param functionName Name of the function as a string (should match the prop name e.g. onDelete)
+   * @param title Display name of the button
+   * @param expectedValue The parameters the function is expected to receive. If there are multiple, pass them in an array and set multipleExpectedValues to true
+   * @param keyboardText Text you may want to enter after clicking the button
+   * @param multipleExpectedValues True if the function expects multiple parameters
+   */
+  const testContextMenuAction = async (
+    functionName: string,
+    title: RegExp,
+    expectedValue: any,
```
mofojed, line 64: Instead of `any`, use `unknown`.

packages/file-explorer/src/FileListContainer.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+    expect(button).toBeInTheDocument();
+
+    await TestUtils.click(user, button);
+
+    if (keyboardText !== undefined) {
+      await user.keyboard(keyboardText);
+    }
+
+    if (expectedValue !== undefined && multipleExpectedValues) {
+      expect(actionFunction).toHaveBeenCalledWith(...expectedValue);
```
mofojed, line 87: Check that `expectedValue` is an array instead of just checking against `undefined`. Then you don't need the `multipleExpectedValues` parameter:
```
if (Array.isArray(expectedValue)) { ... }
```

packages/file-explorer/src/FileListContainer.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const testContextMenuAction = async (
+    functionName: string,
+    title: RegExp,
+    expectedValue: any,
+    keyboardText?: string,
+    multipleExpectedValues = false
+  ) => {
+    const actionFunction = jest.fn();
+    const props = { table, showContextMenu: true };
+    props[functionName] = actionFunction;
```
mofojed, line 70: You can initialize the functionName right in the object init, e.g.:
```
    const props = {
      table,
      showContextMenu: true,
      [functionName]: actionFunction,
    };
```

packages/file-explorer/src/FileTestUtils.ts
```diff
@@ -0,0 +1,98 @@
+import { DirectoryStorageItem, FileStorageItem } from './FileStorage';
+
+class FileTestUtils {
```
mofojed, line 3: I've been moving to a different format for "utils" files, see `ClipboardUtils` as an example.
Instead of declaring all the methods as statics on a class, just export each individual function, then export them all as the default, e.g.
```
export function makeFileName(...) { ... }

export function makeDirName( ... ) { ... }

export default { makeFileName, makeDirName };
```

There's a few reasons why I like it better:
- Consumers can just import the methods they need, and it shortens up their code. Instead of:
```
import FileTestUtils from './FileTestUtils';

const file = FileTestUtils.makeFileName(...);
```
It's now just:
```
import { makeFileName } from './FileTestUtils';

const fileName = makeFileName(...);
```
- Doesn't allow the methods to be re-assigned. Technically, someone could write `FileTestUtils.makeFileName = function() { ... }` which could screw stuff up for every other consumer.
- Clear how the utils are supposed to be used. Right now you could do `const fileTestUtils = new FileTestUtils()` which doesn't make sense; you can't really do anything with the class, you just want the bag of functions.

mofojed (changes requested):

packages/file-explorer/src/FileListContainer.test.tsx
emilyhuxng: I think the parameter is still needed because onDelete expects an array as an argument without spreading it.

packages/file-explorer/src/FileExplorer.test.tsx
```diff
@@ -150,44 +148,16 @@ describe('context menu actions work properly', () => {
...
-    fireEvent.mouseDown(file);
-    fireEvent.mouseMove(directory);
-    fireEvent.mouseUp(directory);
-    // await user.pointer({ keys: '[/MouseLeft]' });
-
-    // expect(file).not.toBeInTheDocument();
-  });
-});
- */
+// TODO: #1081
```
mofojed, line 163: Should also add some context right here:
```
// TODO: #1081 Implement drag and drop tests
```

packages/file-explorer/src/FileList.test.tsx
```diff
@@ -146,5 +138,5 @@ describe('mouse actions', () => {
     );
   });
 
-  // Need to implement drag and drop tests
+  // TODO #1081
```
mofojed, line 141: Context

packages/file-explorer/src/FileListContainer.test.tsx
```diff
@@ -83,20 +85,20 @@ describe('renders correct context menu actions', () => {
...
-      expect(actionFunction).toHaveBeenCalledWith(...expectedValue);
-    } else if (expectedValue === null) {
+    if (expectedValue === null) {
       expect(actionFunction).not.toHaveBeenCalled();
-    } else if (expectedValue !== undefined) {
-      expect(actionFunction).toHaveBeenCalledWith(expectedValue);
-    } else {
+    } else if (expectedValue === undefined) {
       expect(actionFunction).toHaveBeenCalled();
+    } else if (Array.isArray(expectedValue) && multipleExpectedValues) {
```
mofojed, line 92: You shouldn't need the `multipleExpectedValues` parameter at all now

mofojed (changes requested):

mofojed (approved):

PR #1079, opened 2023-02-09
build: Update versions of some dependencies with vulnerabilities

mofojed:
- The only vulnerability left is in jszip, which we still have pinned


mattrunyon (changes requested):
Need to fix test failure

Should there be any corresponding `package.json` updates to this? I see plotly was set to 2.18.1, but wasn't sure if that was what fixed the audit or if it was the dependencies of plotly that needed to be bumped

mattrunyon (approved):

PR #1088, opened 2023-02-15
feat: Add support for clickable links

### LINKED ISSUE #712: Add support for clickable links
As both a user and stakeholder, I'd be very interested in a new Markdown column type that can hold clickable links.  Basically, the column could show text like `Clink this link!` or `News article headline!`.  A user could hover their mouse over a cell, which would show the URL to which the link leads.  If the user clicks on a cell, it opens a new tab/window/whatever in the browser to take them there.

emilyhuxng:
Closes #712 

Installed LinkifyJs for link detection. Refer to original issue for details.

```from deephaven import input_table
from deephaven import dtypes as dht, input_table

my_col_defs = {
    "Integers": dht.int32,
    "Doubles": dht.double,
    "Strings": dht.string,
    "Strings2": dht.string,
}

result = input_table(col_defs=my_col_defs)
```

Run this snippet to create an empty input table and enter links into the cells.

Bundle size is 314.21 KB on the `main` branch and 322.4 KB on this branch.


packages/grid/src/GridRenderer.ts
```diff
@@ -1178,8 +1178,39 @@ export class GridRenderer {
...
+          if (tokenizedText[i].t === 'url') {
+            let { width: linkWidth } = context.measureText(value);
+
+            // Adjust width of line if truncated
+            if (value.endsWith('…')) {
+              linkWidth = context.measureText(
+                value.substring(0, value.length - 1)
+              ).width;
+            }
+            context.fillStyle = '#4878ea';
```
dsmmcken, line 1203: should be a theme variable

packages/grid/src/GridTheme.ts
```diff
@@ -146,6 +147,7 @@ const defaultTheme: GridTheme = Object.freeze({
   backgroundColor: '#000000',
   black: '#000000',
   white: '#ffffff',
+  primary: '#4878ea',
```
dsmmcken, line 150: We want to define it as a specific usage for theming, and not just a general color. 

I would put it adjacent to `textColor` defined as something like `textHyperlinkColor`, and then update the other references to match.

Then in irisgridtheme.module.scss that's where you define theme that the color should be $primary (so just some naming needs to be updated to match on that side).

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3596,6 +3602,42 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      };
+
+      const popperOptions: PopperOptions = {
+        placement: 'bottom',
+      };
+
+      return (
+        <div style={wrapperStyle}>
+          <Tooltip
+            key={Date.now()}
```
mattrunyon, line 3626: Did React complain about this not having a key?

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3596,6 +3602,42 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+            options={popperOptions}
+            ref={this.handleTooltipRef}
+          >
+            <div style={{ textAlign: 'left' }}>
+              {linkHoverDisplayValue} - Click once to follow.
+              <br />
+              Click and hold to select this cell.
+            </div>
```
mattrunyon, line 3634: It might be preferable to do what Google sheets does: the tooltip has links and you click the link in the tooltip. Regular click in the cell just selects the cell

I think that might simplify some things. Also, the link handling might only need to be in `IrisGrid` since `Grid` is supposed to be fairly minimal/no frills

mattrunyon (changes requested):

packages/iris-grid/src/IrisGrid.tsx
dsmmcken: We intentionally follow excel style behaviour for link display and interaction. We discussed both when making the ticket and decided we like the immediacy of being able to directly click a link.

packages/grid/src/GridRenderer.ts
```diff
@@ -1178,8 +1178,39 @@ export class GridRenderer {
         fontWidth,
         truncationChar
       );
+      const tokenizedText = model.tokensForCell(truncatedText);
```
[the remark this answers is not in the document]
emilyhuxng: I think we agreed to just check the truncated text for better performance but we could increase the max length?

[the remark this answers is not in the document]
mattrunyon: If the full link needs to be visible, it adds some limitations like the link can't be very long. Also, I don't see the point in showing the URL in the tooltip if the whole link has to be visible already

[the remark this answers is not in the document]
dsmmcken: Should still link truncated text. For performance could do something it looks like it starts with a <protocol>:// then do full link detection. You might only need to do it until the first space after the truncation point space though as matt said.

You may need to build some test cases to compare performance.

packages/grid/src/GridRenderer.ts
```diff
@@ -1178,8 +1178,39 @@ export class GridRenderer {
...
+        textMetrics.actualBoundingBoxDescent;
+
       if (truncatedText) {
-        context.fillText(truncatedText, textX, textY);
+        // X position of current word
+        let startX = textX;
+
+        // Loop through tokenized text and display differently depending on if it's a url
+        for (let i = 0; i < tokenizedText.length; i += 1) {
+          const { v: value } = tokenizedText[i];
```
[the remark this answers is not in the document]
emilyhuxng: Is it worse for performance if I render the original text and then render the link styling again over the original text?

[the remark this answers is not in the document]
dsmmcken: Yes. Calls to draw text are expensive.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,247 @@
...
+        const left = Math.max(gridX, gridX + startX);
+        const right = Math.min(left + linkWidth, gridWidth - verticalBarWidth);
+
+        // If the block is not a url, continue
+        if (tokenizedText[i].t !== 'url') {
+          startX = right;
+
+          // eslint-disable-next-line no-continue
+          continue;
+        }
```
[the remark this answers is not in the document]
emilyhuxng: Should I just change the `tokensForText` function to be `findLinksInText` or something like that instead and cache the `linkify.find` results instead? I don't think `tokensForText` would be used anywhere else then.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,247 @@
...
+      context,
+      text,
+      textWidth,
+      fontWidth,
+      truncationChar
+    );
+
+    // Check if the truncated text contains a link
+    if (linkify.find(truncatedText, 'url').length !== 0) {
+      const tokenizedText = model.tokensForCell(truncatedText);
```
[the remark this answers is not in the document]
emilyhuxng: Do you mean loop through it and check if any element has type 'url'?

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,247 @@
...
+      const linkHeight = actualBoundingBoxAscent + actualBoundingBoxDescent;
+      // The x position of the word
+      let startX = textX;
+
+      // Consider edge cases
+      const top = Math.max(gridY, gridY + textY - linkHeight / 2);
+      const bottom = Math.min(
+        top + linkHeight,
+        gridHeight - horizontalBarHeight
+      );
```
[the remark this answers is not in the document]
emilyhuxng: I'm not quite sure how to apply it in this case. Would it be something like `_.clamp(top + linkHeight, top, gridHeight-horizontalBarHeight`?

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,247 @@
...
+        const { v: value } = tokenizedText[i];
+        let { width: linkWidth } = context.measureText(value);
+
+        // If the text is not visible (right side is less than gridX)
+        if (startX + linkWidth < gridX) {
+          startX += linkWidth;
+
+          // eslint-disable-next-line no-continue
+          continue;
+        }
```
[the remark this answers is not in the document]
emilyhuxng: Do you mean in `setCursor`? Sorry, could you clarify what you mean?

packages/grid/src/GridRenderer.ts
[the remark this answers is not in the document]
mattrunyon: From the logic I'd think the links might be broken if it wasn't fully visible since it's searching on the `truncatedText`

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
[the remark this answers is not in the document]
mattrunyon: We have `lodash.clamp` installed as a module you can import from. Lodash exports so many things that this makes it cleaner/smaller for what we use (although I think it tree shakes properly now if we were to use all of it).

It depends what your min/max values are. The code in your comment translates to "use the value of `top + linkHeight`, but restrict it to the range `[top, gridHeight - horizontalBarHeight]`"

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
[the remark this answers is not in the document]
mattrunyon: In this check you're calling `linkify.find` in the if and then `tokensForCell` (which if refactored w/ my suggestion also calls `linkify.find`). So double calling find. I don't think the performance should be much issue if the strings are short, but it's just removing a redundant call

So something like

```
const links = model.linksForText(truncatedText);
if (links.length > 0) {
  ...
}
```

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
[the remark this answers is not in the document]
mattrunyon: That sounds reasonable to me. [This comment](https://github.com/deephaven/web-client-ui/pull/1088#discussion_r1110448977) is what I was suggesting I believe.

So change to `linksInText` or `findLinksInText` and cache that instead of `tokensForText`

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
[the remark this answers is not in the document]
mattrunyon: On L129 there's a `if (x >= left && x <= right &&...` check. You could also include this condition there instead of needing this check and the continue.

I would recommend combining some of these into booleans before the check just for readability. It can get tricky to understand all the checks in some cases otherwise

```
const cursorIsInBounds = x >= left && x <= right && y >= top && y <= bottom;
const textIsVisible = startX + linkWidth >= gridX;

if (cursorIsInBounds && textIsVisible) {
  return { stuff };
}
```

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: Sounds good. What about removing the link from the tooltip then? The tooltip just feels very large. And the link text distracts from the interaction explanation which I feel should be the primary focus of a non-interactive tooltip

packages/iris-grid/src/IrisGrid.tsx
emilyhuxng: No, I copy and pasted it from the `cellOverflowTooltip`. Should I remove it?

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  private static currentLinkBox?: LinkInformation;
+
+  /**
+   * Checks if mouse is currently over a link and returns its information if so
+   * @param gridPoint Mouse position information
+   * @param grid The grid
+   * @returns A LinkInformation object if it exists and null otherwise
+   */
+  static isHoveringLink(
```
vbabich, line 39: Naming nitpick - `is...` implies a boolean return value, can we rename this to something like `getHoveringLinkInfo`?

packages/grid/src/GridRenderer.ts
```diff
@@ -1178,8 +1259,35 @@ export class GridRenderer {
...
+      // If it doesn't exist, set lengthOfContent to the minimum between length of the original text and 5000
+      if (lengthOfContent === -1) {
+        lengthOfContent = Math.min(5000, text.length);
+      } else if (lengthOfContent > 5000) {
+        // If the index is greater than 5000, limit it to 5000
+        lengthOfContent = 5000;
+      }
+      const contentToCheckForLinks = text.substring(0, lengthOfContent);
+
+      const links = model.findLinksInText(contentToCheckForLinks);
```
vbabich, line 1274: We should probably consider memoizing based on `modelColumn`/`modelRow` instead of the content strings. Seems expensive to do implicit comparisons on up to 10k strings on each lookup. Then `indexOf` and `substring` could also be cached.

emilyhuxng: How about moving the `indexOf` and `substring` into the cached function instead? So I pass in the original text in the cell and the truncated text. I'm not sure how to do it with just `modelColumn` and `modelRow` because `GridModel` doesn't have a method to get the truncated text of a cell.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+          }
+        }
+      }
+    }
+
+    // There are no links or the mouse is not on a link, restore context and currentLinkBox
+    context.restore();
+    this.currentLinkBox = undefined;
+
+    return null;
```
mattrunyon, line 170: Do this check and early return before the if statement above. Then you can just continue the bulk of the logic without it being nested inside an if. This is a general thing I'd recommend when writing code since it helps reduce nesting

Instead of
```typescript
if (whatWeWant) {
  // Do things
  // Do more things
  // Do more things
  // cleanup and return
}

// Cleanup and return what we don't want
```

Do this so the bulk of the code is less indented
```typescript
if (whatWeDontWant) {
  // Cleanup and return what we don't want
}

// Do things
// Do more things
// Do more things
// cleanup and return
```

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+        grid.startEditing(column, row);
+      }
+    }, 1000);
+
+    return true;
+  }
+
+  onUp(gridPoint: GridPoint, grid: Grid): EventHandlerResult {
+    const link = GridLinkMouseHandler.isHoveringLink(gridPoint, grid);
+    if (link == null || !this.isDown) {
```
mattrunyon, line 234: Add `|| this.isHold` here and then you don't need to check below

Unless there's a reason after a hold that the event should stop propagation (the return true)

mattrunyon (changes requested):

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+      fontWidth,
+      truncationChar
+    );
+
+    let lengthOfContent = text.indexOf(' ', truncatedText.length);
+    if (lengthOfContent === -1) {
+      lengthOfContent = Math.min(5000, text.length);
+    } else if (lengthOfContent > 5000) {
+      lengthOfContent = 5000;
+    }
```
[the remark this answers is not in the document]
vbabich: You're probably right. Would be interesting to do some perf testing on tables with short repeating strings vs long and unique.

packages/grid/src/GridRenderer.ts
```diff
@@ -1132,6 +1133,86 @@ export class GridRenderer {
...
+        value = truncatedText.substring(start);
+        end = truncatedText.length;
+      }
+      // Measure the width of the substring before the link
+      const startX =
+        context.measureText(truncatedText.substring(0, start)).width + textX;
+      const fullLinkWidth = context.measureText(value).width;
+      const truncatedLinkWidth = value.endsWith('…')
+        ? context.measureText(value.substring(0, value.length - 1)).width
+        : fullLinkWidth;
```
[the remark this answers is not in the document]
emilyhuxng: I believe both are needed because `fullLinkWidth` is needed every iteration for the x position of non-link styled text.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,267 @@
...
+        const startX =
+          context.measureText(truncatedText.substring(0, start)).width + textX;
+        // Right side is greater than gridX
+        const textIsVisible = startX + linkWidth >= gridX;
+
+        if (textIsVisible) {
+          // Check if the x position is less than the grid x, then linkWidth should be shifted by gridX - startX
+          linkWidth -= startX < gridX ? gridX - startX : 0;
+
+          const left = Math.max(gridX, gridX + startX);
```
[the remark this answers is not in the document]
emilyhuxng: `startX` is negative when it is off the grid because it's the link width added to `textX` which can be negative.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
emilyhuxng: I think after holding, we should return true because then the event is handled (either the row is selected or is now editable) but do you think it should be false?

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -94,85 +94,74 @@ class GridLinkMouseHandler extends GridMouseHandler {
       truncationChar
     );
```
mattrunyon, line 95: Call `getCachedTruncatedString` instead of `truncateToWidth` (this is the closest github will let me comment)

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
mattrunyon: I'd have to check the behavior if it's not prevented from bubbling. Should probably be fine as is

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -36,7 +36,7 @@ class GridLinkMouseHandler extends GridMouseHandler {
    * @param grid The grid
    * @returns A LinkInformation object if it exists and null otherwise
    */
-  static isHoveringLink(
+  static getHoveringLinkInfo(
```
mattrunyon, line 39: This probably shouldn't be static. It should be an instance method since there will be multiple instances of this handler at one time. While it probably shouldn't cause any issues with incorrect links (only 1 link can ever be hovered at a time), there's potentially some edge case where this being static causes some weird link issue when moving the cursor between grids, but I'm not sure

It's only called from within the class or a descendent class

mattrunyon (changes requested):

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
emilyhuxng: Right, but `IrisGridLinkMouseHandler` also calls the function to sense when to display the tooltip so I'm not sure how to handle that.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+  top: number;
+  right: number;
+  bottom: number;
+  width: number;
+  height: number;
+  value: string;
+  href: string;
+}
+
+class GridLinkMouseHandler extends GridMouseHandler {
```
mofojed, line 24: I'd like to make this more generic, not `Link` specific. Could just name it `TokenMouseHandler` instead, then it'll just trigger when a token is clicked.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
mofojed: For treeboxes, we calculate the size of the box in `GridMetricCalculator` and then stuff it in the `GridMetrics` as `visibleRowTreeBoxes`. I could imagine calculating the `visibleTokenBoxes` in `GridMetricCalculator` as well... though we do only _really_ need it for the hovered token... 

What you can do is refactor this logic into a method in `GridUtils` instead, e.g. 
```
export type TokenBox = BoxCoordinates & { token: Token };

getTokenFromXY(x: Coordinate, y: Coordinate, metrics: GridMetrics): TokenBox | null {
   ...
}
```
Then you can call it from `IrisGridTokenMouseHandler` and `GridTokenMouseHandler`, and they can each store their own instance of `currentTokenBox`. Alternatively, `GridTokenMouseHandler` stores it in it's instance, and `IrisGridTokenMouseHandler` iterates through `grid.mouseHandlers` to find the `GridTokenMouseHandler` instance.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    // If it reaches this point, then the mouse is not hovering a link so restore context and currentLinkBox
+    context.restore();
+    this.currentLinkBox = undefined;
+
+    return null;
+  }
+
+  /**
+   * Set the cursor to pointer when hovering of a link
```
mofojed, line 172:
```suggestion
   * Set the cursor to pointer when hovering over a link
```

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+    event: GridMouseEvent
+  ): EventHandlerResult {
+    this.isDown = true;
+    const link = GridLinkMouseHandler.getHoveringLinkInfo(gridPoint, grid);
+
+    // If a modifier key or shift is down, we don't want to handle it and set isDown to false so onUp won't handle it either
+    if (GridUtils.isModifierKeyDown(event) || event.shiftKey || link == null) {
+      this.isDown = false;
+      return false;
+    }
```
mofojed, line 202: I think you can re-order things here to clean this up:
```suggestion
    const link = GridLinkMouseHandler.getHoveringLinkInfo(gridPoint, grid);

    // If a modifier key or shift is down, we don't want to handle it and set isDown to false so onUp won't handle it either
    if (GridUtils.isModifierKeyDown(event) || event.shiftKey || link == null) {
      return false;
    }
    
    this.isDown = true;
```

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    if (this.timeoutId) {
+      clearTimeout(this.timeoutId);
+    }
```
mofojed, line 206: You don't need to surround with `if`, `clearTimeout` does nothing if you pass in `undefined`:
```suggestion
    clearTimeout(this.timeoutId);
```
(Also the `if` check wouldn't work correctly if `timeoutId` was `0`, since that's a falsey value. It's unlikely `timeoutId` is `0`, but I don't think impossible...)

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+      grid.moveCursorToPosition(column, row);
+      if (
+        isEditableGridModel(model) &&
+        column != null &&
+        row != null &&
+        model.isEditableRange(GridRange.makeCell(column, row))
+      ) {
+        grid.startEditing(column, row);
+      }
+    }, 1000);
```
mofojed, line 225: This should be defined as a `static const` on `GridLinkMouseHandler` instead.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    return true;
+  }
+
+  onUp(gridPoint: GridPoint, grid: Grid): EventHandlerResult {
```
mofojed, line 230: We should handle click behaviour in `onClick` instead of `onUp`, I think. Though we can still have `onUp` reset the `isDown` and `clearTimeout` as well.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    // Link is not null, isDown is true
+    // If isHold is true, then the select already happened so onUp shouldn't do anything
+    if (!this.isHold) {
+      // Cancel the timeout and open a new tab with the link
+      if (this.timeoutId != null) {
+        clearTimeout(this.timeoutId);
+      }
+      this.timeoutId = undefined;
+      window.open(link.href, '_blank');
```
mofojed, line 246: Instead of calling `window.open` directly, we should have an `onTokenClicked` prop set on `Grid` that gets called with the click info and the token.
Then `Grid` can have a default value for that prop that calls `window.open` with the token if it's a `LinkToken`.

Also, we should pass the `noopener` and `noreferrer` window options, since we don't know where this link is going: https://developer.mozilla.org/en-US/docs/Web/API/Window/open
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer

packages/grid/src/GridModel.ts
```diff
@@ -183,6 +185,38 @@ abstract class GridModel<
...
+   * @param text The text in the cell
+   * @param truncatedText The truncated text in the cell (what is visible)
+   * @returns An array with link information
+   */
+  findLinksInVisibleText(
+    text: string,
+    truncatedText: string
+  ): ReturnType<typeof find> {
+    return this.getCachedLinksInVisibleText(text, truncatedText);
+  }
```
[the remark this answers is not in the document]
emilyhuxng: Should the function return tokens for the entire cell then (including cut off text)? Since the truncated string is calculated in `GridRenderer`, I wasn't sure how to only return tokens for visible text without passing in the string.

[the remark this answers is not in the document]
mofojed: Yea, good point. For safety add another parameter for the `maxLength`, eg. `tokensForCell(column: ModelIndex, row: ModelIndex, maxLength?: number): Token[]`.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
emilyhuxng: TypeScript complains about it being undefined.

packages/grid/src/mouse-handlers/GridLinkMouseHandler.ts
mofojed: @emilyhuxng Actually thinking about this a bit more - it probably does make more sense to have this in `GridMetricCalculator`. So a method `getTokenBoxesForVisibleCell(row: VisibleIndex, column: VisibleIndex, state: GridMetricState): TokenBoxes` ... and then it only gets called from the mouse handlers for the cell that is hovered.

Boxes are relative to the cell... then we could memoize in GridMetricCalculator

packages/grid/src/GridModel.ts
```diff
@@ -183,6 +188,45 @@ abstract class GridModel<
...
+  tokensForCell(
+    column: ModelIndex,
+    row: ModelIndex,
+    visibleLength: number
+  ): Token[] {
+    return this.getCachedTokensForCell(column, row, visibleLength);
+  }
+
+  getCachedTokensForCell = memoizeClear(
+    (column: ModelIndex, row: ModelIndex, visibleLength: number): Token[] => {
```
mofojed, line 208: We're memoizing on the wrong thing here. Text can change for a given `column`/`row` combination, so you might get stale results. Should instead memoize on the text. Text could also be the same in multiple cells, and then they'd have the same tokens.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1881,6 +1883,154 @@ export class GridMetricCalculator {
...
+        let { value } = links[i];
+
+        if (end > truncatedText.length) {
+          value = truncatedText.substring(start);
+        }
+
+        let { width: linkWidth } = context.measureText(value);
+        // Measure the width of the substring before the link
+        const startX =
+          context.measureText(truncatedText.substring(0, start)).width + textX;
```
mofojed, line 2001: This is suboptimal - we're measuring the text for all the previous segments which we have already calculated. Consider the case for a string such as `one mike@deephaven.io two emily@deephaven.io three https://deephaven.io` (which gives us three tokens). Before the for loop you're measuring the whole string, then for the 0th token (`mike@deephaven.io`) you measure `mike@deephaven.io` and `one `; then for the 1st token (`emily@deephaven.io`) you're measuring `emily@deephaven.io` and `one mikebender@deephaven.io two `; then for the 2nd token (`https://deephaven.io`) you're measuring `https://deephaven.io`, then `one mike@deephaven.io two emily@deephaven.io three `. So in total you're measuring `one ` and `mikebender@deephaven.io` segments 3 times, `two ` and `emily@deephaven.io` segments twice (not including the time you measure the full string).
If you instead loop through using a `while` (as suggested in the `GridRenderer` when drawing), measuring the segments as you go through them and just storing the current width, you'll only need to measure each segment once. E.g.
```
let tokenIndex = 0;
let textStart = 0;
let textWidth = 0;
while (textStart < truncatedText.length && tokenIndex < tokens.length) {
    const token = tokens[tokenIndex];
    textWidth += context.measureText(truncatedText.substring(textStart, token.start)).width;
    const tokenWidth = context.measureText(token.value);
    ... // Calculate the box and push it to the array

    textStart = nextToken.end + 1;
    textWidth += tokenWidth;
}
```

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    if (linksInCell == null) {
+      this.currentLinkBox = undefined;
+      return false;
+    }
+
+    for (let i = 0; i < linksInCell.length; i += 1) {
+      const { x1: left, x2: right, y1: top, y2: bottom } = linksInCell[i];
+      if (x >= left && x <= right && y >= top && y <= bottom) {
+        this.currentLinkBox = linksInCell[i];
```
mofojed, line 65: Need a check here to ensure the token is a `LinkToken`

mofojed (changes requested):

packages/grid/src/GridRenderer.ts
```diff
@@ -3053,6 +3050,155 @@ export class GridRenderer {
...
+      }
+
+      const {
+        actualBoundingBoxAscent,
+        actualBoundingBoxDescent,
+      } = context.measureText(truncatedText);
+      const linkHeight = actualBoundingBoxAscent + actualBoundingBoxDescent;
+
+      // Consider edge cases
+      const top = Math.max(gridY, gridY + textY - actualBoundingBoxAscent);
```
mofojed, line 3135: The calculations for `top` and `bottom` here seem out of alignment. The bottom part of the link isn't clickable, and the top part extends past the top of the link. I suggest adding some debug code to draw the boxes as well so you can debug easier, and perhaps logging breakpoints here while testing.
Also we have a bit of a discrepancy between `getTextRenderMetrics` and `getTokenBoxesForVisibleCell`; specifically, `getTextRenderMetrics` provides the coordinates in relation to `gridX`/`gridY`, whereas `getTokenBoxesForVisibleCell` translates them by `gridX`/`gridY`. I think it's best to do the same as `getTextRenderMetrics` (e.g. not translate by `gridX`/`gridY`, and whoever calls it must do that translation).

packages/grid/src/GridRenderer.ts
```diff
@@ -3053,6 +3050,155 @@ export class GridRenderer {
...
+    state: GridRenderState
+  ): TokenBox[] {
+    return this.getCachedTokenBoxesForVisibleCell(column, row, state);
+  }
+
+  getCachedTokenBoxesForVisibleCell = memoizeClear(
+    (
+      column: VisibleIndex,
+      row: VisibleIndex,
+      state: GridRenderState
```
mofojed, line 3073: This memoization right now is kind of useless - the `state` object is changed every time the mouse moves. What we _could_ memoize is the position of token boxes relative to the cell; that would require just the text, tokens for the cell, font, and text width as inputs. Would then need to translate in the `getTokenBoxesForVisibleCell` method, e.g.
```
getTokenBoxesForVisibleCell(column, row, state) {
  ...
  const { width: textWidth, x: textX, y: textY } = this.getTextRenderMetrics(state, column, row);
  ... get truncated text
  const tokens = model.tokensForCell(modelColumn, modelRow, truncatedText.length);
  return this.getCachedTokenBoxes(text, tokens, font, textWidth, context).map(tokenBox => ({
    x1: tokenBox.x1 + textX,
    y1: tokenBox.y1 + textY,
    ...
  });
}

getCachedTokenBoxesForVisibleCell = memoizeClear((text, tokens, 
```

packages/grid/src/GridUtils.ts
```diff
@@ -42,16 +42,16 @@ export interface CellInfo {
...
+export type LinkToken = Token & { href: string };
+export type URLToken = Token & { type: 'url' };
+export type EmailToken = Token & { type: 'email' };
+
+export function isLinkToken(token: Token): token is LinkToken {
+  return (token as LinkToken)?.href !== undefined;
+}
 
-export type TokenBox = BoxCoordinates & { token: Token };
+export type TokenBox = BoxCoordinates & { token: LinkToken };
```
mofojed, line 54: Should just be `token: Token`

packages/grid/src/mouse-handlers/GridTokenMouseHandler.ts
```diff
@@ -32,27 +32,30 @@ class GridTokenMouseHandler extends GridMouseHandler {
...
       column,
       row,
-      metricState,
-      renderer,
       renderState
     );
 
     // If there are no links in the cell, return false
-    if (linksInCell == null) {
+    if (linksInCell.length === 0) {
```
mofojed, line 44: Don't need this special case - if `linksInCell.length === 0`, we're going to skep that `for` loop anyway and set `this.currentLinkBox = undefined; return false;` anyway.

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.ts
```diff
@@ -43,25 +44,28 @@ class IrisGridTokenMouseHandler extends GridMouseHandler {
...
+    const linksInCell = renderer.getTokenBoxesForVisibleCell(
       column,
       row,
-      metricState,
-      renderer,
       renderState
     );
 
-    if (linksInCell == null) {
+    if (linksInCell.length === 0) {
```
mofojed, line 55: Don't need this `if` case, we're going to skip the for loop in this case anyway

mofojed (changes requested):

mofojed:
Also @emilyhuxng need to update the unit tests, move the `getOrThrow` tests into `@deephaven/utils` as a test.

packages/grid/src/GridRenderer.ts
```diff
@@ -3009,6 +3050,154 @@ export class GridRenderer {
...
+      column,
+      row
+    );
+
+    const { fontWidths } = metrics;
+
+    // Set the font and baseline and change it back after
+    context.save();
+    context.font = theme.font;
+    context.textBaseline = 'middle';
```
mofojed, line 3088: Instead of setting font/textBaseline again, call `this.configureContext(context, state)` to make sure the context is configured correctly.

Though I went down a bit of a rabbit hole... I was wondering why you needed to set the `textBaseline` considering we never set it to anything else. Digging into it a bit, it looks like the baseline gets reset when we update the width/height of the canvas: https://stackoverflow.com/questions/49094139/canvas-context-gets-reset-when-canvas-width-height-are-changed

We do that in [Grid.updateCanvasScale](https://github.com/deephaven/web-client-ui/blob/46c5e2461ab2344bd62d482ddcfd179c69e81ce0/packages/grid/src/Grid.tsx#L819).

We should really be configuring the context there right after we scale it to ensure it's always got the properties we expect. But `configureContext` takes the `GridRenderState`, which we don't have yet in `updateCanvasScale`... So I think a more complete fix:
- In `Grid.updateCanvas`, add a call to `this.updateRenderState()` after setting the context scale, and then call `this.renderer.configureContext(context, renderState)`.
- In `drawCanvas`, don't call `updateRenderState()` anymore, just use `this.renderState` (since it has already been updated)
- Mark `updateCanvasScale` and `drawCanvas` as `private` methods - we should only really call `requestUpdateCanvas()` or `updateCanvas()` externally.
  - There's a couple places where `updateCanvasScale` is called before calling `updateCanvas()`, those calls can simply be removed from: `componentDidMount()`, `handleResize()`, and `IrisGrid.handleAnimationLoop()`

packages/grid/src/GridRenderer.ts
```diff
@@ -3009,6 +3050,154 @@ export class GridRenderer {
...
+      modelRow,
+      truncatedText.length
+    );
+
+    context.restore();
+
+    // Check if the truncated text contains a link
+    if (tokens.length === 0) {
+      return [];
+    }
```
mofojed, line 3118: Return `EMPTY_ARRAY`

packages/grid/src/GridRenderer.ts
```diff
@@ -3009,6 +3050,154 @@ export class GridRenderer {
...
+    return this.getCachedTokenBoxesForVisibleCell(
+      truncatedText,
+      tokens,
+      theme.font,
+      'middle',
+      linkHeight,
+      context
+    ).map(tokenBox => ({
+      x1: tokenBox.x1 + textX,
+      y1: tokenBox.y1 + (textY - actualBoundingBoxAscent),
```
mofojed, line 3129: I'm still seeing the token box is off by a couple pixels. We can debug together if you want, I'm not sure why it's off.

packages/iris-grid/src/IrisGridTheme.module.scss
```diff
@@ -44,6 +44,7 @@ $header-height: 30px;
   scroll-bar-active-color: $gray-300;
 
   text-color: $white;
+  hyper-link-color: $primary;
```
mofojed, line 47: Hyperlink is one word
```suggestion
  hyperlink-color: $primary;
```

packages/iris-grid/src/IrisGridTheme.ts
```diff
@@ -71,6 +71,7 @@ const theme: Partial<IrisGridThemeType> = Object.freeze({
   scrollBarActiveSelectionTickColor:
     IrisGridTheme['scroll-bar-active-selection-tick-color'],
   textColor: IrisGridTheme['text-color'],
+  hyperlinkColor: IrisGridTheme['hyper-link-color'],
```
mofojed, line 74:
```suggestion
  hyperlinkColor: IrisGridTheme['hyperlink-color'],
```

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.ts
```diff
@@ -0,0 +1,135 @@
...
+      if (
+        x >= left &&
+        x <= right &&
+        y >= top &&
+        y <= bottom &&
+        isLinkToken(tokensInCell[i].token)
+      ) {
+        this.currentLinkBox = translatedTokenBox;
+        return true;
+      }
```
mofojed, line 68: Minor optimization - you could check if it's a link token before translating the token:
```
      if (isLinkToken(tokensInCell[i].token)) {
      
      const translatedTokenBox = GridUtils.translateTokenBox(
        tokensInCell[i],
        metrics
      );
      const { x1: left, x2: right, y1: top, y2: bottom } = translatedTokenBox;
      if (
        x >= left &&
        x <= right &&
        y >= top &&
        y <= bottom
      ) {
        this.currentLinkBox = translatedTokenBox;
        return true;
      }
      }
```

mofojed (changes requested):

packages/grid/src/Grid.tsx
```diff
@@ -791,15 +806,19 @@ class Grid extends PureComponent<GridProps, GridState> {
   }
 
   updateCanvas(metrics = this.updateMetrics()): void {
+    if (!this.canvasContext) throw new Error('canvasContext not set');
+
```
mofojed, line 810: You don't need this - the second line in `updateCanvasScale()` already does this.
```suggestion
```

packages/grid/src/Grid.tsx
```diff
@@ -1731,11 +1705,13 @@ class Grid extends PureComponent<GridProps, GridState> {
     const gridPoint = this.getGridPointFromEvent(event);
     const mouseHandlers = this.getMouseHandlers();
     let cursor = null;
+
```
mofojed, line 1708: Didn't change this code, so I'd get rid of the added blank lines.
```suggestion
```

packages/grid/src/Grid.tsx
```diff
@@ -1731,11 +1705,13 @@ class Grid extends PureComponent<GridProps, GridState> {
...
     for (let i = 0; i < mouseHandlers.length; i += 1) {
       const mouseHandler = mouseHandlers[i];
       const result =
         mouseHandler[functionName] != null &&
         mouseHandler[functionName](gridPoint, this, event);
+
```
mofojed, line 1714:
```suggestion
```

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -53,27 +54,6 @@ export interface GridMetricState {
...
 
-/**
- * Retrieve a value from a map. If the value is not found and no default value is provided, throw.
- * Use when the value _must_ be present
- * @param map The map to get the value from
- * @param key The key to fetch the value for
- * @param defaultValue A default value to set if the key is not present
- * @returns The value set for that key
- */
-export function getOrThrow<K, V>(
```
mofojed, line 64: Instead of just getting rid of this (which would be a breaking API change for anybody using our `@deephaven/grid` package), just re-export the `getOrThrow` function, e.g.
```
export { getOrThrow } from '@deephaven/utils';
```
Otherwise we need to mark this change as breaking.

mofojed (changes requested):
I tested with a table with cells with a lot of links and large links to verify performance, and it seems to be fine:
```
table_size = 1000
largelink_size=100
multilink_size = 10000
t_links = empty_table(table_size).update(["x=i", "link=`https://www.github.com/deephaven/web-client-ui/pull/` + i", "multilink=`http://www.example.com fdsaf d `.repeat(repeat_size)", "largelink= `http://www.example.com/` + `foo/`.repeat(largelink_size)"])
```
However, while testing this I noticed that the tooltip becomes essentially useless when the link is really long, it takes up the full width of the screen and doesn't truncate the link at all:
long-link
@dsmmcken should we have a maxWidth on that tooltip? Wrap or truncate that link?

Side note - it would be kind of cool to have that be an interactive tooltip that you could copy the link from. Not necessarily needed for this to merge though.

dsmmcken:
> a maxWidth on that tooltip? Wrap or truncate that link?

max width + wrap.

dsmmcken:
> Side note - it would be kind of cool to have that be an interactive tooltip that you could copy the link from. Not necessarily needed for this to merge though.

You can already copy the cell though?

mofojed:
> > Side note - it would be kind of cool to have that be an interactive tooltip that you could copy the link from. Not necessarily needed for this to merge though.
> 
> You can already copy the cell though?

Cell could be multiple hyperlinks or have other text not part of the hyperlink though

mofojed (changes requested):
Looks like we're not accounting for moved columns anymore when drawing cell contents:
image
Easy to reproduce, just create the table:
```
table_size = 1000
largelink_size=100
multilink_size = 10000
t_links = empty_table(table_size).update(["x=i", "link=`https://www.github.com/deephaven/web-client-ui/pull/` + i", "multilink=`http://www.example.com fdsaf d `.repeat(multilink_size)", "largelink= `http://www.example.com/` + `foo/`.repeat(largelink_size)"])
```
Then move the `x` column after the `link` column.

Not sure exactly where it's broken.

mofojed (approved):
Excellent work!

mattrunyon:
image

I don't like this tooltip. @dsmmcken should it truncate at some point? Also, the "Click once to follow" is on the same line as the link which is not great for readability even on shorter links like below

image

My recommendation is to remove the link from the tooltip entirely. It should only show the instructions about clicking. We style like a hyperlink, show the cursor like a hyperlink, I think it's fine to omit the URL. Or maybe truncate it.

Or if you really don't want to truncate it, at least move the "Click once to follow" to a separate line and put a blank line in between so there's visual separation

Last thing, I think it should be "Click to follow link" or something. The "once" feels unnecessary in the existing tooltip statement

dsmmcken:
Okay. How about this? We use the browser url preview bar to handle displaying the path, and then remove the url from the tooltip?

image

Proof of concept of placing a temporary "invisible link" over the target area that is under the cursor. Confirmed that the url preview bar updates correctly:
https://codepen.io/dsmmcken/pen/mdGqjZy

mattrunyon:
Ya something like that could work. I just found the tooltip overwhelming w/ long links and I'm not sure how useful the whole link is when it's long. Looks like if you have a really long link, the browser preview will span the width of the window, but truncate at a certain point. At a certain length, I'd say the content of the link becomes way less meaningful. Like TS playground links or Google Drive share links are lengthy, but the data contained in the link is not meaningful to a human after the domain really

image

mofojed:
@mattrunyon it's pretty much exactly the same as the tooltip in Excel for Mac (the "Click once", the wrapping, etc):
excel-hyperlink-tooltip

I don't see a need to hold up this PR for the tooltip. Using the browser preview is kinda neat but I do not like the hack of having an invisible element there. If we really want to change it up, I'd go more the route of a Google sheets kind of tooltip (truncate the URL, make it `interactive` so you can click the copy). But in any case I don't think we should overthink this.

dsmmcken (approved):
Alright, cool :shipit:

mattrunyon:
I know we chose Excel as the behavior to mimic, but I also think their tooltip should be better. The 2nd link you hover in your example doesn't show the full link. The 3rd link does, but doesn't show info on the interaction.

I'll approve, but think we should cleanup the tooltip in a followup.

Just because Excel did it, doesn't mean it is the optimal behavior.

At the very least I would push to put the interaction tips after a blank line so they are clearly visible and distinct from the URL. I don't like that Excel puts them after a hyphen since the point of the tooltip should be to hint how something is used/what it's for.

image

image

mattrunyon (approved):

mattrunyon:
I opened #1144 as a followup for improving the tooltip with the options

PR #1093, opened 2023-02-16
test: Increase code coverage in package utils

### LINKED ISSUE #1089: Increase code coverage in package utils
Increase code coverage in package utils.

emilyhuxng:
Closes #1089 

Increase code coverage in package utils.

### DIFF
--- a/packages/utils/src/ClipboardUtils.test.ts
+++ b/packages/utils/src/ClipboardUtils.test.ts
@@ -19,7 +19,7 @@ describe('Clipboard', () => {
       expect(navigator.clipboard.writeText).toHaveBeenCalledWith('test');
     });
 
-    it('should throw an error when writeText fails and call copyToClipboardExecCommand', async () => {
+    it('should call copyToClipboardExecCommand if writeText fails', async () => {
       Object.assign(navigator, {
         clipboard: {
           writeText: () => {
@@ -27,14 +27,10 @@ describe('Clipboard', () => {
           },
         },
       });
+      document.execCommand = jest.fn(() => true);
       jest.spyOn(navigator.clipboard, 'writeText');
 
-      await expect(copyToClipboard('test')).rejects.toThrowError(
-        'Unable to execute copy command'
-      );
-      expect(navigator.clipboard.writeText).toThrowError(
-        'Could not write text'
-      );
+      await expect(copyToClipboard('test')).resolves.toBeUndefined();
       expect(document.execCommand).toHaveBeenCalledWith('copy');
     });
 
@@ -56,6 +52,7 @@ describe('Clipboard', () => {
       document.execCommand = jest.fn(() => true);
 
       await expect(copyToClipboard('test')).resolves.toBeUndefined();
+      await expect(document.execCommand).toHaveBeenCalledWith('copy');
     });
   });
 });
--- a/packages/utils/src/ClipboardUtils.test.ts
+++ b/packages/utils/src/ClipboardUtils.test.ts
@@ -1,0 +1,19 @@
+import { copyToClipboard, copyToClipboardExecCommand } from './ClipboardUtils';
+
+Object.assign(navigator, {
+  clipboard: {
+    writeText: () => {},
+  },
+});
+
+describe('Clipboard', () => {
+  describe('writeText', () => {
+    jest.spyOn(navigator.clipboard, 'writeText');
+
+    it('should call clipboard.writeText', () => {
+      expect(navigator.clipboard.writeText).toHaveBeenCalledWith('test');
+    });
+
+    it();
+  });
+});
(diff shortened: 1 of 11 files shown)


packages/utils/src/ColorUtils.test.ts
```diff
@@ -0,0 +1,15 @@
+import ColorUtils from './ColorUtils';
+
+describe('isDark', () => {
+  it('returns true if the background is dark', () => {
```
dsmmcken, line 4: You should throw some more tests in here for format of the color string
`#000`
`black`
`rgb(0,0,0)`
`rgba(0,0,0,1)`
`hsl(0, 0%, 0%)`

emilyhuxng: For some reason, `black` doesn't work. Should I omit that test or fix ColorUtils?

dsmmcken: You sure? it works for me?

emilyhuxng: image
It's throwing for me. Did I enter something wrong? I am calling `ColorUtils.isDark('black')`.

packages/utils/src/TestUtils.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const button = screen.getByRole('button');
+    await TestUtils.click(user, button, { ctrlKey: true });
+    expect(clickHandler).toHaveBeenCalledTimes(2);
+    expect(clickHandler.mock.calls[1][0].ctrlKey).toBeTruthy();
+  });
+
+  it('should left click with shift key down', async () => {
+    const button = screen.getByRole('button');
+    await TestUtils.click(user, button, { shiftKey: true });
+    expect(clickHandler).toHaveBeenCalledTimes(3);
```
vbabich, line 69: This makes the tests dependent on the order and number of `it` blocks. Instead, reset all mocks in `beforeEach` and check if the mock has been called once in each `it` block.

packages/utils/src/ClipboardUtils.test.ts
```diff
@@ -0,0 +1,49 @@
...
+            throw new Error('Could not write text');
+          },
+        },
+      });
+      jest.spyOn(navigator.clipboard, 'writeText');
+
+      await expect(copyToClipboard('test')).rejects.toThrowError(
+        'Unable to execute copy command'
+      );
+      expect(navigator.clipboard.writeText).toThrowError(
```
vbabich, line 33: NBD, but this tests the mock defined in the same `it` block. Ok to leave as is if we just want to make sure the mock is functional.

packages/utils/src/ClipboardUtils.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      );
+      expect(document.execCommand).toHaveBeenCalledWith('copy');
+    });
+
+    it('calls copyToClipboardExecCommand if clipboard is undefined', async () => {
+      Object.assign(navigator, {
+        clipboard: undefined,
+      });
+
+      await expect(copyToClipboard('test')).rejects.toThrowError(
```
vbabich, line 44: This could throw or not, depending on the `document.execCommand` mock.
I'd do `document.execCommand = jest.fn(() => false);` before calling `copyToClipboard('test')` here, and add another `it` block with a "happy path" that calls `execCommand` and resolves successfully (doesn't throw) with `document.execCommand = jest.fn(() => true);` before the `copyToClipboard` call.

packages/utils/src/ClipboardUtils.test.ts
```diff
@@ -0,0 +1,49 @@
+import { copyToClipboard } from './ClipboardUtils';
+
+document.execCommand = jest.fn();
```
vbabich, line 3: This mock might need a reset in `beforeEach`

vbabich (changes requested):

packages/utils/src/ClipboardUtils.test.ts
emilyhuxng: I wanted to test the case where `writeText` throws. Should the mock be defined outside?

vbabich: No need to define the mock outside. What I meant in my comment - this line `expect(navigator.clipboard.writeText).toThrowError...` would pass even without calling `copyToClipboard('test')` above:

```
it('passes without calling copyToClipboard ', async () => {
      Object.assign(navigator, {
        clipboard: {
          writeText: () => {
            throw new Error('Could not write text');
          },
        },
      });
      expect(navigator.clipboard.writeText).toThrowError(
        'Could not write text'
      );
    });
```


Again, no big deal, just wanted to point out that `expect(navigator.clipboard.writeText)` tests the mock and not the `copyToClipboard` function.

vbabich (approved):

PR #1102, opened 2023-02-23
docs: Fix VITE_CORE_API_URL in README.md

vbabich:


dsmmcken:
Should this section even be there? Doesn't the default .env.development already set this up correctly?

bmingles:
> Should this section even be there? Doesn't the default .env.development already set this up correctly?

It does look like the combination of `.env` + `.env.development` files handle these 2 env vars. I deleted the `.env.local` files and was able to run things.

vbabich:
> Should this section even be there? Doesn't the default .env.development already set this up correctly?

Might still need this for custom setups. Should I prepend it with something like
```
If your DHC address is different from the default `http://localhost:10000`, edit `.env.local` in each package to contain...
```
?

dsmmcken:
Yes that's better.

README.md
```diff
@@ -34,11 +34,11 @@ We are still using node 16.x and npm 8.x. If you are [using nvm](https://github.
 - `npm run e2e`: Runs the Playwright end-to-end tests locally.
 - `npm run e2e:headed`: Runs end-to-end tests in headed mode debug mode. Useful if you need to debug why a particular test isn't work. For example, to debug the `table.spec.ts` test directly, you could run `npm run e2e:headed -- ./tests/table.spec.ts`.
 
-Edit `.env.local` in each package to contain the following pointing to your local DHC address. These are needed for the session websocket and for things like notebooks to be proxied correctly by Vite.
+If your DHC address is different from the default `http://localhost:10000`, edit `.env.local` in each package to contain the following pointing to your local DHC. These are needed for the session websocket and for things like notebooks to be proxied correctly by Vite.
```
mattrunyon, line 37: This needs to be `.env.development.local` if running locally. The load priority of the files is `.env.{mode}.local`, `.env.{mode}`, `.env.local`, `.env`

So if running in dev mode, the proxy URL won't be overwritten

We should probably just move `VITE_PROXY_URL` into `.env` though. If running `npm run preview` after a build, Vite starts the preview server in production mode. So the proxy URL isn't set and Vite won't proxy the requests

mattrunyon (changes requested):

README.md
vbabich: Moved `VITE_PROXY_URL` to `.env`.

Now `.env.local` is the correct location for the override if we want it to work for both preview and dev, right?

README.md
```diff
@@ -34,11 +34,11 @@ We are still using node 16.x and npm 8.x. If you are [using nvm](https://github.
...
 
 ```
-VITE_CORE_API_URL=http://localhost:10000/jsapi
-VITE_PROXY_URL=http://localhost:10000
+VITE_CORE_API_URL=<jsapi-path>
```
mattrunyon, line 40: I don't think you need this unless the JS API is at a different path than normal on your server. Just the proxy should work. I ran a build and preview locally w/ just the proxy changed and it worked

I think this should be removed from the readme since it's rare somebody would need to change the API URL

mattrunyon (changes requested):

README.md
vbabich: Yep, agreed.

README.md
mattrunyon: Yes, `.env.local` is the right spot and it will work for both

mattrunyon (approved):

PR #1110, opened 2023-02-24
feat: isConfirmDangerProp

### LINKED ISSUE #1109: BasicModal - isConfirmDanger prop
As a developer I want to be able to configure a basic confirmation dialog for actions that a user wouldn't want to take accidentally. 

We should add an optional `isConfirmDanger?: boolean` prop to the `BasicModal` component such that if true:

- The default action is "Cancel"
- The confirmation button should use the "danger" color

image

If the prop is not set or is set to false, the behavior should be as-is:
image

bmingles:
resolves #1109

### DIFF
--- a/packages/components/src/BasicModal.tsx
+++ b/packages/components/src/BasicModal.tsx
@@ -15,6 +15,7 @@ export interface BasicModalProps {
   cancelButtonText?: string;
   confirmButtonText?: string;
   discardButtonText?: string;
+  isConfirmDanger?: boolean;
   children?: React.ReactNode;
   'data-testid'?: string;
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/BasicModal.tsx
```diff
@@ -64,8 +67,12 @@ function BasicModal(props: BasicModalProps) {
   }, [onConfirm, onModalDisable]);
 
   const onOpened = useCallback(() => {
-    confirmButton.current?.focus();
-  }, []);
+    if (isConfirmDanger) {
```
dsmmcken, line 70: I would probably trust the user and keep the focus on the primary action.

vbabich (approved):

PR #1111, opened 2023-02-27
fix: Ordering of subplots

mofojed:
- User is expecting the plots to be ordered 0 at top, increasing downward
- Plotlys bounds coordinate system has 0,0 in the bottom left, increasing upward
- Invert the coordinates so that it appears in plotly correctly
- Add unit tests

### DIFF
--- a/packages/chart/src/ChartUtils.test.ts
+++ b/packages/chart/src/ChartUtils.test.ts
@@ -315,6 +315,79 @@ describe('updating layout axes', () => {
   });
 });
 
+describe('handles subplots and columns/rows correctly', () => {
+  const width = ChartUtils.AXIS_SIZE_PX * 5;
+  const height = ChartUtils.AXIS_SIZE_PX * 10;
+
+  it('handles row location correctly', () => {
+    const axes = ChartTestUtils.makeDefaultAxes();
+    const charts = [
+      ChartTestUtils.makeChart({ axes, row: 0 }),
+      ChartTestUtils.makeChart({ axes, row: 1 }),
+    ];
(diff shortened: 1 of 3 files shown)


packages/chart/src/ChartUtils.test.ts
```diff
@@ -315,6 +315,79 @@ describe('updating layout axes', () => {
...
+      ChartTestUtils.makeChart({ axes, row: 0 }),
+      ChartTestUtils.makeChart({ axes, row: 1 }),
+    ];
+    const figure = ChartTestUtils.makeFigure({ charts, rows: 2 });
+    expect(
+      ChartUtils.getChartBounds(figure, charts[0], width, height)
+    ).toEqual({ bottom: 0.55, top: 1, left: 0, right: 1 });
+    expect(
+      ChartUtils.getChartBounds(figure, charts[1], width, height)
+    ).toEqual({ bottom: 0, top: 0.45, left: 0, right: 1 });
```
mattrunyon, line 334: These tests might be more robust as comparing the values of `left` or `top` instead. Unless you are also wanting to check the spacing which is also fine.

`expect(charts[0].bounds.top > charts[1].bounds.top)` or something like that

mofojed: Good call

mofojed: Actually I do want to check the spacing as well... otherwise we could position them still in the "right" spots, but there's a big gap between them or something. I'll make the spacing dependent on the margin so it's a little more clear.

PR #1114, opened 2023-02-27
test: Increase code coverage in package jsapi-utils

emilyhuxng:
Increase code coverage in package jsapi-utils


__mocks__/dh-core.js
```diff
@@ -466,17 +466,22 @@ class Format {
 }
 
 class FilterCondition {
+  constructor(type, parent) {
```
mofojed, line 469: Doesn't look like you ended up using these, you just defined a new mock in `TableUtils.test.ts`. Revert the changes to this file.

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -33,66 +36,78 @@ function makeColumns(count = 5): Column[] {
...
-  tableSorts = TableUtils.toggleSortForColumn(tableSorts, columns, 3, true);
-  table.applySort(tableSorts);
-
-  expect(table.sort.length).toBe(0);
-});
+type MockFilterCondition = Record<'not' | 'and' | 'or', jest.Mock>;
+function makeFilterCondition(type = ''): MockFilterCondition {
+  return {
+    not: jest.fn(
+      () => makeFilterCondition(`${type}.${FilterType.eq}`) as FilterCondition
```
mofojed, line 43: Shouldn't need the `as FilterCondition` for any of these

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -175,6 +190,30 @@ describe('quick filter tests', () => {
...
+
+    filter[expectedFn].mockReturnValueOnce(expectedResult);
+
+    const result = TableUtils[functionName](column, text);
+
+    if (args.length > 0) {
+      expect(filter[expectedFn]).toHaveBeenCalledWith(...args);
+    } else {
+      expect(filter[expectedFn]).toHaveBeenCalled();
+    }
```
mofojed, line 213: Shouldn't need to split this case up. `...args` will just be an empty array, and calling `.toHaveBeenCalledWith(...args)` just means it will have been called with no arguments, which is actually a stricter check than just `.toHaveBeenCalled()` (which would pass whether it was called with or without arguments).

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -243,7 +282,7 @@ describe('quick filter tests', () => {
     };
 
     dh.FilterCondition = {
-      invoke: jest.fn(type => makeFilterCondition(type)),
+      invoke: jest.fn(type => makeFilterCondition(type) as FilterCondition),
```
mofojed, line 285: Shouldn't need `as FilterCondition` here

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -267,6 +306,381 @@ describe('quick filter tests', () => {
...
+
+      expect(filter.isNull).toHaveBeenCalled();
+      expect(nullResult.not).toHaveBeenCalled();
+      expect(notResult.and).toHaveBeenCalledWith(invokeResult);
+
+      if (args.length > 0) {
+        expect(filter.invoke).toHaveBeenCalledWith(...args);
+      } else {
+        expect(filter.invoke).toHaveBeenCalled();
+      }
```
mofojed, line 442: Shouldn't need to split this out... we should be strict about what it's getting called with.

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -267,6 +306,381 @@ describe('quick filter tests', () => {
...
+        operation,
+        value,
+        timezone
+      );
+
+      if (args.length > 0) {
+        expect(filter[expectedFn]).toHaveBeenCalledWith(...args);
+      } else {
+        expect(filter[expectedFn]).toHaveBeenCalled();
+      }
```
mofojed, line 473: Don't split it out, let's be explicit about what the expected function gets called with.

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -267,6 +306,381 @@ describe('quick filter tests', () => {
...
+        const invokeResult = makeFilterCondition();
+        const expectedResult = makeFilterCondition();
+
+        filter.isNull.mockReturnValueOnce(nullResult);
+
+        filter.invoke.mockReturnValueOnce(invokeResult);
+
+        invokeResult.not.mockReturnValueOnce(notResult);
+
+        (nullResult.or as jest.Mock).mockReturnValueOnce(expectedResult);
```
mofojed, line 624: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -279,7 +693,9 @@ describe('quick filter tests', () => {
     it('handles empty cases', () => {
       const column = makeFilterColumn();
 
+      // @ts-expect-error text is generally a string
       expect(TableUtils.makeQuickNumberFilter(column, null)).toBe(null);
+      // @ts-expect-error text is generally a string
       expect(TableUtils.makeQuickNumberFilter(column, undefined)).toBe(null);
```
mofojed, line 699: Instead of writing these test cases for invalid input (which should be caught by TypeScript anyway), just take out the lines in `makeQuickNumberFilter` where we check `text == null` (since it can never happen).

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -615,8 +1038,10 @@ describe('quick filter tests', () => {
     it('handles invalid cases', () => {
       const column = makeFilterColumn();
 
+      // @ts-expect-error text is generally a string
       expect(() => TableUtils.makeQuickDateFilter(column, null, '')).toThrow();
       expect(() =>
+        // @ts-expect-error text is generally a string
         TableUtils.makeQuickDateFilter(column, undefined, '')
       ).toThrow();
```
mofojed, line 1046: Don't need cases where we pass in `null` or `undefined` as TS shouldn't allow that.

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1168,11 +1593,11 @@ describe('quick filter tests', () => {
 
       filter.isNull.mockReturnValueOnce(nullResult);
 
-      nullResult.not.mockReturnValueOnce(notResult);
+      (nullResult.not as jest.Mock).mockReturnValueOnce(notResult);
 
       filter.invoke.mockReturnValueOnce(invokeResult);
 
-      notResult.and.mockReturnValueOnce(expectedResult);
+      (notResult.and as jest.Mock).mockReturnValueOnce(expectedResult);
```
mofojed, line 1600: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1192,6 +1617,7 @@ describe('quick filter tests', () => {
       const column = makeFilterColumn();
 
       expect(TableUtils.makeQuickTextFilter(column, null)).toBe(null);
+      // @ts-expect-error text is generally a string
       expect(TableUtils.makeQuickTextFilter(column, undefined)).toBe(null);
```
mofojed, line 1621: We should just remove the `null` case in `makeQuickTextFilter`, it never gets called with `null` (or `undefined`)

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1211,7 +1637,9 @@ describe('quick filter tests', () => {
       filter.isNull.mockReturnValueOnce(expectedNullFilter);
 
       const expectedNotFilter = makeFilterCondition();
-      expectedNullFilter.not.mockReturnValueOnce(expectedNotFilter);
+      (expectedNullFilter.not as jest.Mock).mockReturnValueOnce(
```
mofojed, line 1640: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+      expect(TableUtils.makeSelectValueFilter(column, [], false)).toBe(
+        expectedResult
+      );
+      expect(filter.eq).toHaveBeenCalledWith(expectedValue);
+      expect(filter.notEq).toHaveBeenCalledWith(expectedValue);
+      expect(eqResult.and).toHaveBeenCalledWith(notEqResult);
+    };
+
+    it('should return null if there are no selected values and invertSelection is true', () => {
+      const column = {} as Column;
```
mofojed, line 1817: Should probably still create an actual column here and not cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -147,7 +162,7 @@ describe('quick filter tests', () => {
   function makeFilterColumn(type = 'string'): Column {
     const filter = makeFilter();
     const column = IrisGridTestUtils.makeColumn('test placeholder', type, 13);
-    column.filter = jest.fn(() => filter);
+    column.filter = jest.fn(() => filter as FilterValue);
     return column;
   }
```
mofojed, line 167: Much like `MockFilterCondition` is defined above, define `MockColumn` here - that will allow you to get rid of all the casting you have elsewhere. E.g.
```
  type MockColumn = Omit<Column, 'filter'> & { filter(): MockFilter };
  function makeFilterColumn(type = 'string'): MockColumn {
    const filter = makeFilter();
    const column = IrisGridTestUtils.makeColumn('test placeholder', type, 13);
    column.filter = jest.fn(() => filter);
    return column as MockColumn;
  }
```

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+    const testNoSelectedItems = (type: string, expectedValue: unknown) => {
+      const column = makeFilterColumn(type);
+      const filter = column.filter();
+
+      const eqResult = makeFilterCondition();
+      const notEqResult = makeFilterCondition();
+      const expectedResult = makeFilterCondition();
+
+      (filter.eq as jest.Mock).mockReturnValueOnce(eqResult);
+      (filter.notEq as jest.Mock).mockReturnValueOnce(notEqResult);
```
mofojed, line 1805: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+      const column = makeFilterColumn('java.lang.String');
+      const filter = column.filter();
+
+      const isNullResult = makeFilterCondition();
+      const notResult = makeFilterCondition();
+      const notInResult = makeFilterCondition();
+      const expectedResult = makeFilterCondition();
+
+      (filter.notIn as jest.Mock).mockReturnValueOnce(notInResult);
+      (filter.isNull as jest.Mock).mockReturnValueOnce(isNullResult);
```
mofojed, line 1842: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+    it('handles non-empty selected values with null values and invertSelection is false', () => {
+      const column = makeFilterColumn('boolean');
+      const filter = column.filter();
+
+      const isNullResult = makeFilterCondition();
+      const inResult = makeFilterCondition();
+      const expectedResult = makeFilterCondition();
+
+      (filter.isNull as jest.Mock).mockReturnValueOnce(isNullResult);
+      (filter.in as jest.Mock).mockReturnValueOnce(inResult);
```
mofojed, line 1861: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+    });
+
+    it('handles all null values as selected values and invertSelection true', () => {
+      const column = makeFilterColumn('boolean');
+      const filter = column.filter();
+
+      const isNullResult = makeFilterCondition();
+      const expectedResult = makeFilterCondition();
+
+      (filter.isNull as jest.Mock).mockReturnValueOnce(isNullResult);
```
mofojed, line 1877: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+      ).toBe(expectedResult);
+    });
+
+    it('handles all null values as selected values and invertSelection false', () => {
+      const column = makeFilterColumn('boolean');
+      const filter = column.filter();
+
+      const isNullResult = makeFilterCondition();
+
+      (filter.isNull as jest.Mock).mockReturnValueOnce(isNullResult);
```
mofojed, line 1890: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1354,6 +1791,134 @@ describe('quick filter tests', () => {
...
+      ).toBe(isNullResult);
+    });
+
+    it('handles non-empty selected values but no null values and invertSelection is true', () => {
+      const column = makeFilterColumn('int');
+      const filter = column.filter();
+
+      const expectedResult = makeFilterCondition();
+
+      (filter.notIn as jest.Mock).mockReturnValueOnce(expectedResult);
```
mofojed, line 1902: Unnecessary cast

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1535,3 +2100,592 @@ describe('Sorting', () => {
...
+    modelColumn: number,
+    direction: SortDirection,
+    isAbs: boolean,
+    addToExisting: boolean,
+    expectedValue
+  ) => {
+    expect(
+      TableUtils.sortColumn(
+        // @ts-expect-error test what happens when columns is null
+        sorts,
```
mofojed, line 2496: No need to test `null` case

mofojed (changes requested):

mofojed (approved):

PR #1118, opened 2023-02-28
feat: Improve text labels based on suggestions from chatGPT

dsmmcken:
I asked chatGPT to help me brainstorm improvements to some of the labels and hint text based on the Apple Human Interface Guidelines. I then edited them as human to improve them further.

Review and let me know if you think any are worse or weird.

### DIFF
--- a/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
+++ b/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
@@ -614,10 +614,9 @@ export class ColumnSpecificSectionContent extends PureComponent<
 
     return (
       <div className="app-settings-formatting-section" ref={this.containerRef}>
-        <div>Default formatting for matched column names</div>
         <div className="app-settings-menu-description mb-3">
-          Applies a formatting rule to all columns that match a specified name
-          and type.
+          Customize the formatting of table columns by applying rules based on
+          matching both a column name and column type.
         </div>
 
         <TransitionGroup
--- a/packages/code-studio/src/settings/SettingsMenu.tsx
+++ b/packages/code-studio/src/settings/SettingsMenu.tsx
@@ -143,7 +143,7 @@ export class SettingsMenu extends Component<
                   transform="grow-4"
                   className="mr-2"
                 />
-                Default Formatting &amp; Time zone
+                Default Format &amp; Time zone
               </>
             }
             onToggle={this.handleSectionToggle}
@@ -163,7 +163,7 @@ export class SettingsMenu extends Component<
                   transform="grow-4"
                   className="mr-2"
                 />
-                Column Specific Formatting Rules
+                Format by Column Name & Type
               </>
             }
             onToggle={this.handleSectionToggle}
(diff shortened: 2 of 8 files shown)


packages/code-studio/src/settings/SettingsMenu.tsx
```diff
@@ -163,7 +163,7 @@ export class SettingsMenu extends Component<
                   transform="grow-4"
                   className="mr-2"
                 />
-                Column Specific Formatting Rules
+                Format by Column Name & Type
```
mofojed, line 166: We use `&amp;` above, but only `&` here. Should probably be consistent?
```suggestion
                Format by Column Name &amp; Type
```

packages/iris-grid/src/sidebar/AdvancedSettingsMenu.tsx
```diff
@@ -34,9 +34,9 @@ AdvancedSettingsMenu.getAdvancedSettingsTitleForKey = (
 ): string => {
   switch (key) {
     case AdvancedSettingsType.FILTER_CONTROL_CHANGE_CLEARS_ALL_FILTERS:
-      return 'Control filters clear existing table filters';
+      return 'Clear current table filters before applying new filters from a control';
     case AdvancedSettingsType.LINK_CHANGE_CLEARS_ALL_FILTERS:
-      return 'Inbound links clear existing table filters';
+      return 'Clear current table filters before applying new filters from a incoming link filter';
```
mofojed, line 39: I'm not crazy about these wordings, they show up in the menu pretty long.

image

I think I somewhat prefer the older text, but up to you.

dsmmcken, line 39:
```suggestion
      return 'Clear current table filters before applying new filters from an incoming link filter';
```

dsmmcken: I think for advanced settings is probably fine to be more clear because it's a weird option.

mofojed (approved):

packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
```diff
@@ -614,10 +614,9 @@ export class ColumnSpecificSectionContent extends PureComponent<
 
     return (
       <div className="app-settings-formatting-section" ref={this.containerRef}>
-        <div>Default formatting for matched column names</div>
         <div className="app-settings-menu-description mb-3">
-          Applies a formatting rule to all columns that match a specified name
-          and type.
+          Customize the formatting of table columns by applying rules based on
+          matching both a column name and column type.
```
mattrunyon, line 619:
```suggestion
          Customize the formatting of table columns by applying rules based on
          name and type.
```

dsmmcken: That's shorter but not as clear that it matches column names and not table names, and that the requirement to match is on both, which "and" implies. But it could be mis-interpreted to mean you can have rules matching name and rules matching type.

We have space to be more clear in this case.

PR #1119, opened 2023-03-01
fix: Fix the style guide

mofojed:
- Styleguide index wasn't updated to use JS API as a module correctly

### DIFF
--- a/packages/code-studio/src/styleguide/StyleGuideRoot.tsx
+++ b/packages/code-studio/src/styleguide/StyleGuideRoot.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { Provider } from 'react-redux';
 import 'fira';
 import '@deephaven/components/scss/BaseStyleSheet.scss';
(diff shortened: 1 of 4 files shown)


packages/code-studio/src/styleguide/StyleGuideRoot.tsx
```diff
@@ -0,0 +1,31 @@
...
+  useEffect(() => {
+    unregister();
+    DownloadServiceWorkerUtils.registerOnLoaded();
+    MonacoUtils.init({ getWorker: () => new MonacoWorker() });
+
+    // disable annoying dnd-react warnings
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    window['__react-beautiful-dnd-disable-dev-warnings'] = true;
+  }, []);
```
mattrunyon, line 22: This probably doesn't have to be in a `useEffect` really since the root takes no params and has no state, so it shouldn't re-render

mattrunyon (dismissed):

packages/code-studio/src/styleguide/StyleGuideRoot.tsx
mofojed: True... should change all the other `*Root` files as well. I did put it in a `useEffect` as I think that's best practices (if you added a prop later or something), but at the same time - we want these to run before the child renders, and right now the `useEffect` fires _after_ this render returns. So I think more correct to do as you suggest.

PR #1120, opened 2023-03-01
fix: Added date time parsing for conditional formatting

### LINKED ISSUE #1108: Conditional formatting errors when setting an invalid date string
**Description**
Invalid DateTime strings cause conditional formatting errors that result in an unrecoverable state.

NOTE: #1091 and #1106 should fix related issues. Might be good to wait until those are merged before working on this fix.

**Steps to reproduce**
1. Load a table with a DateTime column
2. Table Options -> Conditional Formatting -> Add New Rule
3. Select date column, "is exactly", "blah" for condition
4. Set Style to something other than "No formatting"

**Expected results**
Invalid DateTime strings should not be applied

- Done button should be disabled
- Auto formatting should not be applied
- No error should be shown

**Actual results**
Formatting is auto-applied for invalid date format and a "Unable to open table. Error" is shown

**Versions**

Engine Version: 0.22.0
Web UI Version: 0.30.1
Java Version: 17.0.6
Barrage Version: 0.5.0

bmingles:
Added datetime string parsing / validation.

fixes #1108

### DIFF
--- a/packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
+++ b/packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
@@ -718,7 +718,10 @@ export function isDateConditionValid(condition: DateCondition, value?: string) {
       return true;
 
     default: {
-      const [dateTimeString, tzCode] = (value ?? '').split(' ');
+      const [dateTimeString, ...rest] = (value ?? '').split(' ');
+      // Reconstitute all tokens after the first ' ' in case the user included garbage data at the end
+      // e.g. '2020-01-01 NY blah'
+      const tzCode = rest.join(' ');
 
       try {
         DateUtils.parseDateTimeString(dateTimeString);
--- a/packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
+++ b/packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
@@ -718,19 +718,19 @@ export function isDateConditionValid(condition: DateCondition, value?: string) {
       return true;
 
     default: {
-      const [dt, tz] = (value ?? '').split(' ');
+      const [dateTimeString, tzCode] = (value ?? '').split(' ');
 
       try {
-        DateUtils.parseDateTimeString(dt);
+        DateUtils.parseDateTimeString(dateTimeString);
       } catch (e) {
-        log.debug('Invalid datetime string', dt);
+        log.debug('Invalid datetime string', dateTimeString);
         return false;
       }
 
       try {
-        dh.i18n.TimeZone.getTimeZone(tz);
+        dh.i18n.TimeZone.getTimeZone(tzCode);
       } catch (e) {
-        log.debug('Invalid timezone string', tz);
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
```diff
@@ -717,9 +717,25 @@ export function isDateConditionValid(condition: DateCondition, value?: string) {
     case DateCondition.IS_NOT_NULL:
       return true;
 
-    default:
-      // Proper date validation will be addressed by Issue #1108
-      return value != null && value !== '';
+    default: {
+      const [dt, tz] = (value ?? '').split(' ');
```
vbabich, line 721: One suggestion - use camel case and full words for variable names. We try to follow the Airbnb style guide – https://github.com/airbnb/javascript#naming-conventions
with a few exceptions. `e` as the `catch` block argument is fine, single-character names in loops and maps are also fine, as long as the loop body is fairly short.

In this case, these are local variables and the scope is short so it's probably fine, but I thought I'd mentioned this anyway.

vbabich (approved):

vbabich (dismissed):

packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
```diff
@@ -717,9 +717,25 @@ export function isDateConditionValid(condition: DateCondition, value?: string) {
     case DateCondition.IS_NOT_NULL:
       return true;
 
-    default:
-      // Proper date validation will be addressed by Issue #1108
-      return value != null && value !== '';
+    default: {
+      const [dateTimeString, tzCode] = (value ?? '').split(' ');
```
dsmmcken, line 721: Would `2023-02-23T11:46:31.000000000 NY FOO` pass validation but still fail server size?

vbabich: Good catch.
Also, a dateTime string can have a space in place of the t-separator: `2023-02-23 11:46:31.000000000 NY`.
Regex for the separator in `DateUtils/parseDateTimeString` looks like this: `[tT\s]`.

vbabich: Ignore my comment, looks like the engine expects a `T` in format conditions.

dsmmcken: It would be nice if we could accepted strings anywhere in the UI that takes a datetime with or without the T separator, then normalize it before sending.

bmingles:
@vbabich  It was suggested by @niloc132 that instead of using `convertDateTime`, we could use `autoEpochToNanos` in the Java expression that gets run by the engine. This would require us to parse the datetime in the UI first and then use the nanos value in the expression. This "should" address the issue we still have with the engine only supporting a subset of timezones.

I'm hesitant to do that for this issue given the current solution is already implemented, but I can create a new issue if we think this is a better solution.

vbabich:
@bmingles Agreed, open another issue for this.

vbabich (approved):

PR #1123, opened 2023-03-01
fix: Added conditional formatting to rollback

### LINKED ISSUE #1121: Conditional column formatting - Undo applied style if query engine error occurs
**Description**

If a conditional column condition + style are applied that results in a query engine error, the table gets into a non-recoverable state. 

**Steps to reproduce**

1. Create a ticking table with a DateTime column
```python

mylist = ["mars", "pluto", "saturn"]

    "Id=new Integer(i)",
    "Text=mylist[ (int) (3 * Math.random()) % 3 ]"
])
```

2 Apply a conditional style to a DateTime column with an `is` condition
3. Set the date value to "2020-01-01 America/New_York" (this can be parsed by UI but fails in query engine)

**Expected results**
The table should still be in a usable state. Ideally we can revert the last action on the undo stack or something comparable

**Actual results**
image

**Versions**

Engine Version: 0.22.0
Web UI Version: 0.30.1
Java Version: 17.0.6
Barrage Version: 0.5.0

bmingles:
Added conditional formatting props to rollback

fixes #1121

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -204,6 +204,9 @@ const UNFORMATTED_DATE_PATTERN = `yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS z`;
 function isEmptyConfig({
   advancedFilters,
   aggregationSettings,
+  conditionalFormatEditIndex,
+  conditionalFormatPreview,
+  conditionalFormats,
   customColumns,
   quickFilters,
   reverseType,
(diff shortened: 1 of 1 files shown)


bmingles:
Turns out there is an issue where the format string remains in the input field after the rollback. This could tell the user that their format string worked even though no formatting would be applied which starts becoming similar to showing incorrect data.

Given that fixing the date parsing "should" remove the known use case that can produce this issue + the lift involved to reset the input state, I'm going to close this.

PR #1125, opened 2023-03-02
fix: Fixed rollup divider position

### LINKED ISSUE #1062: Rollup divider is shown at the wrong column edge.
**Description**
Table is grouped by MyChar and MyString, however the border of the rollup is after mychar instead of myString
Wrong:
image

Grouping in the reverse order works correctly. Not sure the cause:
image

**Steps to reproduce**

```

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",
```

1. Create a table using the above code
2. From the table sidebar select rollup rows and group by "mychar" then "mystring"

**Expected results**
Border is after the last rolled up row.

**Actual results**
Border is shown in the incorrect location.

** NOTE **
Waiting for https://github.com/deephaven/deephaven-core/issues/3477 should simplify the work needed on this issue.

**Versions**
Engine Version: 0.21.0
Web UI Version: 0.26.0
Java Version: 11.0.16
Barrage Version: 0.5.0

bmingles:
fixes #1062

### DIFF
--- a/packages/iris-grid/src/IrisGridRenderer.ts
+++ b/packages/iris-grid/src/IrisGridRenderer.ts
@@ -187,14 +187,14 @@ class IrisGridRenderer extends GridRenderer {
       return;
     }
 
-    // The `groupedColumns` array contains 2 items for each grouped column name.
-    // The first half of the array contains 1 copy of each column in display order.
-    // The second half is not sorted in display order. To get the last displayed
-    // group column, we can use the last column in the first half of the array.
-    const modelIndex = groupedColumns.length / 2 - 1;
+    // There seem to be duplicate columns with the same name, so determine the
+    // count of distinct names.
+    const numUniqueNames = new Set(groupedColumns.map(({ name }) => name)).size;
 
-    const columnX = allColumnXs.get(modelIndex);
-    const columnWidth = allColumnWidths.get(modelIndex);
+    const lastVisibleGroupColumnIndex = numUniqueNames - 1;
+
+    const columnX = allColumnXs.get(lastVisibleGroupColumnIndex);
+    const columnWidth = allColumnWidths.get(lastVisibleGroupColumnIndex);
     if (columnX == null || columnWidth == null) {
       return;
     }
--- a/packages/iris-grid/src/IrisGridRenderer.ts
+++ b/packages/iris-grid/src/IrisGridRenderer.ts
@@ -178,7 +178,7 @@ class IrisGridRenderer extends GridRenderer {
     state: IrisGridRenderState
   ): void {
     const { metrics, model, theme } = state;
-    const { groupedColumns, columns } = model;
+    const { groupedColumns } = model;
     const { maxY, allColumnWidths, allColumnXs } = metrics;
     if (
       groupedColumns.length === 0 ||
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -187,12 +187,14 @@ class IrisGridRenderer extends GridRenderer {
...
 
-    const lastGroupedColumn = groupedColumns[groupedColumns.length - 1];
-    const modelIndex = columns.findIndex(
-      c => c.name === lastGroupedColumn.name
-    );
-    const columnX = allColumnXs.get(modelIndex);
-    const columnWidth = allColumnWidths.get(modelIndex);
+    // There seem to be duplicate columns with the same name, so determine the
+    // count of distinct names.
+    const numUniqueNames = new Set(groupedColumns.map(({ name }) => name)).size;
```
dsmmcken, line 192: I would wait for the fix from the JS-API making the de-dupe check unnecessary.

bmingles: @dsmmcken the JS-API has been fixed. I have updated my PR accordingly. I should be a little more resilient for any future changes since it doesn't assume any particular order of the grouped columns now.

packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -187,12 +187,10 @@ class IrisGridRenderer extends GridRenderer {
       return;
     }
 
-    const lastGroupedColumn = groupedColumns[groupedColumns.length - 1];
-    const modelIndex = columns.findIndex(
-      c => c.name === lastGroupedColumn.name
-    );
-    const columnX = allColumnXs.get(modelIndex);
-    const columnWidth = allColumnWidths.get(modelIndex);
+    const lastVisibleGroupColumnIndex = groupedColumns.length - 1;
```
mattrunyon, line 190: I would add a comment along the lines of "This assumes the engine puts the grouped columns at the start of the model, so model and visible index match for grouped columns. If we add freeze support for tree tables or allow moving the grouped columns, this may need to change to get the visible index from the model index" so anybody woh sees this code in the future knows we didn't get the visible from model on purpose

mattrunyon (changes requested):

packages/iris-grid/src/IrisGridRenderer.ts
dsmmcken: that sounds like its waiting to break @mofojed might have opinions.

mattrunyon: It was already broken before if we added support for either of the things I mentioned. `allColumnXs` is keyed on `VisibleIndex` and we were using `ModelIndex`

bmingles: @mattrunyon I've added a comment. It also occurs to me if we do ever support group columns not being the furthest to the left, we'll have some other UI decisions to make since the single vertical line of separation would no longer convey what it does now.

dsmmcken: Frozen columns get a double line and a background.

image

We should probably not crash the table when a frozen column and a rollup are combined, what behaviour happens though I am not sure yet.

bmingles: @dsmmcken  There's definitely some issues with combining frozen + grouping. A couple of example:

- freeze column then group. Frozen columns no longer highlighted and context menu says you can freeze it. If you remove the grouping, frozen column returns.
- Freeze a grouped column causes an error

That said, I don't think these are caused by this PR and likely will require some more testing + requirement definition of how we want to handle various scenarios of grouping + freeze + unfreeze

Any objections to this PR getting merged and us creating a new issue for the above?

mattrunyon: I'm fine with it. I was waiting to see if @mofojed had anything to say in regards to this, but the existing behavior would break if either of those features were implemented anyway, so I think it's fine.

Just for future reference, we typically click the re-request review button next to the reviewers when we're ready for a re-review. That makes the PR show up as "awaiting your review" in VSCode or when filtering the PR list by PRs waiting on your review

mattrunyon (approved):

mofojed (approved):

Issue #833, opened 2022-10-20
Dehydrated grid settings should be based on column names instead of indexes


vbabich:
**Description**

We should use column names instead of indexes in dehydrated grid settings. Otherwise, column shifts can result in grid hydration errors.

**Steps to reproduce**
1. Run this command to create a table:
```
from deephaven import empty_table
t = empty_table(10).update(formulas=["X = `A`"])
```
2. Right-click a value in the column `X`, select Filter By Value -> text is exactly
2. Open Panels menu, click Export Layout
3. Update the table, and shift the columns:
```
from deephaven import empty_table
t = empty_table(10).update(formulas=["I=i", "X = `A`"])
```
4. Open Panels menu, click Import Layout

**Expected results**

Layout restored, table `t` has the filter applied on string column `X`

**Actual results**

Grid hydration fails [will add more details].

mattrunyon:
Similar, but different issue is #643

Issue #643, opened 2022-07-07
Dehydrated grid state uses incorrect column names for moved columns


mattrunyon:
**Description**

The dehydrated grid state converts moved column indexes into moved column names. The idea was we would be ok if the column name order changed from the server when the workspace was applied to the table on a future load.

However, the actual dehydration algorithm does not account for moves when replacing indexes with column names.

**Steps to reproduce**
1. Create a table with columns `A, B, C`
2. Move A to the end (0 -> 2). Now order is `B, C, A`
3. Move C to the beginning (1 -> 0). Now order is `C, B, A`

The moved columns array should be `[{ from: 0, to: 2 }, { from: 1, to: 0 }]`

**Expected results**

Dehydrated grid state contains moved columns `[{ from: 'A', to: 'C' }, { from: 'C', to: 'B' }]`

**Actual results**

Dehydrated grid state contains moved columns `[{ from: 'A', to: 'C' }, { from: 'B', to: 'A' }]`

**Additional details and attachments**

This will require a workspace migration to not break existing workspaces. We could just opt to keep the visible indexes which is the intended use of the moved columns. We already store a copy of the column names, so if the server had a different order, we could do some reconciliation on hydrate.

**Versions**
 - Deephaven: 0.14

PR #1126, opened 2023-03-03
refactor: Replace usage of Column.index with column name

### LINKED ISSUE #965: Remove uses of JS API column index

bmingles:
resolves #965

BREAKING CHANGE: Removed index property from dh.types Column type. IrisGridUtils.dehydrateSort now returns column name instead of index. TableUtils now expects column name instead of index for functions that don't have access to a columns array.

### DIFF
--- a/packages/iris-grid/src/IrisGridUtils.test.ts
+++ b/packages/iris-grid/src/IrisGridUtils.test.ts
@@ -157,8 +157,8 @@ describe('sort exporting/importing', () => {
     const table = makeTable({ columns, sort });
     const exportedSort = IrisGridUtils.dehydrateSort(sort);
     expect(exportedSort).toEqual([
-      { column: 3, isAbs: false, direction: 'ASC' },
-      { column: 7, isAbs: true, direction: 'DESC' },
+      { columnName: columns[3].name, isAbs: false, direction: 'ASC' },
+      { columnName: columns[7].name, isAbs: true, direction: 'DESC' },
     ]);
 
     const importedSort = IrisGridUtils.hydrateSort(table.columns, exportedSort);
--- a/packages/iris-grid/src/IrisGridUtils.test.ts
+++ b/packages/iris-grid/src/IrisGridUtils.test.ts
@@ -5,7 +5,10 @@ import { DateUtils } from '@deephaven/jsapi-utils';
 import type { AdvancedFilter } from './CommonTypes';
 import { FilterData } from './IrisGrid';
 import IrisGridTestUtils from './IrisGridTestUtils';
-import IrisGridUtils, { DehydratedSort } from './IrisGridUtils';
+import IrisGridUtils, {
+  DehydratedSort,
+  LegacyDehydratedSort,
+} from './IrisGridUtils';
 
 function makeFilter() {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -17,7 +20,7 @@ function makeColumns(count = 30) {
 
   for (let i = 0; i < count; i += 1) {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const column = new (dh as any).Column({ index: i, name: `${i}` });
+    const column = new (dh as any).Column({ index: i, name: `name_${i}` });
     columns.push(column);
   }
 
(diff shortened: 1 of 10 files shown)


packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -790,11 +790,11 @@ class IrisGridUtils {
     return (
       sorts
         .map(sort => {
-          const { column: columnIndex, isAbs, direction } = sort;
+          const { columnName, isAbs, direction } = sort;
           if (direction === TableUtils.sortDirection.reverse) {
             return dh.Table.reverse();
           }
-          const column = IrisGridUtils.getColumn(columns, columnIndex);
+          const column = IrisGridUtils.getColumnByName(columns, columnName);
```
mofojed, line 797: This is correct, but we're going to need a migration strategy for data that was stored in the old format (testing against Enterprise). Can just check against the sort value to see if `columnName` is provided, or instead define the `DehydratedSort` type as `column: ModelIndex | ColumnName`, then have a typeof check and look up the column name if a `ModelIndex` is provided.

An example is how we handle `movedColumns`: https://github.com/deephaven/web-client-ui/blob/d6c8a988d62157abfb8daadbff5db3eaef21a247/packages/iris-grid/src/IrisGridUtils.ts#L236
@mattrunyon implemented that migration behaviour in https://github.com/deephaven/web-client-ui/pull/642/files (though it should be typed to a `ColumnName` instead of a `string`).

mofojed (changes requested):

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -98,7 +98,7 @@ export type DehydratedUserColumnWidth = [ColumnName, number];
 export type DehydratedUserRowHeight = [number, number];
 
 export type DehydratedSort = {
-  column: ModelIndex;
+  columnName: ColumnName;
```
bmingles, line 101: @mofojed Based on your comment, does this mean I should change this type to

`column: ColumnName | ModelIndex` ?

packages/iris-grid/src/IrisGridUtils.ts
mattrunyon: The `movedColumns` wasn't really a migration I'd say. It was adding behavior (support for moving ranges in 1 move), not changing or removing any. This issue removes using `ModelIndex` for the sort column.

We'll still need to make sure the `ModelIndex` order is saved since we rely on that and not names for many things still. I'm not sure if that is accurate, but this ticket does remove the direct uses of `column.index` from the API. We still have some reliance on the column order being maintained by the engine which is a bigger issue to deal with.

Related #833 and #643

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -98,7 +98,7 @@ export type DehydratedUserColumnWidth = [ColumnName, number];
 export type DehydratedUserRowHeight = [number, number];
 
 export type DehydratedSort = {
-  column: ModelIndex;
+  column: ColumnName | ModelIndex;
```
mattrunyon, line 101: I think we should make this `LegacyDehydratedSort` or something since moving forward we don't want `DehydratedSort` to use `ModelIndex` as the column. Then `hydrate` can take `LegacyDehydratedSort` and `dehydrate` return `DehydratedSort` without the `| ModelIndex`

packages/iris-grid/src/IrisGridUtils.ts
bmingles: I've updated the code to support index or column name when importing a sort, but exporting uses column name.

packages/iris-grid/src/IrisGridUtils.ts
bmingles: @mattrunyon I've added the `LegacyDehydratedSort` interface as suggested

packages/iris-grid/src/IrisGridUtils.test.ts
```diff
@@ -156,9 +156,9 @@ describe('sort exporting/importing', () => {
     const sort = [columns[3].sort(), columns[7].sort().abs().desc()];
     const table = makeTable({ columns, sort });
     const exportedSort = IrisGridUtils.dehydrateSort(sort);
-    expect(exportedSort).toEqual([
-      { column: 3, isAbs: false, direction: 'ASC' },
-      { column: 7, isAbs: true, direction: 'DESC' },
+    expect(exportedSort).toEqual<DehydratedSort[]>([
```
mofojed, line 159: We should also add a test that it re-hydrates a `LegacyDehydratedSort` as well.

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -1486,3 +1490,48 @@ describe('range operations', () => {
...
+
+  describe('getSortIndex', () => {
+    // sort, columnName, expected
+    const testCases: [Sort[], ColumnName, number | null][] = [
+      [sortList.empty, 'name_999', null],
+      [sortList.hasThree, 'non-existing', null],
+      [sortList.hasThree, 'name_999', 2],
+    ];
+
+    it.each(testCases)(
```
mofojed, line 1514: TIL. Didn't know there was a `it.each`. Could have used this in a bunch of other cases.

mofojed (changes requested):
We need to mark this as a Breaking Change, since the `TableUtils` API is changing to only take a string for column name.
To mark it as a Breaking change, add the `BREAKING CHANGE: ` footer to the PR description as described in the [Contributing guidelines](https://github.com/deephaven/web-client-ui/blob/main/CONTRIBUTING.md#creating-a-pull-request).

packages/jsapi-utils/src/TableUtils.test.ts
bmingles: @mofojed There's also a string literal syntax supported by `it.each` that can be used to inject test variables. I have rarely used it but worth checking out the docs. https://jestjs.io/docs/api#2-testeachtablename-fn-timeout

mofojed: That's pretty neat! Good to know.

mofojed (commented):
Still need to update the PR description with the BREAKING CHANGE footer. We need to either maintain API compatibility with things exported by our packages (in this case, the functions changed in `TableUtils`), or mark it breaking.

mofojed (approved):

PR #1129, opened 2023-03-03
fix: Aggregations should be available when creating a rollup

mofojed:
- When there is a rollup config, aggregations go through the rollup config
  - https://github.com/deephaven/web-client-ui/blob/a069543812b6c544957ebf664e0918e98a3affbf/packages/iris-grid/src/IrisGrid.tsx#L1288
- Allow aggregations to be accessed when using a rollup config


mattrunyon:
Is there a related ticket in web-client-ui or core?

mattrunyon (approved):

Issue #1077, opened 2023-02-08
Update to Vite 4


mattrunyon:
Vite 4 is out (4.1.1 latest). We should update to it. The new major items are a swc-backed plugin for react which makes build/dev faster, rollup 3, and Vite 4.1 turned off some unused rollup caching which helps reduce the amount of RAM needed for build. Migration should be pretty straight forward and the swc-react plugin should work for us

https://vitejs.dev/guide/migration.html

PR #1131, opened 2023-03-03
refactor: Remove support for "internal" plugins

mofojed:
- We never use this and it was causing a warning in vite


mofojed:
Vite warning may go away with an upgrade, check again after #1077

mattrunyon (approved):

PR #1138, opened 2023-03-07
chore: Action to automatically update deephaven-core when new web versions published

mofojed:
- Trigger the [update-web](https://github.com/deephaven/deephaven-core/blob/main/.github/workflows/update-web.yml) action on `deephaven-core` when publishing new web-client-ui packages is done
- Uses [workflow_dispatch REST API](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28) to trigger it
- Requires a Classic token with `public_repo` permissions to trigger the `deephaven-core` action
  - Token is named `DEEPHAVENINTERNAL_PUBLIC_TOKEN`
  - I've added a [comment on the fine-grained token discussion](https://github.com/community/community/discussions/36441#discussioncomment-5230317) to see if there's a fine-grained version of the token that will suffice in the future


devinrsmith (approved):
I've added secret `DEEPHAVENINTERNAL_PUBLIC_TOKEN`

.github/workflows/publish-packages.yml
```diff
@@ -10,17 +10,29 @@ jobs:
...
+        run: ./node_modules/.bin/lerna publish --no-verify-access from-package --yes
         env:
           NODE_AUTH_TOKEN: ${{ secrets.DEEPHAVENBOT_NPM_TOKEN }}
+      - name: Update deephaven-core
+        run: |
+          curl -L \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.DEEPHAVENINTERNAL_PUBLIC_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/deephaven/deephaven-core/actions/workflows/update-web.yml/dispatches --data '{"ref": "main"}'
```
mattrunyon, line 38: This works, but you could also use https://github.com/octokit/request-action or https://github.com/actions/github-script to do this and it _might_ be more robust to any changes in the github API. If using the general `github-script`, you can use https://octokit.github.io/rest.js/v19#actions-create-workflow-dispatch to create it (They provide the octokit rest API under `github.rest`)

mofojed: Yea, I was just following the [docs for the API](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event). It works and I tested it, didn't consider looking for another action since it was a one off.

Side note, I looked into the request-action tool a little bit, funny story when they had the v2 release :laughing: : https://github.com/semantic-release/semantic-release/issues/1373

PR #1142, opened 2023-03-07
fix: DH-12163 - Column grouping sidebar test failure fixes

mattrunyon:
- Does not select new groups when searching
- Shows validation error if you create a new group and blur without typing anything into the name field
- Prevents Firefox on Mac from getting stuck in a bad state where color pickers can't be opened. The bug occurs when multiple color inputs are on the screen at once and the user clicks between the without closing the previous picker

### DIFF
--- a/packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingGroup.tsx
+++ b/packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingGroup.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState, useCallback } from 'react';
 import classNames from 'classnames';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Button, ThemeExport } from '@deephaven/components';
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingGroup.tsx
```diff
@@ -171,43 +205,47 @@ export default function VisibilityOrderingGroup(
...
-            // Mostly for testing. Chrome seems to not give the hidden input focus
-            // Really would only affect screen readers
-            colorInputRef.current?.focus();
-          }}
+          /**
+           * Toggle to close the picker on Chrome
+           * Prevents Firefox on Mac from getting into a stuck state
+           * Does not close on Firefox b/c the picker stays open when the element is removed
+           */
+          onClick={() => setIsColorInputOpen(val => !val)}
```
mofojed, line 213: This is annoying on Chrome - if I click the colour picker, select a colour, the colour picker disappears... and then if I click the colour dropper to show it again I need to click twice. Not sure if there's a better way.

mattrunyon: Just changed it to click instead of focus. Try that.

With focus, clicking the button made it close on chrome. The behavior prior to this change was clicking always opened on chrome, so this shouldn't change any chrome behavior now. There's no consistent event when a color input is closed (no blur event)

mofojed (approved):

PR #1145, opened 2023-03-09
fix: Added smarter caching for command history fetching

bmingles:
[#325]


packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -24,6 +37,55 @@ export class PouchCommandHistoryTable
...
+  ): Promise<ViewportData<CommandHistoryStorageItem>> {
+    if (this.cache) {
+      log.debug('Fetching from cache', viewport, selector);
+    } else {
+      log.debug('Fetching from PouchDB', selector);
+
+      this.cache = await this.db
+        .find({
+          selector,
+          fields: ['id', 'name'],
```
mofojed, line 78: We're not setting a viewport anymore, just fetching it all at once? How large can this grow/what kind of stress testing did you do? At least deserves a comment here...
I'm hesitant this is unbounded. What happens when someone has a million items in their history? We should at least truncate it at a limit if we're going to do this.

mofojed (changes requested):
In the PR description (the first "comment" on the PR), include details on what you did to test this as well.
To link the PR correctly to the issue, use the syntax `Fixes #325` instead of `[#325]`. More details: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

packages/pouch-storage/src/PouchCommandHistoryTable.ts
dsmmcken: This might be more the tree to bark up for the unfiltered view.

https://pouchdb.com/guides/bulk-operations.html#please-use-alldocs

> Far too many developers overlook this valuable API, because they misunderstand it. When a developer says "my PouchDB app is slow!", it is usually because they are using the slow query() API when they should be using the fast allDocs() API.
> 
> For details on how to effectively use allDocs(), you are strongly recommended to read ["Pagination strategies with PouchDB"](http://pouchdb.com/2014/04/14/pagination-strategies-with-pouchdb.html). For 99% of your applications, you should be able to use allDocs() for all the pagination/sorting/searching functionality that you need.

bmingles: I committed a first pass at fixing the performance using allDocs for non-filtered queries. Also addressed some other issues where we were hitting the db multiple times a second. Tested with 20,000 items in storage. Very speedy when no search applied. A bit of a delay on initial search, but then scrolling is fine. Still need to do some more testing / tweaking.

packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -13,16 +29,131 @@ export class PouchCommandHistoryTable
...
+          limit: DISPLAY_COMMAND_HISTORY_LIMIT,
+          sort: [{ id: 'desc' }],
+        });
+      } else {
+        PouchCommandHistoryTable.cache[this.db.name] = this.db
+          .allDocs({
+            include_docs: true,
+            descending: true,
+            limit: DISPLAY_COMMAND_HISTORY_LIMIT + 1, // +1 for an extra row returned by allDocs
```
mofojed, line 105: So we're still adding to the storage every time with `.put()`, then just limiting the results... not sure I like that. We should either viewport it properly, or truncate if we exceed a limit at time of `put`.

mofojed (changes requested):
Please include details of what steps you did you test this in the PR description.

packages/pouch-storage/src/PouchCommandHistoryTable.ts
bmingles: @mofojed  My thought process of this approach was that we could potentially make the limit configurable in the future without losing any underlying history if a user wanted to increase the limit at the expense of slower performance.

Regarding your suggestions:
1. Setting a limit on how much we store would probably be the simplest solution, but someone would need to make a call on what that limit needs to be
2. For "viewport it properly", I think this would mean we need to define a large window of data to minimize the number of queries required during scrolling. e.g. we could request a few hundred extra rows on top and bottom but then only re-fetch after user has scrolled within a threshold close to the edge of the buffer. Unfortunately I think this will mean that scrolling for filtered data is going to suffer since it seems that any query on a large dataset is going to be slow.

mofojed: Having the limit configurable should really be our last option - we don't want to present unnecessary options to the user. I think I'll just pick an arbitrary number of 2500 commands and that should be plenty. 
We want persistence to be server side at some point anyway, so if we do `allDocs()` with a low threshold of 2000 and upper threshold of 2500 (e.g. once you write the 2500th record, it deletes the top 500 bringing it back down to 2000), then we should be good. Although, even looking into the PouchDB documentation, it doesn't actually delete records, it just adds a tombstone: https://pouchdb.com/api.html#delete_document

So I wonder if deleting even helps with performance, or if you need to compact afterwards... How are you currently testing this/with how many records? Still need to include the testing steps in the PR description.

Side note, read this on IndexedDB performance (from someone who used to work on PouchDB): https://nolanlawson.com/2021/08/22/speeding-up-indexeddb-reads-and-writes/

bmingles: I've tested with 20,000 records so far. I plan to add some more formal details to the top of the PR. Was waiting until I finalize the code

packages/pouch-storage/src/PouchCommandHistoryStorage.ts
```diff
@@ -25,7 +25,7 @@ export class PouchCommandHistoryStorage implements CommandHistoryStorage {
   }
 
   async getTable(language: string): Promise<PouchCommandHistoryTable> {
-    return new PouchCommandHistoryTable(language);
+    return this.getUpdateTable(language);
```
bmingles, line 28: Seems this will not do what I was hoping. I'll need to figure out a different way to disable db change events while bulk deleting.

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -138,22 +140,36 @@ export class PouchStorageTable<T extends StorageItem = StorageItem>
 
   private currentViewportData?: ViewportData<T>;
 
-  constructor(databaseName: string) {
-    this.db = new PouchDB<T & PouchStorageItem>(`${DB_PREFIX}${databaseName}`);
+  constructor(
+    databaseName: string,
+    dbOptions?: PouchDB.HttpAdapter.HttpAdapterConfiguration
```
mofojed, line 145: This isn't the correct overload that we're using - guessing you Ctrl+Click'ed through to the definition, and came across this constructor with `HttpAdapterConfiguration`, which is the last overload defined because it is referenced last in the [@types/pouchdb-browser](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/62061c318fa41cc80b099cabb6f8a5c8552da873/types/pouchdb-browser/index.d.ts#L10).

We don't use `adapter: 'http'` though, we're just using the default adapter ([IndexedDB in the browser](https://pouchdb.com/adapters.html)), which is the default overload (which in TS for some reason appears as 4/4 if you type `new PouchDB(...)` ... I wonder why it shows them in the reverse order of declaration).
image

In any case, this should just use `PouchDB.Configuration.DatabaseConfiguration`, where adapter isn't defined.
```suggestion
    dbOptions?: PouchDB.Configuration.DatabaseConfiguration
```

packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -3,26 +3,258 @@ import {
...
-          ]
-        : []
+  constructor(language: string, private cache: PouchCommandHistoryCache) {
+    super(`CommandHistoryStorage.${language}`, ({
+      // Optimizations to cut down on growing table size. These should be safe
+      // since we don't care about revision history for command history
+      // documents.
+      auto_compaction: true,
+      revs_limit: 1,
+    } as unknown) as PouchDB.HttpAdapter.HttpAdapterConfiguration);
```
mofojed, line 39: We shouldn't be using the `HttpAdapter`, see my other comment on `PouchStorageTable`. No need to cast here.
```suggestion
    super(`CommandHistoryStorage.${language}`, {
      // Optimizations to cut down on growing table size. These should be safe
      // since we don't care about revision history for command history
      // documents.
      auto_compaction: true,
      revs_limit: 1,
    });
```

packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -3,26 +3,258 @@ import {
...
+   * @param selector
+   * @param _sort
+   * @returns Promise to array of command history storage items
+   */
+  // Our current version of eslint + prettier doesn't always like the `override`
+  // keyword. Whenever we upgrade, we should annotate this function with `override`.
+  async fetchViewportData(
+    viewport: StorageTableViewport,
+    selector: PouchDB.Find.Selector,
+    _sort: PouchDBSort
```
mofojed, line 203: One issue with not using the sort at all - we need the reverse of the table in the console input. Right now if you press `Up` in the console, it starts from the first command in your history, not the last one you entered. Need to make sure we at least handle that case.

packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -3,26 +3,258 @@ import {
...
+   */
+  private async fetchData(
+    _selector: PouchDB.Find.Selector
+  ): Promise<
+    PouchDB.Find.FindResponse<CommandHistoryStorageItem & PouchStorageItem>
+  > {
+    // Wrapping this in a setTimeout so that it executes on next call stack.
+    // This is necessary to ensure `this.cache` has been initialized due to some
+    // nuances with property initilization with inherited classes.
+    return PromiseUtils.withTimeout(0, async () => {
```
mofojed, line 112: I don't think this needs to be wrapped in a timeout? What about cache isn't initialized at this point?

mofojed (changes requested):

packages/pouch-storage/src/PouchCommandHistoryTable.ts
```diff
@@ -3,26 +3,258 @@ import {
...
+    // Add this table instance to the table registry
+    if (!this.cache.tableRegistry.has(this.cacheKey)) {
+      this.cache.tableRegistry.set(this.cacheKey, new Set());
+    }
+
+    log.debug('Adding table to registry', this.cacheKey);
+    this.cache.tableRegistry.get(this.cacheKey)?.add(this);
+  }
+
+  private searchText?: string;
```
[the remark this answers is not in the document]
bmingles: This is actually the same issue mentioned above. The searchText = '' doesn't actually get initialized until after the parent class constructor has finished. This means any calls in the constructor that access inherited members don't get the initial value.

packages/pouch-storage/src/PouchCommandHistoryTable.ts
mofojed: Ahh yea that's right. Cool.

mofojed (approved):

mofojed (changes requested):
Sort issue needs to be fixed, don't worry about the timeout thing/changing constructor initialization order.

bmingles:
> Sort issue needs to be fixed, don't worry about the timeout thing/changing constructor initialization order.

Sounds good. I'll also take a look at moving the debounce like you mentioned

packages/console/src/command-history/CommandHistoryViewportUpdater.tsx
```diff
@@ -21,6 +22,8 @@ export type CommandHistoryViewportUpdaterProps = {
   onUpdate: ViewportUpdateCallback<CommandHistoryStorageItem>;
 };
 
+const SET_SEARCH_DEBOUNCE_MS = 500;
```
mofojed, line 25: Typically for user input/typing we use 150, or 250. 500 is a bit long. Should make things feel a bit snappier.
```suggestion
const SET_SEARCH_DEBOUNCE_MS = 150;
```

bmingles: Fixed

mofojed (approved):

PR #1149, opened 2023-03-14
feat: JS API reconnect

### LINKED ISSUE #1140: Support reconnecting to the server
As a user, if I momentarily lose connection to a running server (wifi/vpn drops, laptop lid closes), I would like to continue using the console/app without refreshing the page.

With https://github.com/deephaven/deephaven-core/pull/3502 merged, the `dh.IdeConnection.HACK_CONNECTION_FAILURE` is deprecated, and no longer needed, but still left in place until the web UI is updated to support disconnect/reconnect events.

Tables and figures are functional already, but the console component for the `dh.IdeConnection` does not hint to the user that the connection is lost and commands cannot be run, until the reconnect event occurs.

Note that in some cases the session might have expired or the server restarted and all sessions invalidated. In this case, the disconnect event will still occur, but the JS API cannot yet create a new session (see deephaven/deephaven-core#3501), so instead the `dh.CoreClient.EVENT_RECONNECT_AUTH_FAILED` event will be fired from the `dh.CoreClient` instance.

mofojed:
- Listen for `EVENT_DISCONNECT` and `EVENT_RECONNECT` on the connection, and display a "Reconnecting..." message in the console.
- No longer listen to the `HACK_CONNECTION_FAILURE` (it's been deprecated).
- Listen for the `SHUTDOWN` event and display a message after shutdown.
- Fixes #1140 
Testing steps:
1. Start up server with deephaven-core JS API reconnect changes: https://github.com/deephaven/deephaven-core/pull/3502
2. Use ngrok to start a tunnel to that port, e.g.: `ngrok http 10000`
3. Start up Web UI connecting to that tunnel, e.g.: `VITE_CORE_API_URL=http://acfc-23-233-0-34.ngrok.io/jsapi npm start`
4. Open the Web UI in Firefox, run some commands to make sure initial connection is fine.
5. Press Alt and from the File menu, select "Work Offline"
6. See it transition to a disconnected state. Disconnected message should appear and should not be able to enter new commands
7. Reconnect by deselecting "Work Offline" option from Step 5
8. See it transition to connected state. Should be able to run commands and have the results appear.
9. Kill the server (Ctrl+C). See it transition to a Shutdown state, app unloaded.


packages/dashboard-core-plugins/src/panels/ConsolePanel.test.tsx
```diff
@@ -2,49 +2,98 @@ import React from 'react';
...
+function makeSession(): IdeSession {
+  return ({
     addEventListener: jest.fn(),
     subscribeToFieldUpdates: jest.fn(() => () => null),
     removeEventListener: jest.fn(),
     getTable: jest.fn(),
     getObject: jest.fn(),
     runCode: jest.fn(),
-  };
+  } as unknown) as IdeSession;
```
mattrunyon, line 24: Would be better to just mock all of `IdeSession` if it's not a pain. Shouldn't really affect the tests, but if `IdeSession` changes shape we don't know about it in the tests unless they fail

mattrunyon (changes requested):
I left the window in the disconnected state for 2 minutes and when re-enabling network, the "Attempting to reconnect" stays visible. At some point, it attempts to reconnect, but the credentials are invalid. That info should be propagated to the user and prompt them to reload the page to reauthenticate

That error was an uncaught promise though, so might need to be propagated through the JS API in a cleaner way.

```
Uncaught (in promise) Error: Authentication details invalid
    onEnd http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-internal.js:1
    unary http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-internal.js:1
    rawOnEnd http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-internal.js:1
    rawOnEnd http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-internal.js:1
    onTransportEnd http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-internal.js:1
    $onMessage http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-core.js:19095
    handleEvent_2 http://8f2c-50-24-89-157.ngrok.io/jsapi/dh-core.js:19237
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.test.tsx
mofojed: It's a bit of a pain in the ass. class Table doesn't match the types, missing many methods, and doesn't see the static events (since our typings expect them to be properties instead of static). Cleaning up mock should happen after we start using the actual types published by JS API.
I did a bit of cleanup, but was going down a bit of a yak shaving exercise.

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -184,6 +184,13 @@ function AppInit(props: AppInitProps) {
         setError(`Server shutdown: ${detail ?? 'Unknown reason'}`);
         setDisconnectError(null);
       });
+      connection.addEventListener(
+        dh.CoreClient.EVENT_RECONNECT_AUTH_FAILED,
+        event => {
+          log.warn('Reconnect authentication failed');
+          setError('Reconnect authentication failed');
```
mattrunyon, line 191: @dsmmcken thoughts on a message for this? The scenario is the user was disconnected for longer than the auth token is valid, so when it reconnects the authentication fails. User should be prompted to reload since this just shows the window w/ a warning triangle and message.

Maybe "Authentication failed after reconnecting. Please reload the page to reauthenticate"

mattrunyon (changes requested):

packages/code-studio/src/main/AppInit.tsx
dsmmcken: Ok. Make it actionable. Add a refresh page button then.

dsmmcken (approved):
Approving based on the screenshots you sent.

PR #1150, opened 2023-03-14
refactor: Fix fast refresh invalidations

### LINKED ISSUE #727: Improve HMR with Vite
Switching from CRA to Vite also moves us of of Webpack. Webpack possibly does some heuristics to make HMR function in different cases where Vite does not HMR properly.

Vite does not HMR if there are items it cannot determine are side-effect free. Some items like shortcuts possibly break this with the current singleton implementation and should use a class instead.

Also, HOC may break fast refresh. I am not sure if our current redux setup breaks this since it uses HOC.

https://github.com/vitejs/vite/discussions/4577#discussioncomment-1161007
https://github.com/vitejs/vite/issues/6408

mattrunyon:
Fixes #727 as best we can w/o requiring major changes. HMR works best w/ functional components and that's too big of a change to switch everything to functional components just for HMR.

Added an eslint rule which will warn about things that will almost certainly invalidate HMR. Fixed the warnings it emitted.

Tested locally by changing some displayed text values in some panels and seeing if the page triggered a full reload. I didn't have any specific files/cases that triggered full reloads previously and it seems Vite 4 has made it better on its own. If we start running into cases.

Saving `GridRenderer` doesn't trigger a full page reload (didn't before either), but massively slows the page (also had this behavior prior to this change). The change eventually propagates and refreshes

We should keep an eye on https://github.com/vitejs/vite/issues/12062 which will likely also fix the slow HMR issues on some components. There seems to be duplication of modules in the update list and it can explode at times (like GridRenderer triggers 14 unique modules, but 20k updates consisting of just those 14)

BREAKING CHANGE:
Renamed `renderFileListItem` to `FileListItem`.
Renamed `RenderFileListItemProps` to `FileListItemProps`.
Removed exports for `ConsolePlugin.assertIsConsolePluginProps`, `GridPlugin.SUPPORTED_TYPES`, `FileList.getPathFromItem`, `FileList.DRAG_HOVER_TIMEOUT`, `FileList.getItemIcon`, `Grid.directionForKey`, `GotoRow.isIrisGridProxyModel`, and `Aggregations.SELECTABLE_OPTIONS`. These were all only being consumed within their own file and are not consumed in enterprise


packages/components/src/DateTimeInput.tsx
```diff
@@ -35,13 +36,6 @@ export function fixIncompleteValue(value: string): string {
...
 }
 
-export function addSeparators(value: string): string {
-  const dateTimeMillis = value.substring(0, 23);
-  const micros = value.substring(23, 26);
-  const nanos = value.substring(26);
-  return [dateTimeMillis, micros, nanos].filter(v => v !== '').join('\u200B');
-}
-
 const removeSeparators = (value: string) => value.replace(/\u200B/g, '');
```
mofojed, line 39: Mildly inconsistent - why'd we define this a `const` instead of a `function` ?

packages/file-explorer/src/FileList.tsx
```diff
@@ -66,21 +71,14 @@ export interface FileListProps {
...
-export const DEFAULT_ROW_HEIGHT = 26;
-
 // How long you need to hover over a directory before it expands
-export const DRAG_HOVER_TIMEOUT = 500;
+const DRAG_HOVER_TIMEOUT = 500;
 
 const ITEM_LIST_CLASS_NAME = 'item-list-scroll-pane';
 
-export const renderFileListItem = (
+export function RenderFileListItem(
```
mofojed, line 79: Should we just have this in it's own file as well? `FileListItem` ?

packages/file-explorer/src/FileListContainer.tsx
```diff
@@ -191,7 +191,7 @@ export function FileListContainer(props: FileListContainerProps): JSX.Element {
     (itemProps: FileListRenderItemProps): JSX.Element => {
       const { item } = itemProps;
       if (renameItem && renameItem.filename === item.filename) {
-        return renderFileListItem({
+        return RenderFileListItem({
```
mofojed, line 194: Any reason not to write this part like:
```
  const renderItem = useCallback(
    (itemProps: FileListRenderItemProps): JSX.Element => {
      const { item } = itemProps;
      if (renameItem && renameItem.filename === item.filename) {
        return (
          <RenderFileListItem {...itemProps}>
            <FileListItemEditor
              item={item}
              validate={validateRenameItem}
              onSubmit={handleRenameSubmit}
              onCancel={handleRenameCancel}
            />
          </RenderFileListItem>
        );
      }
      return <RenderFileListItem {...itemProps} />;
    },
    [handleRenameCancel, handleRenameSubmit, renameItem, validateRenameItem]
  );
```

mofojed (dismissed):

packages/file-explorer/src/FileList.tsx
mattrunyon: Probably. I'll move it. First pass was just minimal changes to stop the HMR invalidations

packages/components/src/DateTimeInput.tsx
mattrunyon: Idk that's just how it was. Probably because it was a single line lambda function so it's less code?

packages/file-explorer/src/FileListContainer.tsx
mattrunyon: Don't think so. I left it initially since I just did minimal required changes to stop the HMR invalidations. I updated to this. `renderFileItemList` was a functional component in disguise

mofojed (approved):

PR #1160, opened 2023-03-17
fix: Scrolling horizontally in Linker mode renders empty cells

### LINKED ISSUE #1146: Scrolling table horizontally in the Linker mode renders empty cells
**Description**
Scrolling a table horizontally with the Linker mode active renders empty rows and throws an exception in the browser console.

**Steps to reproduce**

1. Open a table
```
import static io.deephaven.csv.CsvTools.readCsv
source = readCsv("https://media.githubusercontent.com/media/deephaven/examples/main/MetricCentury/csv/metriccentury.csv")
```
2. Shrink the window horizontally to only show 3 or 4 columns
3. Press cmd+L to open Linker
4. Scroll the table using the trackpad swipe or mouse wheel, not the panel scroll bar.

**Expected results**

Column data visible

**Actual results**

Table renders empty cells
Exception in the browser console:
```
Uncaught Error: Value is null or undefined
    at k (Asserts.ts:5:28)
    at Eo.drawMouseColumnHover (IrisGridRenderer.ts:257:5)
    at Eo.drawGridBackground (GridRenderer.ts:580:12)
    at Eo.drawGrid (GridRenderer.ts:273:10)
    at Eo.drawGrid (IrisGridRenderer.ts:134:11)
    at Eo.drawCanvas (GridRenderer.ts:232:10)
    at Eo.drawCanvas (IrisGridRenderer.ts:125:11)
    at js.drawCanvas (Grid.tsx:1641:14)
    at js.updateCanvas (Grid.tsx:799:10)
    at Grid.tsx:789:12
```

**Versions**

 - Web UI Version: 0.31.1

emilyhuxng:
Fixes #1146 

Changed `assertNotNull` to `getOrThrow` and update `hoverSelectColumn` on scroll so the highlighted column will change on scroll.

### DIFF
--- a/packages/iris-grid/src/IrisGridRenderer.ts
+++ b/packages/iris-grid/src/IrisGridRenderer.ts
@@ -255,9 +255,8 @@ class IrisGridRenderer extends GridRenderer {
 
     const { allColumnWidths, allColumnXs, maxY } = metrics;
 
-    const x = allColumnXs.get(hoverSelectColumn);
-    const columnWidth = allColumnWidths.get(hoverSelectColumn);
-    if (x == null || columnWidth == null) return;
+    const x = getOrThrow(allColumnXs, hoverSelectColumn);
+    const columnWidth = getOrThrow(allColumnWidths, hoverSelectColumn);
 
     context.fillStyle = theme.linkerColumnHoverBackgroundColor;
     context.fillRect(x, 0, columnWidth, maxY);
--- a/packages/iris-grid/src/IrisGridRenderer.ts
+++ b/packages/iris-grid/src/IrisGridRenderer.ts
@@ -257,8 +257,7 @@ class IrisGridRenderer extends GridRenderer {
 
     const x = allColumnXs.get(hoverSelectColumn);
     const columnWidth = allColumnWidths.get(hoverSelectColumn);
-    assertNotNull(x);
-    assertNotNull(columnWidth);
+    if (x == null || columnWidth == null) return;
 
     context.fillStyle = theme.linkerColumnHoverBackgroundColor;
     context.fillRect(x, 0, columnWidth, maxY);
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -257,8 +257,7 @@ class IrisGridRenderer extends GridRenderer {
 
     const x = allColumnXs.get(hoverSelectColumn);
     const columnWidth = allColumnWidths.get(hoverSelectColumn);
-    assertNotNull(x);
-    assertNotNull(columnWidth);
+    if (x == null || columnWidth == null) return;
```
mofojed, line 260: I'm concerned how we got into this state. The reason there were asserts there is it should _never_ be the case that we have a `hoverSelectColumn` but no matching columnX/columnWidth in the metrics. How are we getting into this state?

mofojed (changes requested):
Need to update the PR description to link it to the ticket it fixes (#1146).
I'd also like an investigation how we're getting into this state, as it indicates there are possibly other issues.

packages/iris-grid/src/IrisGridRenderer.ts
emilyhuxng: `onWheel` in `IrisGridColumnSelectMouseHandler` doesn't update `hoverSelectColumn` so if you were hovering column 0 for example, as you scroll, `hoverSelectColumn` stays `0` but eventually `allColumnXs` and `allColumnWidths` will not have `0` as a key anymore as column 0 isn't in view. Should `hoverSelectColumn` be updated in `onWheel` as well? This would result in it highlighting while the user is scrolling rather than needing to move cursor afterwards.

mofojed: Yes - we're updating the cursor, we should update the hover column as well.

mofojed (changes requested):

mofojed (approved):
PR description will need to be updated before merging (it no longer matches the actual change).

Issue #1157, opened 2023-03-16
DH-14484: Go to value displays Equals, acts like Contains


mofojed:
Reported on Enterprise DH-14484. 

in code-studio, execute the simple query;

```
t = emptyTable(1000000).update("A=`a`+ii", "Int=ii")
```

open the “Go to value” bar. do a search for column “A” “Contains” “a2”. change the search to column “Int”, “5000”, verify that you are now shown the proper row.

change the search back to column “A”; NOTE that the search-type now says “Equals”, and the value is “a5000”. search for the “next” value (down arrow or enter in the search-box); you are now brought to a row where “A” does not “Equals” “a5000”. though “Equals” is displayed, it is acting as “Contains”, which is what we had originally selected (before searching for “Int”)

PR #1162, opened 2023-03-20
fix: Goto Value Skips Rows on String Column, Displays Incorrect Filter, and `shift+enter` Doesn't go to Previous

### LINKED ISSUE #1156: DH-14483 Go to value skips valid rows
Referencing JIRA DH-14483

in code-studio, execute the simple query;

```
t = emptyTable(1000000).update("A=`a`+ii", "Int=ii")
```

open the “Go to value” bar. do a search for “A” “Contains” “a2”. while still in the “a2” box, press enter. we would expect the following rows to be selected for each “enter”;

```
a2
a20
a21
a22
a23
a24
... etc
```

but it seems that valid matching rows are being skipped

Zhou-Ziheng:
fixes #1156, #1157


mofojed (changes requested):

packages/components/src/MaskedInput.tsx
```diff
@@ -386,7 +389,9 @@ const MaskedInput = React.forwardRef<HTMLInputElement, MaskedInputProps>(
         );
         return;
       }
-
+      if (key === 'Enter') {
+        onSubmit?.(event);
```
mofojed, line 393: Need `return` after this

mofojed (changes requested):

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -100,11 +102,11 @@ function GotoRow({
     }
   };
 
-  const handleGotoValueKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+  const handleGotoValueKeySubmit = (e: KeyboardEvent<HTMLInputElement>) => {
```
mofojed, line 105: Could use `useCallback` here

mofojed (approved):

PR #1170, opened 2023-03-22
fix: Handling no columns

### LINKED ISSUE #1169: IrisGrid throws error if given table with no columns
**Description**

IrisGrid throws a .js error if given a table with no columns.

This is the underlying cause of
https://deephaven.atlassian.net/browse/DH-14436

**Steps to reproduce**
Create a table with no columns

```python
x = empty_table(10)
```
**Expected results**
Table panel should load without error but show no columns or rows

**Actual results**
image

**Versions**

Engine Version: 0.22.0
Web UI Version: 0.31.1
Java Version: 17.0.6
Barrage Version: 0.5.0

bmingles:
Added graceful handling to IrisGrid for tables with no columns.

fixes #1169

### DIFF
--- a/packages/iris-grid/src/IrisGrid.test.tsx
+++ b/packages/iris-grid/src/IrisGrid.test.tsx
@@ -10,7 +10,7 @@ class MockPath2D {
   addPath = jest.fn();
 }
 
-window.Path2D = MockPath2D;
+window.Path2D = (MockPath2D as unknown) as new () => Path2D;
 
 const VIEW_SIZE = 5000;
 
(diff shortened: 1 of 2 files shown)


mofojed (approved):

dsmmcken:
Are there other places we need to check for columns[0] being used in unsupported ways within iris grid?

bmingles:
> Are there other places we need to check for columns[0] being used in unsupported ways within iris grid?
@dsmmcken In the IrisGrid component I only see 1 other reference to `partitionTable.columns[0]`. The block containing this code is already in a try / catch.

I do see other `iris-grid` related components that have `columns[0]` references. It may be worth looking into these, but I think it will expand the scope + testing overhead of this issue.

dsmmcken:
That's fine. Should probably make a follow-up ticket and audit iris grid to see if there are behaviours expecting columns to not be zero where we might also fail.

bmingles:
I've created #1172 to investigate whether there are additional issues.

PR #1173, opened 2023-03-22
feat: Display workerName and processInfoId in the console status bar

vbabich:
Get optional workerName and processInfoId along with the login options response from the DHE host, display in the console status bar.

### DIFF
--- a/packages/code-studio/src/main/SessionUtils.ts
+++ b/packages/code-studio/src/main/SessionUtils.ts
@@ -1,9 +1,9 @@
-import { SessionWrapper } from '@deephaven/dashboard-core-plugins';
-import dh, {
-  CoreClient,
-  IdeConnection,
-  LoginOptions,
-} from '@deephaven/jsapi-shim';
+import {
+  LoginOptionsResponse,
+  SessionDetails,
+  SessionWrapper,
+} from '@deephaven/dashboard-core-plugins';
+import dh, { CoreClient, IdeConnection } from '@deephaven/jsapi-shim';
 import Log from '@deephaven/log';
 import shortid from 'shortid';
 import NoConsolesError from './NoConsolesError';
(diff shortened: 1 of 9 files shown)


packages/code-studio/src/main/SessionUtils.ts
```diff
@@ -83,15 +92,15 @@ export function createCoreClient(): CoreClient {
...
+export async function requestParentLoginOptions(): Promise<LoginOptionsResponse> {
   if (window.opener == null) {
     throw new Error('window.opener is null, unable to send auth request.');
   }
   return new Promise(resolve => {
     const listener = (
       event: MessageEvent<{
         message: string;
-        payload: LoginOptions;
+        payload: LoginOptionsResponse;
```
mofojed, line 103: I'm not a big fan of overloading the `loginOptions` directly... and then passing those wherever `loginOptions` are used or wherever we need this info. They're not really related...

Could we instead have another function `requestSessionDetailsFromParent` and have the parent respond to that instead?

packages/jsapi-utils/src/MessageUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+    throw new Error('window.opener is null, unable to send request.');
+  }
+  return new Promise((resolve, reject) => {
+    let timeoutId: NodeJS.Timeout;
+    const id = shortid();
+    const listener = (event: MessageEvent<Message<T>>) => {
+      const { data } = event;
+      log.debug('Received message', data);
+      // TODO: checking just the id should be enough, can probably drop the response message check?
+      if (data?.id !== id || data?.message !== response) {
```
mofojed, line 49: Yea I think you're right, if we just look at the `id`, we don't need to bother looking at `response`, and then the signature can just be `requestParentResponse(request, timeout)`... 

Thinking about this a bit further, there's basically three types of messages we can send:
- `request`: Something we expect a `response` for
- `response`: Responding to a `request`
- `notify`: Sending a message we do not expect a response for
Perhaps we should do that instead... or even just have those as the three different message types... `io.deephaven.message.Request`, `io.deephaven.message.Response`, and `io.deephaven.message.Notify`, and then the payload itself has the details...

I'm just trying to think how we'll want to extend this later for multi-window messaging/embedding in iframes. Let's discuss later today.

mofojed (dismissed):

vbabich:
Resolved conflict in `ConsolePanel.test.tsx` and rebased.

PR #1175, opened 2023-03-22
chore: Update readme and add distTag to lerna config

mattrunyon:
I added `distTag` for `latest` builds from `main`, and while updating the `README` I felt it could use some organizing. Tried to cleanup and group things that belonged together. Also added info about the launch config and moved it to the workspace settings in case users need to modify the default without worrying about it being committed. I tried adding a `.vscode/launch.json` and it is ignored now.


README.md
```diff
@@ -40,19 +58,13 @@ If your DHC address is different from the default `http://localhost:10000`, edit
...
-- [@deephaven/components](./packages/components): Component library used within the web UI.
-- [@deephaven/grid](./packages/grid): High-performance grid component used to display large tables of data.
-- [@deephaven/dashboard](./packages/dashboard/): Dashboards used in [@deephaven/code-studio](./packages/code-studio) for displaying and organizing panels.
-- [@deephaven/golden-layout](./packages/golden-layout): Layout framework used in [@deephaven/dashboard](./packages/dashboard/).
+We have a pre-defined launch config that lets you set breakpoints directly in VSCode for debugging browser code. The `Launch Code Studio` config will launch a new Chrome window that stores its data in your repo workspace. With this setup, you only need to install the React and Redux devtool extensions once. They will persist to future launches using the launch config.
```
mofojed, line 63:
```suggestion
We have a pre-defined launch config that lets you set breakpoints directly in VSCode for debugging browser code. The `Launch Deephaven` config will launch a new Chrome window that stores its data in your repo workspace. With this setup, you only need to install the React and Redux devtool extensions once. They will persist to future launches using the launch config.
```

README.md
```diff
@@ -40,19 +58,13 @@ If your DHC address is different from the default `http://localhost:10000`, edit
...
 
-## Contributing
+If you are not using Chrome (e.g. Chromium on Linux), then you add a new configuration to VSCode and copy the launch config from [`settings.json`](./.vscode/settings.json). Then add the `runtimeExecutable` prop to point to your browser executable. VSCode unfortnuately does not merge workspace `settings.launch` with workspace `launch.json`, so if we add more launch configs you would need to copy to your `.vscode/launch.json` file to get the configs.
```
mofojed, line 65: For whatever reason this doesn't work for me (it also didn't before). I'm going to worry about it after I update my OS though (been meaning to do that for a while).

mofojed (changes requested):
Minor typo

README.md
mattrunyon: Ya I don't have a Linux box to test on, so if you get around to figuring it out and there's more needed, then update this later

PR #1176, opened 2023-03-24
refactor: Update types for TablePlugins

mofojed:
- We were just defining a bunch of unknowns, which is bad practice
- Instead define actual types, so makers of TablePlugins can use those types effectively
- Add some details to the AuthPlugins readme
- Did this cleanup while working on https://github.com/deephaven/deephaven-js-plugins/issues/12
- Tested using the `table-example` plugin: https://github.com/deephaven/deephaven-js-plugins/tree/main/plugins/table-example
  - Used the following snippet:
```
from deephaven.jcompat import j_hashmap
t = empty_table(5).update(["x=i"])
t2 = t.j_table.withAttributes(j_hashmap({ "PluginName": "table-example" }))
```
  - Also tested error message is shown if unable to load the table plugin:
image


packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -151,11 +159,74 @@ export interface IrisGridPanelProps {
...
+   */
+  onStateChange: (pluginState: PanelState['pluginState']) => void;
+
+  /**
+   * Current plugin state. Use to load.
+   */
+  pluginState: PanelState['pluginState'];
+}
+
+export type TablePlugin = React.ForwardRefExoticComponent<
```
mattrunyon, line 226: These types should be in their own file instead of in `IrisGridPanel`

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -151,11 +159,74 @@ export interface IrisGridPanelProps {
...
+
+  /**
+   * Current user information
+   */
+  user: User;
+
+  /**
+   * Current user workspace data
+   */
+  workspace: Workspace;
```
mattrunyon, line 211: Do we want to provide these to the plugin or should it be up to the plugin to fetch these from a context? I guess the workspace is in redux, but the user has a context

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -250,6 +251,18 @@ export type FilterMap = Map<
     filterList: FilterData[];
   }
 >;
+
+export interface ContextMenuData {
```
mattrunyon, line 255: Maybe something like `IrisGridContextMenuData` so it's more specific?

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -726,7 +727,7 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
 
       if (column != null) {
         actions.push(
-          onContextMenu({
+          ...onContextMenu({
```
mattrunyon, line 730: Was this a bug from incorrect types? Just making sure it's intentional since it was not spread before

mattrunyon (changes requested):
Should this be `refactor:` instead of `feat:`? Or something else?

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
mofojed: Yea it's already passed in as a prop. Makes sense to deprecate and recommend just pulling it from redux though... and since it wasn't _really_ documented before anyway, I'll just not doc it.

packages/iris-grid/src/IrisGrid.tsx
mofojed: Yea, by virtue of being in the `@deephaven/iris-grid` package, it's already specific to IrisGrid... the `IrisGrid` prefix for all the mouse handlers was essentially redundant and unnecessary.
I'll change this though to be consistent with the redundancy we already have.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
mofojed: Yup

packages/auth-plugins/README.md
```diff
@@ -8,6 +8,120 @@ Authentication plugins for Deephaven. Used by [AuthBootstrap](../app-utils/src/c
...
+  participant S as Server
+  participant P as AuthPlugin
+  participant J as JS API
+  U->>W: Open app
+  W->>S: Load plugin modules
+  S-->>W: PluginModule[]
+  W->>P: Login
+  P->>J: client.login()
+  J-->>P: Login success
+  P-->>W: Login success
```
mattrunyon, line 30: Maybe add something like this for denoting how an auth plugin is chosen? I'm not sure. Kind of tricky to chart some of this

Maybe call AuthPlugin something like Selected AuthPlugin or Supported AuthPlugin? First Supported AuthPlugin?

```
W->>S: get auth types
S-->>W: AuthTypes[]
```

packages/auth-plugins/README.md
```diff
@@ -8,6 +8,120 @@ Authentication plugins for Deephaven. Used by [AuthBootstrap](../app-utils/src/c
...
+sequenceDiagram
+  participant U as User
+  participant W as Web UI
+  participant S as Server
+  participant P as Auth0Plugin
+  participant T as Auth0 Tenant
+  participant J as JS API
+  U->>W: Open app
+  W->>S: Load plugin modules
+  S-->>W: PluginModule[]
```
mattrunyon, line 108: Might not need both pairs of these since they're documented above?

packages/auth-plugins/README.md
```diff
@@ -8,6 +8,120 @@ Authentication plugins for Deephaven. Used by [AuthBootstrap](../app-utils/src/c
...
+  T->>W: Authorization code
+  W->>S: Load plugin modules
+  S-->>W: PluginModule[]
+  W->>P: Login
+  P->>T: Authorization Code + Client ID + Client Secret to /oauth/token
+  T->>T: Validate Authorization Code + Client ID + Client Secret
+  T-->>P: ID Token and Access Token
+  P->>J: client.login(token)
+  J-->>P: Login success
+  P-->>W: Login success
```
mattrunyon, line 122: I think this should be more like this? Auth plugin gives the info back to the Web UI? At least for keycloak we don't call `client.login` from the plugin, it's called from the auth plugin base which is part of the app

```suggestion
  T-->>P: ID Token and Access Token
  P-->>W: Access Token
  W->>J: client.login(token)
  J-->>W: Login success
```

mattrunyon (changes requested):

packages/auth-plugins/README.md
mofojed: Wait no - `AuthPluginBase` is used by `AuthPluginKeycloak`. It is still the responsibility of the `AuthPlugin` to call `client.login`, as we do in `AuthPluginPsk`. AuthPlugins don't _have_ to use the `AuthPluginBase`, it just made some things easier for a few plugins that just needed to provide credentials and a loading screen, no UI for user input.

packages/auth-plugins/README.md
mofojed: I'll shorten though to just "select auth plugin".

mattrunyon (approved):

PR #1178, opened 2023-03-27
feat: Add signatureHelp and hover providers to monaco

mattrunyon:
Corresponds with https://github.com/deephaven/deephaven-core/pull/3607

This adds signatureHelp and hover providers if the JS API supports the corresponding requests

SignatureHelp is provided when you type a method and then type the `(` character. There's also probably a keybinding in monaco to trigger it, but I'm not sure what it is.

Hover provided when you hover anything

Needs to use [my core branch](https://github.com/mattrunyon/deephaven-core/tree/lsp-improvements) to test. After pulling that branch, will need to build/install the Python server using the following commands

```
./gradlew :py-server:assemble :py-embedded-server:assemble
pip install --force-reinstall py/server/build/wheel/deephaven_core-0.23.0-py3-none-any.whl py/embedded-server/build/wheel/deephaven_server-0.23.0-py3-none-any.whl
```

Then start using one of these commands. Groovy provided just to show it doesn't break anything in Groovy. There is no autocomplete in Python notebooks if you are using a Groovy worker.

```
./gradlew server-jetty-app:run # Python
START_OPTS="-Ddeephaven.console.type=groovy" ./gradlew server-jetty-app:run # Groovy
```


packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * Registers a completion provider with monaco for the language and session provided.
+ */
+class MonacoProviders extends PureComponent<
+  MonacoProviderProps,
+  Record<string, never>
+> {
+  /**
+   * Converts LSP CompletionItemKind to Monaco CompletionItemKind
```
mofojed, line 25: Can you add a comment linking to the spec? https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind

I'm guessing there's no LSP constants we can use from somewhere?

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+      case 22:
+        return monacoKinds.Struct;
+      case 23:
+        return monacoKinds.Event;
+      case 24:
+        return monacoKinds.Operator;
+      case 25:
+        return monacoKinds.TypeParameter;
+      default:
+        return monacoKinds.Variable;
```
mofojed, line 83: Shouldn't we throw here?

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+    this.handleCompletionRequest = this.handleCompletionRequest.bind(this);
+    this.handleSignatureRequest = this.handleSignatureRequest.bind(this);
+    this.handleHoverRequest = this.handleHoverRequest.bind(this);
+  }
+
+  componentDidMount(): void {
+    const { language } = this.props;
+
+    this.registeredCompletionProvider = monaco.languages.registerCompletionItemProvider(
+      language ?? '',
```
mofojed, line 99: Shouldn't need these null coalescors, `language` is guaranteed to be set

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+            sortText,
+            filterText,
+            textEdit,
+            insertTextFormat,
+          } = item;
+
+          const { start, end } = textEdit.range;
+
+          // Monaco expects the columns/ranges to start at 1.
+          const range = {
```
mofojed, line 180: Could probably put this in a `lspToMonacoRange` method, similar to `lspToMonacoKind` to clean it up a little bit.

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+      })
+      .catch((error: unknown) => {
+        log.error('There was an error retrieving completion items', error);
+        return defaultResult;
+      });
+
+    return monacoSignatures;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
mofojed, line 300: Don't need to disable eslint here

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+      })
+      .catch((error: unknown) => {
+        log.error('There was an error retrieving completion items', error);
+        return { suggestions: [] };
+      });
+
+    return monacoCompletionItems;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
mofojed, line 219: Don't need to disable eslint here

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -0,0 +1,346 @@
...
+      return null;
+    }
+
+    const params = {
+      textDocument: {
+        uri: `${model.uri}`,
+        version: model.getVersionId(),
+      },
+      // Monaco 1-indexes Position. LSP 0-indexes Position
+      position: {
```
mofojed, line 316: Perhaps `lspToMonacoPosition` method as well.

packages/console/src/monaco/MonacoProviders.test.tsx
mofojed, line 1: Should also include cases where `getHover` and `getSignatureHelp` methods are not implemented to make sure it handles those correctly, and also where `documentation` is just provided as a `string` (as that's what happens on Enterprise still).

mofojed (changes requested):

dsmmcken:
> There's also probably a keybinding in monaco to trigger it, but I'm not sure what it is.

https://code.visualstudio.com/shortcuts/keyboard-shortcuts-macos.pdf

I wonder if it would be nice to expose these in a non-editable fashion in the settings menu or something?
https://github.com/microsoft/monaco-editor/issues/823#issuecomment-470754000

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: I opted not to since kind is optional from the LSP spec. We don't actually provide a kind from groovy completion items. Monaco will show something and I believe it was Variable by default, but I'd have to double check

I'll adjust the param to be `number | undefined`

mattrunyon:
I need to test this with enterprise still to ensure nothing breaks. I don't think anything will, but I might need to add back quotes in the completion item triggers. We trigger completion in `db.t("`, but at one point I had some weird completion items in DHC when triggered on quote. Mostly paths related to where Python was on the server

packages/console/src/monaco/MonacoProviders.test.tsx
mattrunyon: Enterprise doesn't provide anything for documentation on completion items and community didn't before this update

I've added a test for if the documentation is marked as plaintext and markdown. The existing test didn't set documentation

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -84,6 +87,40 @@ class MonacoProviders extends PureComponent<
...
+  }
+
+  /**
+   * Converts a monaco position to an LSP position
+   * Accounts for LSP indexing from 0 and monaco indexing from 1
+   *
+   * @param position The monaco position
+   * @returns The corresponding LSP position
+   */
+  static monacoToLspPosition(position: monaco.IPosition): Position {
```
mofojed, line 116: Would be fairly easy to add tests for these, or add a test that verifies the positions in the provider.

mofojed (approved):

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -93,30 +130,34 @@ class MonacoProviders extends PureComponent<
...
     );
 
-    this.registeredSignatureProvider = monaco.languages.registerSignatureHelpProvider(
-      language ?? '',
-      {
-        provideSignatureHelp: this.handleSignatureRequest,
-        signatureHelpTriggerCharacters: ['(', ','],
-      }
-    );
+    if (session.getSignatureHelp) {
```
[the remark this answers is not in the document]
mattrunyon: The default rules allow it because function does not have a falsy context. The rule defaults are to avoid `if (str)` if str is nullable since that condition covers both null and empty string

If you want to open a ticket to change it, the scenarios that are allowed by default are `allowString`, `allowNumber`, and `allowNullableObject`

https://typescript-eslint.io/rules/strict-boolean-expressions/#options

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: Added tests for the range/position transforms

mofojed (approved):

PR #409, opened 2022-01-31
Conditional formatting

vbabich:
Add Conditional Formatting menu to the Table Options sidebar.

PR #1181, opened 2023-03-29
fix: Conditional formatting not being applied to custom columns

### LINKED ISSUE #1135: Conditional formatting not being applied to custom columns
**Steps to reproduce**

1. Take any or the one given below, and add a new column to it using the custom columns UI feature in the table sidebar. For example a new column named "Test" equal to "MyInt * MyLong"
2. Apply a conditional format using the table sidebar to that new column. For example "Test", "Is greater than" "0" and a color.

```

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",
```        
      

**Expected results**

Color is applied to the column

**Actual results**

Nothing happens. No errors or anything in the console log, just doesn't apply.

**Versions**

Engine Version: 0.22.0
Web UI Version: 0.30.0
Java Version: 17.0.6
Barrage Version: 0.5.0

emilyhuxng:
Fixes #1135 

Removed `getCachedModelColumns` and uses `model.columns` instead.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1233,13 +1233,6 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       getFormatColumns(columns, rules)
   );
 
-  // customColumns arg is needed to invalidate the cache
-  // eslint-disable-next-line no-unused-vars
-  getCachedModelColumns = memoize(
-    (model: IrisGridModel, customColumns: readonly ColumnName[]) =>
-      model.columns
-  );
-
   /**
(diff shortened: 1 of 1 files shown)


mofojed (approved):
Just adding a note to my previous comment from @vbabich 's change for Conditional Formatting: https://github.com/deephaven/web-client-ui/pull/409/files#r812346032
It appears this render loop is no longer the case - I tried adding/removing custom columns and conditional formatting, and didn't get stuck in a render loop (kept Profiler open to see when redraws occurred).
I think this issue was resolved with the changes for #986 or #1013.

PR #1183, opened 2023-03-29
fix: Preview did not draw correctly when dragging Grids

### LINKED ISSUE #1112: drag proxy preview is broken
**Description**

Dragging a panel used to show a preview of the item while dragging. It should re-render the canvas. It looks like it's only rendering a black square, but for some reason also the selection box? Note this has been broken since bard maybe

image

mofojed:
- When we drag something, there is a frame where the width/height is 0
- We didn't update the metrics within the actual animation frame, so it was drawing the wrong thing
- Fix so metrics are always updated in the animation frame, and also to only check the metrics with the previously checked metrics
- Tested with ticking tables stuck to the bottom to ensure they stuck while ticking
- Fixes #1112

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -536,31 +536,7 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     this.requestUpdateCanvas();
 
-    if (!this.metrics || !this.prevMetrics) {
-      return;
-    }
-
-    const { rowCount, columnCount, height, width } = this.metrics;
-    const {
-      rowCount: prevRowCount,
-      columnCount: prevColumnCount,
-      height: prevHeight,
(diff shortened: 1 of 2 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -850,6 +832,37 @@ class Grid extends PureComponent<GridProps, GridState> {
     }
   }
 
+  checkMetricsUpdates() {
```
mattrunyon, line 835: Can you add a comment for why you would call this method over `updateMetrics`? Seems this is actually just scrolling sticky tables. If that's the case, update the method name to be more clear. `scrollIfSticky` maybe?

mattrunyon (changes requested):

mattrunyon (approved):

PR #1186, opened 2023-03-29
fix: Typing for WritableStream

### LINKED ISSUE #803: Fix type error in TableSaver.tsx
**Description**

The class variable `this.WritableStream` is incorrectly typed in TableSaver.tsx

**Steps to reproduce**
1. Go to `TableSaver.tsx`
3. Scroll down to `WritableStream = window.WritableStream ?? ponyfillWritableStream;`

**Expected results**

WritableStream should be typed as 
`WritableStream: window.WritableStream || ponyfillWritableStream;`

**Actual results**

WritableStream is typed as 
`WritableStream = window.WritableStream ?? ponyfillWritableStream;`

mofojed:
- Import in a different file and just export it
- Use implicit typing to work around the issue
- Fixes #803

### DIFF
--- a/packages/iris-grid/src/sidebar/WritableStream.ts
+++ b/packages/iris-grid/src/sidebar/WritableStream.ts
@@ -1,0 +1,6 @@
+import { WritableStream as ponyfillWritableStream } from 'web-streams-polyfill/ponyfill';
+
+// WritableStream is not supported in Firefox (also IE) yet. use ponyfillWritableStream instead
+const WritableStream = window.WritableStream ?? ponyfillWritableStream;
+
+export default WritableStream;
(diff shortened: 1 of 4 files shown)


packages/iris-grid/src/sidebar/WritableStream.ts
```diff
@@ -0,0 +1,6 @@
+import { WritableStream as ponyfillWritableStream } from 'web-streams-polyfill/ponyfill';
+
+// WritableStream is not supported in Firefox (also IE) yet. use ponyfillWritableStream instead
```
dsmmcken, line 3: Support was added in firefox 100, browserlist `defaults` has 102 as the minimum supported version. So we should actually be able to drop this polyfill entirely now, no?

image

image

mofojed: Yee

PR #1189, opened 2023-03-30
fix: Double clicking a file causes the loader to flash incorrectly

### LINKED ISSUE #942: Double clicking a file causes the loader to flash incorrectly
**Description**

Double clicking a file that is not open causes it to briefly open in preview mode and then in full mode. The loading spinner double flashes in and out on the panel that is opening.

Double clicking a file that is open in preview mode triggers a load as well. This shouldn't happen since the file is already open

**Steps to reproduce**
1. Open a code studio
2. Double click in the file explorer on a file that is not currently open
3. Single click a file that is not open so it opens in preview mode
4. Double click the file that is open in preview mode

**Expected results**

2. The first file opens in full mode w/ 1 loading spinner and no flashes
4. The second file opens in full mode with no loading spinner

**Actual results**

2. The first file opens in preview mode w/ loading spinner, displays file, then loading spinner again, then file in full mode
4. The second file opens in full mode after a loading spinner

**Versions**
 - Deephaven UI: 0.22.2

mattrunyon:
Fixes #942

### DIFF
--- a/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
+++ b/packages/dashboard-core-plugins/src/ConsolePlugin.tsx
@@ -446,47 +446,24 @@ export function ConsolePlugin(
       const isFileOpen = fileIsOpen(fileMetadata);
       const isFileOpenAsPreview = fileIsOpenAsPreview(fileMetadata);
 
-      // If the file is already open, just show and focus it if necessary
-      if (isFileOpen) {
-        showFilePanel(fileMetadata);
-        if (shouldFocus) {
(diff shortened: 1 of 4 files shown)


mofojed (changes requested):
This isn't right - if I single-click on `foo.py`, it opens as preview, and then I single-click on `bar.py`, and it just promotes the existing preview rather than opening `bar.py` in the preview.

mattrunyon:
Good catch. Think it's all fixed now

mofojed (approved):

PR #1190, opened 2023-03-30
feat: Double click notebook tab to remove its preview status

mattrunyon:
Something I noticed while working on #1189 is that we didn't support double clicking the tab to promote from preview to normal state like VSCode does. This PR lets you double click the tab of a notebook in preview to change it from preview state to normal state.

Tested double clicking preview notebook, non-preview notebook, and opening other notebooks in preview after promoting via double click

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
@@ -957,9 +957,12 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     });
   }
 
-  handlePanelTabClick(): void {
+  handlePanelTabClick(e: MouseEvent): void {
     log.debug('handlePanelTabClick');
     this.focus();
+    if (e.detail === 2) {
+      this.removePreviewStatus();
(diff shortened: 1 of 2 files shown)


packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -957,9 +957,12 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
     });
   }
 
-  handlePanelTabClick(): void {
+  handlePanelTabClick(e: MouseEvent): void {
```
mofojed, line 960: Random side note - could use `Panel.onTabClicked` prop instead of adding the event listener to glEventHub directly from here, but no big deal

mofojed (approved):

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
mattrunyon: I'll make that change. Seems like a good cleanup and IMO the fewer places we listen for golden-layout events the better

PR #1199, opened 2023-04-03
fix: Fixed generic selector types

### LINKED ISSUE #1197: Fix Redux Generic Types in @deephaven/redux
**Description**
Some of our Redux selectors support generic args that don't actually get used in return types. For example, in
node_modules/@deephaven/redux/dist/selectors.d.ts

```typescript
export declare const getSettings: <State extends RootState = RootState>(
  store: State
) => RootState['workspace']['data']['settings'];
```
We have the generic `State` parameter, but return `RootState` properties. This should be updated so that the return type uses the generic type.

**Expected**
All Redux actions + selectors that support generics should properly use the generics.

bmingles:
Updated redux selectors to properly derive return types from generic input types.

fixes #1197

### DIFF
--- a/packages/redux/src/selectors.ts
+++ b/packages/redux/src/selectors.ts
@@ -1,4 +1,4 @@
-import type { RootState, Storage, WorkspaceSettings } from './store';
+import type { RootState } from './store';
 
 const EMPTY_OBJECT = Object.freeze({});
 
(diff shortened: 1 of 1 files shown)


packages/redux/src/selectors.ts
```diff
@@ -34,80 +34,82 @@ export const getFileStorage = <State extends RootState = RootState>(
 
 export const getWorkspaceStorage = <State extends RootState = RootState>(
   store: State
-): Storage['workspaceStorage'] => getStorage(store).workspaceStorage;
+): State['storage']['workspaceStorage'] => getStorage(store).workspaceStorage;
```
mattrunyon, line 37: These probably don't need to set a default generic value. It should be inferred from the param just fine

packages/redux/src/selectors.ts
```diff
@@ -34,80 +34,82 @@ export const getFileStorage = <State extends RootState = RootState>(
...
 export const getWorkspace = <State extends RootState = RootState>(
   store: State
-): RootState['workspace'] => store.workspace;
+): State['workspace'] => store.workspace;
 
 // Settings
+type Settings<
+  State extends RootState = RootState
+> = State['workspace']['data']['settings'];
+
```
mattrunyon, line 48: Minor, but I would move this above the selectors probably. Just where I tend to put utils/types so they don't blend into implementation

Also, probably doesn't need a default generic value. In this case, it could make it easier to forget and not make the return type generic. `Settings['defaultDateTimeFormat']` would be valid, but always refer to the `RootState` type

mattrunyon (changes requested):

packages/redux/src/selectors.ts
bmingles: So, should I remove the defaults from all similar selectors?

mattrunyon (approved):

PR #1201, opened 2023-04-03
feat: Added isACLEditor prop to Redux state

### LINKED ISSUE #1200: ACL Editor - Redux Types
As a developer I want the Redux types in dh core to reflect state needed to support ACL Editor features. At the very least we need an `isACLEditor` flag to identify a user who should have access to ACL Editor features.

This is a blocker for [DH-14581](https://deephaven.atlassian.net/browse/DH-14581)

bmingles:
resolves #1200

### DIFF
--- a/packages/redux/src/store.ts
+++ b/packages/redux/src/store.ts
@@ -14,6 +14,7 @@ export interface DeephavenPluginModule {}
 export type DeephavenPluginModuleMap = Map<string, DeephavenPluginModule>;
 
 export interface UserPermissions {
+  isACLEditor: boolean;
   isSuperUser: boolean;
   isQueryViewOnly: boolean;
   isNonInteractive: boolean;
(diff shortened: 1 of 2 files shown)


packages/redux/src/store.ts
```diff
@@ -14,6 +14,7 @@ export interface DeephavenPluginModule {}
 export type DeephavenPluginModuleMap = Map<string, DeephavenPluginModule>;
 
 export interface UserPermissions {
+  isACLEditor: boolean;
   isSuperUser: boolean;
   isQueryViewOnly: boolean;
```
mofojed, line 19: Side note - these permissions really only exist/matter for Enterprise users, so we should be extending the interface in Enterprise redux to add these permissions. That hasn't been converted to TS yet though, and it's not doing much harm being added here.

mofojed (approved):

PR #1204, opened 2023-04-04
feat: Relative links

### LINKED ISSUE #1070: Relative UI links
Some use cases, namely proxying based on a fixed prefix, would greatly benefit if the UI was built using relative links. @mofojed has helped me get a PoC up and running, and it works beautifully.

See https://github.com/deephaven/web-client-ui/tree/relative-base-url

This is not a full solution (as notebooks don't work), but demonstrates that `<prefix>/ide/` works.

I'm happy to help a dev get up a simple, or complex, reverse proxy setup to validate any fix.

Note: it's not possible to use different prefixes w/ gRPC, so for muli-instance proxying, the proxy needs to key off some other kind of information. I was able to leverage the `Referer` header to proxy gRPC appropriately in the case of multiple DHC instances.

A simple setup might look like:

docker-compose.yml:
```yml
version: '3'

name: deephaven-caddy

services:
  deephaven:
    image: <deephaven-server-with-relative-ui-link-support>

  reverse-proxy:
    image: caddy:2.6.2
    restart: unless-stopped
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile:ro
```

Caddyfile:
```
test.localhost {
    handle /foo/* {
        uri strip_prefix /foo
        reverse_proxy {
            to h2c://deephaven:10000
        }
    }
    reverse_proxy {
        to h2c://deephaven:10000
    }
}
```

Then ensuring that https://test.localhost/foo/ide/ works.

mofojed:
- Notebooks work
- Links between notebooks work
- Deeplinking notebooks works in dev mode with proxy
  - Does not work in production - would need server to correctly proxy URLs and write the correct document base
- Embed-grid works
- Embed-chart works
- Plugins work
- Fixes #1070 
- Unit tests and e2e tests pass
- Tested in both a dev build, and production build
- Tested following steps in #1070 
- Tested in DnD, ensured the correct paths were being loaded


packages/babel-preset/index.js
```diff
@@ -22,13 +22,21 @@ module.exports = api => ({
...
+      : false,
+    api.env('test')
+      ? [
+          // Also need to handle import.meta.url
+          'babel-plugin-transform-import-meta',
+        ]
+      : false,
+    api.env('test')
+      ? false // The add-import-extension plugin causes Jest to error, but is needed for proper ESM builds
+      : ['babel-plugin-add-import-extension'],
```
mattrunyon, line 39: Let's change this to something like the below so it's more readable/grouped by environment

```
api.env('test') ? [path.resolve(__dirname, 'importMetaEnvPlugin'), 'babel-plugin-transform-import-meta'] : false,
!api.env('test') ? ['babel-plugin-add-import-extension'] : false,
```

packages/code-studio/src/main/AppRouter.tsx
```diff
@@ -1,19 +1,13 @@
...
 import AppInit from './AppInit';
 
 function AppRouter(): ReactElement {
   return (
     <Router basename={import.meta.env.BASE_URL}>
       <Switch>
-        <Route exact path="/" component={AppInit} />
         <Route path="/notebook/:notebookPath+" component={AppInit} />
-        <Redirect from="*" to="/" />
+        <Route path="*" component={AppInit} />
```
mattrunyon, line 10: Was the redirect causing an issue with the relative paths? It was to not show random URL paths that don't mean anything like `/ide/somethingIMadeUp`

packages/embed-chart/.env
```diff
@@ -1,6 +1,6 @@
 # Location of the iris script and API server
 # Set this value to __mocks__ to use mock server instead
-VITE_CORE_API_URL=/jsapi
+VITE_CORE_API_URL=../../../jsapi
```
mattrunyon, line 3: Would really like to find a better way if possible as this seems brittle

mattrunyon (changes requested):

packages/babel-preset/index.js
mofojed: I agree that is more readable, but it is _not_ the same thing. The [plugins config option](https://babeljs.io/docs/options#plugins) takes an array of type `Array<PluginEntry | Plugin>`. [PluginEntry is defined](https://babeljs.io/docs/options#pluginpreset-entries) as `EntryTarget | [EntryTarget, EntryOptions]`; doing as you suggest would pass in would result in:
```
plugins: [
  [ path.resolve(__dirname, 'importMetaEnvPlugin'), 'babel-plugin-transform-import-meta' ],
  ...
]
e.g. importing entry target `importMetaEnvPlugin` with the entry options `babel-plugin-transform-import-meta`. So then you get:
```
FAIL   @deephaven/code-studio  packages/code-studio/src/main/AppMainContainer.test.tsx
  ● Test suite failed to run

    [BABEL] /home/bender/dev/deephaven/iris-oss/web-client-ui/main/jest.setup.ts: .plugins[0][1] must be an object, false, or undefined
```

I tried using a spread operator, but using that in a ternary apparently screws things up as well:
![image](https://user-images.githubusercontent.com/4505624/231210261-295d2873-38f7-4348-b351-0d915ecb1c47.png)

I tried a few different ways, and this was the way that I hated the least. And I still hate it.

I guess I could put it into a function and that may be more readable, e.g.:
```
function getPlugins(isTestEnv) {
  if (isTestEnv) {
    return [
      // This is needed to replace import.meta.env w/ process in Jest
      // Jest does not play nicely w/ ESM and Vite uses import.meta
      // import.meta is only avaialable in ESM
      path.resolve(__dirname, 'importMetaEnvPlugin'),
      // Also need to handle import.meta.url
      'babel-plugin-transform-import-meta',
    ];
  }
  return [
    // The add-import-extension plugin causes Jest to error, but is needed for proper ESM builds
    'babel-plugin-add-import-extension',
  ];
}

...

  plugins: [
    ...getPlugins(api.env('test')),
    [
      'transform-rename-import',
      {
        // The babel-plugin-add-import-extension adds the .js to .scss imports, just convert them back to .css
        original: '^(.+?)\\.s?css.js$',
        replacement: '$1.css',
      },
    ],
  ].filter(Boolean),
```

... but I still hate it.

File: packages/code-studio/.env [hunk 4]
```diff
@@ -1,25 +1,33 @@
...
+
+## Path to plugins that load a module
+## See https://github.com/deephaven/deephaven-js-plugin-template/
+VITE_MODULE_PLUGINS_URL=../../js-plugins
+
+# Plugin URLs
+## Path to plugins that load/render as a component (e.g. Table plugins)
+## See https://github.com/deephaven/js-plugin-template/ for templates
+# @deprecated, use VITE_MODULE_PLUGINS_URL instead
+VITE_COMPONENT_PLUGINS_URL=./plugins/
```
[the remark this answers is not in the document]
mofojed: Ya we should be able to remove this entirely.

packages/code-studio/.env
```diff
@@ -1,25 +1,33 @@
...
+
+# Path for notebooks and layouts storage on the gRPCStorageService
 VITE_NOTEBOOKS_URL=/notebooks
 VITE_LAYOUTS_URL=/layouts
```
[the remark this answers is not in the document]
mofojed: Because these are being passed into the gRPC storage service, which is through the JS API connection, not as an actual URL.

packages/babel-preset/index.js
mofojed: I'll leave as is

packages/code-studio/.env
```diff
@@ -1,25 +1,24 @@
 # Location of the iris script and API server
 # Set this value to __mocks__ to use mock server instead
-BASE_URL=/ide/
-VITE_CORE_API_URL=/jsapi
+BASE_URL=./
+VITE_CORE_API_URL=../../jsapi
```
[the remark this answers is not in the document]
devinrsmith: It is a responsibility of the server - I expect the majority of users would never need to try and configure it.

That said, there is something that doesn't pass the "sniff" test; b/c I expect this value to essentially always be a constant, as a sibling of `ide/`.

packages/code-studio/.env
```diff
@@ -1,25 +1,33 @@
...
 VITE_NOTEBOOKS_URL=/notebooks
 VITE_LAYOUTS_URL=/layouts
-VITE_ENABLE_LOG_PROXY=true
+
+# Proxy used by Vite
+VITE_PROXY_URL=http://localhost:10000
 VITE_SUPPORT_LINK=https://github.com/deephaven/web-client-ui/
 VITE_DOCS_LINK=https://deephaven.io/core/docs/
-VITE_FAVICON=/favicon-cc-app.svg
+VITE_FAVICON=./favicon-cc-app.svg
```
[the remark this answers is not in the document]
mofojed: Sweet

packages/code-studio/src/main/AppRouter.tsx
mofojed: Yea this doesn't work correctly with the `'./'` at all, because it is just using the current URL path as the "basename".

I think instead, we should be getting the path from the `import.meta.url` (where are assets are). Not sure if the newer version of `react-router-dom` helps with this, but it has many breaking changes: https://reactrouter.com/en/main/upgrading/v5

For now I'll just fix up the base URL here.
Also it doesn't look like the `/notebook/` redirect works anywhere right now.

packages/code-studio/.env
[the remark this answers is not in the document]
mofojed: Okay, I'll make a note that we assume the `jsapi/` directory will be a sibling of the directory where this web UI is hosted (generally `ide/`

packages/code-studio/.env
```diff
@@ -1,25 +1,33 @@
-# Location of the iris script and API server
-# Set this value to __mocks__ to use mock server instead
-BASE_URL=/ide/
-VITE_CORE_API_URL=/jsapi
+# Path at which the app is served
+# https://vitejs.dev/guide/build.html#public-base-path
+BASE_URL=./
+
+# Location of the core JS API and API server
+VITE_CORE_API_URL=../../jsapi
```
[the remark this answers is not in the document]
mofojed: I've made a note that we expect the UI to be served relative to the jsapi. We should then use code splitting to handle the iframe cases and just deprecate embed-grid/embed-chart, then we'll be able to manage our routes a little better.

packages/code-studio/src/main/AppRouter.tsx
mofojed: Also, thinking about this even more - the `BASE_URL` set to `./` would screw up any routing we'd have with paths, since the index.html would be incorrect. e.g if I go to `/ide/notebook/foo.py`, it'll serve up the `index.html`, but the path to the assets will be incorrect, e.g.:
```
<script type="module" crossorigin src="./assets/index-05abad89.js"></script>
```
... Which would try and load `/ide/notebook/assets/index-05abad89.js`, which is of course incorrect.
Hmmm.

devinrsmith:
We discussed having the server parameterize the head base url - can you post a snippet of that section of the index.html @mofojed. From a purity standpoint, it's a bit hacky as now we aren't being a "transparent proxy".

There was also discussion around using query parameters instead of paths, and that could solve the issue.

Or, potentially a service worker.

mofojed:
There's a note in the Create React App docs about this: https://create-react-app.dev/docs/deployment/#serving-the-same-build-from-different-paths

> If you are not using the HTML5 pushState history API or not using client-side routing at all, it is unnecessary to specify the URL from which your app will be served.

We use Vite, but the same principles apply. We are serving up a static application, and client-side routing will not work correctly in this scenario. Either we don't do client-side routing (just using query params to specify paths), or we need server to rewrite the correct `base`. Of the two, just using query params for routing is certainly easier given our current architecture.

@devinrsmith The snippet that would need to be written in the index.html is similar to what we write in Enterprise (though it's written at Web UI Client build time, not dynamically by the server):
```
    <base href="/iriside/" />
```

devinrsmith:
Ok - I'm still interested in the query param route so the proxy / server can be completely transparent about it.

If we need to test out the server/proxy re-writing `<base href=...>`, that's something I'll need to spend cycles on later...

mattrunyon:
We could still do client side routing with the HashRouter, the URLs are just a little bit ugly since they become `/foo/ide#iframe/table`. I think the other downsides are you don't get proper forward/back support in the browser, but we don't really use that anyway

HashRouter lets the bundle stay at a single URL. Everything after the hash is client-side only for displaying routes from the single bundle.

mofojed:
@dsmmcken how about using HashRouter for client side routing?

dsmmcken:
No. Client side routing is a feature of the app.

mattrunyon:
What client routing that we currently use would hash router not support? We don't use the history API in enterprise, so even though there are different URLs, you can't navigate forward/back. So hash router would work identically aside from the URL containing a `#`

mofojed:
This cleans up some of the env vars and adds a `base` tag to the index.html.
It does not complete relative links entirely, since proxy will need to rewrite the base tag, but does not put us in a worse position. Still exploring proxy rewriting the base tag as an option.

packages/code-studio/src/main/AppRouter.tsx
```diff
@@ -1,13 +1,19 @@
...
 import AppInit from './AppInit';
 
 function AppRouter(): ReactElement {
   return (
     <Router basename={import.meta.env.BASE_URL}>
       <Switch>
+        <Route path="/" component={AppInit} />
         <Route path="/notebook/:notebookPath+" component={AppInit} />
-        <Route path="*" component={AppInit} />
+        <Redirect from="*" to="/" />
```
mattrunyon, line 16: I think this was initially just a wildcard path instead of a redirect for DHE deep linking that didn't require us to put a specific route in the router for each deep linking route

That said, it's probably better to have a route listed for each deep linking route in DHE because it gives us a more centralized location for them. Easier discoverability

packages/code-studio/vite.config.ts
```diff
@@ -54,10 +54,8 @@ export default defineConfig(({ mode }) => {
   // It is possible to add one with a custom middleware though if this list grows
   if (env.VITE_PROXY_URL) {
     [
-      env.VITE_CORE_API_URL,
-      env.VITE_NOTEBOOKS_URL,
-      env.VITE_LAYOUTS_URL,
-      env.VITE_MODULE_PLUGINS_URL,
+      path.resolve(env.BASE_URL, env.VITE_CORE_API_URL),
+      path.resolve(env.BASE_URL, env.VITE_MODULE_PLUGINS_URL),
```
mattrunyon, line 58: Should any of the above proxy elements include the `BASE_URL`? I guess styleguide is fine without it. Are the websocket requests always to the server root?

packages/embed-chart/.env
```diff
@@ -1,9 +1,10 @@
-# Location of the iris script and API server
-# Set this value to __mocks__ to use mock server instead
-VITE_CORE_API_URL=../../../jsapi
+# See the values in [code sudio](../code-studio/.env) for more details
+# We assume embed-chart is served at '/iframe/chart', we need to go up 3 levels
```
mattrunyon, line 2:
```suggestion
# We assume embed-chart is served at '/iframe/chart'
```

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -100,70 +100,6 @@ function AppInit(props: AppInitProps) {
   const plugins = usePlugins();
   // General error means the app is dead and is unlikely to recover
   const [error, setError] = useState<unknown>();
-  // Disconnect error may be temporary, so just show an error overlaid on the app
-  const [disconnectError, setDisconnectError] = useState<unknown>();
-  const [isFontLoading, setIsFontLoading] = useState(true);
-
```
mattrunyon: Why are all these lines removed here? Where did they go?

mattrunyon (changes requested):

packages/code-studio/.env
```diff
@@ -1,31 +1,35 @@
...
+# Like the CORE_API_URL, we assume this is served up as a sibling of the code studio
+VITE_MODULE_PLUGINS_URL=../js-plugins
 
 # Plugin URLs
-## Path to plugins that load/render as a component (e.g. Table plugins)
-## See https://github.com/deephaven/js-plugin-template/ for templates
+# Path to plugins that load/render as a component (e.g. Table plugins)
+# See https://github.com/deephaven/js-plugin-template/ for templates
 # @deprecated, use VITE_MODULE_PLUGINS_URL instead
 VITE_COMPONENT_PLUGINS_URL=./plugins/
```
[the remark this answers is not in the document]
mofojed: Yea this should just be removed from Community, it's on Enterprise only. Basically just legacy support for standalone table plugins, but Community moved to plugin modules which support table plugins as well.

packages/code-studio/src/main/AppRouter.tsx
mofojed: I added the wildcard path with a previous change: https://github.com/deephaven/web-client-ui/pull/1204/commits/d51bd7d96963bee2969415da1d39c556f2bc3b16
But that no longer applies, so I'm switching this back.

packages/code-studio/vite.config.ts
mofojed: We always connect to the server root for GRPC.
I did break the styleguide though.

packages/code-studio/src/index.tsx
```diff
@@ -17,18 +17,19 @@ const AppBootstrap = React.lazy(async () => {
   return { default: module.AppBootstrap };
 });
 
+const baseURL = new URL(import.meta.env.BASE_URL, import.meta.url);
```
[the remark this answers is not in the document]
mofojed: Yuuup. Crud, I must have been thinking about this wrong. Proxy/Jetty might be able to re-write the `base` tag in `index.html`, but they wouldn't be able to re-write this...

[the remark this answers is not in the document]
mofojed: Also `import.meta.url` returns a long strange path in vite dev mode: /ide/@fs/home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages

packages/code-studio/src/main/AppInit.tsx
mattrunyon: Ok I'll check the full diff. I normally use the "changes since last review" option when I re-review

packages/code-studio/src/index.tsx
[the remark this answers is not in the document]
mattrunyon: That makes sense I guess b/c Vite serves the files from routes w/ `@fs` in dev mode. So I don't think `import.meta.url` will work reliably in this case. I think it's more intended for [static assets in Vite](https://vitejs.dev/guide/assets.html#new-url-url-import-meta-url)

packages/code-studio/src/main/AppRouter.tsx
```diff
@@ -7,12 +7,16 @@ import {
 } from 'react-router-dom';
 import AppInit from './AppInit';
 
+const basename = new URL(document.baseURI).pathname;
```
mattrunyon, line 10:
```suggestion
const basename = document.location.pathname;
```

packages/code-studio/src/main/AppRouter.tsx
```diff
@@ -7,12 +7,16 @@ import {
...
   return (
-    <Router basename={import.meta.env.BASE_URL}>
+    <Router basename={basename}>
       <Switch>
         <Route exact path="/" component={AppInit} />
-        <Route path="/notebook/:notebookPath+" component={AppInit} />
+        <Route
+          path={`/${import.meta.env.VITE_ROUTE_NOTEBOOKS}:notebookPath+`}
+          component={AppInit}
+        />
```
mattrunyon, line 19: No hash router? I don't think this will work without jetty rewriting the base tag

It works for me in dev mode, but not production

packages/code-studio/src/styleguide/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <base href="#BASE_URL#" />
```
mattrunyon, line 4: Redundant if base url is `./`

packages/code-studio/vite.config.ts
```diff
@@ -26,13 +26,14 @@ export default defineConfig(({ mode }) => {
...
 
+  const baseURL = new URL(env.BASE_URL, `http://localhost:${port}/`);
   // These are paths which should be proxied to the core server
   // https://vitejs.dev/config/server-options.html#server-proxy
   const proxy = {
     // Proxy styleguide here instead of as a route in our app router
     // That way, it is not included in the production build
-    '/styleguide': {
-      target: `http://localhost:${port}/src/styleguide/index.html`,
+    '^/styleguide': {
```
mattrunyon, line 35: You don't need the caret here really. Not a regex and matches start of the path. Shouldn't hurt anything though. [Docs](https://vitejs.dev/config/server-options.html#server-proxy)

packages/code-studio/vite.config.ts
```diff
@@ -49,21 +50,27 @@ export default defineConfig(({ mode }) => {
...
-      env.VITE_LAYOUTS_URL,
-      env.VITE_MODULE_PLUGINS_URL,
-    ].forEach(p => {
-      proxy[p] = {
+    // Some paths need to proxy to the engine server
+    // Vite does not have a "any unknown fallback to proxy" like CRA
+    // It is possible to add one with a custom middleware though if this list grows
+    [env.VITE_CORE_API_URL, env.VITE_MODULE_PLUGINS_URL].forEach(p => {
+      const route = new URL(p, baseURL).pathname;
+      proxy[`^${route}`] = {
```
mattrunyon, line 59:
```suggestion
      proxy[route] = {
```

mattrunyon (changes requested):

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -129,11 +129,11 @@ function AppInit(props: AppInitProps) {
           const storageService = client.getStorageService();
           const layoutStorage = new GrpcLayoutStorage(
             storageService,
-            import.meta.env.VITE_LAYOUTS_URL ?? ''
+            import.meta.env.VITE_STORAGE_PATH_LAYOUTS ?? ''
```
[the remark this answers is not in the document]
mofojed: It was handy when we hadn't solidified the path yet. Would still need it somewhere global.

packages/code-studio/src/main/AppRouter.tsx
mofojed: No hash router... I didn't like it always putting a hash in the URL. Also trickier to resolve paths...

packages/code-studio/.env
```diff
@@ -1,24 +1,34 @@
...
+# Like the CORE_API_URL, we assume this is served up as a sibling of the code studio
+VITE_MODULE_PLUGINS_URL=../js-plugins
+
+# Path for notebooks and layouts storage on the gRPCStorageService
+# Note these are not URLs, these are file system paths on the server in the gRPCStorageService
+VITE_STORAGE_PATH_NOTEBOOKS=/notebooks
+VITE_STORAGE_PATH_LAYOUTS=/layouts
+
+# Any routes we define here
+VITE_ROUTE_NOTEBOOKS=notebooks/
```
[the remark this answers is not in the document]
mofojed: Still need to have it globally somewhere. Where would you suggest?

packages/embed-grid/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <base href="#BASE_URL#" />
```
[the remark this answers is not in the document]
mofojed: You can configure the base to be something else though, so possible to still build with an env var set and have the base tag be set.
We use `/ide/` in dev mode, for example.

packages/code-studio/.env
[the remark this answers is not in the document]
mattrunyon: I guess this is fine. Just feels odd to have `notebooks/` and a `/notebooks` in here for slightly different uses

This is only used in the router and in `ConsolePlugin`, so I'm curious if there's a way to use it more cleanly in both. Not worth trying to figure out now though.

I think it's only used for linking from a markdown notebook right now. Might not need to have the URL if we know the file will just be relative to the `GrpcFileService` base

packages/code-studio/src/main/AppInit.tsx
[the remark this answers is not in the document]
mattrunyon: It's only used here. So we have an env variable to define a constant value used in 1 place specific to the app. Potentially useful if we have cases where we need to change it or proxy it. But right now we do neither

Fine to leave for now, just seemed a roundabout way to define what is now a constant value used in 1 location

packages/embed-grid/index.html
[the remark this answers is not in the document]
mattrunyon: Ok makes sense. Also as you said it makes a Jetty rewrite easier since Jetty can use sed or something to just replace this tag instead of having to find where to insert it

mattrunyon (approved):

packages/code-studio/.env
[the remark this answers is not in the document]
mofojed: Yea the reason I had two is because the GrpcFileService does not necessarily have the same path as our route. They are different things, after all.

PR #1208, opened 2023-04-05
feat: usePanelRegistration hook

### LINKED ISSUE #1207: usePanelRegistration hook
As a developer, I'd like to minimize boilerplate code when registering plugin panels. A `usePanelRegistration` could be used to abstract registration / de-registration.

bmingles:
resolves #1207

### DIFF
--- a/packages/dashboard/src/DashboardPlugin.ts
+++ b/packages/dashboard/src/DashboardPlugin.ts
@@ -16,7 +16,7 @@ export type WrappedComponentType<
 export type PanelComponentType<
   P extends PanelProps = PanelProps,
   C extends ComponentType<P> = ComponentType<P>
-> = ComponentType<P> | WrappedComponentType<P, C>;
+> = (ComponentType<P> | WrappedComponentType<P, C>) & { COMPONENT?: string };
 
 export function isWrappedComponent<
   P extends PanelProps,
--- a/packages/dashboard/src/DashboardPlugin.ts
+++ b/packages/dashboard/src/DashboardPlugin.ts
@@ -1,4 +1,4 @@
-import { Component, ComponentType } from 'react';
+import { Component, ComponentType, NamedExoticComponent } from 'react';
 import { ConnectedComponent } from 'react-redux';
 import GoldenLayout from '@deephaven/golden-layout';
 import type {
@@ -8,6 +8,11 @@ import type {
 } from '@deephaven/golden-layout';
 import PanelManager from './PanelManager';
 
+export type PanelFunctionComponentType<P> = NamedExoticComponent<P> & {
+  COMPONENT?: string;
+  TITLE?: string;
+};
+
 export type WrappedComponentType<
   P extends PanelProps,
   C extends ComponentType<P>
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/layout/usePanelRegistration.ts
```diff
@@ -0,0 +1,51 @@
...
+  registerComponent: DashboardPluginComponentProps['registerComponent'],
+  ComponentType: PanelComponentType<P, C> | React.NamedExoticComponent<P>,
+  hydrate?: PanelHydrateFunction<P>,
+  dehydrate?: PanelDehydrateFunction
+) {
+  // Look for static `COMPONENT` or `displayName` attribute
+  const name =
+    'COMPONENT' in ComponentType
+      ? ComponentType.COMPONENT
+      : ComponentType.displayName;
```
mofojed, line 31: I think you can just use null coalescing to handle this:
```suggestion
  const name = ComponentType.COMPONENT ?? ComponentType.displayName;
```

mofojed (approved):
Just a minor suggestion for using null coalescing, but looks great!

packages/dashboard/src/layout/usePanelRegistration.ts
bmingles: I actually did that originally, but the type system complains when you access `ComponentType.COMPONENT` without verifying it's there

bmingles: nvm, fixed it in the types

packages/dashboard/src/layout/usePanelRegistration.ts
```diff
@@ -0,0 +1,49 @@
...
+  PanelComponentType,
+  PanelDehydrateFunction,
+  PanelHydrateFunction,
+  PanelProps,
+} from '../DashboardPlugin';
+
+/**
+ * Registers a given panel component. Also runs a `useEffect` that will
+ * automatically de-register then panel on unmount.
+ * @param registerComponent
```
mattrunyon, line 13: We might not want this as part of the hook. Bender and I chatted about plugins the other day and there's a lot of rough edges and things that need to be cleaned up around their implementation. Basically we _should_ be able to write our existing plugins all as external plugins in an ideal world

hydrate and dehydrate are the same. IMO they are logic that belong in the component, not at the registration point. Our existing hydration code is in AppMainContainer which separates it from where the hydration is actually used.

I don't think they're changes that should be in this PR, but just something to note that this will probably change signatures as we address those points

mofojed (approved):
@mattrunyon hydration has to be available at the app level, but we should be moving widget panels to just use the default hydration when possible. `metadata` for data about the widget, `fetch` to get the object, `panelState` for storing any previous state set by the user.

packages/dashboard/src/DashboardPlugin.ts
```diff
@@ -8,6 +8,11 @@ import type {
 } from '@deephaven/golden-layout';
 import PanelManager from './PanelManager';
 
+export type PanelFunctionComponentType<P> = NamedExoticComponent<P> & {
+  COMPONENT?: string;
+  TITLE?: string;
```
mofojed, line 13: Since this is used here and in `PanelComponentType`, may want to separate it out and add some JSDocs about what each is for

mofojed (approved):

PR #1211, opened 2023-04-06
feat(@deephaven/components): Custom React Spectrum Provider

### LINKED ISSUE #1210: React Spectrum Theme Provider
Supports Enterprise DH-14581

As a developer, I want to be able to use React Spectrum and to customize css variables to be consistent with Deephaven styles.

bmingles:
Custom React Spectrum Provider component.

resolves #1210


packages/components/src/SpectrumProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children,
+}: React.PropsWithChildren<SpectrumProviderProps>): JSX.Element {
+  return (
+    <Provider
+      UNSAFE_className={cl(
+        styles.component,
+        styles[`component-${colorScheme}`]
+      )}
+      colorScheme={colorScheme}
+      theme={defaultTheme}
```
mofojed, line 24: Should the theme be a prop passed in as well that defaults to `defaultTheme`?

packages/components/package.json
```diff
@@ -45,6 +45,7 @@
     "shortid": "^2.2.16"
   },
   "peerDependencies": {
+    "@adobe/react-spectrum": "^3.x",
```
mofojed, line 48: I think this needs to be flagged as a breaking change? Any consumer of `@deephaven/components` will need to include `@adobe/react-spectrum` now, right? Or is it just if you try and use `SpectrumProvider`

bmingles: As far as I can tell, this doesn't cause any issues until attempting to consume something with a Spectrum dependency (currently only SpectrumProvider).

bmingles: @mofojed I'm not 100% sure whether this is a breaking change or not.
- I uninstalled @adobe/react-spectrum from enterprise and things still work (presumably due to local install in my community location
- I uninstalled from commmunity locally, but then the npm run start-community fails with compilation errors

It's quite possible in the real case where community has been deployed, enterprise will be fine if there are no references to SpectrumProvider, but I can't find a good way to actually test this theory.

I could see marking this as a breaking change to be safe unless you have any other ideas of how we can test it.

bmingles: I did some more testing, and it seems this is a breaking change regardless if SpectrumProvider is used or not. I've updated the description footer accordingly.

packages/components/src/SpectrumUtils.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('themeDHDefault', () => {
+  it('should merge Spectrum default with DH custom styles', () => {
+    const { global, light, dark, medium, large } = defaultTheme;
+
+    expect(themeDHDefault).toEqual({
+      global,
+      light: {
+        ...light,
+        'dh-spectrum-theme--light': 'mock.light',
```
mattrunyon, line 12: You should be able to remove the mock and just expect this to equal `dh-spectrum-theme--light`. The css identity-obj-proxy just creates an identity object. I.e. returns the requested key as the value for anything you try to pull from the object. Might be `--dh-spectrum-theme--light`, but it shouldn't really need the mock

packages/components/src/SpectrumUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+ *   },
+ * }
+ */
+/* eslint-disable import/prefer-default-export */
+export const themeDHDefault = {
+  global,
+  light: {
+    ...light,
+    ...lightDH,
+  },
```
mattrunyon, line 46: Do the themes have any nested objects? If so, need to deep merge these

mattrunyon (changes requested):
Update the breaking change to be 3rd person and specify which package needs to add react spectrum in the comment.

Can also do `feat(@deephaven/components)` in the title I believe to indicate the scope of the change

packages/components/src/SpectrumUtils.test.ts
bmingles: The issue is that the implementation doesn't actually request any specific keys. It use a spread operator which seems to yield an empty object from the proxy. The test will pass, but it won't actually prove the dh props got added.

packages/components/src/SpectrumUtils.ts
bmingles: I don't believe that they do. They seem to just be key value maps

packages/components/src/SpectrumProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export interface SpectrumProviderProps {
+  colorScheme?: 'dark' | 'light';
+  theme?: ProviderProps['theme'];
+  /**
+   * Since this provider may be used to wrap components that are not at the root
+   * of the application, we may want to leave the background color alone so that
+   * components can "inherit" parent component background colors.
+   */
+  transparentBackground?: boolean;
```
mofojed, line 16: Any reason to only include a `transparentBackground` flag and not a `UNSAFE_style` prop to match the `Provider` ?

mofojed (dismissed):

packages/components/src/SpectrumProvider.tsx
bmingles: Looking at this more closely, I'm wondering if the custom provider is providing any real value. Maybe it makes more sense to just do

```typescript
<Provider
  UNSAFE_style={{ backgroundColor: 'transparent' }}
  colorScheme="dark"
  theme={themeDHDefault}
>
```

bmingles: I think we only need the `themeDHDefault` and don't need the custom provider at all. This will be used for wrapping the ACL panels in Enterprise. My assumption is that we will soon be creating more Spectrum based components in @deephaven/components, so having the themeDHDefault be defined there makes the most sense to me.

bmingles: @mofojed  I went ahead and removed the custom `SpectrumProvider` (I kept the `SpectrumUtils`)

packages/components/src/SpectrumUtils.ts
```diff
@@ -0,0 +1,54 @@
+import { defaultTheme } from '@adobe/react-spectrum';
```
mofojed, line 1: If all we're doing is using the `defaultTheme`, I'd rather just add `@react-spectrum/default-theme` as a dependency and get rid of the peer dependency/breaking change.
We're using the ACL Editor as a testing ground for trying out the spectrum components, I don't think the first step should be requiring consumers of our `@deephaven/components` package to add spectrum as a dependency as well.

mofojed (changes requested):

packages/components/src/SpectrumUtils.ts
bmingles: I've made the change

mofojed (approved):

mattrunyon (approved):

PR #1212, opened 2023-04-06
feat:  Table rendering support for databars

### LINKED ISSUE #1151: Table rendering support - registering/using custom cell renderers
`GridModel` will need to have a way of specifying a new type of renderer on a per cell basis. Currently we just support rendering text, so that should be the default cell renderer, and then allow for registering/specifying different renderers. We can have a `TextCellRenderer` as the default, and `BarGaugeRenderer` packaged in `@deephaven/grid`, but not a default.

- [ ] Add a `rendererForCell(column: ModelIndex, row: ModelIndex): CellRendererType` and `type CellRendererType = string`
- [ ] Refactor text cell rendering to a separate class `TextCellRenderer` and use that from `GridRenderer` when rendering a cell by default (or if renderer not specified)
- [ ] Allow registering a different types of renderers that `GridRenderer.drawCellContents` will automatically call if there's a renderer specified for that cell
- [ ] Create `BarGaugeRenderer`, and an extension to `GridModel` for the details required for `BarGaugeRenderer`

We should be able to test this out in the style guide using a mock (similar to `MockGridModel`).

emilyhuxng:
Closes #1151 

The example can be seen at the style guide. 

`IrisGridRenderer` and `GridRenderer` now use different cell renderers depending on the type.


packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -0,0 +1,614 @@
...
+    context: CanvasRenderingContext2D,
+    leftColor: Oklab,
+    rightColor: Oklab,
+    x: number,
+    y: number,
+    width: number,
+    height: number
+  ) {
+    let currentColor = leftColor;
+    for (let currentX = x; currentX <= x + width; currentX += 1) {
```
mofojed, line 586: I think the issue we've observed with the black lines in the gradient at certain zooms is some sort of interpolation effect when drawn at half pixels... or something.
Anyway, I think it's fixed if you just advance by half a pixel instead of a full pixel:
```suggestion
    for (let currentX = x; currentX <= x + width; currentX += 0.5) {
```
Not sure if there's another underlying cause for it. But that seems to resolve it.

packages/grid/src/GridTheme.ts
```diff
@@ -221,6 +225,11 @@ const defaultTheme: GridTheme = Object.freeze({
   // Divider colors between the floating parts and the grid
   floatingDividerOuterColor: '#000000',
   floatingDividerInnerColor: '#cccccc',
+
+  // Databar
+  zeroLineColor: '#6c757d',
+  positiveColor: '#9cdb6c',
+  negativeColor: '#f25a80',
```
mofojed, line 232: Where did you get these colours from? The specific colours we use should be defined in `IrisGridTheme`, and I _think_ should be the same as `positive-number-color` and `negative-number-color` (which map to `$red` and `$green`).
For default GridTheme, we should just be using some basic colours (since we're not dependent on the Deephaven theme), something like:
```suggestion
  // Databar
  zeroLineColor: '#888888',
  positiveColor: '#00ffff',
  negativeColor: '#ff0000',
```

packages/grid/src/mouse-handlers/GridTokenMouseHandler.ts
```diff
@@ -36,7 +37,10 @@ class GridTokenMouseHandler extends GridMouseHandler {
     }
 
     const renderState = grid.updateRenderState();
-    const tokensInCell = renderer.getTokenBoxesForVisibleCell(
+    const textCellRenderer = renderer.getCellRenderer(
+      'text'
+    ) as TextCellRenderer;
+    const tokensInCell = textCellRenderer.getTokenBoxesForVisibleCell(
```
mofojed, line 43: Should be calling `model.rendererForCell` and then get the `renderer.getCellRenderer(type)`, and just return if it's not a text type.

packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -71,6 +74,10 @@ class IrisGridRenderer extends GridRenderer {
     return isAdvancedFilterValid && isQuickFilterValid;
   }
 
+  protected textCellRenderer = new IrisGridTextCellRenderer();
+
+  // protected dataBarCellRenderer = new DataBarCellRenderer();
+
```
mofojed, line 80: Unused

packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -151,25 +158,93 @@ class IrisGridRenderer extends GridRenderer {
...
-        const originalFont = context.font;
-        context.font = `italic ${originalFont}`;
-        const displayValue = value === null ? 'null' : 'empty';
-        super.drawCellContent(context, state, column, row, displayValue);
-        context.font = originalFont;
-        return;
-      }
+
+    const cellRenderer = this.getCellRenderer('text');
+    cellRenderer.drawCellContent(context, state, column, row);
```
mofojed, line 169: Should be getting the cell renderer type from the model, then getting the renderer and using that

packages/iris-grid/src/IrisGridRenderer.ts
```diff
@@ -151,25 +158,93 @@ class IrisGridRenderer extends GridRenderer {
...
     }
-    super.drawCellContent(context, state, column, row);
+    context.fill(icon);
+
+    context.restore();
+  }
+
+  getCellRenderer(rendererType: CellRendererType): CellRenderer {
+    return this.textCellRenderer;
   }
```
mofojed, line 248: Shouldn't need to override this function, just use the parent which already returns `this.dataBarCellRenderer` or `this.textCellRenderer`

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.ts
```diff
@@ -43,7 +44,10 @@ class IrisGridTokenMouseHandler extends GridMouseHandler {
     }
 
     const renderState = grid.updateRenderState();
-    const tokensInCell = renderer.getTokenBoxesForVisibleCell(
+    const textCellRenderer = renderer.getCellRenderer(
+      'text'
+    ) as TextCellRenderer;
+    const tokensInCell = textCellRenderer.getTokenBoxesForVisibleCell(
```
mofojed, line 50: Get cell render type from the model, then get the renderer. Return if it's not a TextCellRenderer.

Ideally we'd put `getTokenBoxesForVisibleCell` into a new interface, and then `TextCellRenderer` would implement that, e.g.:
```
interface TokenBoxesCellRenderer {
  getTokenBoxesForVisibleCell(...): ...;
}

isTokenBoxesCellRenderer(cellRenderer: CellRenderer): cellRenderer is TokenBoxesCellRenderer {
  return (cellRenderer as TokenBoxesCellRenderer)?.getTokenBoxesForVisibleCell !== undefined;
}
```

And then here it would be something like:
```
const cellRendererType = model.rendererForCell(column, row);
const cellRenderer = renderer.getCellRenderer(cellRendererType);
if (!isTokenBoxesCellRendrerer(cellRenderer)) {
  return;
}
...
```

packages/grid/src/CellRenderer.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable class-methods-use-this */
+import { getOrThrow } from '@deephaven/utils';
+import { isExpandableGridModel } from './ExpandableGridModel';
+import { VisibleIndex, Coordinate, BoxCoordinates } from './GridMetrics';
+import { GridRenderState } from './GridRenderer';
```
mofojed, line 5: I recall the motivation for why I wanted to export all the grid types to their own files, but that may have been a bit extreme for what we need here.
The circular import is here, CellRenderer imports GridRenderer, which also imports the CellRenderer type.
Instead, move `EditingCell` and `GridRenderState` out to a `GridRendererTypes.ts` file, and just import `GridRenderState` from there for both, then you no longer have a circular dependency.

packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -0,0 +1,615 @@
...
+
+      const positiveColor = model.positiveColorForCell(modelColumn, modelRow);
+      const negativeColor = model.negativeColorForCell(modelColumn, modelRow);
+      const positiveColorOklab = linearSRGBToOklab(hexToRgb(positiveColor));
+      const negativeColorOklab = linearSRGBToOklab(hexToRgb(negativeColor));
+      const middleColor: Oklab = lerpColor(
+        negativeColorOklab,
+        positiveColorOklab,
+        0.5
+      );
```
mofojed, line 193: Double check with @dsmmcken how we should be getting this middle colour, or if it should be defined separately or something. Right now the gradient looks a little washed compared to the examples.

dsmmcken: I was thinking if the user calls databar with a single positive or negative color, it's not a gradient.

If they want a red-yellow-green, from -100 to 100 gradient they do negative_color=["red", "yellow"],  positive_color: ["yellow, "green"]. There is no "mid" color, each gradient is it's own half. Specify one color get solid, specify multiple, that part is gradient.

packages/grid/src/MockDataBarGridModel.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  textAlignForCell(column: number, row: number): CanvasTextAlign {
+    return this.textAlignments.get(column) ?? DEFAULT_TEXT_ALIGNMENT;
+  }
+
+  rendererForCell(column: ModelIndex, row: ModelIndex): CellRendererType {
+    return 'databar';
+  }
+
+  barColorForCell(column: ModelIndex, row: ModelIndex): string {
```
mofojed, line 151: With Don's comment, this should be:
```suggestion
  barColorForCell(column: ModelIndex, row: ModelIndex): GridColor | [GridColor, GridColor] {
```
And then add TSDocs in the interface definition about how specifying multiple colours gives it a gradient.

Also note that I still think this should be combined with all the options for the databar (e.g. `this.dataBarForCell`)

packages/grid/src/CellRenderer.ts
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable class-methods-use-this */
+import { getOrThrow } from '@deephaven/utils';
+import { isExpandableGridModel } from './ExpandableGridModel';
+import { VisibleIndex, Coordinate, BoxCoordinates } from './GridMetrics';
+import { GridRenderState } from './GridRendererTypes';
+import { GridColor } from './GridTheme';
+
+export type CellRendererType = 'text' | 'databar';
```
mofojed, line 8: We should be consistent, name it `'dataBar'` (since data bar is two words, and we're using that casing everywhere else).
Also, upon reflection, I like the name `CellRenderType` better.
```suggestion
export type CellRendererType = 'text' | 'dataBar';
```

packages/iris-grid/src/IrisGridTextCellRenderer.ts
```diff
@@ -0,0 +1,170 @@
...
+      textWidth,
+      fontWidth,
+      model.truncationCharForCell(modelColumn, modelRow)
+    );
+    context.restore();
+
+    return text !== '' && truncatedText !== text;
+  }
+}
+
```
mofojed, line 169: Override the drawTreeMarker method here and draw it with the icons

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -525,7 +525,16 @@ class IrisGridTableModelTemplate<
         return '*';
       }
     }
-    return text ?? '';
+
+    if (text === null) {
```
mofojed, line 529: Guessing this means exporting behaviour was consistent after this change? Excellent.

packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -0,0 +1,564 @@
...
+    } = metrics;
+
+    const isFirstColumn = column === firstColumn;
+    const rowHeight = getOrThrow(allRowHeights, row);
+    const modelRow = getOrThrow(modelRows, row);
+    const modelColumn = getOrThrow(modelColumns, column);
+    const rowY = getOrThrow(allRowYs, row);
+    const textAlign = model.textAlignForCell(modelColumn, modelRow);
+    const text = model.textForCell(modelColumn, modelRow);
+    const value = Number(text);
```
mofojed, line 58: We should be using `model.valueForCell` here, or even getting the value from `dataBarCellOptions`.
It's possible we have numbers that javascript can't handle so well, and we definitely do not want the renderer assuming what the exact value is from parsing the text.

packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -0,0 +1,564 @@
...
+    const { x: textX, y: textY } = GridUtils.getTextRenderMetrics(
```
mofojed, line 59: I think the text is drawing a little bit high - it doesn't appear to be vertically centered in the cell. I have not debugged why that is.

mofojed (changes requested):

packages/grid/src/DataBarCellRenderer.ts
emilyhuxng: I thought `valueForCell` is only a method on `IrisGrid` models.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1763,6 +1764,10 @@ export class GridMetricCalculator {
       cellHorizontalPadding * 2
     );
 
+    if (isDataBarGridModel(model)) {
+      return columnWidth + 90;
+    }
```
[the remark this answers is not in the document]
emilyhuxng: I put it in the `iterateAllItems` but it will add 90 each time a dataBar cell is found. Could it be added afterwards? Instead of `isDataBarGridModel(model)`, it would be `if (renderType === 'dataBar')`.

packages/grid/src/DataBarCellRenderer.ts
mofojed: Okay in that case you should return this value as a number already as part of the dataBarOptions returned by the model.

packages/grid/src/GridMetricCalculator.ts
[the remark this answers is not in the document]
emilyhuxng:
```
row => {
        const modelRow = this.getModelRow(row, state);
        const cellRenderType = model.renderTypeForCell(modelColumn, modelRow);

        if (text) {
          const cellPadding = cellHorizontalPadding * 2;
          columnWidth = Math.max(
            columnWidth,
            text.length * fontWidth + cellPadding
          );
        }

        if (cellRenderType === 'dataBar') {
          columnWidth += 90;
        }
      }
```

Do you mean like this?

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1745,13 +1745,20 @@ export class GridMetricCalculator {
...
           columnWidth = Math.max(
             columnWidth,
             text.length * fontWidth + cellPadding
           );
         }
+
+        if (cellRenderType === 'dataBar') {
+          // columnWidth = Math.max(columnWidth, columnWidth + 90);
+          hasDataBar = true;
+        }
```
mofojed, line 1761: Ah yea, I'd rather have the logic contained in here, rather than setting `hasDataBar` and adding it at the end. Reason being that as we add different cell renderers, the logic will apply on a cell by cell basis, and we may want to (/should) delegate the metric calculation to the renderer (or a `CellMetricCalculator` registered for that `renderType`). We should also define `dataBarHorizontalPadding` within the theme.
For now we can just do something like:
```
        const cellRenderType = model.renderTypeForCell(modelColumn, modelRow);

        let cellWidth = 0;
        if (text) {
          const cellPadding = cellHorizontalPadding * 2;
          cellWidth = text.length * fontWidth + cellPadding;
        }

        if (cellRenderType === 'dataBar') {
          cellWidth += dataBarHorizontalPadding;
        }
        columnWidth = Math.max(columnWidth, cellWidth);    
```

packages/iris-grid/src/IrisGridDataBarCellRenderer.ts
```diff
@@ -0,0 +1,78 @@
...
+  vsTriangleRight,
+  IconDefinition,
+} from '@deephaven/icons';
+import { IrisGridRenderState } from './IrisGridRenderer';
+
+const ICON_SIZE = 16;
+const ICON_NAMES = Object.freeze({
+  CARET_DOWN: 'caretDown',
+  CARET_RIGHT: 'caretRight',
+});
```
mofojed, line 17: This icon stuff is now copied in multiple places, let's refactor it so we can use it from the same places. In fact, we can actually cache the same icon across all instances of grids (right now we have separate icons cached for each grid, and even each cell renderer).
This code was also written in old JS times, we can take advantage of TypeScript to make it more well defines.
Create a new file called `IrisGridIcons.ts`, and add something like the following:
```
export const ICON_SIZE = 16;

export type IconName = 'sortUp'  | 'sortDown' | 'caretDown' | 'caretRight' | 'cellOverflow';

const iconMap = new Map<IconName, IconDefinition>([
  ['sortUp', dhSortUp],
  ...
]);

const makeIcon = memoize((name: IconName) => {
    const path = Array.isArray(faIcon.icon[4])
      ? faIcon.icon[4][0]
      : faIcon.icon[4];
    const icon = new Path2D(path);
    const scaledIcon = new Path2D();
    const scaleMatrix = {
      a: ICON_SIZE / faIcon.icon[0],
      d: ICON_SIZE / faIcon.icon[1],
    };
    scaledIcon.addPath(icon, scaleMatrix);
   return scaledIcon
}, { max: 1000 });

export function getIcon(name: IconName): Path2D {
  return makeIcon(name);  
}
```

Then get rid of all the icons stuff from inside the renderers, and just do something like:
```
import { getIcon } from './IrisGridIcons';

...
getIcon('caretDown')
```

packages/code-studio/src/styleguide/grid-examples/DataBarExample.tsx
```diff
@@ -98,6 +98,9 @@ function DataBarExample() {
 
   // RTL gradient with multiple colors
   data.push(columnData.slice());
+
+  // Both data bar and text
+  data.push(columnData.slice());
```
mofojed, line 103: I'd add two columns just so we get that checkerboard

mofojed (changes requested):

packages/grid/src/DataBarCellRenderer.ts
emilyhuxng: Not sure why this works but this does seem to solve the problem, however it is like a hacky solution: [https://stackoverflow.com/a/63239887/20005358](https://stackoverflow.com/a/63239887/20005358)

mofojed: Hmm. It looks like the vertical position is actually the same as it is currently...
I'll defer to @dsmmcken for this, but I think for the data bars in particular I'd bump them down a couple of pixels, just to vertically center it. Not sure if our regular text cell rendering is off as well.

Did a bit of debugging by drawing a red line where the `textX`/`textY` is:
image

Did this just be adding the following in:
```
    context.strokeStyle = 'red';
    context.beginPath();
    context.moveTo(textX - 50, textY);
    context.lineTo(textX + 50, textY);
    context.stroke();
    context.closePath();
```

So we're using the `textBaseline = 'middle'`, which draws the text with the middle: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline
That includes descenders, but numbers don't ever descend past the baseline, which is possibly why it looks so high/off.

The [example you linked](https://stackoverflow.com/a/63239887/20005358) is using the bounding box descent, which we don't really care about (since we know numbers don't descend).

What we could do instead is set `textBaseline = 'bottom';`, then we position the text to vertically center it; we're using a monospace font so all numbers should be the same height and we only need to measure once (i.e. just measure `'1234567890'`) and can cache that result.
Then we can position it within the cell on the bottom baseline, so something like `cellTextY = (rowHeight - textHeight) * 0.5`

packages/iris-grid/src/IrisGridTextCellRenderer.ts
```diff
@@ -0,0 +1,194 @@
...
+      : getIcon('caretRight');
+    const iconX = columnX + x1 - 2;
+    const iconY = rowY + y1 + 2.5;
+
+    context.fillStyle = color;
+    context.textAlign = 'center';
+    context.translate(iconX, iconY);
+    context.fill(markerIcon);
+    context.restore();
+  }
```
mofojed, line 191: Could put this in a util method that's shared between IrisGridTextCellRenderer and IrisGridDataCellRenderer

mofojed (commented):
Looking great! As discussed, will not be merging until start of May.

mofojed (approved):

mofojed (approved):

PR #1214, opened 2023-04-10
feat: Improve plugin load error handling

mattrunyon:
Loads plugins individually instead of failing to register if any plugin fails to load. Previously if any plugin failed to load for any reason then no plugins were registered.

Does some validation on if the manifest file exists and the structure of it so that the errors are more descriptive. Does not throw a 2nd error about `Unexpected token <` if there is no manifest file

### DIFF
--- a/packages/code-studio/src/main/AppInit.tsx
+++ b/packages/code-studio/src/main/AppInit.tsx
@@ -73,6 +73,10 @@ async function loadPlugins(): Promise<DeephavenPluginModuleMap> {
       `${import.meta.env.VITE_MODULE_PLUGINS_URL}/manifest.json`
     );
 
+    if (!Array.isArray(manifest.plugins)) {
+      throw new Error('Plugin manifest JSON does not contain plugins array');
+    }
+
     log.debug('Plugin manifest loaded:', manifest);
     const pluginPromises: Promise<unknown>[] = [];
     for (let i = 0; i < manifest.plugins.length; i += 1) {
--- a/packages/code-studio/src/main/AppInit.tsx
+++ b/packages/code-studio/src/main/AppInit.tsx
@@ -74,7 +74,7 @@ async function loadPlugins(): Promise<DeephavenPluginModuleMap> {
     );
 
     log.debug('Plugin manifest loaded:', manifest);
-    const pluginPromises: Promise<DeephavenPluginModule>[] = [];
+    const pluginPromises: Promise<unknown>[] = [];
     for (let i = 0; i < manifest.plugins.length; i += 1) {
       const { name, main } = manifest.plugins[i];
       const pluginMainUrl = `${
@@ -89,7 +89,7 @@ async function loadPlugins(): Promise<DeephavenPluginModuleMap> {
       const module = pluginModules[i];
       const { name } = manifest.plugins[i];
       if (module.status === 'fulfilled') {
-        pluginMap.set(name, module.value);
+        pluginMap.set(name, module.value as DeephavenPluginModule);
       } else {
         log.error(`Unable to load plugin ${name}`, module.reason);
       }
--- a/packages/code-studio/src/main/AppInit.tsx
+++ b/packages/code-studio/src/main/AppInit.tsx
@@ -80,10 +80,10 @@ async function loadPlugins(): Promise<DeephavenPluginModuleMap> {
     log.debug('Plugin manifest loaded:', manifest);
     const pluginPromises: Promise<unknown>[] = [];
     for (let i = 0; i < manifest.plugins.length; i += 1) {
-      const { name, main } = manifest.plugins[i];
+      const { main } = manifest.plugins[i];
       const pluginMainUrl = `${
         import.meta.env.VITE_MODULE_PLUGINS_URL
-      }/${name}/${main}`;
+      }/${main}`;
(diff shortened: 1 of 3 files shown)


packages/code-studio/src/plugins/PluginUtils.tsx
```diff
@@ -58,21 +58,21 @@ class PluginUtils {
...
   ): Promise<{ plugins: { name: string; main: string }[] }> {
-    return new Promise((resolve, reject) => {
-      const request = new XMLHttpRequest();
-      request.addEventListener('load', () => {
-        try {
-          const json = JSON.parse(request.responseText);
-          resolve(json);
-        } catch (err) {
-          reject(err);
+    return fetch(jsonUrl).then(res => {
```
mofojed, line 61: Write this with await to be consistent, rather than mixing .then chaining with async/await.
```
const res = await fetch(jsonUrl);
...
```

packages/code-studio/src/plugins/PluginUtils.tsx
```diff
@@ -58,21 +58,21 @@ class PluginUtils {
...
-      request.addEventListener('error', err => {
-        reject(err);
-      });
-      request.open('GET', jsonUrl);
-      request.send();
+        return { plugins: [] };
+      }
+
+      try {
+        return res.json();
```
mofojed, line 71: `res.json()` returns a promise, so the `try`/`catch` block here does nothing. Need to `await` it, and should log the error itself as well... but actually, we shouldn't return `{ plugins: [] }` in this case anyway, we should just `throw`. `loadPlugins()` already needs to handle a promise rejection, and this is a good case to just reject the promise. I could also argue not being able to load the manifest is still an error and just throw that, just returning `{ plugins: [] }` if the manifest doesn't exist (possibly with logging a warning.

mofojed (changes requested):

packages/code-studio/src/plugins/PluginUtils.tsx
mattrunyon: Good catch

As for the other part, I think I initially did it so there was only 1 error in the log if the manifest 404ed. I can throw and it will at least be a better 2nd error than it was before w/ the parsing error message

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -72,21 +73,30 @@ async function loadPlugins(): Promise<DeephavenPluginModuleMap> {
...
     log.debug('Plugin manifest loaded:', manifest);
-    const pluginPromises = [];
+    const pluginPromises: Promise<unknown>[] = [];
     for (let i = 0; i < manifest.plugins.length; i += 1) {
-      const { name, main } = manifest.plugins[i];
+      const { main } = manifest.plugins[i];
       const pluginMainUrl = `${
         import.meta.env.VITE_MODULE_PLUGINS_URL
-      }/${name}/${main}`;
+      }/${main}`;
```
mofojed, line 86: Why are you making this a breaking change? Don't, I don't want the churn. See discussion we're having with Devin about packaging the JS with python for delivery, and perhaps that invalidates the manifest entirely, or changes the format there, but I'd rather the breaking changes be when that gets implemented instead of breaking it twice.

mofojed: Just put the manifest.json file in the js-plugins folder in the `plugins` folder

mattrunyon: Ya I thought of that halfway through and didn't go back and edit. Either way we need better documentation around it

mofojed: For production cases, yes the [core docs](https://deephaven.io/core/docs/how-to-guides/configuration/js-plugins/#quickstart) are best, and that extracts the files and generates the manifest.json. We're not updating that tooling.
For dev mode, yes we need to update the readme in deephaven-js-plugins, likely instruct to have it just in the plugins directory instead, and the name just being the location of the package.json.

With python wrapped stuff, we'll likely just get rid of the manifest.json entirely or have a new format.

mofojed (changes requested):

packages/code-studio/src/plugins/PluginUtils.tsx
```diff
@@ -58,22 +58,15 @@ class PluginUtils {
...
-      });
-      request.addEventListener('error', err => {
-        reject(err);
-      });
-      request.open('GET', jsonUrl);
-      request.send();
-    });
+    const res = await fetch(jsonUrl);
+    if (!res.ok) {
+      throw res.statusText;
```
mofojed, line 63:
```suggestion
      throw new Error(res.statusText);
```

packages/code-studio/vite.config.ts
```diff
@@ -131,6 +131,7 @@ export default defineConfig(({ mode }) => {
       },
     },
     optimizeDeps: {
+      force: true,
```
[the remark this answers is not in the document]
mattrunyon: Something I did locally b/c I needed to bust the Vite dependency cache and forgot to remove

mofojed (changes requested):
Would recommend throwing an error instead of a string

packages/code-studio/src/plugins/PluginUtils.tsx
mattrunyon: Ya good call

mofojed (approved):

PR #1217, opened 2023-04-11
fix: Fix OneClick links not filtering plots

### LINKED ISSUE #1198: Linker with one click plots doesn't filter plot
**Description**

Attempting to filter a one_click plot with value supplied via linker doesn't filter the selectable_dataset. Issue present in DHC and DHE, will require hotfix.

Reported by @rbasralian in Silverheels 1.20230131.113 on Slack.

**Steps to reproduce**

1. Use the following query:
    ```
    from deephaven.plot.selectable_dataset import one_click
    
            "MyString=new String(`a`+(char) ((i%3)+97))",
            "MyInt=new Integer(i)",
            "MyLong=new Long(i % 3)",
            "MyFloat=new Float(i+i/10)",
    
    
    toc = one_click(simple_ticking, by=["MyString"], require_all_filters=True)
    p = Figure().plot_xy(series_name="Distance", t=toc, x="Timestamp", y="MyInt").show()
    
    ```
2. Create a linker between the plot and the "MyString" column in the table.
3. Double click a row in the table simple_ticking to filter the plot by that value

image

**Expected results**

Plot is updated to show a result.

**Actual results**

Nothing happens.

Adding an input filter panel set to the "MyString" column and entering a value like `aa` does work.

**Versions**

Silverheels: 1.20230131.113 

and DHC:
Engine Version: 0.24.0
Web UI Version: 0.34.0
Java Version: 17.0.6
Barrage Version: 0.5.0

vbabich:
- Update `ChartPanel` to get the filter value from `filterList`
- Hide operator selection for filter source and chart links
- Don't allow linking multiple columns from the same table to a chart target
- Fix issue with link in progress not being reset after creating a new link

Fixes #1198

### DIFF
--- a/packages/dashboard-core-plugins/src/linker/Linker.tsx
+++ b/packages/dashboard-core-plugins/src/linker/Linker.tsx
@@ -355,10 +355,12 @@ export class Linker extends Component<LinkerProps, LinkerState> {
         }
         case 'chartLink': {
           const existingLinkEnd = isReversed === true ? start : end;
+          const existingLinkStart = isReversed === true ? end : start;
           log.debug('creating chartlink', { existingLinkEnd, start, end });
           // Don't allow linking more than one column per source to each chart column
           const linksToDelete = links.filter(
-            ({ end: panelLinkEnd }) =>
+            ({ end: panelLinkEnd, start: panelLinkStart }) =>
+              panelLinkStart?.panelId === existingLinkStart.panelId &&
               panelLinkEnd?.panelId === existingLinkEnd.panelId &&
               panelLinkEnd?.columnName === existingLinkEnd.columnName &&
               panelLinkEnd?.columnType === existingLinkEnd.columnType
--- a/packages/dashboard-core-plugins/src/linker/Linker.tsx
+++ b/packages/dashboard-core-plugins/src/linker/Linker.tsx
@@ -353,6 +353,19 @@ export class Linker extends Component<LinkerProps, LinkerState> {
           this.deleteLinks(linksToDelete);
           break;
         }
+        case 'chartLink': {
+          const existingLinkEnd = isReversed === true ? start : end;
+          log.debug('creating chartlink', { existingLinkEnd, start, end });
+          // Don't allow linking more than one column per source to each chart column
+          const linksToDelete = links.filter(
+            ({ end: panelLinkEnd }) =>
+              panelLinkEnd?.panelId === existingLinkEnd.panelId &&
+              panelLinkEnd?.columnName === existingLinkEnd.columnName &&
+              panelLinkEnd?.columnType === existingLinkEnd.columnType
+          );
+          this.deleteLinks(linksToDelete);
+          break;
+        }
         case 'tableLink':
           // No-op
           break;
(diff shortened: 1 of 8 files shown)


packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -353,6 +353,19 @@ export class Linker extends Component<LinkerProps, LinkerState> {
...
+        case 'chartLink': {
+          const existingLinkEnd = isReversed === true ? start : end;
+          log.debug('creating chartlink', { existingLinkEnd, start, end });
+          // Don't allow linking more than one column per source to each chart column
+          const linksToDelete = links.filter(
+            ({ end: panelLinkEnd }) =>
+              panelLinkEnd?.panelId === existingLinkEnd.panelId &&
+              panelLinkEnd?.columnName === existingLinkEnd.columnName &&
+              panelLinkEnd?.columnType === existingLinkEnd.columnType
+          );
```
mofojed, line 365: Shouldn't we be able to link from two different sources to the same destination? Then whatever source is selected last gets applied.

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -786,20 +785,22 @@ export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
...
       const filterValueMap = new Map(state.filterValueMap);
-
-      filterMapParam.forEach(({ columnType, value }, columnName) => {
+      filterMapParam.forEach(({ columnType, filterList }, columnName) => {
         const column = columnMap.get(columnName);
         if (column == null || column.type !== columnType) {
           return;
         }
+        if (filterList.length < 1) {
+          log.error('Invalid filterMap, filterList is empty', filterMapParam);
```
mofojed, line 800: This is just going to return for this specific param (`{ columnType, filterList }`) and the `.forEach` is going to continue. You should either include the `columnType` in the log line to indicate exactly what is failing, or throw here to get out of the `.forEach`, or use a different for loop type so you can just return.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
vbabich: We don't throw or break out of the loop when the column is null in the condition above, I thought it was ok to just skip the `filterMap`.
Should I throw in the condition above as well?

mofojed: I'm fine with skipping it, but the error logging should be consistent. In the above block, we're not logging anything; in this block, we're logging the entire `filterMapParam`, and not the `columnType` or `columnName` (which is unique to this iteration of the loop).
I'd just do a log debug:
```
log.debug('Ignoring empty filterList for column', columnName);
```

packages/dashboard-core-plugins/src/linker/Linker.tsx
vbabich: Correct, fixed.

PR #1220, opened 2023-04-12
refactor: Refactor grid types into a types file

### LINKED ISSUE #1222: Refactor types in GridRenderer into a new file
Move types in `GridRenderer` to a new file to reduce circular dependencies and clean up imports.

emilyhuxng:
Closes #1222 

Moved types spread across many files in package `grid` into `GridTypes` to reduce circular dependencies.


packages/grid/src/GridTypes.ts
```diff
@@ -0,0 +1,498 @@
+import GridModel from './GridModel';
+import GridRange from './GridRange';
```
mofojed, line 2: This is actually creating a circular import right now; GridRange imports GridTypes, and GridTypes imports GridRange. Likewise, GridModel imports GridTypes and vice versa.
What was the original circular import we were trying to avoid with this change? I think need to provide a bit more guidance here on what cleanup is needed

mofojed (changes requested):

packages/grid/src/GridTypes.ts
emilyhuxng: We were originally trying to avoid `CellRenderer` importing `GridRenderState` from `GridRenderer`, which would also import `CellRenderer`.

mofojed: Right, see my comment: https://github.com/deephaven/web-client-ui/pull/1212#discussion_r1166971643
I think that's really all that's needed

emilyhuxng:
The changes here are no longer needed. Opened a new [PR](https://github.com/deephaven/web-client-ui/pull/1234) with the needed changes.

Issue #1221, opened 2023-04-12
DH-14630: ACL Editor - usePromiseFactory hook


bmingles:
Supports Enterprise DH-14630

As a developer, I want to be able to use React hooks to connect DH tables with React Spectrum. This will be needed for ACL Editor support.

bmingles:
There are some other hooks to associate with this issue

PR #1226, opened 2023-04-13
feat: usePromiseFactory hook

bmingles:
Enterprise DH-14630
Community #1221 

Custom React hook for using Promise factory functions.

### DIFF
--- a/packages/react-hooks/src/usePromiseFactory.ts
+++ b/packages/react-hooks/src/usePromiseFactory.ts
@@ -1,5 +1,5 @@
 /* eslint-disable import/prefer-default-export */
-import React from 'react';
+import { useState, useCallback, useEffect } from 'react';
 
 /**
  * Configuration options for `usePromiseFactory` hook.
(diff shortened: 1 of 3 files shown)


packages/react-hooks/src/usePromiseFactory.ts
```diff
@@ -0,0 +1,83 @@
...
+  const [isLoading, setIsLoading] = React.useState(false);
+
+  const loadPromise = React.useCallback(async () => {
+    setIsLoading(true);
+
+    try {
+      const resolved = await promiseFactory(...args);
+      setData(resolved);
+      setError(null);
+      return resolved;
```
mattrunyon, line 59: Why return here? Are there use cases where you want to call this imperatively?

```ts
const { data, reload } = usePromiseFactory(...);
async function handleClick() {
  const val = await reload();
  // Do something w/ val you can't as cleanly in the main render?
}
```

You wouldn't be able to tell if the promise failed aside from the data being null. And the re-render would have the up-to-date data

mattrunyon (changes requested):

packages/react-hooks/src/usePromiseFactory.ts
bmingles: Agreed. I removed the return

mattrunyon (approved):

PR #1227, opened 2023-04-14
feat: Added support for null in useTableListener

### LINKED ISSUE #1228: useTableListener support null
As a developer I'd like for `useTableListener` to support `null` (it already supports `undefined`) for the `eventEmitter` arg.

bmingles:
resolves #1228

### DIFF
--- a/packages/jsapi-components/src/useTableListener.test.ts
+++ b/packages/jsapi-components/src/useTableListener.test.ts
@@ -1,0 +1,48 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { Evented } from '@deephaven/jsapi-types';
+import useTableListener from './useTableListener';
+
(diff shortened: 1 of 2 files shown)


mattrunyon:
Is this just a nicety from the types perspective? As in most places we would call this we'd need to do `useTableListener(table ?? undefined)` right now? Or is there a distinct reason for null value?

Looks like there's other hooks (`useTable`, `useTableColumn`) that would want the same treatment if that's the case, but the few use cases the table type is `Table | undefined`

bmingles:
This just happened to be for a use case I am working on another PR. Also, it was a chance to add test coverage to this one hook but yes, this is just a convenience to avoid the coalescing.

vbabich (approved):

PR #1230, opened 2023-04-17
feat: DH-14630 useViewportData + supporting utils

### LINKED ISSUE #1221: DH-14630: ACL Editor - usePromiseFactory hook

bmingles:
Supports Enterprise DH-14630

This PR contains a number of custom hooks + utils to support windowed viewports for the ACL Editor.

resolves #1221


packages/jsapi-components/src/useSelectDistinctTable.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export default function useSelectDistinctTable(
+  table: Table | TreeTable | null,
+  ...columnNames: string[]
+) {
+  const selectDistinct = useCallback(
+    async () => table?.selectDistinct(table.findColumns(columnNames)) ?? null,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [table, ...columnNames]
+  );
```
mofojed, line 19: Guess you needed to disable the exhaustive deps because you're spreading the columnNames... any reason to not just pass in column names as an immutable array, and then use that?

packages/jsapi-utils/src/ViewportDataUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+  return table == null || isClosed(table) ? 0 : table.size;
+}
+
+/**
+ * Check if a given table is closed. Tree tables don't have an `isClosed` prop,
+ * so will always return false.
+ * @param table The table to check if it is closed.
+ */
+export function isClosed(table: Table | TreeTable): boolean {
+  if ('isClosed' in table) {
```
mofojed, line 128: I'd add a following ticket in deephaven-core to add the `isClosed` property to `TreeTable`. Especially since it already has a `closed` variable: https://github.com/deephaven/deephaven-core/blob/226a428be5e746a2f1fbf07264a742769631c4c2/web/client-api/src/main/java/io/deephaven/web/client/api/tree/JsTreeTable.java#L358
... and that `closed` variable never gets set. So something seems amiss there. This check is good though.

packages/jsapi-utils/src/ViewportDataUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @param row Row to map to an item.
+ * @param columns Columns to map.
+ * @returns A key / value object for the row.
+ */
+export function defaultRowDeserializer<T>(
+  row: ViewportRow,
+  columns: Column[]
+): T {
+  return columns.reduce((result, col) => {
+    // eslint-disable-next-line no-param-reassign
```
mofojed, line 89: Wish there was a way to disable this eslint just in reduce functions for that particular parameter.

mofojed (changes requested):
Extremely minor cleanup, some comments.

packages/jsapi-components/src/useInitializeViewportData.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Initializes a ListData instance that can be used for windowed views of a
+ * Table. The list must always contain a KeyedItem for ever record in the table,
+ * so it is pre-populated with empty items that can be updated with real data as
+ * the window changes.
+ *
+ * IMPORTANT: this will create an empty KeyedItem object for every row in the
+ * source table. This is intended for "human" sized tables such as those used in
+ * admin panels. This is not suitable for "machine" scale with millions+ rows.
+ * @param table The table that will be used to determine the list size.
```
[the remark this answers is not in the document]
mofojed: Nevermind, this is just the initial viewport - ignore.

packages/jsapi-components/src/useSelectDistinctTable.ts
bmingles: hmm, that hadn't occurred to me, but it seems that puts more burden on the caller vs just being able to

`useSelectDistinctTable(table, 'blah', 'bleh')`

I can add a comment to clarify unless you feel strongly about passing in an array

packages/jsapi-utils/src/ViewportDataUtils.ts
bmingles: I created https://github.com/deephaven/deephaven-core/issues/3714 to follow up on this.

packages/jsapi-components/src/useSelectDistinctTable.ts
mofojed: Yea I don't feel too strongly about it.

mofojed (approved):

PR #1244, opened 2023-04-24
feat: Logging out

mofojed:
- Write auth plugin components as wrappers
  - Allows plugins to always be mounted and control the transition to loading the app
  - Needed to ensure keycloak plugin can wipe token on logout
- Refactor auth plugins to just use `AuthPluginBase`
  - There was a lot of the same code, just refactored it into one common base
  - Also added a `useContextOrThrow` hook
- Added login screen for `AuthPluginPsk`
  - Took the `LoginAnimation` from Enterprise, renamed to `RandomAreaPlotAnimation`
  - Added a `Login` and `LoginForm` components to allow others to reuse if desired
  - Also stores the pre-shared key as a cookie after successful login
- Add broadcast channel for notifying other tabs about login/logout
  - Notifies after successful login, or when logout button is pressed
- Added permission for `canLogout` (since you can't logout when anonymous
- Add context providers for overriding user info and permissions (used by auth plugins)


packages/components/src/RandomAreaPlotAnimation.tsx
mofojed, line 1: This is just the LoginAnimation copied over from Enterprise, I did not make any changes to it other than the name.

mofojed:
Getting an error when trying to import from `auth-keycloak`:
```
@deephaven/js-plugin-auth-keycloak: build started...
@deephaven/js-plugin-auth-keycloak: ✓ 6190 modules transformed.
@deephaven/js-plugin-auth-keycloak: Could not resolve "./Login.css" from "../../node_modules/@deephaven/auth-plugins/dist/Login.js"
@deephaven/js-plugin-auth-keycloak: file: /home/bender/dev/deephaven/deephaven-js-plugins/node_modules/@deephaven/auth-plugins/dist/Login.js
```

packages/auth-plugins/src/AuthPluginPsk.tsx
```diff
@@ -1,66 +1,216 @@
...
+  log.debug2('clearWindowToken');
+  const params = new URLSearchParams(window.location.search);
+  params.delete(PSK_QUERY_PARAM_KEY);
+
+  let queryString = `${params}`;
+  if (queryString.length > 0) {
+    queryString = `?${queryString}`;
+  }
+  const newPath = `${window.location.pathname}${queryString}${window.location.hash}`;
+  window.history.replaceState(null, '', newPath);
```
mattrunyon, line 36:
```suggestion
  const url = new URL(window.location);
  url.searchParams.delete(PSK_QUERY_PARAM_KEY);

  window.history.replaceState(null, '', url.toString());
```

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -132,9 +139,13 @@ function AppInit(props: AppInitProps) {
...
 
-          const loadedWorkspace = await workspaceStorage.load({
-            isConsoleAvailable: sessionWrapper !== undefined,
-          });
+          const [configs, loadedWorkspace] = await Promise.all([
+            client.getServerConfigValues(),
+            workspaceStorage.load({
+              isConsoleAvailable: sessionWrapper !== undefined,
+            }),
+          ]);
```
mattrunyon, line 147: Just making sure either of these erroring should lead to the app rendering an error message. Should be caught by the error boundary I think?

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -957,6 +952,7 @@ export class AppMainContainer extends Component<
           }
           subtitle="Please check your network connection."
         />
+        {/* TODO: Reconnect auth fail - part of the plugin?? AppBootstrap? */}
```
mattrunyon, line 955: Ticket for this?

packages/jsapi-components/src/RefreshTokenUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const RefreshTokenContext = createContext<RefreshToken | null>(null);
+
+/**
+ * Read the refresh token from our cookie store
+ * @returns RefreshToken if it exists, null otherwise
+ */
+export function readRefreshToken(): RefreshToken | null {
+  const cookieToken = Cookies.get(REFRESH_TOKEN_KEY);
+  return cookieToken != null ? JSON.parse(cookieToken) : null;
```
mattrunyon, line 16: This may throw if the token fails to parse as JSON (i.e. a non-JSON string). Have you checked what happens if that's the case? I'm guessing this _shouldn't_ happen since we're setting the cookie typically, but want to make sure it doesn't just break everything (or at least breaks in a somewhat reasonable way)

packages/jsapi-components/src/RefreshTokenUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Store the provided refresh token as a cookie
+ * @param token The refresh token to store
+ */
+export function storeRefreshToken(token: RefreshToken | null) {
+  if (token != null) {
```
mattrunyon, line 24: Could just do this to move the bulk of the code up 1 indent

```
if (token == null) {
  Cookies.remove(REFRESH_TOKEN_KEY);
  return;
}

const cookieToken = ...
```

packages/jsapi-components/src/useBroadcastChannel.ts
```diff
@@ -0,0 +1,39 @@
...
+  const channel = useMemo(() => new BroadcastChannel(name), [name]);
+  useEffect(
+    () => () => {
+      channel.close();
+    },
+    [channel]
+  );
+
+  useEffect(() => {
+    channel.onmessage = (event: MessageEvent) => {
```
mattrunyon, line 25: You can also do `channel.addEventListener('message', (event) => {})`. They co-exist w/ the `onmessage` it seems although you can only set `onmessage` to 1 handler

Maybe add a comment here that (I'm assuming) this is intentionally **NOT** `addEventListener` so that you can set the new listener without needing to remove the old listener. Or change to cleanup the event listener from the effect if you switch to `addEventListener`

mattrunyon (changes requested):

jest.config.base.cjs
```diff
@@ -26,6 +26,10 @@ module.exports = {
     ),
     // Handle monaco worker files
     '\\.worker.*$': 'identity-obj-proxy',
+    'react-transition-group': path.join(
+      __dirname,
+      './__mocks__/react-transition-group.js'
+    ),
```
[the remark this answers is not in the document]
mofojed: Yea if we don't have this here the mock doesn't get applied and all the tests fail. Guess it's how we have jest set up.
The automatic mocking happens for `__mocks__` folders next to the node_modules folders of our packages, but not at the root it would appear.
I'm not too invested in figuring it out beyond that, perhaps we switch to `vitest` at some point.

package-lock.json
```diff
@@ -26828,6 +26862,7 @@
         "prettier": "^2.2.1"
       }
     },
+    "packages/react-hook": {},
```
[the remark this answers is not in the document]
mofojed: Hmm. Doing an `npm install` doesn't show any differences... but you're right, this is an error and there's other parts in the package-lock which have a similar error, e.g.:
```
    "packages/jsapi-components/node_modules/@deephaven/react-hooks": {
      "resolved": "packages/log",
      "link": true
    },
```
We fixed a similar issue with #1247 , #1248... it's worrying that `npm install` isn't identifying any issues though.
Might just delete package-lock.json and regenerate, since it does not look right...

packages/app-utils/src/components/AppBootstrap.test.tsx
```diff
@@ -19,13 +20,42 @@ jest.mock('../plugins', () => ({
...
 
+const mockChannel = {
+  postMessage: jest.fn(),
+};
+jest.mock('@deephaven/jsapi-components', () => ({
+  ...jest.requireActual('@deephaven/jsapi-components'),
+  RefreshTokenBootstrap: jest.fn(({ children }) => children),
+  useBroadcastChannel: jest.fn(() => mockChannel),
+  useBroadcastLoginListener: jest.fn(),
+}));
```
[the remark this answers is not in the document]
mofojed: Yea exactly. Was only used in two spots... and I wasn't sure I wanted a global mock for jsapi-components mocking out the broadcast listener, or if we could do a global mock for `BroadcastChannel`, or something.

packages/jsapi-components/src/useBroadcastLoginListener.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  BroadcastLoginMessage,
+  BroadcastLogoutMessage,
+  isBroadcastLoginMessage,
+  isBroadcastLogoutMessage,
+  Message,
+} from '@deephaven/jsapi-utils';
```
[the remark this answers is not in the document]
mofojed: Yea - I agree, `jsapi-utils`, `jsapi-components`, and `jsapi-bootstrap` (perhaps jsapi-types as well) can likely all be combined into just `jsapi`. I'll leave that as a future change. The biggest difference right now is that `jsapi-utils` doesn't depend on React, and it depends on `jsapi-shim`. The boundary for jsapi-types makes sense on it's own, but it could just as easily be in the same package.
`jsapi-shim` should be deprecated, which we should do after @vbabich 's de-globalization change is ready to go.

packages/jsapi-utils/src/MessageUtils.ts
```diff
@@ -10,17 +10,69 @@ export const LOGIN_OPTIONS_REQUEST =
 export const SESSION_DETAILS_REQUEST =
   'io.deephaven.message.SessionDetails.request';
 
+/**
+ * Use a BroadcastChannel for sending messages between tabs, such as when the user logs out.
+ * Something that isn't clear from the docs - for inter-browser communication, you must use `postMessage` and `onmessage`.
+ * `BroadcastChannel` is an `EventTarget`, and does support `addEventListener`/`dispatchEvent` etc... but that's only within the same instance of that object!
+ * https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API
+ */
```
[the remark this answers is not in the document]
mofojed: Yea I like that for themes/user preferences. It might be able to work for this as well, I was just porting over the existing functionality we've got from Enterprise.

packages/jsapi-utils/src/MessageUtils.ts
```diff
@@ -10,17 +10,69 @@ export const LOGIN_OPTIONS_REQUEST =
...
+export interface BroadcastLogoutMessage extends Message<void> {
+  message: typeof BROADCAST_LOGOUT_MESSAGE;
 }
 
 export interface Response<T> {
   id: string;
   payload: T;
 }
 
+export function isMessage<T>(obj: unknown): obj is Message<T> {
```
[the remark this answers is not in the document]
mofojed: No default in the `Message` interface. You're right this should just be `Message<unknown>` though, other checks for a specific payload type should check the `message` type.

packages/react-hooks/src/useContextOrThrow.ts
```diff
@@ -0,0 +1,20 @@
+import React, { useContext } from 'react';
+
+/**
+ * Fetch a given context or throw an error if it is null.
+ * @param context Context to use
+ * @param message Message to throw if context is null
```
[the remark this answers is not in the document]
mofojed: These comments were brought to you by Copilot. Added " or undefined".

packages/redux/src/store.ts
```diff
@@ -24,13 +24,17 @@ export interface UserPermissions {
   canCreateQueryMonitor: boolean;
   canCopy: boolean;
   canDownloadCsv: boolean;
+  canLogout: boolean;
```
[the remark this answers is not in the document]
mofojed: > Every other auth method should have some sort of logout, no?
Not sure that's true - we don't _know_ what every other auth logout method could be. I think it's alright as a permission.

packages/auth-plugins/src/AuthPluginBase.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: AuthPluginBaseProps): JSX.Element {
+  const client = useClient();
+  const [error, setError] = useState<unknown>();
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+
+  useEffect(() => {
+    let isCanceled = false;
+    function verifyNotCanceled() {
+      if (isCanceled) {
+        throw new CanceledPromiseError('Login canceled.');
```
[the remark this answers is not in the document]
mofojed: Everywhere it's called is wrapped in a `try`/`catch`

packages/auth-plugins/src/AuthPluginBase.tsx
```diff
@@ -0,0 +1,84 @@
...
+        const loginOptions = await getLoginOptions();
+        verifyNotCanceled();
+
+        log.info('Logging in...');
+        await client.login(loginOptions);
+        verifyNotCanceled();
+
+        setIsLoggedIn(true);
+      } catch (e) {
+        if (!isCanceled) {
```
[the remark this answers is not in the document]
mofojed: No, I don't want it to throw here, this is where it catches the canceled error. Perhaps I should just inline the `verifyNotCanceled` method above and get rid of that method?

packages/auth-plugins/src/AuthPluginParent.tsx
```diff
@@ -1,54 +1,40 @@
...
 import { AuthPlugin, AuthPluginProps } from './AuthPlugin';
+import AuthPluginBase from './AuthPluginBase';
+import {
+  UserPermissionsOverride,
+  UserPermissionsOverrideContext,
+} from './UserContexts';
 
 const log = Log.module('AuthPluginParent');
 
+const permissionsOverrides: UserPermissionsOverride = { canLogout: false };
```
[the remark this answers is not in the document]
mofojed: Not at this time, anyway. This is used in DnD in silverheels, and also when opening an IDE up for a community worker from Enterprise. Logging out from the child would need to logout of the parent as well, and I wasn't sure we wanted that behaviour. As you can't logout from the child right now, leaving it as is.

packages/auth-plugins/src/AuthPluginPsk.tsx
mofojed: Geez I really took the stupid way with this one, good cleanup.

packages/code-studio/src/main/AppInit.tsx
mofojed: Yea it'll be caught just below and error shown.

packages/code-studio/src/main/AppMainContainer.tsx
mofojed: Filed #1256, and I'll just remove the comment from here. It was a note for later, and an omission when implementing #1149.

packages/jsapi-components/src/RefreshTokenUtils.ts
mofojed: Yea this would be a problem on Enterprise as well.

Agreed, shouldn't be a problem, but I should wrap in a try catch anyway.

packages/jsapi-components/src/useBroadcastChannel.ts
mofojed: Yes, `addEventListener` does not do what you would expect and would be wrong. I have a comment about this in `MessageUtils`, but I'll move that comment over here instead as it's more relevant here:
```
/**
 * Use a BroadcastChannel for sending messages between tabs, such as when the user logs out.
 * Something that isn't clear from the docs - for inter-browser communication, you must use `postMessage` and `onmessage`.
 * `BroadcastChannel` is an `EventTarget`, and does support `addEventListener`/`dispatchEvent` etc... but that's only within the same instance of that object!
 * https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API
 */
```

mattrunyon: I tried `addEventListener` with `postMessage` and it seemed to work. Just FYI

mofojed: Weird. I wonder why I previously made that comment, as I can find no evidence now that addEventListener ever behaved differently. Will just use `addEventListener` and remove the comment.

package-lock.json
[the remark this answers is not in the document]
mofojed: I must have screwed something up with it while merging previous changes. In any case, I just copied the `package.json` from `main` and ran an `npm install` again, and now it looks correct.

packages/jsapi-utils/src/MessageUtils.ts
```diff
@@ -127,7 +124,7 @@ export async function requestParentResponse<T>(
       }
       window.clearTimeout(timeoutId);
       window.removeEventListener('message', listener);
-      resolve(data.payload);
+      resolve(data.payload as T);
```
mattrunyon, line 127: I think I'd rather this util return `Promise<unknown>` since it doesn't do any validation. Then it's up to consumption to either validate it or cast it if they don't care about the validation really

Just makes the util more clear

mattrunyon (changes requested):

packages/auth-plugins/src/Login.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { RandomAreaPlotAnimation } from '@deephaven/components';
+import Logo from './logo.png';
```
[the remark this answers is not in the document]
mofojed: It is something we do with DHE, which has some server side configuration: https://github.com/deephaven-ent/iris/blob/d6950c95398cbe9a400eca8ba90e11201dcdae7c/web/server-frontend/src/main/java/com/illumon/iris/web/server/AbstractIrisWebServer.java#L170
We don't have that same configuration on the DHC side.

That being said, I'll wire it up the same on the client. Just means the logo is in 3 public paths (ide, embed-grid, embed-chart) instead of just being packaged with auth plugins.

And yes, a custom auth plugin would be able to customize whatever.

mattrunyon (approved):
Looks good. Just confirming we intentionally remove the PSK from the URL after PSK auth is complete? Still refreshes fine, just making sure that's intentional since I thought that before this change the key stayed in the URL

PR #1249, opened 2023-04-25
fix: Move @deephaven/redux to be a dependency instead

mofojed:
- Issue when trying to do an `npm install` in another project that depended on @deephaven/dashboard, as the peerDependency would not resolve the "file:../redux" path
  - Not sure if lerna or npm should automatically be replacing those at publish time?
- Was adding redux as a peerDep because we only want one instance of the redux store; otherwise weird bugs can occur that are difficult to track down
- It was already the case with @deephaven/dashboard-core-plugins that it was just a dependency, not a peerDependency. Enterprise handled it just by doing deduplication
- Just move @deephaven/redux to be a dependency instead of a peerDependency

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25697,12 +25697,12 @@
         "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
         "@deephaven/jsapi-types": "file:../jsapi-types",
         "@deephaven/log": "file:../log",
+        "@deephaven/redux": "file:../redux",
         "@paciolan/remote-component": "2.13.0",
         "@paciolan/remote-module-loader": "^3.0.2",
         "fira": "mozilla/fira#4.202"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


bmingles (approved):
Looks good to me

mofojed:
[Published an alpha package](https://github.com/deephaven/web-client-ui/actions/runs/4801948245/jobs/8544803088), `0.37.1-redux-peer-dep.0`. Verified that I was able to complete an `npm install` successfully on Enterprise using that alpha version of the packages, e.g.:
```
  "dependencies": {
    "@deephaven/chart": "0.37.1-redux-peer-dep.0",
    "@deephaven/components": "0.37.1-redux-peer-dep.0",
    "@deephaven/console": "0.37.1-redux-peer-dep.0",
    "@deephaven/dashboard": "0.37.1-redux-peer-dep.0",
    ....
```

PR #1257, opened 2023-05-01
DH-14630: ACL Editor Hooks

bmingles:
Supports enterprise DH-14630


packages/react-hooks/src/useFormWithDetachedSubmitButton.ts
```diff
@@ -0,0 +1,86 @@
...
+ * const { formProps, submitButtonProps } = useFormWithDetachedSubmitButton(preventDefaultFormSubmit);
+ *
+ * <form {...formProps}></form>
+ * <button {...submitButtonProps} type="submit">Submit</button>
+ *
+ * @param preventDefaultFormSubmit Optionally disable default form submit behavior.
+ * @returns props that can be spread on a form component + a submit button component.
+ */
+export default function useFormWithDetachedSubmitButton(
+  preventDefaultFormSubmit = false
```
mofojed, line 65: Would we ever actually want this to be `false`? Perhaps should default to `true`?

packages/react-hooks/src/useFormWithDetachedSubmitButton.ts
```diff
@@ -0,0 +1,86 @@
...
+  return {
+    formProps: {
+      id: formId,
+      onSubmit: preventDefaultFormSubmit ? preventDefault : undefined,
+    },
+    submitButtonProps: {
+      form: formId,
+      ref: submitButtonRef,
+    },
+  };
```
mofojed, line 85: Should memoize these props to prevent unnecessary re-rendering by consumers, e.g.
```
const formProps = useMemo(() => ({ id: formId, onSubmit: ... }), [ formId, preventDefaultFormSubmit, preventDefault]);
```

mofojed (changes requested):

packages/react-hooks/src/useFormWithDetachedSubmitButton.ts
bmingles: I flipped this to be `enableDefaultFormSubmitBehavior` so that the default is disabled

mofojed (approved):

packages/jsapi-components/src/useTableClose.test.ts
```diff
@@ -32,3 +32,17 @@ it.each([closedTable, null, undefined])(
...
+
+  const { rerender } = renderHook(t => useTableClose(t), {
+    initialProps: table,
+  });
+
+  expect(table.close).not.toHaveBeenCalled();
+
+  rerender(nextTable);
+
+  expect(table.close).toHaveBeenCalled();
```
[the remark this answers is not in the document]
bmingles: Fixed and pushed

mofojed (approved):

PR #1258, opened 2023-05-02
feat: De-globalize JSAPI in Chart package

vbabich:
- De-globalize JSAPI in the Chart package.
- Update `dashboard-core-plugins` to match the Chart package changes.

BREAKING CHANGE:
- `ChartUtils` class now needs to be instantiated with a JSAPI object, most of the methods converted from static to instance methods.
- All `ChartModelFactory` methods require JSAPI object as the first argument.
- `FigureChartModel` constructor requires JSAPI object as the first argument.


packages/chart/src/ChartModelFactory.ts
```diff
@@ -1,4 +1,4 @@
-import dh, { Figure, Table } from '@deephaven/jsapi-shim';
+import { dhType, Figure, Table } from '@deephaven/jsapi-shim';
```
mofojed, line 1: Import types from `@deephaven/jsapi-types` instead, e.g.:
```suggestion
import type { dh as dhType, Figure, Table } from '@deephaven/jsapi-types';
```

We should be able to remove dependency on `@deephaven/jsapi-shim` entirely.

packages/chart/src/ChartUtils.test.ts
```diff
@@ -557,22 +559,22 @@ it('creates range breaks from holidays correctly', () => {
...
+    expect(chartUtils.getAxisPropertyName(dh.plot.AxisType.SIZE)).toBe(null);
+    expect(chartUtils.getAxisPropertyName(dh.plot.AxisType.Z)).toBe(null);
   });
 });
 
 describe('getPlotlyChartMode', () => {
   const { LINE, SCATTER } = dh.plot.SeriesPlotStyle;
-  const { getPlotlyChartMode } = ChartUtils;
+  const getPlotlyChartMode = chartUtils.getPlotlyChartMode.bind(chartUtils);
   test('scatter plots', () => {
```
mofojed, line 578: Side note, this would be a good example of using `test.each`, it would be something like:
```
test.each([
  [ SCATTER, null, null, 'markers' ],
  [ SCATTER, null, false, undefined ],
  ...
  [ LINE, true, true, 'lines+markers' ],
])('returns the expected plotly type for %s %s %s', (areLinesVisible, areShapesVisible, result) => {
  expect(getPlotlyChartMode(SCATTER, areLinesVisible, areShapesVisible)).toBe(result);
});
```

packages/chart/src/ChartUtils.ts
```diff
@@ -5,12 +5,13 @@ import {
   TableColumnFormatter,
   TableUtils,
 } from '@deephaven/jsapi-utils';
-import dh, {
+import {
```
mofojed, line 8: Change the import to be from `@deephaven/jsapi-types`

packages/chart/src/ChartUtils.ts
```diff
@@ -884,142 +408,640 @@ class ChartUtils {
...
+      (seriesData as Partial<OhlcData>).increasing = {
+        line: { color: theme.ohlc_increasing },
+      };
+      (seriesData as Partial<OhlcData>).decreasing = {
+        line: { color: theme.ohlc_decreasing },
+      };
+    } else if (plotStyle === dh.plot.SeriesPlotStyle.PIE) {
+      seriesData.textinfo = 'label+percent';
+
+      // TODO Open DefinitelyTyped/Plotly PR to mark family and size as optional
```
mofojed, line 975: Looks like we never opened a PR for this...

packages/chart/src/FigureChartModel.ts
```diff
@@ -120,6 +129,12 @@ class FigureChartModel extends ChartModel {
     this.stopListeningFigure();
   }
 
+  getCachedChartUtils = memoize((dh: dhType) => new ChartUtils(dh), { max: 1 });
+
+  getChartUtils(): ChartUtils {
+    return this.getCachedChartUtils(this.dh);
+  }
+
```
mofojed, line 137: Why bother with memoization etc.? The dh is passed in once, and we'll never update it again... I'd just make it a member variable, e.g.

```
this.dh = dh;
this.chartUtils = new ChartUtils(dh);
```

packages/dashboard-core-plugins/src/ChartPlugin.tsx
```diff
@@ -37,9 +37,10 @@ export function ChartPlugin(props: ChartPluginProps): JSX.Element | null {
       }
 
       const metadata = { name, figure: name };
+      const makeApi = () => dh;
```
mofojed, line 40: Instead of a `makeApi`, we should use contexts, and just `useApi` from here, e.g.
```
const dh = useApi();
```

Also just wrapping the child panels with the context provider as well. #1229 will make that simpler.

Also - `dh` seems to be the global instance here. I'm not sure why it's not causing a compilation error? Perhaps remove the dependency on `jsapi-shim` in `dashboard-core-plugins` and then see...

packages/dashboard-core-plugins/src/panels/ChartPanel.test.tsx
```diff
@@ -72,6 +72,7 @@ function makeChartPanelWrapper({
   glContainer = makeGlComponent(),
   glEventHub = makeGlComponent(),
   columnSelectionValidator = undefined,
+  makeApi = () => Promise.resolve(dh),
```
mofojed, line 75: This doesn't match `ChartPlugin`, which just returns `dh` and not a Promise...

packages/embed-chart/src/App.tsx
```diff
@@ -58,7 +58,11 @@ function App(): JSX.Element {
           // Create the `ChartModel` for use with the `Chart` component
           log.debug(`Creating model...`);
 
-          const newModel = await ChartModelFactory.makeModel(undefined, figure);
+          const newModel = await ChartModelFactory.makeModel(
```
mofojed, line 61: Should use `const dh = useApi()` above, and not depend on `jsapi-shim`, instead pulling the types from `jsapi-types`.

mofojed (changes requested):

packages/dashboard-core-plugins/src/ChartPlugin.tsx
vbabich: I'm splitting the changes into separate PRs, this PR de-globalizes the API in the Chart package with minimal changes in the consumers.
Are we planning to land #1229 in vermilion?

mofojed: Disregard that comment about #1229, it won't be relevant for this anyway, since it'll be on a per panel basis, and we need to load the API asynchronously. This method is fine.

packages/dashboard-core-plugins/src/panels/ChartPanel.test.tsx
vbabich: Updated `ChartPlugin` to return `Promise.resolve(dh)`.

packages/dashboard-core-plugins/src/index.test.tsx
```diff
@@ -16,16 +18,18 @@ import {
...
-        <FilterPlugin />
-        <GridPlugin hydrate={() => undefined} />
-        <ChartPlugin hydrate={() => undefined} />
-        <ConsolePlugin />
-        <LinkerPlugin />
-        <MarkdownPlugin />
-        <PandasPlugin hydrate={() => undefined} />
-      </Dashboard>
-    </Provider>
+    <ApiContext.Provider value={dh}>
```
vbabich, line 21: @mofojed I'm getting the API in `ChartPlugin` via the `useApi()` hook. What's the right way to test this? Is passing a global via `ApiContext` fine, or what would be a cleaner way?

packages/chart/src/ChartUtils.test.ts
vbabich: Yep, much cleaner. Updated the tests.

packages/dashboard-core-plugins/src/index.test.tsx
mofojed: Yes this way of testing is fine. You can just move `@deephaven/jsapi-shim` to the `devDependencies`

vbabich: Moved it `devDependencies` in the Chart package, will do the same in `dashboard-core-plugins` after converting to `useApi`.

packages/chart/src/ChartUtils.ts
```diff
@@ -185,187 +184,6 @@ class ChartUtils {
```
mofojed, line 184: I had a thought. Right now we're doing `new ChartUtils(dh)` in a bunch of places; which could result in multiple `ChartUtils` objects that are pretty much the same. Not the biggest deal, _but_, we _could_ provide a static method to get an existing chart utils object based on the API... and just call `ChartUtils.get(dh)` (or a util function `getChartUtils(dh)`) instead. Could use [WeakMaps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#caching) to they get cleaned up automatically if the API gets garbage collected as well.

Probably wouldn't change much in terms of overhead, but thought I'd mention it.

packages/code-studio/src/styleguide/Charts.tsx
```diff
@@ -1,30 +1,19 @@
...
-    this.state = {
-      model: new MockChartModel(),
-    };
-  }
-
-  render(): React.ReactElement {
-    const { model } = this.state;
+function Charts(): ReactElement {
+  const dh = useApi();
+  const [model] = useState(new MockChartModel(dh));
```
mofojed, line 7: Should be in the habit of using a function to initialize state if it's creating an object or otherwise doing something "expensive".
```suggestion
  const [model] = useState(() => new MockChartModel(dh));
```

packages/dashboard-core-plugins/package.json
```diff
@@ -32,7 +32,9 @@
     "@deephaven/grid": "file:../grid",
     "@deephaven/icons": "file:../icons",
     "@deephaven/iris-grid": "file:../iris-grid",
+    "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
     "@deephaven/jsapi-shim": "file:../jsapi-shim",
```
mofojed, line 36: Removing this in a future update... looks like the only thing that will need to be updated is the `DropdownFilterPanel`

packages/dashboard-core-plugins/tsconfig.json
```diff
@@ -34,6 +34,9 @@
     {
       "path": "../iris-grid"
     },
+    {
+      "path": "../jsapi-bootstrap"
+    },
     {
       "path": "../jsapi-shim"
     },
```
mofojed, line 42: We should really have some sort of eslint/rule to test that these match the package.json... need to add jsapi-types here.
Can also remove all these extra new lines, they're not necessary and makes the file more readable without them imo.

packages/embed-chart/package.json
```diff
@@ -20,7 +20,7 @@
     "@deephaven/chart": "file:../chart",
     "@deephaven/components": "file:../components",
     "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
-    "@deephaven/jsapi-shim": "file:../jsapi-shim",
+    "@deephaven/jsapi-types": "file:../jsapi-types",
```
mofojed, line 23: Update tsconfig.json removing shim and adding types

mofojed (changes requested):

packages/chart/src/ChartUtils.ts
vbabich: I don't have enough information about this, who should have opened a PR?

mofojed: It would have been from when Tony was doing TS conversion. If you could create a ticket for it just with this comment and the link to the pie.d.ts, that would be great.

mofojed (approved):

PR #1259, opened 2023-05-02
feat: Update @vscode/codicons to v0.0.33

dsmmcken:
Update adds several new icons see https://github.com/microsoft/vscode-codicons/releases/tag/0.0.33 for changes.

They renamed circle-large-outline to circle-large for naming consistency. I've fixed our usage of vsCircleLargeOutline to vsCircleLarge for the renamed icon (no change in visuals).

BREAKING CHANGE: `vsCircleLargeOutline` icon renamed to `vsCircleLarge`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,7 +91,7 @@
         "@types/react-window": "^1.8.5",
         "@types/shell-quote": "^1.7.1",
         "@types/shortid": "0.0.29",
-        "@vscode/codicons": "0.0.32",
+        "@vscode/codicons": "0.0.33",
         "chokidar-cli": "^2.1.0",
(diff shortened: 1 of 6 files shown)


mofojed (changes requested):
We should still add Breaking Change to this PR, as we're no longer exporting `vsCircleLargeOutline` from `@deephaven/icons`

mofojed:
Another option would be to re-export it with the old name to reduce breakage? But may be more hassle than it's worth.

dsmmcken:
better to let it break and update the other side I think, the name change makes sense.

mofojed (approved):

PR #1261, opened 2023-05-03
fix: DH-14657 Better disconnect handling

mofojed:
- In some cases, the API disconnects and the reconnects almost immediately afterwards, causing the "Reconnecting..." modal to appear and be distracting
- Rather than have the API lie about whether it's connected or not, UI debounces when it displays the modal, just blocking interaction with a transparent modal for the duration of the disconnection if it reconnects immediately
- Also added logging so we can see when connected/disconnected, in case this is further observed
- Tested using the steps in #1149 , and also manually setting the `debounceMs` to `5000` to be able to react to it

### DIFF
--- a/packages/components/src/modal/DebouncedModal.test.tsx
+++ b/packages/components/src/modal/DebouncedModal.test.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { DEFAULT_DEBOUNCE_MS } from '@deephaven/react-hooks';
 import { act, render, screen } from '@testing-library/react';
 import DebouncedModal from './DebouncedModal';
 import Modal from './Modal';
@@ -10,6 +9,7 @@ const children = (
     <div>{mockChildText}</div>
   </Modal>
 );
+const DEFAULT_DEBOUNCE_MS = 250;
 
 beforeAll(() => {
   jest.useFakeTimers();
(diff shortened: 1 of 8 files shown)


packages/react-hooks/src/useDebouncedValue.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Debounces a value. Returns the value after the debounce time has elapsed.
+ * @param value Value to debounce
+ * @param debounceMs Amount of time to debounce
+ * @returns The debounced value
+ */
+export function useDebouncedValue<T>(
+  value: T,
+  debounceMs = DEFAULT_DEBOUNCE_MS
```
mattrunyon, line 13: Why provide a default here if we're not using it?

I think it might make more sense for the hook to not include a default since it's more of a util. Just like lodash debounce has no default

packages/react-hooks/src/useDebouncedValue.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
+
+export const DEFAULT_DEBOUNCE_MS = 250;
+
+/**
+ * Debounces a value. Returns the value after the debounce time has elapsed.
```
mattrunyon, line 6:
```suggestion
 * Debounces a value.
 * Returns the initial value immediately.
 * Returns the latest value after no changes have occurred for the debounce duration.
```

packages/react-hooks/src/useDebouncedValue.ts
```diff
@@ -0,0 +1,28 @@
...
+) {
+  const [debouncedValue, setDebouncedValue] = useState<T>(value);
+  useEffect(() => {
+    const timeoutId = setTimeout(() => {
+      setDebouncedValue(value);
+    }, debounceMs);
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [value, debounceMs]);
```
mattrunyon, line 23: Does this trigger 2 renders (not necessarily draws) when first called? I think React is smart enough to see `setDebouncedValue(sameValue)` as a no-op

packages/react-hooks/src/useDebouncedValue.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+it('should return the initial value after the debounce time has elapsed', () => {
+  const value = 'mock value';
+  const { result, rerender } = renderHook(() => useDebouncedValue(value));
+  expect(result.current).toBe(value);
+  rerender();
+  act(() => {
+    jest.advanceTimersByTime(DEFAULT_DEBOUNCE_MS);
+  });
+  expect(result.current).toBe(value);
```
mattrunyon, line 26: Also `expect(result.all.length).toBe(1)` I think?

packages/components/src/modal/DebouncedModal.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <DebouncedModal isOpen={false}>{children}</DebouncedModal>
+    );
+    expect(screen.queryByTestId('debounced-modal-backdrop')).toBeNull();
+    expect(screen.queryByText(mockChildText)).toBeNull();
+
+    act(() => {
+      rerender(<DebouncedModal isOpen>{children}</DebouncedModal>);
+    });
+    expect(screen.queryByTestId('debounced-modal-backdrop')).not.toBeNull();
+    expect(screen.queryByText(mockChildText)).toBeNull();
```
mattrunyon, line 34: `.toBeInTheDocument()` or w/e that matcher is called. Most expect calls in these tests could swap to that matcher

mattrunyon (changes requested):

packages/react-hooks/src/useDebouncedValue.test.ts
mofojed: No, it should be 2 here - 1 for the initial render, 1 for the rerender. But yea I'll add the check, good call.

packages/components/src/modal/DebouncedModal.test.tsx
mofojed: Always forget about that one.

packages/react-hooks/src/useDebouncedValue.ts
mofojed: It does only render once.

packages/react-hooks/src/useDebouncedValue.ts
mofojed: Yeah fair, I'll pull out the default.

mattrunyon (approved):

PR #1263, opened 2023-05-04
feat: remove click handler setting onTabContentFocusIn

dsmmcken:
OnTabContentFocusIn is used to set a class on the tab to highlight which tab currently has focus so the user knows where any keyboard input or shortcuts used will be applied. This was being set by both focusin and any clicks inside the tab, with clicks triggering an addtional call to focus.

Both handlers were commited at the same time, but I can't think of any reason we were actually setting the focus on click. The click itsellf should do that. My best guess is tabindex=-1 wasn't on the parent at the time (during development as it was also added in the same commit) or maybe browser support at the time? It all came in under one commit.

I have tested in chrome (win & mac), firefox (win) and safari (mac) and clicking on or anywher within any panel (tables, plots, one clicks, console, notebooks, log panel, command history, file exporer) all still set the focus as expected.

Additionally I addressed a firefox edge case around focus and an item being removed from the dom.

### DIFF
--- a/packages/golden-layout/src/controls/Tab.ts
+++ b/packages/golden-layout/src/controls/Tab.ts
@@ -184,7 +184,9 @@ export default class Tab {
     // being erroneously marked as focused if the user's DOM element
     // is removed and they click another tab. To prevent this, we
     // remove existing "lm_focusin" classes first.
-    $('lm_focusin').removeClass('lm_focusin');
+    $('.lm_focusin', this._layoutManager.root.element).removeClass(
+      'lm_focusin'
+    );
 
     this.element.addClass('lm_focusin');
   }
--- a/packages/golden-layout/src/controls/Tab.ts
+++ b/packages/golden-layout/src/controls/Tab.ts
@@ -83,11 +83,6 @@ export default class Tab {
       this.contentItem.container._contentElement
         .on('focusin', this._onTabContentFocusIn)
         .on('focusout', this._onTabContentFocusOut);
-      this.contentItem.container._contentElement[0].addEventListener(
-        'click',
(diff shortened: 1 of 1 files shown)


packages/golden-layout/src/controls/Tab.ts
```diff
@@ -181,25 +176,16 @@ export default class Tab {
...
-      // happening in proper order. Can use HTMLElement.focus() to avoid.
-      this.contentItem.container._contentElement[0].focus(); // [0] needed to use dom focus, not jquery method
-    }
+    // Ensure only one tab is marked as having focus at a time.
+    // In Firefox, if the focused element is removed from the DOM,
+    // the focusout event won't trigger. This can result in two tabs
+    // being erroneously marked as focused if the user's DOM element
+    // is removed and they click another tab. To prevent this, we
+    // remove existing "lm_focusin" classes first.
+    $('lm_focusin').removeClass('lm_focusin');
```
mofojed, line 187: Can you scope this to within the layout manager root? Also shouldn't it be `.lm_focusin`? You sure we need this?
```suggestion
    $('.lm_focusin', this._layoutManager.root).removeClass('lm_focusin');
```

mofojed (changes requested):

packages/golden-layout/src/controls/Tab.ts
dsmmcken: `root.element`, yeah. 

You can replicate the issue in firefox only by clicking the table sidebar menu button on a table, then clicking inside another table, and both will show as selected.

highlight

mofojed (approved):

PR #1266, opened 2023-05-04
build: Add dist-tag v0.31 for the v0.31 branch

mofojed:


.github/workflows/publish-packages.yml
```diff
@@ -21,6 +21,6 @@ jobs:
       - run: npm run build
       # Need the --no-verify-access access flag since we use an automation token. Otherwise publish step fails
       # https://github.com/lerna/lerna/issues/2788
-      - run: ./node_modules/.bin/lerna publish --no-verify-access from-package --yes
+      - run: ./node_modules/.bin/lerna publish --no-verify-access from-package --yes --dist-tag v0.31
```
mattrunyon, line 24: This is technically already included in `lerna.json` so it's not necessary here unless you want to be explicit here. If so, we should remove from `lerna.json`

mattrunyon: And is in the readme to update that field in `lerna.json` https://github.com/deephaven/web-client-ui#creating-a-hotfixlts-branch

mofojed: Right I forgot you took care of this already!

PR #1277, opened 2023-05-08
refactor: Make chart easier to consume for plotly express plugin

mattrunyon:
This allows the plotly express plugin to pass its own plotly instance into ChartPanel so that it can bundle more of plotly than our default charting does.

Also cleaned up some types, loosened and deprecated `VariableTypeUnion` since we should just be using `string` in those cases, and added some 3D plotting theming and buttons


packages/chart/src/plotly/Plot.ts
```diff
@@ -0,0 +1,7 @@
+// Import the react customizable bundle for building, only pull in the modules we need
+// This reduces the build size. Plotly has a lot of modules we don't need/use.
+// https://github.com/plotly/react-plotly.js#customizing-the-plotlyjs-bundle
+import createPlotlyComponent from 'react-plotly.js/factory.js';
+import Plotly from './Plotly';
+
+export default createPlotlyComponent(Plotly);
```
mofojed, line 7: Verified this works from docs/Docusaurus?

mattrunyon: Reverted to the old version. Something is weird w/ the docs and the ESM/CJS compilation. In addition we lazy load it in the docs so that adds another layer

packages/chart/src/Chart.tsx
```diff
@@ -257,8 +260,10 @@ export class Chart extends Component<ChartProps, ChartState> {
         modeBarButtons: [
           customButtons,
           ['toImage'],
-          ['zoom2d', 'pan2d'],
-          ['zoomIn2d', 'zoomOut2d', 'autoScale2d', 'resetScale2d'],
+          is3d ? ['zoom3d', 'pan3d'] : ['zoom2d', 'pan2d'],
+          is3d
+            ? ['orbitRotation', 'tableRotation', 'resetCameraDefault3d']
+            : ['zoomIn2d', 'zoomOut2d', 'autoScale2d', 'resetScale2d'],
```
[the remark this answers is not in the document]
mattrunyon: Shows controls for both now if both exist

mofojed (dismissed):

mofojed (approved):

PR #1279, opened 2023-05-08
build: Update to Vite 4.3.5

### LINKED ISSUE #1209: Update to Vite 4.3.0
When Vite 4.3.0 releases, it should contain a performance improvement that can vastly help our HMR in certain cases from https://github.com/vitejs/vite/pull/12658

Upgrade to that once available. Currently 4.3.0-beta.2 is released, so should be 4.3.0 within the next few weeks hopefully.

mattrunyon:
Fixes #1209

Pulled the Vite dependency to the root since we use the same version in any apps like we do with TypeScript

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,6 +92,7 @@
         "@types/react-window": "^1.8.5",
         "@types/shell-quote": "^1.7.1",
         "@types/shortid": "0.0.29",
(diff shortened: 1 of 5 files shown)


mattrunyon:
I'll still wait b/c there will be a package-lock merge conflict

PR #1285, opened 2023-05-10
chore: test: Change import to import type, jsapi-shim to jsapi-types where possible

vbabich:
- Change `import` to `import type`, `jsapi-shim` to `jsapi-types` where possible.
- Move `jsapi-shim` to `devDependencies`.
- Convert `ChartTestUtils` static methods to instance methods, fix unit tests.

### DIFF
--- a/packages/chart/src/ChartTestUtils.ts
+++ b/packages/chart/src/ChartTestUtils.ts
@@ -1,10 +1,11 @@
-import dh, {
+import dh from '@deephaven/jsapi-shim';
+import type {
   Axis,
   Chart,
   Figure,
   Series,
   SeriesDataSource,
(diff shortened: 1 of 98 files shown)


packages/chart/src/ChartTestUtils.ts
```diff
@@ -1,10 +1,16 @@
-import dh, {
+import {
+  AxisFormatType,
+  AxisPosition,
+  AxisType,
+  dh as DhType,
```
mofojed, line 5: Should be under `import type`

mofojed (changes requested):
Nice. Just one spot to address!

mofojed (dismissed):

mofojed (approved):

Issue #1283, opened 2023-05-10
Long column name does not wrap correctly in Tooltip


mofojed:
**Description**
Related to #1276

When creating a column with a very long name, on hover, the tooltip does not wrap the name so the name goes off the tooltip (and eventually the screen).

**Steps to reproduce**
1. Run the following code:
```
from deephaven import new_table
from deephaven.column import string_col, int_col

result = new_table([
    int_col("I1", [1, 2, 3]),
    int_col("I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I2I23", [4, 5, 6]),
])
```
2. Hover over the column header `l2...`

**Expected results**
2. The long name is handled gracefully in the tooltip, perhaps by line wrapping

**Actual results**
2. The long table name is not handled gracefully on the tooltip

**Additional details and attachments**

<img width="1438" alt="Screenshot 2023-05-08 at 3 01 58 PM" src="https://user-images.githubusercontent.com/10480451/236927038-b9ffc7ed-2cb1-4397-91c5-af3f18b8facb.png">

**Versions**

Engine Version: 0.24.0
Web UI Version: 0.34.0
Java Version: 11.0.17
Barrage Version: 0.5.0

PR #1290, opened 2023-05-10
fix: add word-break to long column names in column tooltip

dsmmcken:
resolves: #1283

### DIFF
--- a/packages/iris-grid/src/ColumnStatistics.scss
+++ b/packages/iris-grid/src/ColumnStatistics.scss
@@ -12,6 +12,7 @@
 .column-statistics-title {
   padding-top: $spacer-1;
   font-weight: 500;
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/ColumnStatistics.scss
```diff
@@ -23,8 +24,6 @@
   color: $gray-400;
   font-weight: normal;
   user-select: none;
-  text-overflow: ellipsis;
-  overflow: hidden;
```
mofojed, line 27: Why'd you remove this truncation?

dsmmcken: I don't think we actually want to hide the column type.

mofojed (approved):

PR #1291, opened 2023-05-11
feat: add contains ignore case in go to row

### LINKED ISSUE #1274: Add Contains Ignore Case option to Seek Row functionality
We currently have Seek Row functionality in grids for searching for text within a column. There are only a few options there, we should add the Contains Ignore Case option as well.

image

Context: https://illumon.slack.com/archives/CAUG7MC2D/p1683398883944149

ethanalvizo:
Closes #1274


packages/iris-grid/src/GotoRow.tsx
```diff
@@ -281,6 +286,7 @@ function GotoRow({
             <div className="goto-row-input">
               <input
```
mofojed, line 287: Forgot to respond in Slack to this question, but you _can_ use `idi` if you generate the id using `shortid` or something similar. See `DropdownFilter.tsx` for an example of doing that. It's probably best to have the label linked with the input correctly as you had in Slack.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3334,7 +3336,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
             selectedColumn,
             dh.ValueType.STRING,
             inputString,
-            isEquals,
+            isIgnoreCase,
```
mofojed, line 3339: Good catch on this one.

packages/iris-grid/src/GotoRow.test.tsx
mofojed, line 1: Great work adding the tests! Did you check coverage using `npm test -- --coverage` ?

packages/iris-grid/src/GotoRow.tsx
ethanalvizo: I was thinking of applying this label change but for the "Go to value" row there's only one label and 2 inputs (column selector and value to search). In the scenario that the column is strings there actually ends up being 3 inputs. Should I just be linking the label with one of those inputs? If so, which one should it be?

packages/iris-grid/src/GotoRow.test.tsx
ethanalvizo: Just ran it now I'm getting about 60-70% coverage so the file is yellow. Should I be aiming for 100 here?

packages/iris-grid/src/GotoRow.tsx
mofojed: For one label on multiple inputs, it would be more correct to use a [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) and [legend](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend) instead of a label. Something like:
```
<fieldset>
  <legend>Go to value</legend>
  ...
  <select aria-label="Filter type" ... />
  <input aria-label="Value input" .... />
</fieldset>
```
Then you should be able to do `screen.getByRole('group', { name: 'Go to value' })` to get the `fieldset`, or get the individual inputs by `screen.getByLabelText("Filter type")`
Docs for [accessibility here](https://www.w3.org/TR/html-aam-1.0/#fieldset-element-accessible-name-computation).

mofojed:
recheck

packages/iris-grid/src/GotoRow.test.tsx
mofojed: 100% is what we strive for; any increase in coverage is great though.
Once you trigger the CLA bot again (I think you need to comment `recheck`), and CLA passes, code cov should run and it'll link to a full report for the project, so you can see which areas have the most coverage.

ethanalvizo:

ethanalvizo:
recheck

mofojed (dismissed):

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -844,7 +844,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       gotoValueError: '',
 
       gotoValueSelectedColumnName: model.columns[0]?.name ?? '',
-      gotoValueSelectedFilter: FilterType.eq,
+      gotoValueSelectedFilter: FilterType.eqIgnoreCase,
```
mofojed, line 847: Did @dsmmcken say we should update the default selected filter?

mofojed (changes requested):

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -314,15 +331,15 @@ function GotoRow({
...
+            <fieldset
               className={classNames('goto-row-wrapper', {
                 'is-inactive': isGotoRowActive,
               })}
               onClick={() => setIsGotoRowActive(false)}
               onFocus={() => setIsGotoRowActive(false)}
               role="group"
             >
-              <div className="goto-row-text">Go to value</div>
+              <legend className="goto-row-text">Go to value</legend>
```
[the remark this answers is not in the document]
ethanalvizo: When using the fieldset and legend components, the following input values keep going to the next line. I've tried adding flex-wrap: nowrap and white-space: nowrap but it doesn't override the behaviour for some reason

image

packages/iris-grid/src/IrisGrid.tsx
ethanalvizo: Yea I might have misinterpreted what he meant but I thought whenever we search something in the UI the default should be case insensitive. If we're querying for something though using the query language that should be case sensitive.

packages/iris-grid/src/GotoRow.tsx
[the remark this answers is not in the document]
ethanalvizo: I wrapped the legend in a div tag for now and that fixes the behaviour but I think that loses the link that we wanted it to have as like a label/input

packages/iris-grid/src/IrisGrid.tsx
dsmmcken: Yeah. Equals case insensitive as the default would be more consistent with the rest of our ui

packages/iris-grid/src/GotoRow.tsx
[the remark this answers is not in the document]
dsmmcken: I wouldn't bother.

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -86,10 +87,10 @@ function GotoRow({
     ({ columns } = model.table);
   }
 
-  const res = 'Row number';
-
   const { rowCount } = model;
 
+  const gotoRowInputId = `goto-row-input-${shortid()}`;
```
mofojed, line 92: Right now this id is updating on every render. You should memoize this, something like:
```
  const gotoRowInputId = useMemo(() => `goto-row-input-${shortid()}`, []);
```

Otherwise the ID changes with every re-render/anytime a prop changes, which is unnecessary.

mofojed (changes requested):

mofojed (approved):

PR #1292, opened 2023-05-11
feat: De-globalize JSAPI in Console package

vbabich:
- Remove global `dh` references in the Console package, update consumers.
- Add `SET_API` reducer.

BREAKING CHANGE:
- Components `IrisGrid`, `Chart`, `ChartBuilder`, `AdvancedFilterCreator`, `GotoRow`, `IrisGridModelUpdater`, `TableCSVExporter` get the JSAPI reference from the `model` prop. `dh` prop removed.
- `makeApi` props in `IrisGridPanel` and `ChartPanel` removed.
- Components `Console`, `ConsoleMenu`, `ConsoleStatusBar` now require the JSAPI instance in the `dh` prop.
- `ConsoleUtils`: static methods `isTableType`, `isWidgetType`, `isOpenableType`, `isFigureType`, `isPandas` require JSAPI instance passed in the first argument.
- `SessionUtils`: static methods `createSessionWrapper`, `loadSessionWrapper` require JSAPI instance passed in the first argument.
- Class `IrisGridModel` requires JSAPI instance passed in the constructor args.
- Components `DashboardLayout`, `ObjectIcon` has to be wrapped in `ApiContext.Provider` passing the JSAPI instance.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26163,6 +26163,7 @@
       "dependencies": {
         "@deephaven/components": "file:../components",
         "@deephaven/golden-layout": "file:../golden-layout",
+        "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
         "@deephaven/log": "file:../log",
         "@deephaven/react-hooks": "file:../react-hooks",
         "@deephaven/redux": "file:../redux",
@@ -28409,6 +28410,7 @@
       "requires": {
         "@deephaven/components": "file:../components",
         "@deephaven/golden-layout": "file:../golden-layout",
+        "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
         "@deephaven/log": "file:../log",
         "@deephaven/mocks": "file:../mocks",
         "@deephaven/react-hooks": "file:../react-hooks",
--- a/package-lock.json
+++ b/package-lock.json
@@ -26122,7 +26122,7 @@
         "@deephaven/chart": "file:../chart",
         "@deephaven/components": "file:../components",
         "@deephaven/icons": "file:../icons",
-        "@deephaven/jsapi-shim": "file:../jsapi-shim",
+        "@deephaven/jsapi-bootstrap": "file:../icons",
         "@deephaven/log": "file:../log",
         "@deephaven/storage": "file:../storage",
         "@deephaven/utils": "file:../utils",
@@ -26140,6 +26140,7 @@
         "shell-quote": "^1.7.2"
       },
       "devDependencies": {
+        "@deephaven/jsapi-shim": "file:../jsapi-shim",
         "@deephaven/mocks": "file:../mocks",
         "@deephaven/tsconfig": "file:../tsconfig"
       },
@@ -26150,6 +26151,10 @@
         "react": "^17.x",
         "react-dom": "^17.x"
       }
+    },
+    "packages/console/node_modules/@deephaven/jsapi-bootstrap": {
+      "resolved": "packages/icons",
+      "link": true
     },
     "packages/dashboard": {
       "name": "@deephaven/dashboard",
@@ -28367,6 +28372,7 @@
         "@deephaven/chart": "file:../chart",
         "@deephaven/components": "file:../components",
         "@deephaven/icons": "file:../icons",
+        "@deephaven/jsapi-bootstrap": "file:../icons",
         "@deephaven/jsapi-shim": "file:../jsapi-shim",
         "@deephaven/log": "file:../log",
         "@deephaven/mocks": "file:../mocks",
@@ -28385,6 +28391,17 @@
         "popper.js": "^1.16.1",
         "prop-types": "^15.7.2",
         "shell-quote": "^1.7.2"
+      },
+      "dependencies": {
+        "@deephaven/jsapi-bootstrap": {
+          "version": "file:packages/icons",
+          "requires": {
+            "@fortawesome/fontawesome-common-types": "^6.1.1",
+            "svg-parser": "^2.0.4",
+            "svg-path-tools": "^1.0.0",
+            "svgo": "^3.0.2"
+          }
+        }
       }
     },
     "@deephaven/dashboard": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -25834,6 +25834,7 @@
         "babel-plugin-transform-rename-import": "^2.3.0"
       }
     },
+    "packages/bootstrap": {},
     "packages/chart": {
       "name": "@deephaven/chart",
       "version": "0.38.0",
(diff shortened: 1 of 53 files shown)


packages/console/src/Console.tsx
```diff
@@ -974,6 +974,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
       >
         <div className="console-pane" ref={this.consolePane}>
           <ConsoleStatusBar
+            dh={dh}
```
mofojed, line 977: Need to pull this from context or pass in as prop

mofojed (changes requested):

packages/console/src/Console.tsx
mattrunyon: You can consume a context from a class component btw. It's just complicated if you need to consume multiple contexts. `MyComponent.contextType = Context`

https://legacy.reactjs.org/docs/context.html#classcontexttype

mofojed (changes requested):

package-lock.json
```diff
@@ -25834,6 +25834,7 @@
         "babel-plugin-transform-rename-import": "^2.3.0"
       }
     },
+    "packages/bootstrap": {},
```
mofojed, line 25837: Hmm this shouldn't be here...

package-lock.json
```diff
@@ -26151,13 +26153,18 @@
         "react-dom": "^17.x"
       }
     },
+    "packages/console/node_modules/@deephaven/jsapi-bootstrap": {
+      "resolved": "packages/bootstrap",
```
mofojed, line 26157: This is still pointing to the wrong directory for some reason... Maybe revert all changes to the package-lock, and just run `npm install` again ? E.g.:
```
git co main -- package-lock.json
npm install
```

packages/console/tsconfig.json
```diff
@@ -7,23 +7,13 @@
...
-      "path": "../chart"
-    }
+    { "path": "../components" },
+    { "path": "../jsapi-bootstrap" },
+    { "path": "../jsapi-types" },
+    { "path": "../jsapi-shim" },
+    { "path": "../log" },
+    { "path": "../storage" },
+    { "path": "../utils" },
+    { "path": "../chart" }
```
mofojed, line 17: Hm.. I'd put chart at the top for alphabetical, and it also looks like we're missing `"../icons"` here (which we depend on in the package.json). Wonder why we haven't gotten an error with that :thinking: 
We really need to write some tests that make sure this stuff is in sync.

packages/dashboard/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@deephaven/components": "file:../components",
     "@deephaven/golden-layout": "file:../golden-layout",
+    "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
```
mofojed, line 27: Need to add this to `../dashboard/tsconfig.json`

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -121,17 +124,21 @@ export function DashboardLayout({
...
-              glContainer={glContainer}
-              glEventHub={glEventHub}
-            >
-              {/* eslint-disable-next-line react/jsx-props-no-spreading */}
-              <CType {...props} ref={ref} />
-            </PanelErrorBoundary>
-          </Provider>
+          // Enterprise should be able to override the JSAPI
+          // for each panel via the props
+          <ApiContext.Provider value={dh ?? defaultDh}>
```
mofojed, line 131: `dh` is already `DhType`, if we need null-coalescing here, then the prop should be `dh?: DhType` likely instead. Or even just `PanelProps`, which is correctly typed.

mofojed (changes requested):

mofojed (approved):
Looks good to me, think @mattrunyon may still want a look.

packages/redux/src/reducers/api.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * JSAPI instance
+ */
+import { SET_API } from '../actionTypes';
+import { replaceReducer } from './common';
+
+export default replaceReducer(SET_API, null);
```
mofojed, line 7: If we set this to `undefined` instead of `null`, do we still need to make the changes to the formatters?

vbabich: No, the formatters changes aren't related to the Settings menu bug, changing `null` to `undefined` wouldn't make a difference. On a second thought, those changes weren't necessary, I've reverted the [Fix optional param initialization](https://github.com/deephaven/web-client-ui/pull/1292/commits/2ff1ed0a52f19ef8a20fef05f9b37c1a907bf387) commit.

packages/chart/src/ChartModel.ts
```diff
@@ -28,11 +29,14 @@ class ChartModel {
 
   static EVENT_LOADFINISHED = 'ChartModel.EVENT_LOADFINISHED';
 
-  constructor() {
+  constructor(dh: DhType) {
+    this.dh = dh;
```
mofojed, line 33: I'm not big on this, but I think this reveals a separate issue... our `ChartModel` right now isn't abstract, as it requires the Deephaven API (and also outputs as Plotly, which could be a further abstraction but is alright for now...).
The issue is that `FigureChartModel` needs the formatter settings, and `Formatter` itself depends on our DH API. In a perfect world, `Formatter` interface wouldn't be dependent on the DH API (as that's an implementation detail), and the chart model itself would manage it's own formatting (and listening for formatter updates) rather than `Chart.tsx` handling the settings change and setting it on the `ChartModel`...

That being said, now's not the time to change that abstraction, model was already tied to the JS API because of the formatter settings. So leave it like this, but something to consider for the future...

packages/iris-grid/src/sidebar/ChartBuilder.tsx
```diff
@@ -116,7 +115,8 @@ class ChartBuilder extends PureComponent<ChartBuilderProps, ChartBuilderState> {
     this.handleXAxisChange = this.handleXAxisChange.bind(this);
     this.sendChange = this.sendChange.bind(this);
 
-    const { dh, model } = props;
+    const { model } = props;
+    const { dh } = model;
     const { columns } = model;
```
mofojed, line 120:
```suggestion
    const { columns, dh } = model;
```

mofojed (commented):
Still reviewing the Enterprise side/functionality, so not approving yet. Only one minor cleanup comment.

packages/iris-grid/src/sidebar/ChartBuilder.tsx
mattrunyon, line 120: Combine these destructure assignments

mattrunyon (commented):
One minor thing I saw. Looks fine to me though

packages/chart/src/ChartModel.ts
vbabich: Agreed, makes sense.

mofojed (approved):

PR #1294, opened 2023-05-12
feat: Added new icons and added composition example to styleguide

dsmmcken:
Added dhOrganizationAdd icon for ACL editor and dhAddSmall for composition if needed somewhere else. Also added composition example to styleguide.


bmingles (approved):
Changes look good to me. 

I did notice that our icon search seems to expect a dash cased string instead of camelCase. I don't think this is related to this change at all, so approving.

PR #1297, opened 2023-05-15
chore: consolidate git ignore files

### LINKED ISSUE #1236: Consolidate gitignore files
We have a gitignore in all packages which is identical and could be consolidated into 1 root gitignore. The gitignore in packages should be reserved for overrides of the default

From the [gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format), if there is no separator at the beginning or middle of a pattern, it can match at any level. So just having `node_modules` at the root will ignore any folder called `node_modules` at any level. For `.env` files we should be able to have a `.env*local` that ignores all local env files

To test we should run `npm run build`, `npm run e2e`, and `CI=true npm test` (generates coverage). Ensure the only tracked changes by git are the gitignore files

ethanalvizo:
Closes #1236


.gitignore
```diff
@@ -17,18 +17,15 @@
...
+.env*local
 .project
 .settings/
 .eslintcache
 .stylelintcache
 lerna-debug.log
 Lerna-Profile-*.json
 
-/public/vs
+public/vs
```
mattrunyon, line 28: Not sure this is actually used anywhere (might have been a long time ago). Should be able to remove it

packages/app-utils/.gitignore
```diff
@@ -1,31 +1,3 @@
 # See https://help.github.com/ignore-files/ for more about ignoring files.
```
mattrunyon, line 1: Delete the files that are now empty or just comments

.gitignore
```diff
@@ -1,14 +1,14 @@
...
 
 # testing
-/coverage
+coverage
 
 # production
-/build
-/dist
+build
+dist
```
mattrunyon, line 11: Minor thing that shouldn't change how the ignore works: add a trailing slash to the things we expect to be directories. Git will ignore at any depth as long as there's no leading or middle slash, but trailing should be fine

packages/components/.gitignore
```diff
@@ -1,32 +1,9 @@
...
-# dependencies
-/node_modules
-
-# testing
-/coverage
-
 # production
-/build
-/dist
 /css
```
mattrunyon, line 4: You should be able to ignore `css` at the root since we tend to not write CSS directly (we use scss) or include any folders called `css`

packages/dashboard-core-plugins/.gitignore
```diff
@@ -1,34 +1,12 @@
...
-/public/vs
-
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-
-src/**/*.css
 
 cypress/screenshots
 cypress/videos
```
mattrunyon, line 12: You can remove these too and delete this file. We don't use cypress in testing any more

packages/golden-layout/.gitignore
```diff
@@ -1,6 +1,3 @@
-node_modules
 .idea
 npm-debug.log
```
mattrunyon, line 2: These shouldn't be needed here

Side note, add `.idea` to the root gitignore in case someone opens the project in IntelliJ or Webstorm. That's the Jetbrains project settings folder similar to `.vscode`

packages/golden-layout/.gitignore
```diff
@@ -1,6 +1,3 @@
...
-.DS_Store
-dist
 css
```
mattrunyon, line 3: Can remove this if we just ignore `css` from the root. Then the whole file can be deleted

packages/grid/.gitignore
```diff
@@ -1,22 +1,9 @@
 
 # See https://help.github.com/ignore-files/ for more about ignoring files.
 
-# dependencies
-node_modules
-
 # builds
-build
-dist
 .rpt2_cache
```
mattrunyon, line 5: Should be able to delete this

packages/grid/.gitignore
```diff
@@ -1,22 +1,9 @@
...
 
 # misc
-.DS_Store
 .env
```
mattrunyon, line 8: Delete this. We shouldn't ignore the non-local `.env` files, but also there is no `.env` in grid. Only in the apps (code-studio, embed-grid, embed-chart)

mattrunyon (changes requested):
Looks like most/all nested gitignore files can be deleted since they're empty or can be empty w/ my suggestions

packages/components/.gitignore
ethanalvizo: image
do these two lines accomplish the same thing? If so I assume I would just keep the 2nd one since it's shorter

mattrunyon: No they don't. Although the 1st won't do anything in the root of our current repo because we don't have a `src` folder in the root. So you can remove the 1st

The first is any file ending in `.css` inside the `src` folder relative to the `.gitignore` location

The second is any folder called `css` at any depth.

The only place we have a tracked CSS file is `packages/icons/docs/style.css`. Everywhere else is `.scss` or generated `css` that we don't track because it's a build artifact.

mattrunyon (changes requested):

mattrunyon (approved):

Issue #1282, opened 2023-05-10
Drag to re-arrange custom columns not working


mofojed:
**Description**

In the Grid Custom Columns side bar, unable to drag to re-arrange columns.

**Steps to reproduce**

1. Create a table, e.g. 
```
from deephaven import empty_table
t = empty_table(100).update("x=i")
```
2. In the Custom Columns menu in the table side bar, add two custom columns, e.g. TEST1, TEST2, with just `1` and `2` as values
3. Drag the handle to re-arrange the custom columns

**Expected results**

3. The columns should be dragged and re-arrange

**Actual results**

3. Columns do not get dragged

**Additional details and attachments**
![Peek 2023-05-10 09-25](https://github.com/deephaven/web-client-ui/assets/4505624/98269d76-0b07-4449-8fb3-8d0148351d02)

**Versions**

Engine Version: 0.25.0
Web UI Version: 0.38.0
Java Version: 11.0.18
Barrage Version: 0.5.0
OS: Ubuntu 22.04
Browser: Chrome 112.0.5615.49

PR #1299, opened 2023-05-15
fix: drag to re-arrange custom columns not working

dsmmcken:
resolves: #1282

Broken by #1013. Can't use prop spreading with our Button component, added support for spreading.
Needs silverheels cherry-pick.


mattrunyon:
Any reason to not allow prop spreading any unknown props to our button? There's probably just some specific event handler we don't explicitly add to our button that react beautiful dnd uses

mattrunyon:
This is one of the exceptions to the rule I think. The button component has a bunch of extra HTML props that it can/should be able to accept (it should be a superset of the base button). So it would be silly to explicitly add every prop instead of spreading those we don't handle in some specific way

mattrunyon (approved):

PR #1300, opened 2023-05-15
fix: Search icon styleguide using prefixed string

dsmmcken:
Changes from searching the kebab case icon name to the component name.

Also strip spaces and dashes before searching, so kebab case search still works.


bmingles (approved):
Changes look good. I verified search now works with spaces, dashes, and is case-insensitive

PR #1303, opened 2023-05-16
style: underlay for maximized panels

### LINKED ISSUE #1166: Add an underlay to maximized panels
We currently add an underlay to modals which dims the background content so the modal is more visually distinguished. We should do the same for panels that are maximized. The minimize button can be hard to find and I think the underlay would help with that.

ethanalvizo:
Closes #1166

### DIFF
--- a/packages/dashboard/src/layout/GoldenLayout.scss
+++ b/packages/dashboard/src/layout/GoldenLayout.scss
@@ -47,10 +47,9 @@ $lm-splitter-active-color: $gray-300;
 }
 
 .lm_maximised {
-  border: 15px solid rgba($black, 0.25);
+  border: 15px solid rgba($black, 0.85);
   height: 100% !important;
   width: 100% !important;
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/layout/GoldenLayout.scss
```diff
@@ -50,6 +50,7 @@ $lm-splitter-active-color: $gray-300;
   border: 15px solid rgba($black, 0.25);
```
dsmmcken, line 50: Oh, hey this line is actually adding an underlay already, and maybe should just be turned up?

Try 0.85?

ethanalvizo: Ooh yes you're right good catch! I forgot it's kind of reversed. To decrease the visibility of the background I gotta increase the opacity of this border

dsmmcken (approved):

PR #1310, opened 2023-05-18
test: add e2e test for select distinct values

### LINKED ISSUE #1305: E2E Test: Select Distinct Values

ethanalvizo:
Closes #1305


tests/utils.ts
```diff
@@ -1,6 +1,11 @@
 import { Locator, Page } from '@playwright/test';
 import shortid from 'shortid';
 
+export enum TableTypes {
```
ethanalvizo, line 4: Made a new enum since I think the other e2e tests that I need to write will need different types of tables to fully test their functionality. If this ends up being a small enum list I will just move over the unique command strings into table.spec.ts instead

tests/utils.ts
```diff
@@ -20,9 +25,24 @@ export function generateVarName(prefix = 'v'): string {
...
+  tableName = generateVarName('t'),
+  type = TableTypes.Number
+): string {
+  switch (type) { 
+    case TableTypes.StringAndNumber:
+      return `from deephaven import new_table
+from deephaven.column import string_col, double_col
+
+${tableName} = new_table([
+double_col("Doubles", [3.1, 5.45, -1.0, 1.0, 3.0, 4.20]),
```
ethanalvizo, line 38: formatting looks odd on the string but if I have any indents it causes errors when it is entering the command into the console. Not sure how I could improve readability here or if its necessary. I could make it all single line to avoid any indents but that would create super long strings

tests/table.spec.ts-snapshots/tests-complex-table-operations-can-select-distinct-values-1-chromium-darwin.png
mofojed, line 1: As we noted, these snapshots indicate some other issue, as you can see the column name is being truncated here, even though it shouldn't be. Still need to get to the root cause of that... 
This is with the Mac snapshot as well, so this is a different issue than the `npm run e2e:update-ci-snapshots` command (running in an older version Docker contain, focal) is generating snapshots that don't match CI in GitHub Actions.

This might be kind of a difficult (and interesting) bug to debug. We can take a look at it sometime tomorrow, but you should be able to _see_ this bug happen in headed mode, if you set a breakpoint when we try and take the snapshot (should see it's truncated). Or even just running in a browser manually, we should be able to reproduce the issue.

tests/utils.ts
mofojed: Yes looks fine. We could also do something like import the template string from a file name, but then we'd need to search/replace for the table name, I think this is decent enough.

PR #1312, opened 2023-05-18
fix: makeApiContextWrapper and createMockProxy

### LINKED ISSUE #1311: TestUtils - makeApiContextWrapper console error and mock proxy improvements
- makeApiContextWrapper currently shows a console error due to ApiContextWrapper not being wrapped in forwardRef
- The proxy created by createMockProxy prints `undefined` when used in matchers. This could use some cleanup

bmingles:
- makeApiContextWrapper - wrapped in forwardRef to remove console.errr
- createMockProxy - now prints object properties instead of `undefined` in jest matchers. Proxied mock methods now include a mockName() call to give clearer output

fixes #1311

### DIFF
--- a/packages/jsapi-components/src/HookTestUtils.tsx
+++ b/packages/jsapi-components/src/HookTestUtils.tsx
@@ -1,18 +1,11 @@
-import React, { forwardRef, ReactNode } from 'react';
+import React, { ReactNode } from 'react';
 import { ApiContext } from '@deephaven/jsapi-bootstrap';
 import type { dh as DhType } from '@deephaven/jsapi-types';
 
-export function makeApiContextWrapper<TProps, TRef = unknown>(dh: DhType) {
-  return forwardRef<TRef, TProps>(function ApiContextWrapper(
-    {
-      children,
-    }: {
-      children?: ReactNode;
-    },
-    _ref
-  ) {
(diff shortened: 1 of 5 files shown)


packages/jsapi-components/src/HookTestUtils.tsx
```diff
@@ -1,11 +1,18 @@
...
-export function makeApiContextWrapper(dh: DhType) {
-  return function ApiContextWrapper({ children }: { children?: ReactNode }) {
+export function makeApiContextWrapper<TProps, TRef = unknown>(dh: DhType) {
+  return forwardRef<TRef, TProps>(function ApiContextWrapper(
+    {
+      children,
+    }: {
+      children?: ReactNode;
+    },
+    _ref
```
vbabich, line 12: I wonder what's the root cause of the issue? Why does `renderHook` require the ref? `_ref` argument isn't even passed anywhere, could this be a bug in `renderHook` itself?

bmingles: I suspect it uses it internally for something but not sure

bmingles: nvm, I was wrong. Real cause of the original issue is that `ref` needs to be excluded from the auto proxy

vbabich (approved):

PR #1319, opened 2023-05-24
fix: truncated column headers

### LINKED ISSUE #1318: Truncated Column Headers
**Description**

When creating tables with headers that aren't single characters, it is sometimes truncated rather than autosizing the column widths to fit the whole word.
**Steps to reproduce**
Create a new table with a column title a couple characters long (ex. 'strings' or 'doubles')

1. Paste the following code into the console
```
from deephaven.column import string_col, double_col

result = new_table([
    string_col("Strings", ["Creating", "New", "Tables", "Tables", "New", "Creating"])
])
```

**Expected results**

image

**Actual results**
image

**Versions**

 - Web UI Version: 0.24.2
 - OS: Mac
 - Browser: Edge

ethanalvizo:
Closes #1318 

Fixes bug in font pre-loading created in [previous PR](https://github.com/deephaven/web-client-ui/pull/1180)

### DIFF
--- a/packages/app-utils/src/components/FontBootstrap.tsx
+++ b/packages/app-utils/src/components/FontBootstrap.tsx
@@ -20,7 +20,7 @@ export type FontBootstrapProps = {
  * FontBootstrap component. Handles preloading fonts.
  */
 export function FontBootstrap({
-  fontClassNames = ['fira-sans-regular', 'fira-sans-bold', 'fira-mono'],
+  fontClassNames = ['fira-sans-regular', 'fira-sans-semibold', 'fira-mono'],
   children,
 }: FontBootstrapProps) {
   const [isLoaded, setIsLoaded] = useState(false);


mofojed (approved):
Just noting this class name needed to match the class name in `BaseStyleSheet.scss`. There is a [comment in that file](https://github.com/deephaven/web-client-ui/blob/7bcf326441d86421276c4bb66eaa63d8ded9a61f/packages/components/scss/BaseStyleSheet.scss#L847) that is very outdated.

Not saying you need to change this, but if we wanted to be even more "foolproof", we wouldn't just be using class names that are defined in another package... maybe `FontBootstrap` instead of being passed in a bunch of class names, we'd pass in a bunch of styles to preload and then we'd just load some elements with those styles set (bypassing css/class names). That's out of scope for this patch though.

PR #1321, opened 2023-05-24
fix: failing linter test from de-globalize PR

dsmmcken:


packages/console/src/csv/CsvInputBar.tsx
```diff
@@ -8,7 +8,7 @@ import React, {
 import classNames from 'classnames';
 import type { JSZipObject } from 'jszip';
 import { Button, Checkbox } from '@deephaven/components';
-import { IdeSession, Table } from '@deephaven/jsapi-shim';
+import { IdeSession, Table } from '@deephaven/jsapi-types';
```
vbabich, line 11:
```suggestion
import type { IdeSession, Table } from '@deephaven/jsapi-types';
```

packages/console/package.json
```diff
@@ -48,7 +49,6 @@
     "react-dom": "^17.x"
   },
   "devDependencies": {
-    "@deephaven/jsapi-shim": "file:../jsapi-shim",
```
vbabich, line 51: `jsapi-shim` should be in `devDependencies` since it's only used in tests.

vbabich (changes requested):

PR #1322, opened 2023-05-25
refactor: Clean up golden-layout css

dsmmcken:
We maintain our golden-layout package, bake our theme customizations directly into the dark theme. Eliminates some css. Used e2e testing to prevent visual regressions, the two changes showing are intentional.

Had to exclude the ThemeExport from the karma tests, it didn't like the scss import.


packages/dashboard/src/PanelPlaceholder.scss
```diff
@@ -4,9 +4,13 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  padding: $spacer-3;
   min-height: 100%;
   max-height: 100%;
   text-align: center;
   overflow: hidden;
+
+  // add padding on a iner wrapper so it is included in max height of parent
```
mofojed, line 12:
```suggestion
  // add padding on an inner wrapper so it is included in max height of parent
```

packages/golden-layout/karma.conf.cjs
```diff
@@ -24,7 +24,14 @@ module.exports = function (config) {
     },
 
     browserify: {
+      debug: true,
       plugin: ['esmify'],
+      // ignored because it doesn't understand the scss import in GoldenLayoutThemeExport
+      configure: function (bundle) {
+        bundle.on('prebundle', function () {
+          bundle.ignore('./dist/GoldenLayoutThemeExport.js');
```
mofojed, line 32: Can't you ignore the scss file instead? EDIT: Ah but then the js file would puke. Nevermind

packages/golden-layout/package.json
```diff
@@ -10,6 +10,7 @@
   "source": "src/index.ts",
   "type": "module",
   "dependencies": {
+    "@deephaven/components": "file:../components",
```
mofojed, line 13: If it's really just for the theme, would be nice at some point to extract the theme from `@deephaven/components` to just `@deephaven/theme` or something.

mofojed (changes requested):

mofojed (approved):

PR #1324, opened 2023-05-26
test: add e2e tests for conditional format

### LINKED ISSUE #1316: E2E Test: Conditional Formatting

ethanalvizo:
Closes #1316


tests/table.spec.ts
```diff
@@ -139,6 +139,176 @@ test.describe('tests complex table operations', () => {
...
+      page.locator('.iris-grid .iris-grid-loading-status')
+    ).toHaveCount(0);
+
+    const tableOperationsMenu = page.locator(
+      'data-testid=btn-iris-grid-settings-button-table'
+    );
+    await tableOperationsMenu.click();
+
+    // Wait for Table Options menu to show
+    await expect(page.locator('.table-sidebar')).toHaveCount(1);
```
mofojed, line 184: This is boilerplate code used by all your sidebar tests, you can just put this into a [beforeEach](https://playwright.dev/docs/api/class-test#test-before-each) within a [describe](https://playwright.dev/docs/api/class-test#test-describe-1) block for the sidebar tests, e.g.:
```
test.describe('table sidebar', () => {
  test.beforeEach(async ({page}) => {
    // ... open the table and sidebar
  });

  test('conditional formatting', async () => { ... });
  test('organize columns', async () => { ... });
  ...
});
```

tests/table.spec.ts
```diff
@@ -139,6 +139,176 @@ test.describe('tests complex table operations', () => {
...
+    await page.keyboard.type('3');
+    await page.getByRole('button', { name: 'No formatting' }).click();
+    await page.getByRole('button', { name: 'Positive' }).click();
+
+    const delayForGridRender = 500;
+    const conditionSelect = page.locator('#condition-select');
+
+    // Is Equal To
+    await page.getByRole('button', { name: 'Rows' }).click();
+    await page.waitForTimeout(delayForGridRender);
```
mofojed, line 201: We do not want to waitForTimeout, the [Playwright Docs explicitly discourage it's use](https://playwright.dev/docs/api/class-page#page-wait-for-timeout).

Part of the reason we don't want to is that it's flakey, e.g. it could take more than 500ms, and then this test would fail; or (what's more likely the case here) is that 500ms is way longer than it takes for the grid to re-render, so now our test is taking longer to run than necessary.

Instead, we should be asserting on things we can see on the page. There's a loading bar with text whenever we apply table operations (at least, there should be), so we can check the load is done by doing something like:
```
    // Wait until it's done loading
    await expect(
    ).toHaveCount(0);
```
If that's not working, then we need to double-check the loading bar is appearing correctly in each case, or investigate what else is going on.

tests/table.spec.ts
```diff
@@ -139,6 +139,176 @@ test.describe('tests complex table operations', () => {
...
+    await expect(page.locator('.table-sidebar')).toHaveCount(1);
+
+    // Open Conditional Formatting Panel
+    await page.locator('data-testid=menu-item-Conditional Formatting').click();
+
+    // Create New Formatting Rule
+    await page.getByRole('button', { name: 'Add New Rule' }).click();
+    await page.getByPlaceholder('Enter value').click();
+    await page.keyboard.type('3');
+    await page.getByRole('button', { name: 'No formatting' }).click();
```
mofojed, line 193: Wonder if there's a better way to select this... right now it's looking for a button that says "No formatting", when really the instructions we want are "click the dropdown labelled 'Style'". Perhaps we can update our accessibility labels/roles to make it more intuitive.

tests/table.spec.ts
```diff
@@ -139,6 +139,176 @@ test.describe('tests complex table operations', () => {
...
+
+    // Delete
+    await page.getByRole('button', { name: 'Done' }).click();
+    await page.getByRole('button', { name: 'Delete rule' }).click();
+
+    await expect(page.locator('.iris-grid-column')).toHaveScreenshot();
+
+    // Cancel
+    await page.getByRole('button', { name: 'Add New Rule' }).click();
+    await page.locator('#condition-select').first().selectOption('is-not-null');
```
mofojed, line 308: Wasn't you that added it, but the div shouldn't have an id `condition-select`. It should use a `shortid` generated id if it needs to use an `id` at all, and we should instead be using a label name or `aria-label` or `data-testid` to identify the element in our tests instead.
`id` should be unique in the whole document: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

mofojed (changes requested):

tests/table.spec.ts
ethanalvizo: Yea it seems that changing the conditional formatting rules doesn't trigger the `startLoading` function in IrisGrid.tsx and that's needed to show the loading status element. Adding in the implementation now but wanted to make sure this omission wasn't intended in the original PR #409

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2318,6 +2319,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         invertSearchColumns
       );
       this.setState({ searchFilter });
+      this.stopLoading();
```
mofojed, line 2322: As mentioned on other PR, don't want stopLoading here.

tests/table.spec.ts
```diff
@@ -128,69 +183,100 @@ test.describe('tests complex table operations', () => {
...
+  test('can conditional format', async () => {
+    // Open Conditional Formatting Panel
+    await page.locator('data-testid=menu-item-Conditional Formatting').click();
 
-    // Check snapshot
+    // Setup New Formatting Rule
+    await page.getByRole('button', { name: 'Add New Rule' }).click();
+    await page.locator('.style-editor').click();
+    await page.getByRole('button', { name: 'Positive' }).click();
+    await page.getByRole('button', { name: 'Done' }).click();
```
mofojed, line 243: Doing `test.step` to break it up here would be optimal

tests/table.spec.ts
```diff
@@ -128,69 +183,100 @@ test.describe('tests complex table operations', () => {
...
+    await expect(
+      page.locator('.iris-grid .iris-grid-loading-status')
+    ).toHaveCount(0);
 
     // Check snapshot
     await expect(page.locator('.iris-grid-column')).toHaveScreenshot();
+  });
 
-    await page.keyboard.press('Backspace');
+  test('can conditional format', async () => {
```
mofojed, line 235: Should test changing which column conditional formatting is applied on as well

mofojed (changes requested):

mofojed (approved):

PR #1330, opened 2023-05-29
feat: Improve golden layout tab overflow drop down behaviour

dsmmcken:
- Header now has overflow hidden - which fixes an issue where the controls causing overflow of the viewport
- Since header has overflow, the dropdown menu is now moved to root while opened
- Add logic to flip menu to expand upward when there's no space below
- Add long name to test layout to check truncation behaviour
- Close menu when item is picked up based on a drag from the dropdown
- Resize dropdown when window is resized

Example of incorrect overflow:
Screenshot 2023-05-26 111108


packages/golden-layout/src/controls/Header.ts
```diff
@@ -654,13 +660,14 @@ export default class Header extends EventEmitter {
...
   // enables synthetic keyboard navigation of the list
   _handleFilterKeydown(e: JQuery.TriggeredEvent) {
-    if (this.dropdownKeyIndex === -1) return;
-
     if (e.key === 'Escape') {
       this._hideAdditionalTabsDropdown();
       return;
     }
 
+    //escape should always close, even if list is empty
```
mofojed, line 668: Comment should have a space after `//` to be consistent with other comments.

That being said, I think comment here could be improved. This comment should should be before the escape key check above, then you could have a comment here saying something like `// a negative dropdownKeyIndex means the list is empty` (if that's the case, I'm inferring that from this comment saying something about the list being empty though I don't see why the list could have items in it but the dropdownKeyIndex isn't set for some reason).

packages/golden-layout/src/controls/Header.ts
```diff
@@ -773,31 +780,60 @@ export default class Header extends EventEmitter {
     this.tabDropdownContainer.hide();
     this.isDropdownShown = false;
 
+    // put it back in the header dom
```
mofojed, line 783: _Why_ are we putting it back in the header DOM? Why don't we just create the container when we show the list and then remove it completely when it's hidden?

mofojed (changes requested):

packages/golden-layout/src/controls/Header.ts
dsmmcken: list can be empty when it's filtered to something non-matching.

packages/golden-layout/src/controls/Header.ts
dsmmcken: Which would be easier if this was react and not jquery 🤷

mofojed (approved):

PR #1338, opened 2023-05-31
refactor: Clean up golden-layout css

mattrunyon:
Reverts #1334 

Applies #1322 w/ a fix for enterprise builds

Still seems to need the ignore in `karma.config.cjs` though. Probably because the golden-layout package tests use old libraries


dsmmcken (approved):
Was the change to simply move it adjacent to the .js rather than in the /scss folder?

mattrunyon:
Also had to add `./src:./dist` to the `build:sass` for golden-layout so that the module would be compiled to `GoldenLayout.module.css`

packages/golden-layout/karma.conf.cjs
```diff
@@ -25,6 +25,12 @@ module.exports = function (config) {
 
     browserify: {
       plugin: ['esmify'],
+      // ignored because it doesn't understand the scss import in GoldenLayoutThemeExport
+      configure: function (bundle) {
+        bundle.on('prebundle', function () {
+          bundle.ignore('./dist/GoldenLayoutThemeExport.js');
```
mofojed, line 31: Not sure I full understand this. The `./dist/GoldenLayoutThemeExport.js` file that's outputted has a `css` import, not an `scss` import - so what's being ignored here specifically?

mofojed (approved):

mofojed (commented):
@dsmmcken Are all the updates screenshots correct? Looks like things shifted horizontally by a couple pixels

packages/golden-layout/karma.conf.cjs
mattrunyon: This was from the original PR, but I think it is skipping this file instead of trying to bundle it with browserify which I guess doesn't like the CSS import? Idk, but I tried removing it and the tests failed.

mattrunyon:
The screenshots should be from the original PR. I just reverted the revert commit then added the fix on top of that.

PR #1340, opened 2023-05-31
fix: panels menu should only open downwards

dsmmcken:
Noticed it would shift left sometimes if window is really short. Only allow it to open downwards.


mofojed (approved):

mofojed (changes requested):
You marked this `feat`, should be `fix`

mofojed (approved):

PR #1342, opened 2023-06-01
fix: DateWrapper fails on first day of month

### LINKED ISSUE #1341: TableUtils DateWrapper test logic is incorrect on first day of month
**Description**
TableUtils DateWrapper test logic is wrong on first/last day of month. 

**Steps to reproduce**
Run the `TableUtils.test.ts` file and the 'should return a DateWrapper object if columnType is date' unit test will fail

**Expected results**
Yesterday date range is returned

**Actual results**
Yesterday date was returning as NaN

Same issue as #1335

ethanalvizo:
Closes #1341


mofojed (approved):
Great catch!

PR #1348, opened 2023-06-02
feat: improvements to null and empty strings filters in grid

### LINKED ISSUE #1243: Confusing quick filter behavior around string column "null"
I'm unable to do a literal search for the string "null" using the string "null" in the web UI quick filter - it assumes I want to search for the `null` value. This is due to an implementation detail in `io.deephaven.engine.table.impl.select.MatchFilter` when using `MatchFilter#strValues`:

```java
Object convertStringLiteral(String str) {
    if (str.equals("null")) {
        return null;
    }
    ...
}
```

It may be that this is "as-desired", but it's a potential gotcha as it's different than every other string.

image

I see the importance of wanting to filter to `null` values, but I wonder if there might be a better way to expose this functionality? I wonder if an explicit option "is null" would be better? (This could also be a good place to introduce an "is empty" filter.)

I'm able to workaround this issue by going to advanced filters and explicitly selecting the "null" value.

image

As another note, the values in the advanced filters box might want to use the text rendering that the grid cell uses, so the user can tell the difference between `null` and "".

dsmmcken:
Fixes #1243

-  advanced styling for null and empty strings
- ~quick filter input styling for null and empty strings~
- context menu styling for null and empty strings
- escape the null when using filter by menu
- escape any operator when using filter by menu
- New special empty string context menu
- Fixed broken startswith/endwith
- Allow empty string search as just "=" or "!="

Any filter containing a backslashes however is still broken in the jsapi https://github.com/deephaven/deephaven-core/issues/3912


packages/iris-grid/src/FilterInputField.scss
```diff
@@ -62,4 +62,9 @@
       }
     }
   }
+
+  &.null-style .iris-grid-input-field {
+    font-style: italic;
+    color: $gray-300;
+  }
```
mofojed, line 69: Doesn't seem to be used anywhere
```suggestion
```

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -426,6 +427,26 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+      } else if (value === '') {
+        // empty string gets a special menu
+        if (quickFilters.get(modelColumn)) {
+          filterMenu.actions.push({
+            title: 'And',
+
+            actions: this.emptyStringFilterActions(
+              column,
+              valueText,
+              value,
```
mofojed, line 439: Why do we bother passing in `valueText` and `value`? We already know what the value and the text is, it's the empty string, as specified in the function name. No need to pass it in.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -983,12 +1009,16 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
           columnIndex,
           IrisGridContextMenuHandler.getQuickFilterCondition(
             filter,
-            column.filter().contains(filterValue),
+            column
+              .filter()
+              .isNull()
+              .not()
```
mofojed, line 1015: This could've been a separate PR fixing just this case.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -1518,6 +1559,90 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
     return actions;
   }
 
+  emptyStringFilterActions(
+    column: Column,
+    valueText: string,
+    value: unknown,
```
mofojed, line 1565: No need to pass in `valueText` and `value`, we know what it is

packages/jsapi-utils/src/TableUtils.test.ts
```diff
@@ -523,6 +523,91 @@ describe('quick filter tests', () => {
...
+      ['!', '\\!'],
+    ])(
+      'given %p, escape quick filter should return %p',
+      (testCase, expectedResult) => {
+        expect(TableUtils.escapeQuickTextFilter(testCase)).toBe(expectedResult);
+      }
+    );
+  });
+
+  describe('unescapeQuickTextFilter', () => {
```
mofojed, line 568: This is just the flip side, right? You could just add another expect in the previous test.each, might make it more clear that we expect this to flip and flip back.
```
      'given %p, escape quick filter should return %p',
      (unescaped, escaped) => {
        expect(TableUtils.escapeQuickTextFilter(unescaped)).toBe(escaped);
        expect(TableUtils.unescapeQuickTextFilter(escaped)).toBe(unescaped);
      }
```

mofojed (changes requested):
Good overall, some cleanup

packages/jsapi-utils/src/TableUtils.test.ts
dsmmcken: There's a couple of unique tests either side, for checking already escaped and already unescaped strings,  I could consolidate the one that are equal though.

mofojed (changes requested):
Can you add the "Fixes #1243" to the description so it links properly? Then I'll approve.

mofojed (approved):

PR #1353, opened 2023-06-07
fix: Catch errors when emitting events to prevent breaking entire layout

### LINKED ISSUE #1352: Golden Layout can break if a panel throws while dragging
**Description**

Discovered while investigating https://github.com/deephaven/deephaven-js-plugins/issues/26, the entire layout breaks when a panel throws while dragging. Add some safeguards to prevent one erroneous panel from breaking the entire layout.

**Steps to reproduce**
1. Follow the steps in https://github.com/deephaven/deephaven-js-plugins/issues/26

**Expected results**
1. Panel should drop correctly, layout shouldn't be broken (though that particular panel may be displaying incorrectly)

**Actual results**
1. Layout breaks, panel can't drop

image

mofojed:
- Added a try/catch around all events that are emitted from Golden Layout
- Added @deephaven/log to golden layout so we can log the error
- Tested using the steps in https://github.com/deephaven/deephaven-js-plugins/issues/26
- Fixes #1352

### DIFF
--- a/packages/golden-layout/src/utils/EventEmitter.ts
+++ b/packages/golden-layout/src/utils/EventEmitter.ts
@@ -1,3 +1,7 @@
+import Log from '@deephaven/log';
+
+const log = Log.module('EventEmitter');
+
 /**
  * A generic and very fast EventEmitter
  * implementation. On top of emitting the
@@ -65,7 +69,11 @@ class EventEmitter {
     if (subs) {
       for (let i = 0; i < subs.length; i++) {
         const ctx = subs[i].ctx || {};
-        subs[i].fn.apply(ctx, args);
+        try {
+          subs[i].fn.apply(ctx, args);
+        } catch (e) {
+          log.error('Error while emitting event:', e);
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


packages/golden-layout/src/utils/EventEmitter.ts
```diff
@@ -65,7 +65,11 @@ class EventEmitter {
     if (subs) {
       for (let i = 0; i < subs.length; i++) {
         const ctx = subs[i].ctx || {};
-        subs[i].fn.apply(ctx, args);
+        try {
+          subs[i].fn.apply(ctx, args);
+        } catch (e) {
+          console.error('Error while emitting event:', e);
```
mattrunyon, line 71: Does `subs[i]` contain any useful info here? Or `subs[i].fn`?

mattrunyon (approved):

packages/golden-layout/src/utils/EventEmitter.ts
mattrunyon: Looks like the stack trace here contains the proper info

PR #1357, opened 2023-06-08
fix: Ran npm audit fix on packages

mofojed:


package-lock.json
```diff
@@ -25483,9 +25507,9 @@
       }
     },
     "node_modules/vite": {
-      "version": "4.3.5",
-      "resolved": "https://registry.npmjs.org/vite/-/vite-4.3.5.tgz",
-      "integrity": "sha512-0gEnL9wiRFxgz40o/i/eTBwm+NEbpUeTWhzKrZDSdKm6nplj+z4lKz8ANDgildxHm47Vg8EUia0aicKbawUVVA==",
+      "version": "4.3.9",
```
mattrunyon, line 25510: Should we bump this in our package.json? Looks like it's the only package in this audit that we explicitly list

mattrunyon (approved):

PR #1361, opened 2023-06-08
test: add e2e tests for advanced settings

### LINKED ISSUE #1356: E2E Test: Advanced Settings

ethanalvizo:
Closes #1356


tests/table-operations.spec.ts
```diff
@@ -490,3 +493,117 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+
+  await test.step('create input filter', async () => {
+    await page.getByRole('button', { name: 'Controls' }).click();
+
+    const inputFilter = page.getByRole('button', { name: 'Input Filter' });
+    const target = page.getByText('Command History');
+    const dropIndicator = page.locator('.lm_dragProxy');
+    await dragComponent(inputFilter, target, dropIndicator);
+
+    await page.getByTestId('btn-page-close').click();
```
mofojed, line 526: I think this should be in the previous step - it's related to close the side bar, not to adding an input filter

tests/table-operations.spec.ts
```diff
@@ -490,3 +493,117 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+
+  await test.step('use input filter', async () => {
+    const stringColumnOpt = '10';
+    await page.getByRole('combobox').selectOption(stringColumnOpt);
+    await page.getByRole('button', { name: 'Save' }).click();
+
+    await expect(page.getByPlaceholder('Enter value...')).toHaveCount(1);
+    await page.getByPlaceholder('Enter value...').fill('~a9');
+    await page.getByPlaceholder('Enter value...').click();
+    // await page.keyboard.type('~a9');
```
mofojed, line 549: I don't think you need these two lines

tests/table-operations.spec.ts
```diff
@@ -490,3 +493,117 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+    await page.keyboard.press('Enter');
+
+    await waitForLoadingDone(page);
+
+    // await expect(page.locator('.input-filter-value-card')).toHaveScreenshot();
```
mofojed, line 554: Delete unused code

tests/table-operations.spec.ts
```diff
@@ -490,3 +493,117 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+  });
+
+  await test.step('drag table to new panel', async () => {
+    // opening up this menu makes it easier to drag to that corner
+    await page
+      .getByTestId('btn-iris-grid-settings-button-table')
+      .nth(1)
+      .click();
+
+    const table = page.locator('.lm_tabs').nth(2).locator('.lm_tab').first();
```
mofojed, line 575: Can you get the tab by the table name instead?

mofojed (changes requested):

tests/table-operations.spec.ts
dsmmcken: `page.locator('.lm_tab').filter({ has: page.getByText('test-z') });` is what I did in the golden-layout test.

tests/table-operations.spec.ts
```diff
@@ -490,3 +493,117 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+    await page.getByRole('button', { name: 'Linker' }).click();
+
+    const firstStringCol = page.locator('.iris-grid .grid-wrapper').first();
+    await firstStringCol.click({ position: { x: 20, y: 10 } });
+    const secondStringCol = page.locator('.iris-grid .grid-wrapper').nth(1);
+    await secondStringCol.click({
+      position: { x: 20, y: 10 },
+    });
+
+    await page.getByRole('button', { name: 'Done' }).click();
```
mofojed, line 596: Should probably restrict this to within the linker dialog

tests/table-operations.spec.ts
mofojed, line 1: So to be a little bit more clear on what these advanced options are setting... you should be selecting one column with the Input Filter, and a different column with the linker to verify this is actually testing these advanced settings.
The advanced settings when enabled will clear _all_ the filters on the table before applying the new filter; otherwise they just apply the new filter on the column in addition to the current filter.
So I think to fully test these options (both off and on), you would do the following:
- Create tables A and B
- Add an Input Filter that's wired to the `String` column
- Add a link from table A to table B on the `Int` column
- Set the input filter to filter the tables. Now both tables should be filtered on the `String` column
- Double click the first table. Now the table B should be filtered on both the `String` and `Int` column. This is verifying that the Advanced Setting for incoming links in it's default state (off) is working correctly.
- Change the input filter. Verify that Table B is now filtered on the new value and still filtered on the `Int` column. This is verifying that the Advanced Setting for new filters from a control in it's default state (off) is working correctly.
- Change the advanced setting for from a control on Table B to On.
- Change the input filter value. Now table B should just be filtered on the new `String` filter, the `Int` filter should be removed.
- Click the link from the first table. Table B should be filtered on `String` and `Int` now (verifying the advanced setting for from an incoming link is still off and is distinct from the other setting)
- Change the advanced filter setting from a link to On
- Click the link from the first table. Table B should now be filtered only on `Int`.

I think that covers all the possible cases for advanced settings.

tests/table-operations.spec.ts
ethanalvizo: Was thinking about this as well but the way tables are generated in the table tests give them unique and random names each time. I can rewrite it to force a table name onto the ones I created but I assume there's a reason we wanted unique names for all the tables created here?

image

mofojed: You can create the random table name ahead of time to pass into the function and use that

tests/table-operations.spec.ts
```diff
@@ -89,6 +89,14 @@ async function changeCondFormatHighlight(page: Page) {
   await waitForLoadingDone(page);
 }
 
+async function artificialWait(page: Page) {
```
mofojed, line 92: Add some comments on why we need this... basically that sometimes the loading indicator appears so quickly playwright misses it.

I'm still looking for a way to see if you can increase the retry frequency for a test to get around this artificialWait (as I assume that's what the problem is).

mofojed (changes requested):
Some minor cleanup... I really do want to get to the bottom of that artificialWait thing

tests/table-operations.spec.ts
```diff
@@ -89,6 +90,21 @@ async function changeCondFormatHighlight(page: Page) {
...
+ * Acts as a small time buffer so that the table has time to update before the screenshot.
+ * Can be deleted in the future if the timing is figured out with the polling intervals for advanced settings.
+ * Without it the snapshot is taken before the table is updated.
+ * Attempt was made to check for loading status bar with custom polling intervals but no luck
+ * @param page
+ */
+async function artificialWait(page: Page) {
+  const tableOperationsMenu = page
+    .locator('data-testid=btn-iris-grid-settings-button-table')
+    .nth(1);
```
mofojed, line 103: Don't think this should be hardcoded to 1, should pass in as a param

tests/table-operations.spec.ts
```diff
@@ -490,3 +509,165 @@ test('rollup rows and aggregrate columns', async ({ page }) => {
...
+    await page.getByRole('button', { name: 'Save' }).click();
+
+    await expect(page.getByPlaceholder('Enter value...')).toHaveCount(1);
+    await page.getByPlaceholder('Enter value...').click();
+    await page.keyboard.type('>1000');
+    await expect(page.getByPlaceholder('Enter value...')).toHaveValue('>1000');
+
+    await waitForLoadingDone(page);
+    await waitForLoadingDone(page, 1);
+    await artificialWait(page);
```
mofojed, line 595: Really the wait should be before the waitForLoadingDone, since that will fire immediately.

mofojed (changes requested):
Very minor changes, then let's get it in

mofojed (approved):

PR #1363, opened 2023-06-08
fix: Cannot add control from Controls menu with click

mofojed:
- We were defaulting the value to `null` instead of `undefined`
- A golden-layout refactor explicitly checked for `undefined` instead of `null`
- For some reason the type error in `FilterPlugin` and also `AppMainContainer` wasn't throwing an error when building?? Will need to investigate why not
- Loosen restriction to just be `!= null` instead of `!== undefined` as that is the intent


dsmmcken:
Is it the same event that is erroring if I look at GridPlugin.tsx?

dsmmcken (approved):
Works but event typing around this is messed up.

PR #1365, opened 2023-06-09
revert: adding back "Table rendering support for databars (#1212)"

ethanalvizo:


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -496,7 +496,7 @@ class IrisGridTableModelTemplate<
     }
   }
 
-  textValueForCell(x: ModelIndex, y: ModelIndex): string | null {
+  textValueForCell(x: ModelIndex, y: ModelIndex): string | null | undefined {
```
ethanalvizo, line 499: this function is the only thing I changed from the original commit

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -2011,7 +2025,7 @@ class IrisGridTableModelTemplate<
     }
   }
 
-  editValueForCell(x: ModelIndex, y: ModelIndex): string | null {
+  editValueForCell(x: ModelIndex, y: ModelIndex): string | null | undefined {
```
ethanalvizo, line 2028: this is the other

PR #1370, opened 2023-06-12
feat: Console output test util

### LINKED ISSUE #1369: Test Util - Disable console methods
As a developer, I would like a test util that makes it easy to disable console log in unit tests. This will help with certain cases where 3rd party code includes `console.error` statements + some other internal use cases where we may want to disable for certain tests.

bmingles:
Console output test util

resolves #1369

### DIFF
--- a/packages/utils/src/TestUtils.test.tsx
+++ b/packages/utils/src/TestUtils.test.tsx
@@ -20,16 +20,29 @@ describe('asMock', () => {
 });
 
 describe('disableConsoleOutput', () => {
+  const arg0 = 'mock arg';
+  const allMethodNames = ['log', 'warn', 'error', 'info', 'debug'] as const;
+
+  it('should disable all methods if given no method names', () => {
+    TestUtils.disableConsoleOutput();
+
+    allMethodNames.forEach(methodName => {
+      // eslint-disable-next-line no-console
+      console[methodName](arg0);
+
+      // eslint-disable-next-line no-console
+      expect(console[methodName]).toHaveBeenCalledWith(arg0);
+    });
+  });
+
   it.each([
     [['log']],
     [['warn']],
     [['error']],
     [['info']],
     [['debug']],
-    [['log', 'warn', 'error', 'info', 'debug']],
+    [allMethodNames],
   ] as const)('should mock console method implementations: %s', methodNames => {
-    const arg0 = 'mock arg';
-
     TestUtils.disableConsoleOutput(...methodNames);
 
     methodNames.forEach(methodName => {
--- a/packages/utils/src/TestUtils.test.tsx
+++ b/packages/utils/src/TestUtils.test.tsx
@@ -19,6 +19,29 @@ describe('asMock', () => {
   expect(someFunc('a,b,c')).toEqual(3);
 });
 
+describe('disableConsoleOutput', () => {
+  it.each([
+    [['log']],
+    [['warn']],
+    [['error']],
+    [['info']],
+    [['debug']],
+    [['log', 'warn', 'error', 'info', 'debug']],
+  ] as const)('should mock console method implementations: %s', methodNames => {
+    const arg0 = 'mock arg';
+
+    TestUtils.disableConsoleOutput(...methodNames);
+
+    methodNames.forEach(methodName => {
+      // eslint-disable-next-line no-console
+      console[methodName](arg0);
+
+      // eslint-disable-next-line no-console
+      expect(console[methodName]).toHaveBeenCalledWith(arg0);
+    });
+  });
+});
+
 describe('findLastCall', () => {
   it('should return undefined if call not matched', () => {
     const fn = jest.fn<void, [string, number]>();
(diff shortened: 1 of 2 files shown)


packages/utils/src/TestUtils.ts
```diff
@@ -52,6 +52,24 @@ class TestUtils {
     fn: (...args: TArgs) => TResult
   ): jest.Mock<TResult, TArgs> => (fn as unknown) as jest.Mock<TResult, TArgs>;
 
+  /**
+   * Selectively disable logging methods on `console` object. Uses spyOn so that
+   * changes will be reverted after leaving the test scope that it is set in. If
+   * no method names are given, all will be disabled.
+   */
```
mattrunyon, line 59: I would maybe change this to a single param that takes an array and then clarify the intention w/ `@param methodNames The console methods to disable`

packages/utils/src/TestUtils.ts
```diff
@@ -52,6 +52,24 @@ class TestUtils {
...
+  static disableConsoleOutput = (
+    ...methodNames: ('log' | 'warn' | 'error' | 'info' | 'debug')[]
+  ): void => {
```
mattrunyon, line 62: This will let you actually disable all console methods if you wanted to (there are 24 methods on console)

```suggestion
  static disableConsoleOutput(
    methodNames: (keyof typeof console)[]
  ): void {
```

packages/utils/src/TestUtils.ts
```diff
@@ -52,6 +52,24 @@ class TestUtils {
...
+    if (methodNames.length === 0) {
+      // eslint-disable-next-line no-param-reassign
+      methodNames = ['log', 'warn', 'error', 'info', 'debug'];
```
mattrunyon, line 65: Might as well mock everything?

```suggestion
      methodNames = Object.getOwnPropertyNames(console).filter(name => typeof console[name] === 'function');
```

packages/utils/src/TestUtils.test.tsx
```diff
@@ -19,6 +19,42 @@ describe('asMock', () => {
...
 
+describe('disableConsoleOutput', () => {
+  const arg0 = 'mock arg';
+  const allMethodNames = ['log', 'warn', 'error', 'info', 'debug'] as const;
+
+  it('should disable all methods if given no method names', () => {
+    TestUtils.disableConsoleOutput();
+
+    allMethodNames.forEach(methodName => {
+      // eslint-disable-next-line no-console
```
mattrunyon, line 30: You can just wrap the entire test block w/ this instead

```
/* eslint-disable no-console */
// Code here that will have no-console turned off
/* eslint-enable no-console */
// Code here will lint no-console rule again
```

mattrunyon (changes requested):

mattrunyon:
Do you have a specific example for using this right now? Is it just to disable 3rd party logging like react warning about unmounting component lifecycles? Our log package should be silent during tests by default

packages/utils/src/TestUtils.ts
bmingles: > 

There are a few prop type warning / errors in existing tests which might be good candidates. The specific use case though is in `setupTests.js` I added a `console.error` to surface hook errors that are otherwise swallowed. For cases where we expect an error, this is a way to disable.

bmingles:
>

There are a few prop type warning / errors in existing tests which might be good candidates. The specific use case though is in `setupTests.js` I added a `console.error` to surface hook errors that are otherwise swallowed. For cases where we expect an error, this is a way to disable.

packages/utils/src/TestUtils.ts
bmingles: I did actually look at just using the `keyof Console`, but there's 1 prop that it doesn't like, so end up having to do `keyof Omit<Console, 'Console'>`.

packages/utils/src/TestUtils.test.tsx
bmingles: I'm aware of that format, but still ends up being 2 eslint comments. Didn't really see a benefit for this particular use case

packages/utils/src/TestUtils.ts
bmingles: Good catch. I will add the `@param methodNames`. Is there a strong reason not to use the rest params though? Looks like it works fine either way.

image

packages/utils/src/TestUtils.test.tsx
bmingles: actually nvm. I see there were more

packages/utils/src/TestUtils.ts
mattrunyon: The main benefit would be adding params in the future is easier if needed, but I don't expect this to be used too often.

I think part of my tendency also stems from before `...rest` existed and you had to use `arguments` as a magic value if you wanted to use variadic functions

packages/utils/src/TestUtils.ts
bmingles: I was able to get this working

packages/utils/src/TestUtils.ts
bmingles: agreed

packages/utils/src/TestUtils.ts
bmingles: I'll probably just add the jsdoc on this one and leave the rest param unless you feel strongly otherwise

mattrunyon (approved):

PR #1383, opened 2023-06-21
fix: AuthPluginParent wasn't working when embedded in an iframe

### LINKED ISSUE #1373: AuthPluginParent should work when embed is opened in an iframe
**Description**

Discovered while @jnumainville was testing deephaven-ipywidgets in DnD Jupyter. We were hoping to use the AuthPluginParent to provide the authentication details, but that only works when the window is opened in another tab, not in an iframe.

Providing a zip that contains an html file reproducing the issue when you are running Deephaven locally at localhost:10000

**Steps to reproduce**

1. Open the attached iframe-auth-provider.html example
2. Open the developer console

**Expected results**
2. Developer console should have a `message received` printed out, and the iframe should time out after 10s (since it tries to use the auth plugin provider

**Actual results**
2. AuthPluginParent doesn't get used, it's just trying to use the AuthPluginAnonymous.

The error is because we are only checking `window.opener`, instead of also checking `window.parent`: https://github.com/deephaven/web-client-ui/blob/626de830ba4f580c90b0d0e2ee51ce8fd0452ad9/packages/auth-plugins/src/AuthPluginParent.tsx#L52

We also need to make sure we send the message to `window.parent` as well in this case, instead of `window.opener` (which is null): https://github.com/deephaven/web-client-ui/blob/626de830ba4f580c90b0d0e2ee51ce8fd0452ad9/packages/jsapi-utils/src/MessageUtils.ts#L134
[iframe-auth-provider.zip](https://github.com/deephaven/web-client-ui/files/11820138/iframe-auth-provider.zip)

mofojed:
- Check for window.opener _or_ window.parent when using AuthPluginParent
- Reorder plugin priority to use AuthPluginParent first if specified
- Tested using an example html page that both embeds and pops open a new tab
- Fixes #1373 

To test:
1. Start up deephaven-core with PSK set to `hello`: `./gradlew server-jetty-app:run -Ppsk=hello`
2. Start up Web UI: `npm start`
3. Create a table named `t` in the Web UI at https://localhost:4000:
```
t = empty_table(50).update(["x=i%5", "y=(i*i)%5"])
```
4. Open the HTML file in the provided zip.
[iframe-auth-provider-local.zip](https://github.com/deephaven/web-client-ui/files/11820375/iframe-auth-provider-local.zip)
5. Ensure the table appears correctly in the iframe. Click the "Open in new tab" link, and ensure the table opens there correctly as well.

Also tested that going to localhost:4010/?name=t&authProvider=parent directly did not try and use the parent plugin, because it wasn't opened by another window.

### DIFF
--- a/packages/jsapi-utils/src/MessageUtils.test.ts
+++ b/packages/jsapi-utils/src/MessageUtils.test.ts
@@ -1,3 +1,4 @@
+import { TestUtils } from '@deephaven/utils';
 import {
   makeMessage,
   makeResponse,
@@ -7,11 +8,11 @@ import {
 
 it('Throws an exception if called on a window without parent', async () => {
   await expect(requestParentResponse('request')).rejects.toThrow(
-    'window.opener is null, unable to send request.'
+    'window parent is null, unable to send request.'
   );
 });
 
-describe('requestParentResponse', () => {
+describe('requestParentResponse with opener', () => {
   let addListenerSpy: jest.SpyInstance;
   let removeListenerSpy: jest.SpyInstance;
   let listenerCallback;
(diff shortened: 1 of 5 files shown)


packages/jsapi-utils/src/MessageUtils.test.ts
```diff
@@ -80,3 +81,78 @@ describe('requestParentResponse', () => {
...
+
+  afterEach(() => {
+    addListenerSpy.mockRestore();
+    removeListenerSpy.mockRestore();
+    windowParentSpy.mockRestore();
+    mockPostMessage.mockClear();
+    messageId = undefined;
+  });
+
+  it('Posts message to parent and subscribes to response', async () => {
```
vbabich, line 116: Are these `it` blocks the same as above? Can we maybe reuse them? wrap in a function or something.

vbabich (approved):

packages/jsapi-utils/src/MessageUtils.test.ts
mofojed: Good call, refactored

vbabich (approved):

PR #1387, opened 2023-06-26
build: Some TypeScript issues were not being identified correctly

### LINKED ISSUE #1368: Some TypeScript issues are not breaking the build
**Description**

It's been observed there are some errors in TypeScript that are shown in VS Code, but are not breaking the build. An example is in `GridPlugin.tsx`, around the `dragEvent`: https://github.com/deephaven/web-client-ui/blob/a5d4e21af6c8b8b31205c85bf4c7917c48319d72/packages/dashboard-core-plugins/src/GridPlugin.tsx#L81

image

mofojed:
- Remove `types` declaration from `tsconfig.json`
  - Looking at the documentation, setting this ignores all other `@types/*` and only uses the types you specify globally: https://www.typescriptlang.org/tsconfig#types
- Use a `DragListenerEvent` type that is used in `@deephaven/golden-layout` for handling drags
  - Only pick the attributes that are actually used from `JQuery.TriggeredEvent`
  - Now both React and JQuery events can be passed safely
- Converted the remaining styleguide examples from jsx to tsx
- Remove dependency on `@deephaven/tsconfig.json` from each package
  - Only needed from the root itself
- Fixes #1368


mattrunyon:
I'm not sure removing types is the right answer, but maybe we need to add jquery to types? By removing types, all `@types/*` are now visible globally which probably isn't what we want. Although as more libraries move to native TS, I think this becomes less of a problem

Or maybe we do want it this way. It seems to only expose `declare` statements from @types from what I can tell, so it shouldn't affect most of the types we have.

With this change you can have `React.DragEvent` as a type without any imports related to react. Which might not be bad. If you just have `DragEvent` it will pick the browser version though since react types export in a namespace

mofojed:
@mattrunyon I haven't noticed any weirdness of types. I'd rather use the default settings than override when possible.
When you say "all `@types/*` are now visible globally", what does that mean? The types are not visible on the global scope, you still need to import the correct module when using it.

mattrunyon:
@mofojed All things in the @types directory in node_modules

For example, Jest (which we want) declares things that are global in `node_modules/@types/jest`. That's why you can just call `beforeAll` or `describe` in any TS file and there's no complaint.

React adds its types to the global namespace, but doesn't declare that functions just exist AFAIK. With our existing setup, you cannot say `React.DragEvent` as a type without importing React into the file. With the change, you can. But that's probably fine considering that it's namespaced (`DragEvent` still refers to DOM event, not React's event) and it's only a type. I don't think any of the libraries (except Jest) declare functions/values that exist globally

mofojed:
@mattrunyon  I haven't noticed anything weird compared to previous behaviour.

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -468,7 +468,7 @@ class LayoutUtils {
     replaceConfig?: Partial<ItemConfigType>;
     createNewStack?: boolean;
     focusElement?: string;
-    dragEvent?: DragSourceEvent;
+    dragEvent?: DragEvent;
```
mattrunyon, line 471: W/ the change to `types` in tsconfig, we could just use `React.DragEvent` here without adding an import. Not sure how you feel about that

packages/golden-layout/tsconfig.json
```diff
@@ -5,5 +5,6 @@
     "outDir": "dist/"
   },
   "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.js", "src/**/*.jsx"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules"],
+  "references": [{ "path": "../components" }]
```
mattrunyon, line 9: Good catch. Although it's only used for scss so probably why it never caused an issue

mattrunyon (approved):

packages/dashboard/src/layout/LayoutUtils.ts
mofojed: Same difference to me

packages/golden-layout/tsconfig.json
mofojed: True

PR #1390, opened 2023-06-28
feat: disable column sorting on unsupported types

### LINKED ISSUE #1380: Columns without a sortable attribute should disable sort behaviours
Blocked by deephaven/deephaven-core#1914. 

Once columns have sortable attributes the UI should disable certain sorting behaviours for grid.
1. Disable "Sort by" and "Add additional sort" menu options in the column header context menu
image
2. Disable the clicking of a column header from triggering a sort
3. Add "`<icon=dhSortSlash>` Not sortable"  as text to the column tooltip, below where the column frozen text appears.
image

ethanalvizo:
Closes #1380

### DIFF
--- a/packages/iris-grid/src/AdvancedFilterCreator.test.tsx
+++ b/packages/iris-grid/src/AdvancedFilterCreator.test.tsx
@@ -54,7 +54,7 @@ function makeAdvancedFilterCreatorWrapper(
       selectedValues: [],
     },
     model: irisGridTestUtils.makeModel(),
-    column: irisGridTestUtils.makeColumn(),
+    column: irisGridTestUtils.makeColumn('0'), // needs to match column name in makeModel so that columnIndex can be found when rendering
     formatter: new Formatter(dh),
   }
 ) {
--- a/packages/iris-grid/src/AdvancedFilterCreator.tsx
+++ b/packages/iris-grid/src/AdvancedFilterCreator.tsx
@@ -202,7 +202,7 @@ class AdvancedFilterCreator extends PureComponent<
     );
     this.valuesTablePromise
       .then(valuesTable => {
-        if (valuesTable.columns[0].isSortable) {
+        if (valuesTable.columns[0].isSortable ?? false) {
           const sort = valuesTable.columns[0].sort().asc();
           valuesTable.applySort([sort]);
         }
@@ -385,7 +385,7 @@ class AdvancedFilterCreator extends PureComponent<
    */
   sortTable(direction: SortDirection, addToExisting = false): void {
     const { column, onSortChange } = this.props;
-    if (column.isSortable) {
+    if (column.isSortable ?? false) {
       onSortChange(column, direction, addToExisting);
     }
   }
@@ -568,7 +568,7 @@ class AdvancedFilterCreator extends PureComponent<
                 onClick={this.handleSortDown}
                 icon={dhSortAmountDown}
                 tooltip={`Sort ${column.name} Descending`}
-                disabled={!column.isSortable}
+                disabled={!(column.isSortable ?? false)}
               />
               <Button
                 kind="ghost"
@@ -580,7 +580,7 @@ class AdvancedFilterCreator extends PureComponent<
                   <FontAwesomeIcon icon={dhSortAmountDown} rotation={180} />
                 }
                 tooltip={`Sort ${column.name} Ascending`}
-                disabled={!column.isSortable}
+                disabled={!(column.isSortable ?? false)}
               />
             </div>
           </div>
--- a/packages/iris-grid/src/AdvancedFilterCreator.tsx
+++ b/packages/iris-grid/src/AdvancedFilterCreator.tsx
@@ -214,11 +214,7 @@ class AdvancedFilterCreator extends PureComponent<
     );
     this.valuesTablePromise
       .then(valuesTable => {
-        const columnIndex = model.getColumnIndexByName(
-          valuesTable.columns[0].name
-        );
-        assertNotNull(columnIndex);
-        if (model.isColumnSortable(columnIndex)) {
+        if (valuesTable.columns[0].isSortable ?? true) {
           const sort = valuesTable.columns[0].sort().asc();
           valuesTable.applySort([sort]);
         }
--- a/packages/iris-grid/src/AdvancedFilterCreator.tsx
+++ b/packages/iris-grid/src/AdvancedFilterCreator.tsx
@@ -202,9 +202,10 @@ class AdvancedFilterCreator extends PureComponent<
     );
     this.valuesTablePromise
       .then(valuesTable => {
-        const sort = valuesTable.columns[0].sort().asc();
-        valuesTable.applySort([sort]);
-
+        if (valuesTable.columns[0].isSortable) {
+          const sort = valuesTable.columns[0].sort().asc();
+          valuesTable.applySort([sort]);
+        }
         this.setState({ valuesTable });
       })
       .catch(error => {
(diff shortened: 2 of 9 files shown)


packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -565,6 +568,7 @@ class AdvancedFilterCreator extends PureComponent<
                 onClick={this.handleSortDown}
                 icon={dhSortAmountDown}
                 tooltip={`Sort ${column.name} Descending`}
+                disabled={!column.isSortable}
```
mofojed, line 571: @dsmmcken I wasn't sure about in Advanced Filters - are we just disabling these buttons? Or should it say "Not sortable" with that icon?

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2410,7 +2410,9 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       modelColumn,
       addToExisting
     );
-    this.updateSorts(sorts);
+    if (model.columns[columnIndex].isSortable) {
+      this.updateSorts(sorts);
+    }
```
mofojed, line 2415: We should add a `log.debug` in an `else` statement here indicating we didn't sort because the column wasn't sortable.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -291,6 +291,14 @@ abstract class IrisGridModel<
     return false;
   }
 
+  /**
+   * @param index The column index to check
+   * @returns Whether the column is sortable
+   */
+  isColumnSortable(index: ModelIndex): boolean {
```
mofojed, line 298: You don't need these methods on `IrisGridModel`, as you can just access `isSortable` on the `Column` itself.

I guess it is strange that there is an `isColumnFrozen` method, rather than accessing that on the `Column` itself. I would prefer to make that one the odd one out though.

mofojed (changes requested):

packages/iris-grid/src/AdvancedFilterCreator.tsx
dsmmcken: I recommended disable, but could also change the tooltip to "Column is not sortable"

packages/jsapi-types/src/dh.types.ts
```diff
@@ -558,6 +558,7 @@ export interface Column {
   readonly constituentType: string;
 
   readonly isPartitionColumn: boolean;
+  readonly isSortable: boolean;
```
mofojed, line 561: Need this to be optional because we do not have it implemented on Enterprise.
```suggestion
  readonly isSortable?: boolean;
```

packages/iris-grid/src/IrisGridModel.ts
ethanalvizo: The IrisGridProxyModel complains if the isColumSortable method is erased from here

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -203,7 +203,7 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
       model.getColumnHeaderParentGroup(modelIndex, 0) === undefined &&
       !(isExpandableGridModel(model) && model.hasExpandableRows);
     const isColumnFrozen = model.isColumnFrozen(modelIndex);
-    const isColumnSortable = column.isSortable;
+    const isColumnSortable = column.isSortable ?? false;
```
mofojed, line 206: We want to default to `true` if the `isSortable` property isn't present. Otherwise we won't be able to sort at all on Enterprise!

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -270,7 +270,7 @@ class IrisGridTableModel extends IrisGridTableModelTemplate<Table, UIRow> {
   }
 
   isColumnSortable(modelIndex: ModelIndex): boolean {
-    return this.columns[modelIndex].isSortable;
+    return this.columns[modelIndex].isSortable ?? false;
```
mofojed, line 273: Need to default to `true`

packages/iris-grid/src/IrisGridModel.ts
mofojed: Right, you'd need to remove it from `IrisGridProxyModel` as well.
Making this property optional is annoying though, as it does require adding the `?? true` everywhere that accesses it... so perhaps leave this method in, and make the other changes to default to `?? true` that I mentioned (instead of `?? false`) and opt to use `model.isColumnSortable(column)` instead of the property on the column directly when possible.

mofojed (changes requested):

packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -202,7 +214,11 @@ class AdvancedFilterCreator extends PureComponent<
...
       .then(valuesTable => {
-        if (valuesTable.columns[0].isSortable ?? false) {
+        const columnIndex = model.getColumnIndexByName(
+          valuesTable.columns[0].name
+        );
+        assertNotNull(columnIndex);
+        if (model.isColumnSortable(columnIndex)) {
           const sort = valuesTable.columns[0].sort().asc();
           valuesTable.applySort([sort]);
         }
```
mofojed: In this particular case, you'd still want to check the `valuesTable` column... the `valuesTable` is generated from the model, but the capabilities will not match the original model.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2410,7 +2410,11 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       modelColumn,
       addToExisting
     );
-    this.updateSorts(sorts);
+    if (model.columns[columnIndex].isSortable ?? false) {
```
mofojed, line 2413: The `const sorts = TableUtils.toggleSort ...` should be inside this if block, before the `updateSorts`.. no point creating the sort if it's not sortable.
Also should be checking `model.isColumnSortable` here.

mofojed (changes requested):

packages/iris-grid/src/AdvancedFilterCreator.test.tsx
```diff
@@ -54,7 +54,7 @@ function makeAdvancedFilterCreatorWrapper(
       selectedValues: [],
     },
     model: irisGridTestUtils.makeModel(),
-    column: irisGridTestUtils.makeColumn(),
+    column: irisGridTestUtils.makeColumn('0'), // needs to match column name in makeModel so that columnIndex can be found when rendering
```
ethanalvizo, line 57: There's probably a more fool-proof way to achieve this?

ethanalvizo: Maybe I should be handling that null/undefined case differently in AdvancedFilterCreator.tsx

mofojed: I think this is fine. Anything else you'd do would be a bit long-winded and unnecessary. You could define the columns beforehand, for example, so something like:
```
const columns = irisGridTestUtils.makeColumns(3, 'test');
...
function makeAdvancedFilterCreatorWrapper(
....
    } = {
        model: irisGridTestUtils.makeModel(irisGridTestUtils.makeTable(columns)),
        column: column[0],
```

There's still a lot of tests here disabled, we should really look at re-enabling them.

mofojed (approved):

PR #1391, opened 2023-06-30
feat: Outdent code when running selection from a notebook

### LINKED ISSUE #1326: DevX feedback: IDE could handle indents better
This is more of a DevX feedback:  I noticed that when I arbitrarily select a few lines within a Python function to run just those lines, I hit
```
Type: <class 'IndentationError'>
Value: unexpected indent (<string>, line 1)
```
type of errors. While that’s technically correct, PyCharm and other IDEs got really good at doing the smart thing in such cases, and handle it automatically.

Not sure if that’s an easy fix but could help improve the DevX of your DH UI editor.

mattrunyon:
Fixes #1326 

To test, put code in a notebook and run a selection where every line is indented. For example, run the function body of this code

```python
def my_func():
  a = 2
  b = 3
  c = a + b
  print(c)
```

Could be made into a util if we want to use it elsewhere. I looked in the monaco documentation to see if it had anything built-in for this and couldn't find any methods for it

### DIFF
--- a/packages/console/src/notebook/ScriptEditor.tsx
+++ b/packages/console/src/notebook/ScriptEditor.tsx
@@ -138,7 +138,15 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     const wholeLineRange = range
       .setStartPosition(startLineNumber, startLineMinColumn)
       .setEndPosition(endLineNumber, endLineMaxColumn);
-    return model?.getValueInRange(wholeLineRange);
+
+    // Un-indent lines in case a user selected only lines within a block to run
+    const lines = model?.getValueInRange(wholeLineRange).split('\n');
+    const minIndent = Math.min(
+      ...lines.map(line => line.length - line.trimStart().length)
+    );
+    const value = lines.map(line => line.slice(minIndent)).join('\n');
+
+    return value;
   }
 
   handleEditorInitialized(innerEditor: editor.IStandaloneCodeEditor): void {
--- a/packages/console/src/notebook/ScriptEditor.tsx
+++ b/packages/console/src/notebook/ScriptEditor.tsx
@@ -5,7 +5,7 @@ import React, { Component, ReactElement, RefObject } from 'react';
 import { LoadingOverlay, ShortcutRegistry } from '@deephaven/components';
 import Log from '@deephaven/log';
 import type { IdeSession } from '@deephaven/jsapi-types';
-import { assertNotNull } from '@deephaven/utils';
+import { assertNotNull, TextUtils } from '@deephaven/utils';
 import { editor, IDisposable } from 'monaco-editor';
 import Editor from './Editor';
 import { MonacoProviders, MonacoUtils } from '../monaco';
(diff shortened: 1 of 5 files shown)


packages/console/src/notebook/ScriptEditor.tsx
```diff
@@ -138,7 +138,15 @@ class ScriptEditor extends Component<ScriptEditorProps, ScriptEditorState> {
     const wholeLineRange = range
       .setStartPosition(startLineNumber, startLineMinColumn)
       .setEndPosition(endLineNumber, endLineMaxColumn);
-    return model?.getValueInRange(wholeLineRange);
+
+    // Un-indent lines in case a user selected only lines within a block to run
+    const lines = model?.getValueInRange(wholeLineRange).split('\n');
+    const minIndent = Math.min(
```
mofojed, line 144: Using a `reduce` would probably be slightly more performant.

mofojed (approved):

dsmmcken (dismissed):
```py
def my_func():
    # run from here
    a = 2
    b = 3

    def nested(x,y):
        return (a + b)

    c = nested(a, b)
    print(c)
    # to here
    return c
```

Failure case

mattrunyon:
@dsmmcken Good catch. The newline is what wasn't accounted for (indent size is 0). Maybe I should pull this out into a util and write a few unit tests

packages/console/src/notebook/ScriptEditor.tsx
mattrunyon: Yes it would. Although probably not that important since it's run 1 time, but I can't pass up free performance!

[Link](https://codeburst.io/javascript-finding-minimum-and-maximum-values-in-an-array-of-objects-329c5c7e22a2)

packages/utils/src/TextUtils.test.ts
```diff
@@ -85,3 +85,31 @@ describe('sort text', () => {
...
+    {
+      input: '\n\n  a\n\n    b\n  c',
+      output: '\n\na\n\n  b\nc',
+      case: 'blank lines in the code',
+    },
+    {
+      input: '',
+      output: '',
+      case: 'empty string',
+    },
```
mofojed, line 111: While we're adding cases, should have one with no indents on any line, and ones with trailing whitespace (which this outdenting shouldn't affect), and some with `\t` characters as well.

packages/utils/src/TextUtils.ts
```diff
@@ -58,6 +58,24 @@ class TextUtils {
...
     return 0;
   }
+
+  /**
+   * Outdent (opposite of indent) a string so the lowest indent level is 0
+   * Assumes each line is indented with the same characters (i.e. no mixed tabs/spaces)
+   * @param code The code to outdent
+   * @returns A code block where the minimum indent level of a line is 0
+   */
+  static outdentCode(code: string) {
```
mofojed, line 68: I'd almost rather put this in `ScriptEditorUtils`, as this function is more specific to a code block than it is to just plain text.

mofojed (changes requested):

mofojed (approved):

PR #1394, opened 2023-07-03
feat: "Extends" TypeScript util type

### LINKED ISSUE #1393: TypeScript utility type: Extends
As a developer I'd like to be able to assert that a union type is a subset of another. A simple "Extends" utility type as suggested here should work:

```
type Extends<T, U extends T> = U;
```

bmingles:
TypeScript util type "Extends"

resolves #1393

### DIFF
--- a/packages/utils/src/TypeUtils.ts
+++ b/packages/utils/src/TypeUtils.ts
@@ -11,7 +11,7 @@
  * // Compiler will complain
  * type NotDirection = Extends<Direction, 'blah'>
  */
-export type Extends<T, U extends T> = U;
+export type Extends<T, U extends T> = U extends T ? U : never;
 
 /**
  * Util type to extract the value from an object.
--- a/packages/utils/src/TypeUtils.ts
+++ b/packages/utils/src/TypeUtils.ts
@@ -6,10 +6,10 @@
  * declare type Direction = 'north' | 'south' | 'east' | 'west'
  *
  * // This works
- * type Down = Subset<Direction, 'south'> // 'south'
+ * type Down = Extends<Direction, 'south'> // 'south'
  *
  * // Compiler will complain
- * type NotDirection = Subset<Direction, 'blah'>
+ * type NotDirection = Extends<Direction, 'blah'>
  */
 export type Extends<T, U extends T> = U;
 
--- a/packages/utils/src/TypeUtils.ts
+++ b/packages/utils/src/TypeUtils.ts
@@ -1,4 +1,19 @@
 /**
+ * Util type to create a "subtype" of T. Useful for creating subsets of union
+ * types.
+ *
+ * e.g.
+ * declare type Direction = 'north' | 'south' | 'east' | 'west'
+ *
+ * // This works
+ * type Down = Subset<Direction, 'south'> // 'south'
+ *
+ * // Compiler will complain
+ * type NotDirection = Subset<Direction, 'blah'>
+ */
+export type Extends<T, U extends T> = U;
+
+/**
  * Util type to extract the value from an object.
  *
  * e.g. Given


packages/utils/src/TypeUtils.ts
```diff
@@ -1,3 +1,18 @@
+/**
+ * Util type to create a "subtype" of T. Useful for creating subsets of union
+ * types.
+ *
+ * e.g.
+ * declare type Direction = 'north' | 'south' | 'east' | 'west'
+ *
+ * // This works
+ * type Down = Subset<Direction, 'south'> // 'south'
```
mattrunyon, line 9: Assuming these examples should be Extends

packages/utils/src/TypeUtils.ts
```diff
@@ -1,3 +1,18 @@
...
+ *
+ * // Compiler will complain
+ * type NotDirection = Subset<Direction, 'blah'>
+ */
+export type Extends<T, U extends T> = U;
```
mattrunyon, line 14: What's the use case you're trying to solve with this? Would the built-in Extract utility work?

mattrunyon (changes requested):

packages/utils/src/TypeUtils.ts
bmingles: sigh, yes

packages/utils/src/TypeUtils.ts
bmingles: Extract allows non-existent literal types but produces a never. Extends will actually complain if you pass an invalid type

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCiIbAHYBMDOAPACoA0cAqnKFnvnMQHxwC8lA3AFCe7ADGAGwCGUVEhRwAIgEtRfGNIjZWcAOTZoMABaq4AHzX4IAV226Dq4EPwxzagO7Abq7gHpXcPhAC2YYdOx8TnFUAEEVDFoCQhk5BSVyVSNTLRBVRjcPXAgnbFV4L19-bGDkVAAhCMwoIXkY2X547ETk7TTGIA

packages/utils/src/TypeUtils.ts
bmingles: fixed

packages/utils/src/TypeUtils.ts
mattrunyon: Just complains in different spots. Looks like w/ this `Extends` it complains at the type definition and not usages. With `Extract` complains at usages b/c it is `never`.

If you wanted best of both worlds you could use `type Extends<T, U extends T> = U extends T ? U : never`

I'm guessing this is aimed at union types mostly since interfaces can just use `extends` keyword in their definition

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCiIbAHYBMDOAPACoA0cAqnKFnvnMQHxwC8lA3AFCe7ADGAGwCGUVEhRwAIgEtRfGNIjZWcAOTZoMABaq4AHzX4IAV226Dq4EPwxzagO7Abq7gHpXcPhAC2YYdOx8TnFUAEEVDFoCQhk5BSVyVSNTLRBVRjcPXAgnbFV4L19-bGDkVAAhCMwoIXkY2X547ETk7TSMzgAzY2x5RWVvRAAxHr4ACiEALjhQ8gAjafKASmmANwhpXDgAb044fbghADoA3hAAeU6xgCJWrWulrgO4OZO8UEubu4euAF9OIA

bmingles: Good suggestion. Fixed

mattrunyon (approved):

PR #1398, opened 2023-07-05
feat: DH-14538: Export InputEditor and added options

### LINKED ISSUE #1397: DH-14538: Export InputEditor
As a developer, I want to be able to use the InputEditor component in Enterprise.

bmingles:
- Exporting InputEditor from @deephaven/iris-grid
- Added optional properties `className` and `placeholder`

resolves #1397

Supporting DH-14538

### DIFF
--- a/packages/iris-grid/src/sidebar/InputEditor.tsx
+++ b/packages/iris-grid/src/sidebar/InputEditor.tsx
@@ -4,6 +4,8 @@ import classNames from 'classnames';
 import './InputEditor.scss';
 
 interface InputEditorProps {
+  className?: string;
+  placeholder?: string;
   value: string;
   onContentChanged: (value?: string) => void;
   editorSettings: Partial<monaco.editor.IStandaloneEditorConstructionOptions>;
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/sidebar/InputEditor.tsx
```diff
@@ -175,9 +178,13 @@ export default class InputEditor extends Component<
           data-testid="custom-column-formula"
         />
         {isEditorEmpty && !value && (
-          <div className="editor-placeholder text-muted">Column Formula</div>
+          <div className="editor-placeholder text-muted">
+            {placeholder ?? 'Column Formula'}
```
mofojed, line 182: I don't think we should show anything for placeholder if specified, then `CustomColumnInput` should be passing in "Column Formula", instead of having it in this component. Within the context of `InputEditor`, knowledge of what is being input doesn't make sense.

mofojed (changes requested):

packages/iris-grid/src/sidebar/InputEditor.tsx
bmingles: Fixed

mofojed (approved):

PR #1400, opened 2023-07-06
fix: Use user permissions for iframes instead of query parameters

### LINKED ISSUE #1337: Features available for iframes are not the same as the ide by default
canCopy and canDownloadCsv options are false by default for iframes, as opposed to true by default for the ide. Ideally, there should be parity between the defaults between these use cases.

mofojed:
- Add UserUtils for fetching the User.
- Add ServerConfigBootstrap and UserBootstrap to bootstrap those objects into context.
- Remove query parameters for controlling user function in iframes; instead it is defined on the server.
- Fixes #1337.

Tested by setting the following in the dh-defaults.prop and verified the functionality matched with what I set:
```
diff --git a/props/configs/src/main/resources/dh-defaults.prop b/props/configs/src/main/resources/dh-defaults.prop
index 502ad119c4..65080d03db 100644
--- a/props/configs/src/main/resources/dh-defaults.prop
+++ b/props/configs/src/main/resources/dh-defaults.prop
@@ -55,7 +55,7 @@ authentication.anonymous.warn=true
 authentication.client.configuration.list=AuthHandlers
 
 # List of configuration properties to provide to authenticated clients, so they can interact with the server.
-client.configuration.list=java.version,deephaven.version,barrage.version,http.session.durationMs
+client.configuration.list=java.version,deephaven.version,barrage.version,http.session.durationMs,internal.webClient.appInit.canCopy,internal.webClient.appInit.canDownloadCsv
 # Version list to add to the configuration property list. Each `=`-delimited pair denotes a short name for a versioned
 # jar, and a class that is found in that jar. Any such keys will be made available to the client.configuration.list
 # as <key>.version.
@@ -64,3 +64,6 @@ client.version.list=deephaven=io.deephaven.engine.table.Table,barrage=io.deephav
 
 # Specifies additional setup to run on threads that can perform table operations with user code. Comma-separated list, instances must be of type io.deephaven.util.thread.ThreadInitializationFactory
 thread.initialization=io.deephaven.server.console.python.DebuggingInitializer
+
+# internal.webClient.appInit.canCopy=false
+internal.webClient.appInit.canDownloadCsv=false
\ No newline at end of file
```


packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export function getUserFromConfig(
+  serverConfig: Map<string, string>,
+  overrides: Partial<Omit<User, 'permissions'>> = {},
+  permissionsOverrides: Partial<UserPermissions> = {}
+): User {
+  function getValue(key: string): string | undefined {
+    return getAppInitValue(serverConfig, key);
+  }
+  function getBooleanValue(key: string, defaultValue = true): boolean {
```
mattrunyon, line 31: Maybe require the default value here? You provide it explicitly everywhere anyway

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (defaultValue) {
+      return !(getValue(key) === 'false');
+    }
+    return getValue(key) === 'true';
```
mattrunyon, line 35: Maybe we should warn about incorrect values here?

```ts
const value = getValue(key);
if (value === 'true') {
  return true;
}
if (value === 'false') {
  return false;
}
if (value !== '') {
  // Warn about unsupported config value and that we're falling back to default
}
return defaultValue;
```

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+  const name = getValue('name') ?? '';
+  const operateAs = getValue('operateAs') ?? name;
+  const groups = getValue('groups')?.split(',') ?? [];
+  const isACLEditor = getBooleanValue('isACLEditor', false);
+  const isSuperUser = getBooleanValue('isSuperUser', false);
+  const isQueryViewOnly = getBooleanValue('isQueryViewOnly', false);
+  const isNonInteractive = getBooleanValue('isNonInteractive', false);
```
mattrunyon, line 43: Wish we had named this `isInteractive`. Double negatives are annoying. "Not non-interactive". Oh well

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+      return !(getValue(key) === 'false');
+    }
+    return getValue(key) === 'true';
+  }
+  const name = getValue('name') ?? '';
+  const operateAs = getValue('operateAs') ?? name;
+  const groups = getValue('groups')?.split(',') ?? [];
+  const isACLEditor = getBooleanValue('isACLEditor', false);
+  const isSuperUser = getBooleanValue('isSuperUser', false);
+  const isQueryViewOnly = getBooleanValue('isQueryViewOnly', false);
```
mattrunyon, line 42: Why are these in community? Aren't they DHE or DnD only?

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+  const isNonInteractive = getBooleanValue('isNonInteractive', false);
+  const canCopy = getBooleanValue('canCopy', true);
+  const canDownloadCsv = getBooleanValue('canDownloadCsv', true);
+  const canCreateDashboard = getBooleanValue('canCreateDashboard', true);
+  const canCreateCodeStudio = getBooleanValue('canCreateCodeStudio', true);
+  const canCreateQueryMonitor = getBooleanValue('canCreateQueryMonitor', true);
+  const canUsePanels = getBooleanValue('canUsePanels', true);
```
mattrunyon, line 49: Same question about why in community

mattrunyon (changes requested):

packages/auth-plugins/src/UserUtils.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    serverConfig.set('internal.webClient.appInit.canUsePanels', 'false');
+    serverConfig.set('internal.webClient.appInit.canCreateDashboard', 'false');
+    serverConfig.set('internal.webClient.appInit.canCreateCodeStudio', 'false');
+    serverConfig.set(
+      'internal.webClient.appInit.canCreateQueryMonitor',
+      'false'
+    );
+    serverConfig.set('internal.webClient.appInit.canCopy', 'false');
+    serverConfig.set('internal.webClient.appInit.canDownloadCsv', 'false');
+    serverConfig.set('internal.webClient.appInit.canLogout', 'false');
```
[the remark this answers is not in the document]
mofojed: Ya, Copilot wrote this all for me ;)

packages/auth-plugins/src/UserUtils.ts
mofojed: Yea, they are. I didn't consider removing them, but I think we may as well.

Also looks like there is a bit of cleanup that could be done on the Enterprise side as well: https://github.com/deephaven-ent/iris/blob/ac1c2eb8c2c9be9114ddc58d22a9477104d6d26f/web/client-ui/src/permissions/PermissionsUtils.ts#L11
There's already an `EnterpriseUserPermissions` interface declared, so we're declaring it twice.

packages/auth-plugins/src/UserUtils.ts
mofojed: Removing it along with the other Enterprise props

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Retrieve a user object provided the server config and overrides
+ * @param serverConfig Server config map
+ * @param overrides Override values for the user
+ * @param permissionsOverrides Override specific permissions for the user
+ * @returns The user object
+ */
+export function getUserFromConfig(
+  serverConfig: Map<string, string>,
+  overrides: Partial<Omit<User, 'permissions'>> = {},
+  permissionsOverrides: Partial<UserPermissions> = {}
```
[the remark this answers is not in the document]
mofojed: Has to do with how plugins can override stuff. Previously plugins were only allowed to override the permissions. https://github.com/deephaven/js-plugin-template/#login-plugin
I guess it could be changed to use a deep partial here instead, just already had the two overrides split up.

packages/auth-plugins/src/UserUtils.ts
```diff
@@ -28,24 +31,24 @@ export function getUserFromConfig(
...
+    }
+    if (value !== undefined) {
+      log.warn(`Unexpected value for ${key}: ${value}`);
     }
-    return getValue(key) === 'true';
+    return defaultValue;
   }
   const name = getValue('name') ?? '';
   const operateAs = getValue('operateAs') ?? name;
   const groups = getValue('groups')?.split(',') ?? [];
```
mattrunyon: Are these part of DHC? Or will they be? I'm not sure if these are used in DnD

mofojed: Not used in DHC. I didn't want to make any breaking changes as part of this change.
The `User` interface is used in Enterprise.

mattrunyon: https://github.com/deephaven/web-client-ui/issues/1401

mattrunyon (approved):

PR #1403, opened 2023-07-10
fix: tree and rollup table default to non-sortable

### LINKED ISSUE #1402: Tree and rollup tables are not sortable
**Description**

After #1390 tree and rollup tables are not sortable.

The definition added to `IrisGridTableModel` should be in `IrisGridTableModelTemplate` because `IrisGridTreeTableModel` is using the default implementation which just always returns false.

We should also change the default to true probably since the engine should just fail the sort and tell the UI in those cases. According to [this comment](https://github.com/deephaven/deephaven-core/issues/1914#issuecomment-1028294907), the bug where the error was thrown instead of reported as a message to the UI is fixed, so should be fine to default to let the user try to sort (although the default implementation in `IrisGridModel` won't be hit in our uses)

**Steps to reproduce**

1. Create a tree or rollup table.
2. Try to sort a column that you would normally be able to sort

**Expected results**

3. Column sorts

**Actual results**

3. Column is not sortable in the UI

**Versions**

 - Engine Version: 0.26.0
 - Web UI Version: 0.44.0

ethanalvizo:
Closes #1402

### DIFF
--- a/packages/iris-grid/src/AdvancedFilterCreator.test.tsx
+++ b/packages/iris-grid/src/AdvancedFilterCreator.test.tsx
@@ -54,7 +54,7 @@ function makeAdvancedFilterCreatorWrapper(
       selectedValues: [],
     },
     model: irisGridTestUtils.makeModel(),
-    column: irisGridTestUtils.makeColumn(),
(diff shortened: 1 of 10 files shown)


mattrunyon:
I think something is off with your git history on this? A bunch of changes from the original PR that should already be in main.

You should only need the last commit on its own branch based on main

PR #1412, opened 2023-07-18
feat: Add support for Category plots with error bars

mofojed:
- Wasn't adding the error bars in this case
- Needed for https://github.com/deephaven/deephaven-core/issues/2156 to work
- Needs https://github.com/deephaven/deephaven-core/pull/4202
- Tested with the snippet in the ticket to create the plot:
```
from deephaven.plot import PlotStyle
from deephaven.column import int_col, string_col

source = new_table([
    string_col("Categories", ["A", "B", "C"]),
    int_col("Categories2", [1, 2, 3]),
    int_col("Values", [9, 10, 11]),
    int_col("LowerError", [8, 9, 10]),
    int_col("UpperError", [10, 11, 12])
])

v2_plot = Figure().plot_cat(series_name="Categories Plot v2", t=source, category="Categories2", y="Values", y_low="LowerError", y_high="UpperError").show()
```
- Also tested OHLC and existing Bar plots from the core/docs examples to ensure they looked the same

### DIFF
--- a/packages/chart/src/ChartUtils.ts
+++ b/packages/chart/src/ChartUtils.ts
@@ -196,7 +196,8 @@ class ChartUtils {
   static getPlotlyErrorBars(
     x: number[],
     xLow: number[],
-    xHigh: number[]
+    xHigh: number[],
+    theme = ChartTheme
   ): ErrorBar {
     const array = xHigh.map((value, i) => value - x[i]);
     const arrayminus = xLow.map((value, i) => x[i] - value);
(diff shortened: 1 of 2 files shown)


mattrunyon (approved):
Tested it w/ the example and looks to be working

@dsmmcken The error bar color is a little low contrast on our default bar colors at least. Not sure what your thoughts are, but I don't think it needs to be part of this change if it needs to be changed at all
newplot

PR #1417, opened 2023-07-21
build: Update to lerna 7.x

### LINKED ISSUE #1235: Add provenance to published npm packages
see: https://github.blog/changelog/2023-04-19-npm-provenance-public-beta/

Not sure if lerna publish supports the flag or not.

mofojed:
- Removed `useWorkspaces` flag (no longer used)
- Removed 'useNx' flag (NX seems to work in the latest)
- Add provenance flag to .npmrc
- Fixes #1235
- Tested `npm start`, `npm test`, and `npm run build` steps and ensured they all worked correctly

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,4 @@
 {
-  "packages": ["packages/*"],
   "version": "0.44.1",
   "command": {
     "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,5 @@
 {
+  "useNx": false,
   "version": "0.44.1",
   "command": {
     "publish": {
--- a/lerna.json
+++ b/lerna.json
@@ -1,7 +1,5 @@
 {
   "packages": ["packages/*"],
-  "useWorkspaces": true,
-  "useNx": false,
   "version": "0.44.1",
   "command": {
     "publish": {
--- a/.github/workflows/publish-alpha.yml
+++ b/.github/workflows/publish-alpha.yml
@@ -16,6 +16,7 @@ jobs:
     permissions:
       contents: read
       packages: write
+      id-token: write
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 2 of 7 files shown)


lerna.json
```diff
@@ -1,7 +1,5 @@
 {
   "packages": ["packages/*"],
```
mattrunyon, line 2: This should be removed. In the 7.0.0 changelog I noticed it says that the package manager workspace config will be used by default, so this field is now an override of the npm workspaces settings

mattrunyon (changes requested):

lerna.json
mofojed: I am not (too) ashamed to admit it took me ~15 minutes to figure out I just needed to remove the `./` from the `workspaces` field.

mattrunyon:
Have you tried publishing an alpha version from this branch? Should probably do that to ensure the provenance stuff works

mattrunyon:
There is also an issue w/ caching and build. I ran a build, changed `vite.config.ts`, then ran build again and it used the cached build. Probably something I didn't get right when I tried setting up nx the first time.

And another issue w/ nx is it causes `.env` files to load differently and doesn't play nicely with Vite. `npm start` doesn't load `.env.development` b/c nx loads `.env` first, so Vite does not override previously set env variables. The LogProxy is on in dev mode (should be off) and the base URL is relative and not `/ide/` like `.env.development` specifies

See https://github.com/deephaven/web-client-ui/pull/1011#issuecomment-1404361931 for previous issues

mattrunyon:
Need to add `provenance=true` to `.nvmrc` or `"publishConfig": { "provenance": true }` to `package.json` to enable it

mattrunyon (approved):
Looks like the action change will need to merge before we can validate this publishes properly, but looks like it should based on the failed publish saying it needs the `id-token` permission

PR #1424, opened 2023-07-28
fix: inconsistent drag on e2e tests for webkit

### LINKED ISSUE #1360: Playwright drag not working if operation is too quick on webkit
**Description**
In the E2E tests for table operations, the last step 'Drag' has the following code:
`await dragComponent(
      dragButton,
      panelAbove,
      dropIndicator,
      0,
      browser === 'webkit' ? 1000 : undefined
    );`

This is because the drag behaviour is inconsistent for webkit when generating the linux snapshots if the number for the "steps" param is any lower. The steps param defines how many intermediate mouse move events there while dragging to the destination

**Steps to reproduce**
1. Type the following command into your terminal:
`npm run e2e:headed  -- ./tests/table-operations.spec.ts:340`
2. Play through the test and the drag should work just fine with the 2nd column 'Drag' appearing before the 1st column 'Test'
3. Lower the steps from 1000 to 100 
4. Now regenerate the ci snapshots using the following command:
`npm run e2e:update-ci-snapshots'`

**Expected results**
image

**Actual results**
image

**Versions**

Engine Version: 0.25.0
Web UI Version: 0.40.4
Java Version: 17.0.7
Barrage Version: 0.5.0

ethanalvizo:
Closes #1360


mofojed (changes requested):
This needs to be rebased and cleaned up

PR #1426, opened 2023-07-28
Added containsIgnoreCase to FilterValue interface

bmingles:
Added containsIgnoreCase to FilterValue interface
#resolves 1425


mattrunyon (changes requested):
While you're here, there's also `matches` and `matchesIgnoreCase` in the JS AP that I saw. Add those. Same signatures as `contains`

mattrunyon (approved):

PR #1437, opened 2023-08-03
build(@deephaven/icons): Properly package icons and remove unnecessary files in dist

mattrunyon:
While updating packages consumed in the docs, I noticed our icons `package.json` was off. It specified it was a module type (ESM), but exported CJS as `.js`. This is incorrect as those will be interpreted as ESM by Node. Instead we should be exporting a `.cjs` file and a `.js` or `.mjs` as the main file.

Added an exports section to the `package.json` to indicate which file is for ESM and which is for CJS. We still need to distribute CJS for Jest tests.

Removed unnecessary files from the distribution. We don't use imports like `import vsTrash from '@deephaven/icons/vsTrash';` and the ESM tree shakes, so I removed the individual files exported. They also would have been imported from `/icons/dist` which is bad practice. Removed the SVG files from the files for the published package as well since they aren't used by anything in production.

BREAKING CHANGE: Any imports/aliasing to `@deephaven/icons/dist` should be removed and just read the package contents normally (e.g. DHE jest and vite configs for using community packages locally). See the changes to vite and jest configs in this change for how to update


mofojed (commented):
Publish an alpha and make sure Enterprise still works with the published alpha?

mattrunyon:
Tested with an alpha and enterprise builds/tests/looks correct.

I put this as a breaking change mostly as a message to update the DHE community packages paths in Vite and Jest (for using the `-community` scripts). Right now they import from the dist folder directly. I tested these locally w/ the updated configs (matching config in this PR) and they work too

mofojed (approved):

PR #1448, opened 2023-08-15
fix: Upgrade Monaco to ^0.41.0

### LINKED ISSUE #1191: Notebook mouse wheel and new LSP feature
WRT https://github.com/deephaven/web-client-ui/pull/1178 in the context of a **notebook**:

When I've got a hover pop-up and I use the mouse wheel to scroll in the pop-up, everything works as expected (the mouse wheel action is confined to the limits of the pop-up).

When I'm typing and get a pop-up and I use the mouse wheel to scroll in the pop-up, the mouse wheel action effects the position of my notebook (causing the pop-up to move as well). This is not what I expect to happen. (Is this the "signatureHelp" feature?)

### LINKED ISSUE #1445: Typing newline then space in InputEditor Component Crashes Chrome
**Description**

The <InputEditor/> component in use in the CustomColumn editor + the ACL editor crashes Chrome when typing a new line followed by typing a leading space on that line.

This was discovered in Enterprise (see DH-15438), but we suspect the fix will need to be made to Community.

**Steps to reproduce**
In an <InputEditor/> instance (currently in CustomColumns + Table ACL filter editro), type:
```
ab
```
then
```
a
b
```
then
```
a
 b
 ```
**Expected results**
The editor should add a leading space to the 2nd line

3. New page to load
-->

**Actual results**
Chrome crashes with an "Aw Snap" message

**Versions**

Version 1.20230512.060beta ()
Web Client (ef239cea)
Java Eclipse Adoptium 17.0.7

### PR DESCRIPTION
**Pull request by bmingles**
Fixes #1445, fixes #1191 Fixing Monaco bug by upgrading to latest ^0.41.0

Supports DH-15438

BREAKING CHANGE: Monaco will need to be upgraded to ^0.41.0 in Enterprise to ensure compatibility

**Tests Performed**

- Console Input
    - `Cmd+F` does nothing
    - Intellisense can be closed via `Esc`
- Log tab
    - `Esc` does not close find input
    - `Esc` does clear selection when focus is in the log content
- Code Editor
    - Verified that newline with leading space no longer crashes the browser tab
      ```
      a
       a
      ```
   - Wrote some Python code. Intellisense, syntax highlighting, and general typing experience seemed as expected
   - Execute full code + selected code successfully

### DIFF
--- a/packages/console/src/log/LogView.tsx
+++ b/packages/console/src/log/LogView.tsx
@@ -9,6 +9,7 @@ import ConsoleUtils from '../common/ConsoleUtils';
 import LogLevel from './LogLevel';
 import './LogView.scss';
 import LogLevelMenuItem from './LogLevelMenuItem';
+import { MonacoUtils } from '../monaco';
 
 interface LogViewProps {
   session: IdeSession;
@@ -226,36 +227,30 @@ class LogView extends PureComponent<LogViewProps, LogViewState> {
       wordWrap: 'on',
     });
 
-    // When find widget is open, escape key closes it.
-    // Instead, capture it and do nothing. Same for shift-escape.
-    this.editor.addCommand(monaco.KeyCode.Escape, () => undefined);
-    this.editor.addCommand(
-      // eslint-disable-next-line no-bitwise
-      monaco.KeyMod.Shift | monaco.KeyCode.Escape,
-      () => undefined
-    );
+    // Override default Monaco keybindings for `escape` and `shift-escape`
+    [
+      [monaco.KeyCode.Escape],
+      [monaco.KeyMod.Shift | monaco.KeyCode.Escape], // eslint-disable-line no-bitwise
+    ].forEach(keybindings => {
+      // Monaco default behavior is for escape key to close the find widget.
+      // Instead, capture it and do nothing. Same for shift-escape.
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      MonacoUtils.disableKeyBindings(this.editor!, keybindings);
 
-    // Restore regular escape to clear selection, when editorText has focus.
-    this.editor.addCommand(
-      monaco.KeyCode.Escape,
-      () => {
-        const position = this.editor?.getPosition();
-        assertNotNull(position);
-        this.editor?.setPosition(position);
-      },
-      'findWidgetVisible && editorTextFocus'
-    );
-
-    this.editor.addCommand(
-      // eslint-disable-next-line no-bitwise
-      monaco.KeyMod.Shift | monaco.KeyCode.Escape,
-      () => {
-        const position = this.editor?.getPosition();
-        assertNotNull(position);
-        this.editor?.setPosition(position);
-      },
-      'findWidgetVisible && editorTextFocus'
-    );
+      // Restore regular escape to clear selection, when editorText has focus.
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      this.editor!.addAction({
+        id: 'clear-selection-on-escape',
+        label: '',
+        keybindings: [monaco.KeyCode.Escape],
+        keybindingContext: 'findWidgetVisible && editorTextFocus',
+        run: () => {
+          const position = this.editor?.getPosition();
+          assertNotNull(position);
+          this.editor?.setPosition(position);
+        },
+      });
+    });
   }
 
   destroyMonaco(): void {
--- a/packages/console/src/log/LogView.tsx
+++ b/packages/console/src/log/LogView.tsx
@@ -232,14 +232,14 @@ class LogView extends PureComponent<LogViewProps, LogViewState> {
       [monaco.KeyCode.Escape],
       [monaco.KeyMod.Shift | monaco.KeyCode.Escape], // eslint-disable-line no-bitwise
     ].forEach(keybindings => {
+      assertNotNull(this.editor);
+
       // Monaco default behavior is for escape key to close the find widget.
       // Instead, capture it and do nothing. Same for shift-escape.
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      MonacoUtils.disableKeyBindings(this.editor!, keybindings);
+      MonacoUtils.disableKeyBindings(this.editor, keybindings);
 
       // Restore regular escape to clear selection, when editorText has focus.
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      this.editor!.addAction({
+      this.editor.addAction({
         id: 'clear-selection-on-escape',
         label: '',
         keybindings: [monaco.KeyCode.Escape],
--- a/packages/console/src/monaco/MonacoUtils.ts
+++ b/packages/console/src/monaco/MonacoUtils.ts
@@ -341,7 +341,7 @@ class MonacoUtils {
   }
 
   // Tracks whether removeConflictingKeybindings() has been called
-  private static conflictingKeybindingsRemoved = false;
+  static conflictingKeybindingsRemoved = false;
 
   /**
    * Remove any keybindings which are used for our own shortcuts.
--- a/packages/console/src/monaco/MonacoUtils.ts
+++ b/packages/console/src/monaco/MonacoUtils.ts
@@ -363,9 +363,10 @@ class MonacoUtils {
       }
     );
 
-    // Mac has a system shortcut tied to Cmd+H, so we can't override it
+    // Cmd+H is used to focus Community console history in Windows + Linux.
+    // An alternate shortcut is used for Mac, so no need to override it
+    // (See ConsoleShortcuts.ts)
     if (!MonacoUtils.isMacPlatform()) {
-      // Console focus history in Community (see ConsoleShortcuts.ts)
       monaco.editor.addKeybindingRule({
         /* eslint-disable-next-line no-bitwise */
         keybinding: monaco.KeyMod.CtrlCmd | monaco.KeyCode.KeyH,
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: packages/console/src/log/LogView.tsx [hunk 1]
```diff
@@ -226,36 +227,30 @@ class LogView extends PureComponent<LogViewProps, LogViewState> {
...
-    );
+    // Override default Monaco keybindings for `escape` and `shift-escape`
+    [
+      [monaco.KeyCode.Escape],
+      [monaco.KeyMod.Shift | monaco.KeyCode.Escape], // eslint-disable-line no-bitwise
+    ].forEach(keybindings => {
+      // Monaco default behavior is for escape key to close the find widget.
+      // Instead, capture it and do nothing. Same for shift-escape.
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      MonacoUtils.disableKeyBindings(this.editor!, keybindings);
```
mofojed, line 238: You could use an `assertNotNull(this.editor)` above instead of doing a non-null assertion; the advantage of that is you would just need to do it once above these statements and don't need to disable the warning in the multiple spots here.
NOt a big deal either way

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -339,91 +340,60 @@ class MonacoUtils {
...
       {
-        windows: 'ctrl+D',
-        mac: 'meta+D',
+        keybinding: monaco.KeyMod.CtrlCmd | monaco.KeyCode.KeyD,
+        command: null,
       },
+      // Console focus history in Community (see ConsoleShortcuts.ts)
       {
-        windows: 'ctrl+H',
+        keybinding: monaco.KeyMod.WinCtrl | monaco.KeyCode.KeyH,
```
mofojed, line 370: This isn't working correctly on Windows/Linux. You need to remove `monaco.KeyMod.WinCtrl | monaco.KeyCode.KeyH` on Mac, but `monaco.KeyMod.CtrlCmd | monaco.KeyCode.KeyH` on Windows/Linux. We use Ctrl+H on Windows and Ctrl+Option+H for this shortcut on mac:
```
    shortcut: [MODIFIER.CTRL, KEY.H],
    macShortcut: [MODIFIER.CTRL, MODIFIER.OPTION, KEY.H], // cmd+H is a system level shortcut on Mac, so use something else
```

With this change pressing Ctrl+H while in the console input brings up the Find/Replace widget:
image

Granted this keyboard shortcut is kind of janky already, since it only works if the Command History panel is visible, but may as well not make it worse.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -339,91 +340,60 @@ class MonacoUtils {
...
-   * Based on the following comment to remove a keybinding and let the event bubble up
-   * https://github.com/microsoft/monaco-editor/issues/287#issuecomment-331447475
-   * The issue for an API for this has apparently been open since 2016. Link below
-   * https://github.com/microsoft/monaco-editor/issues/102
-   * @param editor The editor to remove the keybinding from
-   * @param keybinding The key string to remove. E.g. 'ctrl+C' for copy on Windows
+   * Creates actions with a `noop` run function to disable specified keybindings.
+   * Note that this will swallow the events. To disable default keybindings in a
+   * way that allows events to propagate upward, see `removeConflictingKeybindings`
+   * @param editor Editor to add the actions to
```
mofojed, line 381:
```suggestion
   * @param editor Editor to disable keybindings for
```

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -339,91 +340,60 @@ class MonacoUtils {
...
+    // This function adds the keybindings to the global registry, so we only
+    // want to do it once
+    if (MonacoUtils.conflictingKeybindingsRemoved) {
+      return;
+    }
+
+    MonacoUtils.conflictingKeybindingsRemoved = true;
+
+    /* eslint-disable no-bitwise */
+    monaco.editor.addKeybindingRules([
```
mofojed, line 362: I was kind of confused about difference between this and `disableKeybindings`. May want to add a note in the docs stating it applies globally for all monaco editor instances.

packages/console/src/notebook/Editor.tsx
```diff
@@ -94,7 +94,11 @@ class Editor extends Component<EditorProps, Record<string, never>> {
       ...settings,
     };
     assertNotNull(this.container);
+
+    MonacoUtils.removeConflictingKeybindings();
```
mofojed, line 98: We could just call this from `MonacoUtils.init`, and then don't need to check in the function if it's been called more than once.

packages/iris-grid/src/sidebar/CustomColumnBuilder.test.tsx
```diff
@@ -100,7 +100,10 @@ test('Ignores deleted formulas on save', async () => {
   // There is an issue with populating the custom columns and then editing the existing column
   // RTL/monaco aren't playing nicely and it won't edit the existing text
   // This test instead creates the new text, saves, then removes it to test the same behavior
-  jest.useFakeTimers();
+  jest.useFakeTimers({
+    // Monaco makes a call to performance.mark(), so we need to leave it in-tact
```
mofojed, line 104:
```suggestion
    // Monaco makes a call to performance.mark(), so we need to leave it intact
```

mofojed (changes requested):
Just the command history shortcut no longer working.

packages/console/src/monaco/MonacoUtils.ts
bmingles: To clarify, I was under the impression that WinCtrl + KeyH wasn't a thing on Mac so no need to remove it? It didn't appear to be handled at all for the Mac case in the original code

mofojed (dismissed):

packages/console/src/monaco/MonacoUtils.ts
mattrunyon: WinCtrl is windows key (meta key) on windows/Linux and Control on Mac

CtrlCmd is control on windows/Linux and cmd on Mac

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -341,89 +344,57 @@ class MonacoUtils {
...
+   * Note that this will swallow the events. To disable default keybindings in a
+   * way that allows events to propagate upward, see `removeConflictingKeybindings`
+   * @param editor Editor to disable keybindings for
+   * @param keybindingsList List of keybinding tuples to disable
    */
-  static removeKeybinding(
+  static disableKeyBindings(
     editor: monaco.editor.IStandaloneCodeEditor,
-    keybinding: string | undefined
+    ...keybindingsList: number[][]
```
mattrunyon, line 387: Minor, but I'm not a big fan of rest args when the type is known like this. It won't matter in this case, but there is a negative impact on memory usage for calls w/ rest args compared to an array arg when there's a large number of arguments. I also find the hover tooltip less clear because 1 level of the array actually belongs to the rest param, so I never actually use a 2D array as a consumer

It's also harder to add another arg in the future if we needed it. Maybe we want to add a `global = false` param. Now everywhere using `disableKeyBindings` must be rewritten to support that

`disableKeyBindings(editor, [[1], [2], [3]])` != `disableKeyBindings(editor, [1], [2], [3])` (although they would do the same in this case since monaco would just register 1 action w/ 3 possible keybindings I believe)

mattrunyon (changes requested):

packages/console/src/monaco/MonacoUtils.ts
bmingles: Ah, I misinterpreted "WinCtrl" to be "ctrl on Windows"

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -341,89 +344,57 @@ class MonacoUtils {
...
-   * https://github.com/microsoft/monaco-editor/issues/287#issuecomment-331447475
-   * The issue for an API for this has apparently been open since 2016. Link below
-   * https://github.com/microsoft/monaco-editor/issues/102
-   * @param editor The editor to remove the keybinding from
-   * @param keybinding The key string to remove. E.g. 'ctrl+C' for copy on Windows
+   * Creates actions with a `noop` run function to disable specified keybindings.
+   * Note that this will swallow the events. To disable default keybindings in a
+   * way that allows events to propagate upward, see `removeConflictingKeybindings`
+   * @param editor Editor to disable keybindings for
+   * @param keybindingsList List of keybinding tuples to disable
```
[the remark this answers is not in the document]
bmingles: Good call. I've simplified to `number[]` arg

packages/console/src/monaco/MonacoUtils.ts
bmingles: @mattrunyon I've changed to a `number[]` arg no longer rest param.

Not sure I follow on the `global = false` param? How would that be consumed?

bmingles: Ah, I think I understand now you were just given an example of adding a new arg, not suggesting I add a global arg?

mattrunyon (changes requested):

packages/console/src/monaco/MonacoUtils.ts
mattrunyon: Yep just an example

mattrunyon (approved):

mattrunyon:
@dsmmcken This enables bracket colorization. Did you want to open a ticket to change the default bracket colors? The keys are [here](https://code.visualstudio.com/api/references/theme-color) under Bracket pair colorization

I also noticed the command history does not colorize the brackets, but does have colorization. Odd

packages/console/src/monaco/MonacoUtils.ts
mofojed: Naming stuff is hard :stuck_out_tongue:

mofojed (approved):

PR #1450, opened 2023-08-17
fix: Heap usage request throttling

### LINKED ISSUE #1439: UI becomes unresponsive after suspending computer for a while
**Description**

After leaving the Deephaven UI open while the computer is suspended for a while, UI is unresponsive and server logs are flooded with `No AuthenticationRequestHandler registered for type Bearer`.

**Steps to reproduce**
1. Open the Web UI in the browser
2. Suspend/Sleep your computer for ~15 minutes
3. Wake up computer and look at Web UI

**Expected results**
3. Web UI should be working, or should have a popup asking to re-authenticate

**Actual results**
3. Web UI is not working, and there are errors in the logs. If left suspended long enough, it will be completely unresponsive. You may need to wait a few minutes before the "Authentication failed" message appears, prompting you to refresh (assuming the UI hasn't locked up completely).

In the UI logs, there are a bunch of errors from the `HeapUsage` component:
```
[HeapUsage] Unable to get heap usage Error: Authentication details invalid
    at Object.onEnd (dh-internal.js:1:724077)
    at dh-internal.js:1:55831
    at dh-internal.js:1:42819
    at Array.forEach (<anonymous>)
    at e.rawOnEnd (dh-internal.js:1:42781)
    at e.onTransportEnd (dh-internal.js:1:42488)
    at Object.$onMessage (dh-core.js:20455:35)
    at MultiplexedWebsocketTransport$3methodref$onMessage$Type.handleEvent_2 [as handleEvent]
```
Right now this component just polls, we should probably be smarter about how we queue up the requests instead of just using `setInterval`, so we don't accidentally queue up a bunch of events; also should handle the error better.

On the server there are a bunch of logs repeated:
```
2023-08-04T14:40:51.529Z | qtp1871312485-127    |  INFO | i.d.s.s.SessionService    | No AuthenticationRequestHandler registered for type Bearer
2023-08-04T14:41:01.529Z | qtp1871312485-127    |  INFO | i.d.s.s.SessionService    | No AuthenticationRequestHandler registered for type Bearer
2023-08-04T14:41:07.072Z | qtp1871312485-127    |  INFO | i.d.s.s.SessionService    | No AuthenticationRequestHandler registered for type Bearer
2023-08-04T14:41:07.090Z | qtp1871312485-113    |  INFO | i.d.s.s.SessionService    | No AuthenticationRequestHandler registered for type Bearer
```

Not sure why that is getting triggered repeatedly, but takes a while for the `EVENT_RECONNECT_AUTH_FAILED` to fire.

**Versions**

Engine Version: 0.26.1
Web UI Version: 0.44.1
Java Version: 17.0.7
Barrage Version: 0.6.0

bmingles:
fixes #1439 - Heap usage code now guarantees that new requests won't be scheduled until all previous requests are complete. 

I tested by shutting my Mac for ~2 hours and returning. Timings looks something like this:
image

It seems that requests still happen occasionally while sleeping. On Wake up, there were a few ticks resulting in "Unable to get heap usage" console errors. Then a prompt showed up indicating credentials had expired at which point requests start succeeding again.


packages/react-hooks/src/useAsyncInterval.ts
```diff
@@ -0,0 +1,72 @@
...
+    trackingRef.current.count += 1;
+    trackingRef.current.started = now;
+
+    log.debug(
+      `tick #${trackingRef.current.count}.`,
+      elapsedSinceLastTick,
+      'ms elapsed since last tick.'
+    );
+
+    await callback();
```
mofojed, line 43: If the component unmounts during this callback, I think the timer will continue on - it returns from this callback, and there's nothing to stop it from calling `window.setTimeout` at the end.
Should set an `isCancelled` ref when unmounted, or you could overload the `setTimeoutRef` instead.

mofojed (changes requested):

packages/react-hooks/src/useAsyncInterval.ts
```diff
@@ -64,6 +69,7 @@ export function useAsyncInterval(
     setTimeoutRef.current = window.setTimeout(tick, targetIntervalMs);
 
     return () => {
+      isCancelledRef.current = true;
```
mofojed, line 72: This should be in it's own `useEffect`, with no dependencies so it's only called on `unmount`. This will "cancel" it anytime there's a change to `callback` or to `targetIntervalMs`, which would not be correct.

mofojed (changes requested):

packages/react-hooks/src/useAsyncInterval.ts
bmingles: good catch. I split out a separate hook

mofojed (approved):

mofojed:
Mentioning #28

PR #1453, opened 2023-08-18
fix: flaky e2e tests

ethanalvizo:
Fixes:

- code not pasting/typing into editor
- deleting tables between tests
- menus not closing between steps
- using dblclick over drag when possible

Does not fix:
- table disconnect observed on DHC version 0.27.1


tests/utils.ts
```diff
@@ -120,6 +120,10 @@ export async function pasteInMonaco(
       element.dispatchEvent(clipboardEvent);
     }, text);
   }
+
+  if (text.length) {
```
mofojed, line 124:
```suggestion
  if (text.length > 0) {
```

tests/table-operations.spec.ts
```diff
@@ -105,13 +105,21 @@ async function artificialWait(page: Page, tableNumber = 0) {
   await page.getByTestId('btn-page-close').first().click();
 }
 
+const tableName = generateVarName('t');
 test.beforeEach(async ({ page }) => {
   await page.goto('');
 
+  page.on('console', msg => {
+    if (msg.type() === 'error') console.log(`Error text: "${msg.text()}"`);
+  });
```
mofojed, line 114: Do we need to do this? It's already logged in the console and can get it from the browser logs

mofojed (changes requested):

tests/table-operations.spec.ts
ethanalvizo: No not necessarily I just found it convenient having it show up in your terminal as it goes through the test as well

mattrunyon: You don't get the browser logs from playwright unless you do this I think

mofojed (approved):

tests/utils.ts
```diff
@@ -154,6 +154,10 @@ export async function pasteInMonaco(
     // Sanity check the paste happened
     await expect(locator.locator('textarea')).not.toBeEmpty();
   }
+
+  if (text.length > 0) {
+    await expect(locator.locator('textarea')).not.toBeEmpty();
+  }
```
mofojed, line 160: This is duplicated, no need for this block
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #1456, opened 2023-08-22
feat: Consolidate and normalize plugin types

### LINKED ISSUE #1454: Normalize and centralize plugin types and configs
Change plugins to expect a default exported object (or function if there's anything we think should be available at runtime to the config) containing things like `component`, `supportedTypes`, etc. Add TS types for the different plugin config types

mattrunyon:
Fixes #1454. Needs #1451

Still some cleanup that could be done around `DashboardPlugin`. I think instead of static metadata on the component (`displayName` is ok since it's part of React) those items should probably be part of the config object.


packages/plugin/src/PluginTypes.ts
```diff
@@ -0,0 +1,88 @@
+export const PluginType = Object.freeze({
+  AUTH_PLUGIN: 'AuthPlugin',
+  DASHBOARD_PLUGIN: 'DashboardPlugin',
+  TABLE_PLUGIN: 'TablePlugin',
+});
+
+export type LegacyDashboardPlugin = { DashboardPlugin: React.ComponentType };
+
+export type LegacyAuthPlugin = {
```
mofojed, line 9: Add JSDocs and `@deprecated` to the legacy types.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -639,13 +645,11 @@ export class AppMainContainer extends Component<
     const pluginModule = plugins.get(pluginName);
     if (
       pluginModule != null &&
-      (pluginModule as { TablePlugin: ReactElement }).TablePlugin != null
+      (isTablePlugin(pluginModule) || 'TablePlugin' in pluginModule)
```
mofojed, line 648: You could rewrite this if block to be a little bit clearer, I think:
```
if (pluginModule != null) {
  if (isTablePlugin(pluginModule)) {
    return pluginModule.component;
  }
  if ('TablePlugin' in pluginModule) {
    return pluginModule.TablePlugin as unknown as TablePlugin;
  }
}
```

mofojed (changes requested):
Will need to update template: https://github.com/deephaven/deephaven-js-plugin-template/pull/1

mattrunyon:
I'd like to pull out DashboardPlugin too, but the dependency there on PanelManager is tricky w/ redux and that change feels like it would warrant its own PR

packages/plugin/src/PluginTypes.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @deprecated Use DashboardPlugin instead
+ */
+export type LegacyDashboardPlugin = { DashboardPlugin: React.ComponentType };
+
+/**
+ * @deprecated Use AuthPlugin instead
+ */
+export type LegacyAuthPlugin = {
```
mofojed, line 17: This isn't quite right - right now filtering isn't picking up the keycloak auth plugin.
I think it should be something like:
```
export type LegacyAuthPlugin = {
  AuthPlugin: {
    Component: ...,
    isAvailable: ...
  }
};
```

packages/app-utils/src/plugins/PluginUtils.tsx
```diff
@@ -115,7 +82,10 @@ export async function loadModulePlugins(
       const module = pluginModules[i];
       const { name } = manifest.plugins[i];
       if (module.status === 'fulfilled') {
-        pluginMap.set(name, module.value);
+        pluginMap.set(
+          name,
+          'default' in module.value ? module.value.default : module.value
```
mofojed, line 87: Would prefer wrapping this in a type predicate, e.g. `isLegacyPlugin`

packages/app-utils/src/plugins/PluginUtils.tsx
```diff
@@ -147,28 +117,45 @@ export function getAuthHandlers(
...
-    [...pluginMap.entries()].filter(
-      ([, plugin]: [string, { AuthPlugin?: AuthPlugin }]) =>
-        isAuthPlugin(plugin.AuthPlugin)
-    ) as [string, { AuthPlugin: AuthPlugin }][]
-  ).map(([name, plugin]) => [name, plugin.AuthPlugin]) as [
-    string,
-    AuthPlugin,
-  ][];
+  const legacyAuthPlugins: AuthPlugin[] = (
+    [...pluginMap.entries()].filter(([, plugin]) => 'Component' in plugin) as [
```
mofojed, line 126: This is incorrect. See my comment on `LegacyAuthPlugin` below, but you're filtering out the `auth-keycloak` plugin.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -632,20 +638,18 @@ export class AppMainContainer extends Component<
...
-      return (
-        pluginModule as {
-          TablePlugin: TablePlugin;
-        }
-      ).TablePlugin;
+    if (pluginModule != null) {
+      if (isTablePlugin(pluginModule)) {
+        return pluginModule.component;
+      }
+      if ('TablePlugin' in pluginModule) {
```
mofojed, line 650: Type predicate `isLegacyTablePlugin`

packages/plugin/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "engines": {
+    "node": ">=16"
+  },
+  "scripts": {
+    "build": "cross-env NODE_ENV=production run-p build:*",
+    "build:babel": "babel ./src --out-dir ./dist --extensions \".ts,.tsx,.js,.jsx\" --source-maps --root-mode upward"
+  },
+  "devDependencies": {},
+  "dependencies": {},
+  "peerDependencies": {},
```
mofojed, line 26: Looks like you're missing some dependencies here? This package definitely references `@deephaven/iris-grid`, `@deephaven/components`, `@deephaven/jsapi-types` from `TablePlugin`.
That's one thing kind of annoying about this - creating an `AuthPlugin`, you don't care about `@deephaven/iris-grid` at all, but you need to import this package which will have a dependency on that.

packages/plugin/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "@deephaven/plugin",
+  "version": "0.46.0",
+  "description": "Deephaven JS Plugin Core",
+  "author": "Deephaven Data Labs LLC",
+  "license": "Apache-2.0",
+  "type": "module",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/deephaven/web-client-ui.git",
+    "directory": "packages/app-utils"
```
mofojed, line 11:
```suggestion
    "directory": "packages/plugin"
```

packages/plugin/src/PluginTypes.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+/**
+ * @deprecated Use TablePlugin instead
+ */
+export type LegacyTablePlugin = {
+  TablePlugin: TablePluginComponent;
+};
+
+export type LegacyPlugin =
```
mofojed, line 29: Add `@deprecated` tag to `LegacyPlugin`

mofojed (changes requested):

packages/plugin/package.json
mattrunyon: Ya they're technically dev dependencies since they aren't needed to run, but the types are required for someone else to install the package and get type support. So that would make them regular dependencies

I can't think of a better way that doesn't require you to install iris-grid or components. Unless we want to duplicate the types across both or pull them into a separate, shared types package which doesn't really make sense for these types IMO.

If we want to consolidate the definitions, I think it's just something we have to accept as a trade-off. We could make packages for each plugin type, but I'm not really a big fan of adding more packages here just to save a few MB of disk space for a development environment

packages/plugin/src/PluginTypes.ts
mattrunyon: It turns out that the auth keycloak plugin is different from the core auth plugins. The core auth plugins match the type I have. Keycloak auth plugin matches the type you proposed

mofojed (approved):

PR #1458, opened 2023-08-24
ci: Run e2e in parallel

mattrunyon:


mofojed:
Side note: we could look at changing the name from "Build and End-to-end Tests" to just "e2e Tests" or something. Seems like the build checks that show up are a bit redundant:
image

mattrunyon:
Doesn't seem to be any faster (in fact it might be a bit slower). Also since the e2e-tests have a different job name in this configuration, they aren't required checks.

I'll close for now. We can revisit this if we hit a point where using a matrix for testing results in a good speed gain

PR #1463, opened 2023-08-27
test: Add firefox support for playwright pasteInMonaco

mattrunyon:
Doesn't seem to be a single solution that works in all 3 browsers in headless mode, but between 2 solutions we can cover pasting in all 3 browsers

### DIFF
--- a/tests/utils.ts
+++ b/tests/utils.ts
@@ -1,4 +1,5 @@
 import { Locator, expect, Page } from '@playwright/test';
+import os from 'node:os';
 import shortid from 'shortid';
 
 export enum TableTypes {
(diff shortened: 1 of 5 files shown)


mofojed (approved):

tests/utils.ts
```diff
@@ -99,18 +100,44 @@ export async function typeInMonaco(
...
+    return tempInput.id;
+  }, text);
+
+  // Copy the contents of the textarea which was selected above
+  await page.keyboard.press(`${modifier}+C`);
+
+  // Remove the textarea
+  await page.evaluate(id => {
+    document.getElementById(id)?.remove();
+  }, inputId);
```
mofojed, line 133: Is this whole chunk only needed for `firefox` ?

Issue #1472, opened 2023-08-30
Quick filter font doesn't match in Chrome


mofojed:
**Description**

The font in the text input field when editing a quick filter does not match the font/text drawn on the canvas, and it looks like the input shifts. This seems to only be reproducible on Chrome; Firefox looks correct (see attachments).

**Steps to reproduce**

1. On a table, open the Quick Filter and enter the number '37'
2. Click off the quick filter, and then click into the quick filter

**Expected results**
2. Text should line up exactly whether the field is in focus or not.

**Actual results**
2. Text shifts and the spacing is different. See videos:

**Additional details and attachments**
On Chrome you can see the text shift quite a bit, the spacing is different:
![font-issue-chrome](https://github.com/deephaven/web-client-ui/assets/4505624/0f940ce6-848c-4c60-bd5e-580f4ea5000b)

On Firefox you can see the font is the same (though it does look like it shifts up/down about a half pixel):
![font-issue-firefox](https://github.com/deephaven/web-client-ui/assets/4505624/dba430bb-dbb8-4802-813f-f1fda7f3fb62)

**Versions**
Engine Version: 0.28.0
Web UI Version: 0.46.0
Java Version: 11.0.15
Barrage Version: 0.6.0
Chrome Version:  114.0.5735.198 (arm64)
Firefox Version: 116.0.3

Note: This is also preventing Playwright from being updated to 1.37.1, as this issue is present there.

PR #1475, opened 2023-08-31
fix: quick filter focus text doesn't match canvas text

dsmmcken:
resolves: #1472 

looks like tnum wasn't being inherited.


mofojed (changes requested):
Weird that this is only a problem with updated Chrome/Playwright?

However, with this fix I'm now  seeing strange behaviour in Firefox:
font-issue-review-firefox
Chrome still shifts by a couple pixels, and can fix that by changing the top padding in `.iris-grid-input-field` to `5px`...
font-issue-review-chrome

dsmmcken:
> Chrome still shifts by a couple pixels, and can fix that by changing the top padding in .iris-grid-input-field to 5px...

That massively mis-aligns it for me in chrome/win the opposite direction.

dsmmcken:
I am guessing we are running up against platform limitations. I adjusted so it's right in chrome/firefox on windows, but I don't think we can get it perfect against the matrix of OS and browsers. Does it look reasonable to you on linux or is it way off still?

mofojed (commented):
Looks perfect for me on Ubuntu in both Firefox and Chrome. Mac it shifts up/down by a pixel or two, but fonts look the same.

I think just rebase off the latest and update snapshots then this is good to go. Thanks!

mofojed (approved):

PR #1476, opened 2023-08-31
feat: bindAllMethods util function

### LINKED ISSUE #1474: Bind all methods util
As a developer, I'd like to be able to bind all methods of a class to itself. This will be less error prone than explicitly binding methods which can get out of sync as new methods are added to classes.

bmingles:
Util functions for binding all methods of an object to itself.
resolves #1474

### DIFF
--- a/packages/utils/src/ClassUtils.ts
+++ b/packages/utils/src/ClassUtils.ts
@@ -1,0 +1,47 @@
+export type MethodName<T> = {
+  [K in keyof T]: T[K] extends (...args: unknown[]) => unknown ? K : never;
+}[keyof T];
+
+/**
+ * Bind all methods in the prototype chain to the instance (excluding those
+ * defined on Object.prototype)
+ * @param instance The instance to bind methods to
+ */
+export function bindAllMethods<T>(instance: T) {
+  const methodNames = getAllMethodNames(instance);
+
+  methodNames.forEach(methodName => {
+    // eslint-disable-next-line no-param-reassign
+    instance[methodName] = (
+      instance[methodName] as (...args: unknown[]) => unknown
+    ).bind(instance) as T[typeof methodName];
+  });
+}
+
+/**
+ * Get all class method names. This will return names of all methods defined in
+ * an object's prototype chain exlcluding those defined on Object.prototype.
+ * @param instance
+ */
+export function getAllMethodNames<T>(instance: T): MethodName<T>[] {
+  const methodNames = new Set<MethodName<T>>();
+
+  let current = instance;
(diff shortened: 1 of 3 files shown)


packages/utils/src/ClassUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param instance The instance to bind methods to
+ */
+export function bindAllMethods<T>(instance: T) {
+  const methodNames = getAllMethodNames(instance);
+
+  methodNames.forEach(methodName => {
+    // eslint-disable-next-line no-param-reassign
+    instance[methodName] = (
+      instance[methodName] as (...args: unknown[]) => unknown
+    ).bind(instance) as T[typeof methodName];
```
mattrunyon, line 17: Since you have to cast here anyway, I don't think the generic type is worth the added complexity

packages/utils/src/ClassUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+ * an object's prototype chain exlcluding those defined on Object.prototype.
+ * @param instance
+ */
+export function getAllMethodNames<T>(instance: T): MethodName<T>[] {
+  const methodNames = new Set<MethodName<T>>();
+
+  let current = instance;
+
+  // Traverse the prototype chain until we get to Object.prototype
+  while (current != null && current !== Object.prototype) {
```
mattrunyon, line 32: Might not want to iterate all the way back through the prototype chain? Not sure if this could cause anything to break in a React class component if it were used in one

packages/utils/src/ClassUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+        typeof current[name as keyof typeof current] === 'function'
+      ) {
+        methodNames.add(name as MethodName<T>);
+      }
+    }
+
+    current = Object.getPrototypeOf(current);
+  }
+
+  return [...methodNames.keys()].sort();
```
mattrunyon, line 46: Why sort?

mattrunyon (changes requested):

packages/utils/src/ClassUtils.ts
bmingles: Yes, I had that thought as well. I could see parameterizing it with a flag or something

packages/utils/src/ClassUtils.ts
bmingles: Was easier for testing, but I could just sort in the test I guess

mattrunyon (approved):

PR #1481, opened 2023-09-05
fix: Console History Not Scrolling to Bottom (DH-14062)

nbauernfeind:
I discovered DH-14062 independently and found that this method occasionally was prematurely returning when the view should stay stuck to the bottom of the contents.

Sometimes `pane.scrollTop` is sometimes 0.5 less than the expected value when tailing the content. This `Math.round` bumps it up. I only saw `xxxxx.0` and `xxxxx.5` values during investigation, but I did not do any further research on the origin of the value or the rounding error.

Here is a sample groovy snippet that reproduces 100% of the time:
```groovy
import io.deephaven.internal.log.LoggerFactory
log = LoggerFactory.getLogger(Object.class
log.error().append(new RuntimeException()).endl()
t = io.deephaven.engine.util.TableTools.timeTable("PT1s")
```

### DIFF
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -614,7 +614,10 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
   scrollConsoleHistoryToBottom(force = false): void {
     const pane = this.consoleHistoryScrollPane.current;
     assertNotNull(pane);
-    if (!force && pane.scrollTop < pane.scrollHeight - pane.offsetHeight) {
+    if (
+      !force &&
+      Math.round(pane.scrollTop) < pane.scrollHeight - pane.offsetHeight
+    ) {
       return;
     }
 
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -616,7 +616,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
     assertNotNull(pane);
     if (
       !force &&
-      Math.round(pane.scrollTop) < pane.scrollHeight - pane.offsetHeight
+      Math.abs(pane.scrollHeight - pane.clientHeight - pane.scrollTop) >= 1
     ) {
       return;
     }


packages/console/src/Console.tsx
```diff
@@ -614,7 +614,10 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
   scrollConsoleHistoryToBottom(force = false): void {
     const pane = this.consoleHistoryScrollPane.current;
     assertNotNull(pane);
-    if (!force && pane.scrollTop < pane.scrollHeight - pane.offsetHeight) {
+    if (
+      !force &&
+      Math.round(pane.scrollTop) < pane.scrollHeight - pane.offsetHeight
```
mofojed, line 619: I think use `Math.ceil` instead to be safe:
```suggestion
      Math.ceil(pane.scrollTop) < pane.scrollHeight - pane.offsetHeight
```
Although - I'm unable to consistently reproduce with the snippet you provided, as I believe it is dependent on the size of the console (which is dependent on browser size etc). I'm not certain this is the "correct" fix. I think instead we should be checking the size within the `requestAnimationFrame` ... as I'm fairly certain the issue is that we're not checking the updated scrollHeight at this point.

mofojed (changes requested):

packages/console/src/Console.tsx
mattrunyon: MDN has an example on checking an element is totally scrolled [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled)

nbauernfeind: Another reproducer is to simply print something and sleep a little in a loop:
```groovy
for (int ii = 0; ii < 100; ++ii) {
    println("test line " + ii)
    Thread.sleep(20)
}
```

The fix I originally gave Don was:
```typescript
pane.scrollTop + 1 < pane.scrollHeight - pane.offsetHeight
```
(noting the `+1`). 

Interestingly the MDN article states:
> scrollTop is a non-rounded number, while scrollHeight and clientHeight are rounded — so the only way to determine if the scroll area is scrolled to the bottom is by seeing if the scroll amount is close enough to some threshold (in this example 1):
> ```
> Math.abs(element.scrollHeight - element.clientHeight - element.scrollTop) < 1;
> ```
it sounds like my original suggestion is the official MDN solution. Should I swap to that instead?

mofojed (approved):

PR #1484, opened 2023-09-05
fix: Change dynamic import to string

mofojed:
- Remove a warning when building with webpack: https://stackoverflow.com/a/73359606


mattrunyon (changes requested):
Add a comment to the code so it's clear the string template of a string is on purpose

mattrunyon (approved):

Issue #1469, opened 2023-08-28
Move common ACL Editor code from Enterprise to Community


bmingles:
The web based ACL Editor was largely coded in the Enterprise codebase. We should move any common code we'd like to share into Community.

Supports Enterprise DH-15527

PR #1490, opened 2023-09-06
chore: DH-15527 Move ACL Editor Code to Community

bmingles:
Moved ACL Editor util code from Enterprise to Community. This should not introduce any functional changes.

Supports #1469 and DH-15527


packages/utils/package.json
```diff
@@ -27,5 +27,8 @@
   "sideEffects": false,
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "@adobe/react-spectrum": "^3.29.0"
```
mofojed, line 32: This seems like a pretty large dependency to add to `@deephaven/utils` which currently has no other dependencies. There's basically two things added to utils that require this:
1. `SpectrumUtils`: That's only used within `@deephaven/components` and `SelectionUtils`, which is also used in `react-hooks`. Should probably move it in there.
2. `WindowedListData`: Used within jsapi-components, jsapi-utils, react-hooks. jsapi-components depends on jsapi-utils which depends on react-hooks, so move `WindowedListData`, so I think `react-hooks` would be a more appropriate place for that.

`@deephaven/utils` is really just for vanilla utility methods, and shouldn't have any large dependencies.

mofojed (changes requested):
Only issue I think is the dependency of `utils` on `@adobe/react-spectrum`. That dependency should be lifted up to `react-hooks` and out of `utils`.

packages/utils/package.json
bmingles: Spectrum dependency removed from @deephaven/utils

packages/jsapi-utils/src/ViewportDataUtils.ts
```diff
@@ -2,7 +2,8 @@ import type { Key } from 'react';
 import clamp from 'lodash.clamp';
 import type { Column, Row, Table, TreeTable } from '@deephaven/jsapi-types';
 import Log from '@deephaven/log';
-import { KeyedItem, WindowedListData } from '@deephaven/utils';
+import type { WindowedListData } from '@deephaven/react-hooks';
```
mofojed, line 5: jsapi-utils depending on react-hooks feels wrong as well. I think moving the dependencies here to jsapi-components instead (those are where React components that use the jsapi live).

mofojed (changes requested):

packages/jsapi-utils/src/ViewportDataUtils.ts
bmingles: @mofojed can you elaborate on this one? jsapi-components depends on jsapi-utils. Which dependencies were you thinking of moving to jsapi-components?

bmingles: I was able to remove the react-hooks dependency by reconfiguring some types

mofojed (approved):

PR #1491, opened 2023-09-06
feat: adds copy file support to file explorer and fixes rename bug

### LINKED ISSUE #185: File Explorer Rename Robustness
**Description**

When renaming files from the file explorer there are a few issues.

As shown in the video:
1. Cannot change capitalization of new file.
2. Once renamed, cannot rename a second time (to achieve intended capitalization).
3. After attempting to rename a second time the right-click context menu is no longer bound to the UI.

**Steps to reproduce**
1. Create a new file.
2. Save with "unintended" name.
3. Try to change capitalization. (Reproduces the first issue.)
4. Compromise by renaming to a temporary name.
5. Attempt to rename a second time; but the file is not renamable. (Reproduces the second issue.)
6. Right click a third time and the context menu has become unbound. (Reproduces the third issue.)

**Expected results**

I expect to be able to change capitalization of a file. When I can't make up my mind, I expect to be able to rename a file more than once. I expect that renaming a file does not break the context menu.

**Actual results**

Each of the three desired actions fail; and the obvious work arounds don't work. (Hard refreshing works.)

https://user-images.githubusercontent.com/1030443/132140365-900d889a-8447-47dc-b594-5764d355cf3e.mov

### LINKED ISSUE #1375: Rename File Explorer "Copy" action to "Copy file"
In the file explorer, if you right click on a notebook, there is a "Copy" button. That is actually "Make a Copy" or just  "Make Copy". I expected "Copy" to then let me paste it somewhere

Google Drive has a "Make a Copy" button which has the same behavior as ours. Creates a copy and names it w/ "Copy of" preceding the file name. We append "-1" to the file name, so similar behavior

### LINKED ISSUE #1488: Add "Copy File" action to File Explorer
As a user, I would like a way to create a copy of a file inside the file explorer. Currently, right clicking a file in the file explorer only gives the "Delete" and "Rename" actions. There is no way to make a duplicate of a file in the file explorer outside of copy and pasting the contents of the file itself. I would like a "Copy File" option to be added so that files can be duplicated inside the file explorer by selecting the action from the right click menu.

dsmmcken:
Fixes #185, Fixes #1375, Fixes #1488 and other issues with File Explorer

Fixes issue where you couldn't rename a renamed item, and then can't use context menu at all on file explorer.

Fixes issue where triggering a tooltip while renaming would re-render and exit edit mode.

Wires up copy, new file, new folder in context menu.

Renames copy -> copy file

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
@@ -7,6 +7,7 @@ import FileExplorer, {
   FileStorageItem,
   FileUtils,
   NewItemModal,
+  isDirectory,
 } from '@deephaven/file-explorer';
 import React, { ReactNode } from 'react';
 import { connect, ConnectedProps } from 'react-redux';
@@ -94,6 +95,7 @@ export class FileExplorerPanel extends React.Component<
     super(props);
 
     this.handleFileSelect = this.handleFileSelect.bind(this);
+    this.handleCopyFile = this.handleCopyFile.bind(this);
     this.handleCreateFile = this.handleCreateFile.bind(this);
     this.handleCreateDirectory = this.handleCreateDirectory.bind(this);
     this.handleCreateDirectoryCancel =
@@ -139,7 +141,7 @@ export class FileExplorerPanel extends React.Component<
     );
   }
 
-  handleCreateDirectory(path?: string): void {
+  handleCreateDirectory(): void {
     this.setState({ showCreateFolder: true });
   }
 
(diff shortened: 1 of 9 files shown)


packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
```diff
@@ -157,6 +159,31 @@ export class FileExplorerPanel extends React.Component<
...
+        return false;
+      }
+    };
+    let newNameExists = await checkNewName();
+    while (newNameExists) {
+      newName = FileUtils.getCopyFileName(newName);
+      // await in loop is fine here, this isn't a parallel task
+      // eslint-disable-next-line no-await-in-loop
+      newNameExists = await checkNewName();
+    }
```
mofojed, line 183: You can simplify this a bit:
```suggestion    
    // await in loop is fine here, this isn't a parallel task
    // eslint-disable-next-line no-await-in-loop
    while (await checkNewName()) {
      newName = FileUtils.getCopyFileName(newName);
    }
```

mofojed (changes requested):
Still a case that fails:
- Right-click a file and select Copy File
- Right-click the new copy and select Delete

It'll try and delete the old file.

Or another case:
- Right-click file, click Delete
- Right-click the now focused file, click Delete

I think when the list updates, we need to update the `focusedFilePath` as well, since in the above cases the selection change isn't firing again.

dsmmcken:
> Still a case that fails:
> 
> * Right-click a file and select Copy File
> * Right-click the new copy and select Delete
> 
> It'll try and delete the old file.
> 
> Or another case:
> 
> * Right-click file, click Delete
> * Right-click the now focused file, click Delete
> 
> I think when the list updates, we need to update the `focusedFilePath` as well, since in the above cases the selection change isn't firing again.

I was firing the selection change again when the viewport updated, but re-using the handleSelectionChange wasn't re-fetching the items because the selected range didn't change. I added a force param, but open to better ways of doing it.

mofojed (approved):
This is fine... I think technically we could just use memoization to do the selected items, but no need to change all that and possibly introduce other bugs.

PR #1495, opened 2023-09-08
fix: Hide "Append Command" button when viewing partition aware table in iframe UI

### LINKED ISSUE #1414: Don't show "Append Command" when `onPartitionAppend` prop is not provided
**Description**

When the `onPartitionAppend` prop is not provided to `IrisGrid`, do not show the `Append Command` in the Partition selector, as it won't do anything.

**Steps to reproduce**
1. Open http://localhost:10000 and create a partition aware table:
```python
from deephaven.column import int_col, double_col, string_col
grades1 = new_table([
    string_col("Name", ["Ashley", "Jeff", "Rita", "Zach"]),
    int_col("Test1", [92, 78, 87, 74]),
    int_col("Test2", [94, 88, 81, 70]),
    int_col("Average", [93, 83, 84, 72]),
    double_col("GPA", [3.9, 2.9, 3.0, 1.8])
])

grades2 = new_table([
    string_col("Name", ["Jose", "Martha", "Mary", "Richard"]),
    int_col("Test1", [67, 92, 87, 54]),
    int_col("Test2", [97, 99, 92, 63]),
    int_col("Average", [82, 96, 93, 59]),
    double_col("GPA", [4.0, 3.2, 3.6, 2.7])
])

write(grades1, "grades/school=elementary/part1.parquet")
write(grades2, "grades/school=middle/part2.parquet")

grades = read("grades")
```
This uses the format of [Nested Parquet Tables](https://deephaven.io/core/docs/conceptual/parquet-formats/#nested-parquet-files) to create a partition aware table.
2. Open the table in the iframe UI: http://localhost:10000/iframe/table?name=grades

**Expected Results**
2. "Append Command" button should not be there, as there is no console session to append the command to

**Actual Results**
2. "Append Command" button is there, and clicking it does nothing

GcWan:
* Currently, viewing a partition aware table in the iframe UI will show the "Append Command" button even though there is no console session to append the command to.
* Changed the partition selector to only render the "Append Command" button when no `onPartitionAppend` prop is provided to IrisGrid. The "Append Command" button should now be available in the IDE but not the iframe UI.

* Resolves #1414


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4421,7 +4422,12 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
                   partition={partition}
                   onChange={this.handlePartitionChange}
                   onFetchAll={this.handlePartitionFetchAll}
-                  onAppend={this.handlePartitionAppend}
+                  onAppend={
+                    onPartitionAppend !==
```
mofojed, line 4426: Instead of this check, I'd remove the default prop and just check that `onPartitionAppend` is undefined.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -205,13 +206,15 @@ class IrisGridPartitionSelector<T> extends Component<
...
           Ignore &amp; Fetch All
         </button>
-        <button
-          type="button"
-          className="btn btn-outline-primary btn-append"
-          onClick={this.handleAppendClick}
-        >
-          Append Command
-        </button>
+        {onAppend !== IrisGridPartitionSelector.defaultProps.onAppend && (
```
mofojed, line 209: Just check that it is undefined and remove the default.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
GcWan: I'm actually getting compiler warnings now since I can't pass in undefined into onAppend due to incompatible types. Should I resolve this by changing the type of onAppend in the props interface or by passing in something like `() => undefined` instead? Alternatively, is there a third solution that is better than both of these?

mofojed: Change the prop in the interface to be optional, e.g.:
```
  onPartitionAppend?: (partitionColumn: Column, value: string) => void;
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2336,7 +2335,9 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     if (partitionColumn == null) {
       return;
     }
-    onPartitionAppend(partitionColumn, value);
+    if (onPartitionAppend) {
+      onPartitionAppend(partitionColumn, value);
+    }
```
mofojed, line 2340: You can just write this with optional chaining:
```suggestion
    onPartitionAppend?.(partitionColumn, value);
```

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -73,7 +72,9 @@ class IrisGridPartitionSelector<T> extends Component<
 
     const { onAppend } = this.props;
     const { partition } = this.state;
-    onAppend(partition);
+    if (onAppend) {
+      onAppend(partition);
+    }
```
mofojed, line 77:
```suggestion
    onAppend?.(partition);
```

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -205,13 +207,15 @@ class IrisGridPartitionSelector<T> extends Component<
...
           Ignore &amp; Fetch All
         </button>
-        <button
-          type="button"
-          className="btn btn-outline-primary btn-append"
-          onClick={this.handleAppendClick}
-        >
-          Append Command
-        </button>
+        {onAppend === undefined || (
```
mofojed, line 210: I'd prefer to write this as the truthy case and use `&&` instead of the falsey case and use `||`, just for consistency with what we do elsewhere in the app, e.g:
```suggestion
        {onAppend !== undefined && (
```

mofojed (changes requested):

mofojed (approved):

Issue #1497, opened 2023-09-08
SessionUtils has a self package import from @deephaven/jsapi-utils


bmingles:
**Description**

`SessionUtils` has an import from its owning package `@deephaven/jsapi-utils`. We should fix this and configure eslint to disallow such imports.

PR #1498, opened 2023-09-08
fix: Forbid self package import

bmingles:
- Custom eslint rule to forbid importing from package name of package containing the module
- Fixed offending SessionUtils

I tested this by running `npm run test:lint`. It showed expected linting error before fixing SessionUtils. Fixing SessionUtils made error go away. Also tested in vscode. Error shows up in the editor as expected.

fixes: #1497

### DIFF
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -14,6 +14,9 @@
     "eslint-config-prettier": "8.3.0",
     "eslint-config-react-app": "7.0.0"
   },
+  "devDependencies": {
+    "eslint-plugin-deephaven": "file:./plugin"
+  },
   "peerDependencies": {
     "@typescript-eslint/eslint-plugin": "^5.46.0",
     "@typescript-eslint/parser": "^5.46.0",
--- a/packages/eslint-config/plugin/no-self-package-import.js
+++ b/packages/eslint-config/plugin/no-self-package-import.js
@@ -1,0 +1,60 @@
+/**
+ * Derive the `@deephaven` package name from the given path.
+ * Note that this assumes that the package folder name uses the  naming
+ * convention as the package.json `name` field.
+ */
+function deriveDeephavenPackageNameFromPath(path) {
+  // In the off chance that there is a packages folder higher up in the path,
+  // find the deepest one.
+  const packageFolderNameI = path.lastIndexOf('packages');
+
+  if (packageFolderNameI === -1) {
+    return null;
+  }
+
+  // Find the path token immediately after the `packages` folder.
+  const [packageFolderName] = path
+    .substring(packageFolderNameI + 'packages'.length + 1)
+    .split('/');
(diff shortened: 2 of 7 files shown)


packages/eslint-config/package.json
```diff
@@ -14,6 +14,9 @@
     "eslint-config-prettier": "8.3.0",
     "eslint-config-react-app": "7.0.0"
   },
+  "devDependencies": {
+    "eslint-plugin-deephaven": "file:./plugin"
+  },
```
mattrunyon, line 19: Looks like you don't need to do it this way. You can import the plugin directly

https://eslint.org/docs/latest/extend/custom-rule-tutorial#step-7-bundle-the-custom-rule-in-a-plugin

bmingles: This doesn't work if you try to add it to an array of plugins since the plugins array only supports string names. Granted the current config only has 1 plugin, this will be more future proof.

That said if you find a way to get it working alongside other plugins, I'd love to see how. I was not able to get it working.

packages/eslint-config/plugin/no-self-package-import.js
```diff
@@ -0,0 +1,60 @@
...
+  // Find the path token immediately after the `packages` folder.
+  const [packageFolderName] = path
+    .substring(packageFolderNameI + 'packages'.length + 1)
+    .split('/');
+
+  if (packageFolderName === '') {
+    return null;
+  }
+
+  return `@deephaven/${packageFolderName}`;
```
mofojed, line 24: One thing I'm not crazy about is this rule is not generic; even when we include this eslint config from Enterprise it won't technically be correct. We should also be using the `path` package to resolve paths rather than manipulating the string directly, as there's probably some edge cases you're missing here (e.g. what if I had Enterprise using this eslint config and it was in a folder like `~/deephaven/packages/app/...`, then it would think the package name was `@deephaven/app`).
Getting the package name from the package.json should be doable, though I do also wonder about performance and how we can cache results (since this will be called on every import in every file, don't want to have to traverse and read package.json every time...). I'd rather check in the fix first without the eslint rule than check it in with specific `@deephaven` logic.

bmingles: This "shouldn't" ever actually be used in enterprise since it doesn't really apply. It is currently consumed via the top level Community package so doesn't get pulled in with the shared config. It really is scoped to @deephaven community packages. Does this alleviate any of your concern?

packages/eslint-config/package.json
bmingles: @mattrunyon @mofojed Thinking this PR may be a cleaner option
https://github.com/deephaven/web-client-ui/pull/1499/files

mofojed:
Superceded by https://github.com/deephaven/web-client-ui/pull/1499

PR #1503, opened 2023-09-11
chore: Enabled eslint rules requiring return types on functions

### LINKED ISSUE #1413: Explicit return type not required in TypeScript anymore
**Description**

Not sure when this changed, but we used to enforce an explicit return type. We're no longer doing that, and we should re-enable enforcing it it.

**Steps to reproduce**
1. In the code, remove the return type from a function

**Expected results**
1. Linter should fail, requiring a return type to be specified

**Actual results**
1. Linter does not fail

bmingles:
- Enabled @typescript-eslint/explicit-module-boundary-types for all .ts files
- Enabled @typescript-eslint/explicit-function-return-type for non-test files with allowExpressions exclusion
- Added a bunch of return types
- Added test folders to eslint ignore rules
- Added util type to TypeUtils `OnlyOneProp`

Note, the only runtime changes that should be in this PR are 2 if statements in MockChartModel
https://github.com/deephaven/web-client-ui/pull/1503/files#r1322021147

resolves #1413


packages/chart/src/MockChartModel.ts
```diff
@@ -154,16 +163,23 @@ class MockChartModel extends ChartModel {
...
   }
 
-  static makeDefaultLayout(dh) {
+  static makeDefaultLayout(dh: DhType): Partial<Layout> {
     const layout = new ChartUtils(dh).makeDefaultLayout(ChartTheme);
     layout.title = 'Chart';
-    layout.xaxis.title = 'Datestamp';
-    layout.yaxis.title = 'Price';
+
+    if (layout.xaxis) {
```
bmingles, line 170: These 2 if statements should be the only 2 runtime code changes in this PR

mattrunyon:
I'm not sure if we want the explicit return on all functions. We previously only had the module boundary rule enabled.

All functions feels too noisy to me and removes some of the good parts of TS inference. Module boundaries being explicit makes it easier to catch/document public API breakages. Any non-module boundaries will only be used within our code and inferred types will all get checked.

bmingles:
> I'm not sure if we want the explicit return on all functions. We previously only had the module boundary rule enabled.
> 
> All functions feels too noisy to me and removes some of the good parts of TS inference. Module boundaries being explicit makes it easier to catch/document public API breakages. Any non-module boundaries will only be used within our code and inferred types will all get checked.

@mattrunyon I didn't quite enable it for all functions. The `allowExpressions` exclusion is enabled, which is the one that would have made things noisy for me. For all of the return types the new rules made me add, I didn't feel like any of them were heavy handed. Most to all of them add clarity to the codebase IMO.

mattrunyon:
I don't think most of the added types were heavy, but it feels like writing `const x: number = 42;` instead of letting TS infer that easily which is one of its benefits. At a glance, there were a bunch of just `: void` or `: Promise<void>`.

I think the rule was removed from the default rulesets we consumed because it goes against the power of TS inference, but module boundaries for a library help us catch breaking the API on accident.

If we set up [api-extractor](https://api-extractor.com/) we could even remove this linter rule and still have confidence we didn't break our external API. There's already a bunch of inferred return types we have to make explicit because they are module boundaries, but the module is not exposed as part of the public API.

bmingles:
To clarify, this shouldn't require adding types to things like `const x: number = 42;`. It only requires return types on functions. I'm a fan of type inference when writing code, but explicit function return types can make code more readable since you know what it returns just by looking at the function. It also catches a class of errors where you intend to return a particular type but have a code path in the function that doesn't actually do that.

That said, I'm not hard sold on having to enable this rule, but I do want to highlight the benefits I see in doing so.

packages/app-utils/src/components/LoginNotifier.tsx
```diff
@@ -5,7 +5,7 @@ import { useEffect } from 'react';
 /**
  * Component that broadcasts a message when mounted. Should be mounted after the user has logged in.
  */
-export function LoginNotifier() {
+export function LoginNotifier(): null {
```
mofojed, line 8: Kind of weird just having a `null` return type here - I was going to suggest `ReactNode` as that includes `null`, but then that freaks out in `AuthBootstrap`. Looks like that issue is fixed in TS 5.1: https://devblogs.microsoft.com/typescript/announcing-typescript-5-1-beta/#decoupled-type-checking-between-jsx-elements-and-jsx-tag-types
If we were >=5.1, might consider using `ReactNode` instead.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -1002,7 +1002,12 @@ export class AppMainContainer extends Component<
   }
 }
 
-const mapStateToProps = (state: RootState) => ({
+const mapStateToProps = (
+  state: RootState
+): Omit<
+  AppMainContainerProps,
+  'match' | 'setActiveTool' | 'updateDashboardData' | 'updateWorkspaceData'
+> => ({
```
mofojed, line 1010: Another way to separate the props is in Linker.tsx. I kind of like the way Linker is done, what do you think? Doesn't need to be changed in this PR.

packages/components/src/BulkActionBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+import type { StyleProps } from '@react-types/shared';
+import commonStyles from './SpectrumComponent.module.scss';
+
+export interface BulkActionBarProps {
+  styleProps?: StyleProps;
+  selectedItemCount: 'all' | number;
+  onClearSelection: () => void;
+}
+
+export function BulkActionBar({
```
mofojed: I think this is from your other PR? Shouldn't be part of this PR...

packages/components/package.json
```diff
@@ -23,6 +23,7 @@
     "build:sass": "sass --embed-sources --load-path=../../node_modules ./src:./dist ./scss/BaseStyleSheet.scss:./css/BaseStyleSheet.css"
   },
   "dependencies": {
+    "@adobe/react-spectrum": "^3.29.0",
```
mofojed: Shouldn't be part of this PR

packages/eslint-config/index.js
```diff
@@ -95,6 +104,7 @@ module.exports = {
       rules: {
         'import/no-extraneous-dependencies': 'off',
         '@typescript-eslint/no-non-null-assertion': 'off',
+        '@typescript-eslint/explicit-function-return-type': 'off',
```
mofojed, line 107: I'm dumb - why is this needed?

packages/iris-grid/src/sidebar/conditional-formatting/ConditionEditor.tsx
```diff
@@ -141,7 +141,7 @@ function getNumberInputs(
   conditionValue?: string,
   startValue?: string,
   endValue?: string
-) {
+): JSX.Element | null {
```
mofojed, line 144: I think `ReactNode` would make more sense here.

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -34,7 +34,20 @@ export type PouchDBSort = Array<
   string | { [propName: string]: 'asc' | 'desc' }
 >;
 
-function makePouchFilter(type: string, value: FilterValue | FilterValue[]) {
+type PouchFilter = OnlyOneProp<{
+  $eq: FilterValue | FilterValue[];
+  $neq: FilterValue | FilterValue[];
```
mofojed, line 39: Wish we could use the PouchDB types here instead - it looks like there's actually a bug here. `$neq` should be `$ne`. Obviously was a bug before this change, but if we used PouchDB's types it would pick that up? https://www.npmjs.com/package/pouchdb-find

In any case, can mark it with a follow up ticket or as a separate smaller PR as this would be a runtime change.

mofojed (changes requested):
Overall supportive of the change. Being explicit seems safer, in cases where you might accidentally return the wrong thing/change the return type inadvertently. You can still use type inference to hover over the method to figure out what TS thinks it is if you haven't filled it in yet, but they we should be explicit and lock it in.
Some things to clean up.

packages/eslint-config/index.js
bmingles: Disabling in tests because there are a lot more offenders there. I figured we can be looser on return types in tests.

packages/components/package.json
bmingles: I think rebase got rid of this

packages/components/src/BulkActionBar.tsx
bmingles: rebase took care of this

packages/pouch-storage/src/PouchStorageTable.ts
bmingles: The types for `$regex` also differ. Pouch types say it's a string, but we are using Regex

bmingles:
@mofojed @mattrunyon I ran these rules on Enterprise just to see how big the breaking change is going to be. Looks like 278 errors vs 0 errors for the @typescript-eslint/explicit-module-boundary-types change. I think I may skip adding the `@typescript-eslint/explicit-function-return-type` rule for now to avoid the breaking change. To Matt's point, the original configuration that we lost was likely the boundary type check, so this at least tackles the scope of the original ticket.

packages/app-utils/src/components/LoginNotifier.tsx
bmingles: It looks like `JSX.Element | null` is currently the return type declared by `React.FC`. Maybe that new 5.1 type will address this, but seems like `<SomeComponent/>` does not work for things like `string` which is included in `ReactNode`. For now I think `JSX.Element | null` will be the most appropriate for component return types.

packages/code-studio/src/main/AppMainContainer.tsx
bmingles: I'm a fan of this approach. Wish I'd noticed it before. I won't address on this PR, as there were a bunch of them.

packages/iris-grid/src/sidebar/conditional-formatting/ConditionEditor.tsx
bmingles: Looks like `JSX.Element | null` is the same type used by `React.FC` so going to stick with it for now

PR #1508, opened 2023-09-12
feat: Update go to row panel's row number with cursorRow

### LINKED ISSUE #1406: Go to row panel should update the row number as you navigate the selection
As you navigate the selection, the row number in the `Go to row` panel should change.

image (20)

GcWan:
The row number in the go to row panel now updates when users click on a row in the table. The value displayed when multiple selections or ranges of rows are selected depends on the value of cursorRow.

Fixes a bug where selecting a row then using the context menu to open the go to row panel would fill in the panel with the incorrect row number.

Resolves #1406.


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2915,6 +2915,9 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     if (copyOperation != null) {
       this.setState({ copyOperation: null });
     }
+    if (this.grid?.state.cursorRow != null) {
+      this.setState({ gotoRow: `${this.grid.state.cursorRow + 1}` });
+    }
```
mofojed, line 2920: Good start, but if you click on a row in the grid, then press and hold down on the keyboard, instead of moving downward through the grid it will move focus to the "Go to value" field which is incorrect. We need to set the row without changing focus.

That looks to be caused by a `useEffect` in `GotoRow`. There isn't much detail in the original [Go to row spec](https://drive.google.com/drive/folders/1qEAoNMTI6lVknZfGy655_VAObb59M9--) or https://github.com/deephaven/web-client-ui/issues/616. Consider the usability of how/when we should focus, and if you are unsure/want some guidance, message in #ddl-team-web channel and we can discuss.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2543,7 +2547,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     const text = IrisGridUtils.convertValueToText(cellValue, type);
     this.setState({
       isGotoShown: !isGotoShown,
-      gotoRow: `${cursorRow}`,
+      gotoRow: `${cursorRow + 1}`,
```
mofojed, line 2550: Nice catch

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -55,343 +57,362 @@ interface GotoRowProps {
...
+            </div>
+          )}
+        </>
+      </IrisGridBottomBar>
+    );
+  }
+);
+GotoRow.displayName = 'GotoRow';
 
+export type GotoRowFocus = { focus: () => void };
```
mofojed, line 417: Rather than defining this down here, define this above the component declaration and add the types to the `forwardRef` call.

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -55,343 +57,362 @@ interface GotoRowProps {
...
-  gotoValue,
-  gotoValueFilter,
-  onGotoValueSelectedColumnNameChanged,
-  onGotoValueSelectedFilterChanged,
-  onGotoValueChanged,
-  onGotoValueSubmit,
-}: GotoRowProps): ReactElement {
-  const gotoRowInputRef = useRef<HTMLInputElement>(null);
-  const gotoValueInputRef = useRef<HTMLInputElement>(null);
+const GotoRow = forwardRef(
```
mofojed, line 60: Add types here
```suggestion
const GotoRow = forwardRef<GotoRowFocus, GotoRowProps>(
```

mofojed (changes requested):

mofojed (approved):

PR #1509, opened 2023-09-12
fix: webpack dynamic import

mofojed:
- Use webpackIgnore to disable url resolving: https://webpack.js.org/loaders/css-loader/#disable-url-resolving-using-the--webpackignore-true--comment
- Tested with https://github.com/deephaven-examples/deephaven-react-app-enterprise/tree/cra-light-dashboard. No warning was emitted from the webpack build step.
- Tested `npm start` still loaded correctly


mattrunyon (changes requested):
Add a comment that the webpack ignore comment is intentional so clients using webpack can consume ApiBootstrap. Don't want it to get accidentally removed since we don't use webpack

mattrunyon (approved):

Issue #1504, opened 2023-09-11
Theming - Theme Provider


bmingles:
As a developer, I'd like to be able to load custom themes via the existing DH plugin system. 

- A theme will be defined as a set of custom CSS variables (Note: the scope of this issue is to load a minimal set of variables. The full themes will be defined + mapped in other issues)
- Load minimal variables for a base dark + light theme
- Load custom overrides of these variables from a new 'ThemePlugin' type
- Preload background and spinner color from last applied theme

@dsmmcken did a bunch of preliminary POC work that is available in a draft PR
https://github.com/deephaven/web-client-ui/pull/1460

PR #1511, opened 2023-09-13
feat: Dynamic theme selection

bmingles:
Draft PR showing dynamic theme selection.

- Theme key for selected theme is stored in local storage
- Loads default + custom theme files from public folder (default_theme.css & custom_theme.css respectively)
- Custom theme files can specify `--dh-base-theme` to inherit from a default theme
- Loader assigns appropriate css class for default + custom theme to body element
- IrisGrid themes replace any css variable references via getComputedStyles + getPropertyValue

This includes a few minimal changes to `BaseStyleSheet.scss`, `goldenlayout-dark-theme.scss`, and `IrisGridTheme.ts` to prove they can consume the theme variables.

NOTES: 
- We will need to determine custom themes dynamically likely from a manifest. This is not implemented yet
- This does not yet include an example of Spectrum theming, but we should just be able to map their custom css props to our `dh-xxxx` ones.

To see it work, you can set the theme key in local storage in the console and then refresh the page. You should see different css classes get attached to the body based on the theme key:

```js
localStorage.setItem('deephaven.themeKey', '') // should load default dark theme

localStorage.setItem('deephaven.themeKey', 'custom-aaa') // should load custom-aaa + default-dark theme

localStorage.setItem('deephaven.themeKey', 'custom-bbb') // should load custom-bbb + default-light theme

localStorage.setItem('deephaven.themeKey', 'custom-ccc') // should load custom-ccc + default-dark theme
```

#1504

### DIFF
--- a/packages/code-studio/index.html
+++ b/packages/code-studio/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <base href="%BASE_URL%" />
@@ -16,6 +16,8 @@
     <link rel="manifest" href="/manifest.json" />
     <link rel="icon" href="%VITE_FAVICON%" />
     <title>Deephaven</title>
+    <link rel="stylesheet" href="/default_theme.css" />
+    <link rel="stylesheet" href="/custom_theme.css" />
   </head>
   <body>
     <noscript> You need to enable JavaScript to run this app. </noscript>
--- a/packages/code-studio/src/index.tsx
+++ b/packages/code-studio/src/index.tsx
@@ -1,11 +1,13 @@
 import React, { Suspense } from 'react';
 import ReactDOM from 'react-dom';
 import '@deephaven/components/scss/BaseStyleSheet.scss';
-import { LoadingOverlay } from '@deephaven/components';
+import { initializeTheme, LoadingOverlay } from '@deephaven/components';
 import { ApiBootstrap } from '@deephaven/jsapi-bootstrap';
 import logInit from './log/LogInit';
 
 logInit();
+
+initializeTheme();
 
 // Lazy load components for code splitting and also to avoid importing the jsapi-shim before API is bootstrapped.
 // eslint-disable-next-line react-refresh/only-export-components
--- a/packages/components/src/theme/ThemeUtils.ts
+++ b/packages/components/src/theme/ThemeUtils.ts
@@ -1,0 +1,89 @@
+export type ThemeVariables = Record<string, string | boolean | number | null>;
+
+export const THEME_LOCAL_STORAGE_KEY = 'deephaven.themeKey';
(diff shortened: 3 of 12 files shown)


packages/code-studio/src/index.tsx
```diff
@@ -1,12 +1,14 @@
...
 import ReactDOM from 'react-dom';
 import '@deephaven/components/scss/BaseStyleSheet.scss';
-import { LoadingOverlay } from '@deephaven/components';
+import { initializeTheme, LoadingOverlay } from '@deephaven/components';
 import { ApiBootstrap } from '@deephaven/jsapi-bootstrap';
 import logInit from './log/LogInit';
 
 logInit();
 
+initializeTheme();
```
dsmmcken, line 10: We might want to inline and call this function in a script tag in the actual html <head>, that way there's no flash of the wrong theme before react loads/initializes.

packages/code-studio/index.html
```diff
@@ -16,6 +16,8 @@
     <link rel="manifest" href="/manifest.json" />
     <link rel="icon" href="%VITE_FAVICON%" />
     <title>Deephaven</title>
+    <link rel="stylesheet" href="/default_theme.css" />
+    <link rel="stylesheet" href="/custom_theme.css" />
```
dsmmcken, line 20: Is this more of a POC or intended implementation of how custom themes would be loaded?

It's probably preferable that they are split into separate files, and then loaded dynamically, such that theme files can be created and shared individually.

bmingles: I imagine the default themes would actually live in the codebase not in public. Those could be split up however.

For the custom ones, it's easiest to just have 1 file and statically load it. I could see the benefit of splitting them up, would just have to see how hard it is to determine the full list before our app code has actually loaded

packages/code-studio/src/index.tsx
bmingles: I'm fine with that. Are you actually seeing a flash?

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -0,0 +1,89 @@
...
+export function replaceCssVariables(
+  targetElement: HTMLElement | null,
+  variables: ThemeVariables
+): ThemeVariables {
+  return Object.entries(variables).reduce((result, [key, value]) => {
+    // eslint-disable-next-line no-param-reassign
+    result[key] =
+      typeof value === 'string'
+        ? value.replace(CSS_PROPERTY_NAME_REGEX, (_match, propertyName) =>
+            getComputedStyle(targetElement ?? document.body).getPropertyValue(
```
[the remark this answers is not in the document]
bmingles: ah yes. I saw that. This is mostly just to show case "where" we would hook the replacement into Iris, but yeah needs to be built out more.

bmingles:
Closing this in favor of #1524

bmingles:
Closing this in favor of #1524

PR #1526, opened 2023-09-19
fix: Right clicking with a custom context menu open should open another context menu

### LINKED ISSUE #1525: Trying to open a context menu with a custom context menu already open does not work
**Description**

If you have a custom context menu already open (like on a table) and right click on something else that should create a custom context menu (like another table), the browser default context menu is opened instead.

**Steps to reproduce**

1. Open a table
2. Right click in the table
3. Custom context menu opens
4. Right click somewhere else in the table

**Expected results**

5. Custom context menu opens

**Actual results**

5. Browser default context menu opens

**Versions**

 - Web UI Version: 0.49.0
 - OS: PopOS 22.04 (Ubuntu-based)
 - Browser: Chromium 117

mattrunyon:
Fixes #1525 

This was bothering me while working on element plugins which should have a panel w/ the custom context menu enabled. Tested on Chrome/Firefox on Linux

### DIFF
--- a/packages/components/src/context-actions/ContextMenuRoot.tsx
+++ b/packages/components/src/context-actions/ContextMenuRoot.tsx
@@ -58,11 +58,46 @@ class ContextMenuRoot extends Component<
   openMenu: React.RefObject<ContextMenu>;
 
   handleContextMenu(e: MouseEvent): void {
-    if (!ContextActionUtils.isContextActionEvent(e)) {
+    if (!this.container.current) {
       return;
     }
 
-    if (!this.container.current) {
+    const parentRect = this.container.current.getBoundingClientRect();
+    const top = e.clientY - parentRect.top;
+    const left = e.clientX - parentRect.left;
+    const { actions } = this.state;
+
+    // Mac and Linux appear to trigger contextmenu events on mousedown vs. mouseup on Windows
+    // Mouseup on Windows triggers blur before contextmenu which effectively does what this path does
+    if (actions != null && e.target === this.container.current) {
+      // re-emit right clicks that hit the context-root blocking layer
+      // while we already have a custom context menu open
+      e.preventDefault();
+
+      // Set actions to null removes the menu
+      // That allows a new menu to be opened on a different element so initial position is set properly
+      // Otherwise the instance of this menu may be reused
+      // A new contextmenu event is triggered on the element at the location the user clicked on the blocking layer
+      this.setState({ actions: null }, () => {
+        const element = document.elementFromPoint(left, top); // x y
+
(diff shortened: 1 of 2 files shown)


dsmmcken:
Please suggest a basic test plan we can use to manually check this, because this has proven tricky. I can do the windows testing.

We need to test manually on the full matrix:
- [ ] mac chrome
- [ ] mac firefox
- [ ] mac safari
- [ ] linux chrome
- [ ] linux firefox
- [ ] windows chrome
- [ ] windows edge
- [ ] windows firefox

dsmmcken:
This introduces a new bug which prevents the native menu showing.

Right clicking on something that would be a native menu previously (like say you are trying to select text and copy text from a previously run command in the console) now it doesn't generate the native menu anymore, and further blocks all interactivity until another left click is performed. I assume we are generating like an empty menu with an overlay or something still when we shouldn't be.

Otherwise passes suggested manual testing on all Windows browsers.

mattrunyon:
Good catch. It's actually an issue w/ just the console (native menu in other places like the maximize button or user settings button works fine).

We set/emit console clear and focus history as context actions, but they get filtered out by the util to turn them into menu items. Easy enough to add another check, but seems like a bug/unintended for those to be passed as context actions.

See https://github.com/deephaven/web-client-ui/blob/main/packages/console/src/Console.tsx#L878

dsmmcken:
Also noted same behaviour in console history panel's search field.

mattrunyon:
Ok should be fixed. I took a quick look back, but looks like those actions never registered properly since the initial fork of DHC. They've always been filtered out. Not sure why they're registered as context menu items really other than to setup the shortcut I guess.

mofojed (approved):
Verified looks good on Mac as well

mofojed (commented):
Actually read Don's comments. Yeah the native menu still doesn't show up on the first right-click after showing our custom menu, unsure if that's something we should handle.
Also the context action items for clear/focus history were intended for keyboard shortcuts. Anything without a title is just a context action item that's triggered by keyboard and shouldn't appear in the menu.

mattrunyon:
@mofojed Still good to merge this? I don't think we can handle showing the right-click after showing our custom menu. The event is getting re-emitted, but the browser doesn't seem to do anything with it. Probably because we listen for the event, but I'm guessing the native browser event can only be triggered by trusted/user-interaction events, not our re-emit.

dsmmcken:
> the native menu still doesn't show up on the first right-click after showing our custom menu
It does on Chrome/Windows.

This is okay on the Windows side.

dsmmcken (approved):

PR #1528, opened 2023-09-20
fix: Copy did not work from embedded iframes

### LINKED ISSUE #1527: Copy does not work from iframe embedded grids
**Description**

If you try and copy from a grid within an embedded iframe, it does not actually copy to clipboard and no error is shown. Also, context menu doesn't work.

**Steps to reproduce**

1. Start up web UI locally
2. Create a table that you can embed in an iframe, e.g.:
```
t = empty_table(20000).update("x=i")
```
3. Create an HTML file embedding the grid in an iframe, e.g.
```
<html>
  <body>
    <h1>Embedded table</h1>
    <iframe
      src="http://localhost:10000/iframe/table/?name=t"
      width="800"
      height="500"
    ></iframe>
  </body>
</html>
```
4. Open the embedded HTML file in your browser
5. Select some cells and press Ctrl+C to copy
6. Try to copy from right click menu

**Expected results**
5. Text should be copied to clipboard
6. Right-click menu should open correctly

**Actual results**
5. Text not copied to clipboard. An error is shown in the console logs, but not in the UI:
```
[IrisGridCopyHandler] copyText error DOMException: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://goo.gl/EuHzyv for more details.
```
6. The grid context menu is not shown

image

**Versions**

Engine Version: 0.29.0
Web UI Version: 0.49.0
Java Version: 11.0.20.1
Barrage Version: 0.6.0
Browser: Chromium 116.0.5845.179

mofojed:
- embedded iframes require the 'clipboard-write' permission to copy in Chrome. Added note to the README
- Display an error correctly if the copy doesn't succeed, and added unit test
- Fixed context menu as well - default menu handler did not return an array as it should've
- Tested in Firefox and Chrome on Linux
- Fixes #1527

### DIFF
--- a/packages/iris-grid/src/IrisGridCopyHandler.test.tsx
+++ b/packages/iris-grid/src/IrisGridCopyHandler.test.tsx
@@ -160,3 +160,35 @@ it('retry option available if fetching fails', async () => {
   expect(copyToClipboard).toHaveBeenCalled();
   expect(screen.getByText('Copied to Clipboard!')).toBeTruthy();
 });
+
+it('shows an error if the copy fails permissions', async () => {
+  const user = userEvent.setup({ delay: null });
+  const error = new Error('Test copy error');
+  mockedCopyToClipboard.mockReturnValueOnce(Promise.reject(error));
+
+  const ranges = GridTestUtils.makeRanges();
+  const copyOperation = makeCopyOperation(ranges);
+  mountCopySelection({ copyOperation });
+
+  await waitFor(() =>
+    expect(copyToClipboard).toHaveBeenCalledWith(DEFAULT_EXPECTED_TEXT)
+  );
+
+  expect(screen.getByText('Fetched 50 rows!')).toBeTruthy();
+
+  mockedCopyToClipboard.mockClear();
+  mockedCopyToClipboard.mockReturnValueOnce(Promise.reject(error));
+
+  const btn = screen.getByText('Click to Copy');
+  expect(btn).toBeTruthy();
+
+  await user.click(btn);
+
+  await waitFor(() =>
+    expect(copyToClipboard).toHaveBeenCalledWith(DEFAULT_EXPECTED_TEXT)
+  );
+
+  expect(
+    screen.getByText('Unable to copy. Verify your browser permissions.')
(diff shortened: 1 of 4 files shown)


packages/iris-grid/src/IrisGridCopyHandler.test.tsx
```diff
@@ -160,3 +160,35 @@ it('retry option available if fetching fails', async () => {
...
+    expect(copyToClipboard).toHaveBeenCalledWith(DEFAULT_EXPECTED_TEXT)
+  );
+
+  expect(screen.getByText('Fetched 50 rows!')).toBeTruthy();
+
+  mockedCopyToClipboard.mockClear();
+  mockedCopyToClipboard.mockReturnValueOnce(Promise.reject(error));
+
+  const btn = screen.getByText('Click to Copy');
+  expect(btn).toBeTruthy();
```
mattrunyon, line 183:
```suggestion
  expect(btn).toBeInTheDocument();
```

packages/iris-grid/src/IrisGridCopyHandler.tsx
```diff
@@ -252,27 +260,20 @@ class IrisGridCopyHandler extends Component<
...
-        this.startHideTimer();
-      },
-      error => {
-        log.error('copyText error', error);
-        this.setState({
-          buttonState: IrisGridCopyHandler.BUTTON_STATES.CLICK_TO_COPY,
-          copyState: IrisGridCopyHandler.COPY_STATES.CLICK_REQUIRED,
-        });
-      }
-    );
```
mattrunyon, line 272: Why was this removed?

mattrunyon (changes requested):

packages/iris-grid/src/IrisGridCopyHandler.tsx
mofojed: It was moved down to where the error is caught instead - since we want to show a different error depending on how the copy action occurs (from the initial Copy action that displayed this bar, or from clicking the explicit "Copy" button after that appears).

mattrunyon (approved):

PR #1532, opened 2023-09-20
fix!: CSS based loading spinner

### LINKED ISSUE #1531: DH-15686: Loading Spinner Stutter
**Description**

The loading spinner stutters a bit when loading the page. It's only really noticeable when the page load takes long enough to see it such as when plugins are being loaded.

Supports DH-15686

spinner-stutter

**Steps to reproduce**

This can be seen locally when plugins are being loaded. A synthetic delay can be introduced in `PluginUtils` right after the manifest is loaded

```
await new Promise(resolve => {
  setTimeout(resolve, 5000);
});
```

**Expected results**

The spinner should rotate seamlessly

**Actual results**

It shifts a small bit and backtracks

**Versions**
0.26.0

bmingles:
- Swapped our LoadingSpinner implementation with a simpler html / CSS only spinner
- Fixes a layout shift after font load caused by `font-size: $iris-panel-icon-font-size` in `.message-icon` of LoadingOverlay
- Fixes loading spinner backtracking due to multiple instances of spinners

**Regression Tests Performed**
- LoadingOverlay - I added a synthetic wait to PluginUtils right after the manifest loading.  Loading overlay is now seamless. No layout shift or backtracking

Verified the following look as they did before

Community
- Styleguide indeterminate and progress spinners 
- Styleguide context menu spinners
- Console "Running" spinner
- AppMainContainer disconnection spinner
- Command history tooltip elapsed spinner
- Advanced filter spinner
- Column statistics spinner
- IrisGridCopyHandler slight adjustment such that spinner + text are centered
- PartitionSelectorSearch spinner
- Commit button in PendingDataBottomBar
- Custom column builder "applying" spinner
- Table CSV exporter spinner

Enterprise
- Enterprise login spinner

```
  await new Promise(resolve => {
      setTimeout(resolve, 2000);
  });
```

fixes #1531

BREAKING CHANGE: Inline LoadingSpinner instances will need to be decorated with `className="loading-spinner-vertical-align"` for vertical alignment to work as before


packages/components/src/LoadingSpinner.scss
```diff
@@ -1,6 +1,42 @@
+.loading-spinner {
+  --primary-color: var(--dh-accent-color, #4c7dee);
+  /* stylelint-disable-next-line alpha-value-notation */
+  --secondary-color: rgba(240, 240, 240, 0.5);
```
dsmmcken, line 4: This needs to be based on a variable. A hardcoded value here would look terrible on a light theme for example.

bmingles: Any thoughts on what it should be tied to?

bmingles: I'll tie it to a semantic color for now and refine it once I get into the color mapping

packages/components/src/LoadingOverlay.scss
```diff
@@ -13,9 +13,14 @@ $iris-panel-scrim-bg: rgba(black, 0.1);
   overflow: hidden;
 
   .message-content {
+    display: flex;
+    flex-direction: column;
     font-size: $iris-panel-message-font-size;
+    gap: 20px;
     .message-icon {
       font-size: $iris-panel-icon-font-size;
+      height: $iris-panel-icon-font-size;
```
bmingles, line 22: I set height + line height explicitly to avoid a layout shift after font load. The previous height was 96px based on 1.5 line-height from body. Converting to flexbox with a gap of 20px seems to match the layout as before but without the layout shift

packages/utils/src/DOMUtils.ts
```diff
@@ -28,4 +28,29 @@ export function identityExtractHTMLElement<T>(
   return maybeHTMLElement instanceof HTMLElement ? maybeHTMLElement : null;
 }
 
-export default { getClosestByClassName, identityExtractHTMLElement };
+/**
+ * Synchronize CSS animations with a given name to the same start time.
+ * @param animationName
+ * @param startTime
```
mofojed, line 34: Params need descriptions. Also as it is now, you could make `startTime` default to `0` (since that's the only value you actually pass in).

packages/utils/src/DOMUtils.ts
```diff
@@ -28,4 +28,29 @@ export function identityExtractHTMLElement<T>(
...
+    .getAnimations()
+    .filter(
+      a => a instanceof CSSAnimation && a.animationName === animationName
+    );
+
+  animations.forEach(a => {
+    // eslint-disable-next-line no-param-reassign
+    a.startTime = startTime;
+  });
+}
```
mofojed, line 50: Not sure if I'm reading this incorrectly, but looking at the example on MDN: https://developer.mozilla.org/en-US/docs/Web/API/Animation/startTime#examples
They're setting the new animation to be the same as the existing start time; here you're resetting all of the animations to the same `startTime` (which is always `0`, since that's the only value you pass in). Wouldn't that mean that an existing animation would jump back when a new one is added? Whereas we should just be syncing the new animation to any existing loading spinner that's already shown.

packages/components/src/context-actions/ContextMenu.tsx
```diff
@@ -594,7 +594,7 @@ class ContextMenu extends PureComponent<ContextMenuProps, ContextMenuState> {
     if (pendingItems.length > 0) {
       pendingElement = (
         <div className="loading">
-          <LoadingSpinner />
+          <LoadingSpinner className="mimic-fa-layers-vertical-align" />
```
mofojed, line 597: Not crazy about needing to provide a long class name in many of these cases... should we have a `verticalAlign` prop or something that maps to a class name instead? Or any other way to do this?

mofojed (changes requested):

packages/components/src/context-actions/ContextMenu.tsx
bmingles: I was thinking eventually we should use flex positioning for all of the cases and gradually get rid of the "hack" class. It's kind of a flag that this is styled in a deprecated way. That said, I'm not a big fan of it either, was just trying to not muck with too much of the existing layout. I'm fine to change this to a prop instead.

packages/utils/src/DOMUtils.ts
bmingles: My observation when I dug into this is that the `startTime` is usually `null` by default (also was not what I expected from the docs). My understanding is that it's kind of a static value that determines the basis of when animations are calculated. Setting explicitly to zero later if the animation is already based on zero shouldn't impact the existing animation.

mofojed: Interesting. Can't say I fully understand it, but it does look like they're in sync and doesn't reset the existing animations, so cool.

packages/components/src/context-actions/ContextMenu.tsx
mofojed: Or just a different class name; like we have `loading-spinner-large`, maybe `loading-spinner-vertical-align`

mofojed (approved):
Side note, I imagine we're going to move to Spectrum's `ProgressCircle` at some point for loading spinner: https://react-spectrum.adobe.com/react-spectrum/ProgressCircle.html
I wonder if that will cause this issue to come up again.

mofojed:
A lot of the tests will need to be updated to look for this new loading spinner... we should have a consistent way of determining that in the tests. data-testid is probably the easiest; accessibility roles like the aria-live don't really apply to the loading spinner itself, just to the element behind it... whatever you think may be best.

bmingles:
`data-testid` should be used the same way it was on the original spinner.

For aria roles, here's what Spectrum uses on their indeterminate spinner. I could see us adding those.

image

mofojed:
Ah yea, using `role="progressbar"` makes sense, that would be better for tests to use than the data-testid. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role
Not all tests were using the data-testid so those failing now can be updated to look for that role instead, and/or you could have a utility function for checking if there's a loading element.

bmingles:
@mofojed all tests have been updated, and e2e are passing as well.

mofojed (approved):

PR #1533, opened 2023-09-21
fix: Render tables partitioned by non-string columns

### LINKED ISSUE #1441: Partition-selection drop-downs don't work properly for non-String columns
**Description**

When using non-`String` partitioning columns (`long` in my example), the partition-selection drop-down incorrectly tries to filter as if the values are strings. It wraps them in quotes, causing `MatchFilter` parsing to reject the literal.

**Steps to reproduce**

Use a command like:
```
t=io.deephaven.parquet.table.ParquetTools.readTable(path)
```
to load a Parquet table that uses metadata files, a `TableDefinition`, or inferred partitioning column types from key=value directory name pairs. If any of the partitioning column types is not a `String`, you should see this error or similar.

Note that the automatic first-partition retrieval fails, as does selecting a partition to filter to. `coalesce()` or `where()` on the table to skip this servers as a workaround.
 
**Expected results**

The table to render properly.

**Actual results**

A table widget with an error loaded. See screeenshot.

Here's the exception stack trace from the server-side, as displayed in the console:
```
heduler-Concurrent-3 | i.d.s.s.SessionService    | Internal Error 'ef498c1c-f351-4b5a-9f2b-da5f85b0e713' java.lang.IllegalArgumentException: Failed to convert literal value <"100"> for column "cnum" of type long
	at io.deephaven.engine.table.impl.select.MatchFilter.init(MatchFilter.java:155)
	at io.deephaven.engine.table.impl.PartitionAwareSourceTable.whereImpl(PartitionAwareSourceTable.java:288)
	at io.deephaven.engine.table.impl.PartitionAwareSourceTable.where(PartitionAwareSourceTable.java:272)
	at io.deephaven.engine.table.impl.PartitionAwareSourceTable.where(PartitionAwareSourceTable.java:35)
	at io.deephaven.server.table.ops.FilterTableGrpcImpl.create(FilterTableGrpcImpl.java:57)
	at io.deephaven.server.table.ops.FilterTableGrpcImpl.create(FilterTableGrpcImpl.java:30)
	at io.deephaven.server.table.ops.TableServiceGrpcImpl$BatchExportBuilder.doExport(TableServiceGrpcImpl.java:689)
	at io.deephaven.server.table.ops.TableServiceGrpcImpl.lambda$batch$5(TableServiceGrpcImpl.java:541)
	at io.deephaven.server.session.SessionState$ExportObject.doExport(SessionState.java:921)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at io.deephaven.server.runner.scheduler.SchedulerModule$ThreadFactory.lambda$newThread$0(SchedulerModule.java:78)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NumberFormatException: For input string: ""100""
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Long.parseLong(Long.java:678)
	at java.base/java.lang.Long.parseLong(Long.java:817)
	at io.deephaven.engine.table.impl.select.MatchFilter$ColumnTypeConvertorFactory$4.convertStringLiteral(MatchFilter.java:229)
	at io.deephaven.engine.table.impl.select.MatchFilter.init(MatchFilter.java:152)
	... 15 more
```

Here's a screenshot of the issue:
Screenshot 2023-08-08 at 9 43 32 AM

**Versions**

Version
0.27.0

georgecwan:
* Partitioned tables were unable to be rendered if the partitioned column type is not `java.lang.String`
* Grid can now render string, numerical, and character columns
* Fixes partitioning columns filter dropdown to correctly display `char` types instead of their ASCII values
* The partition search bar function is now able to search for numbers
* Removes the handleFocus function for `PartitionSelectorSearch` since it made it impossible for the user to edit the search bar after deselecting it
* Fixes #1441

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1921,38 +1921,27 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   }
 
   updatePartition(partition: string, partitionColumn: Column): void {
-    if (partitionColumn.type === 'char' && partition === '') {
+    if (TableUtils.isCharType(partitionColumn.type) && partition === '') {
       return;
     }
 
     const { model } = this.props;
 
-    let partitionFilterValue;
-    if (
-      partitionColumn.type === 'java.lang.String' ||
-      partitionColumn.type === 'char'
-    ) {
-      partitionFilterValue = model.dh.FilterValue.ofString(
-        model.displayString(
+    const partitionText = TableUtils.isCharType(partitionColumn.type)
+      ? model.displayString(
           partition,
           partitionColumn.type,
           partitionColumn.name
         )
-      );
-    } else if (
-      partitionColumn.type === 'java.lang.Boolean' ||
-      partitionColumn.type === 'boolean'
-    ) {
-      partitionFilterValue = model.dh.FilterValue.ofBoolean(
-        typeof partition === 'string'
-          ? TableUtils.makeBooleanValue(partition)
-          : partition
-      );
-    } else {
-      partitionFilterValue = model.dh.FilterValue.ofNumber(partition);
-    }
+      : partition;
+    const partitionFilter = this.tableUtils.makeQuickFilterFromComponent(
+      partitionColumn,
+      partitionText
+    );
 
-    const partitionFilter = partitionColumn.filter().eq(partitionFilterValue);
+    if (partitionFilter === null) {
+      return;
+    }
 
     const partitionFilters = [partitionFilter];
     this.setState({
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1938,7 +1938,6 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       partitionColumn,
       partitionText
     );
-
     if (partitionFilter === null) {
       return;
     }
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1921,10 +1921,25 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   }
 
   updatePartition(partition: string, partitionColumn: Column): void {
+    if (partitionColumn.type === 'char' && partition === '') {
+      return;
+    }
+
     const { model } = this.props;
     const partitionFilter = partitionColumn
       .filter()
-      .eq(model.dh.FilterValue.ofString(partition));
+      .eq(
+        partitionColumn.type === 'java.lang.String' ||
+          partitionColumn.type === 'char'
+          ? model.dh.FilterValue.ofString(
+              model.displayString(
+                partition,
+                partitionColumn.type,
+                partitionColumn.name
+              )
+            )
+          : model.dh.FilterValue.ofNumber(partition)
+      );
     const partitionFilters = [partitionFilter];
     this.setState({
       partition,
@@ -4424,7 +4439,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
                     type: string,
                     stringName: string
                   ) => model.displayString(value, type, stringName)}
-                  columnName={partitionColumn.name}
+                  column={partitionColumn}
                   partition={partition}
                   onChange={this.handlePartitionChange}
                   onFetchAll={this.handlePartitionFetchAll}
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1943,7 +1943,11 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       partitionColumn.type === 'java.lang.Boolean' ||
       partitionColumn.type === 'boolean'
     ) {
-      partitionFilterValue = model.dh.FilterValue.ofBoolean(partition);
+      partitionFilterValue = model.dh.FilterValue.ofBoolean(
+        typeof partition === 'string'
+          ? TableUtils.makeBooleanValue(partition)
+          : partition
+      );
     } else {
       partitionFilterValue = model.dh.FilterValue.ofNumber(partition);
     }
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1926,20 +1926,30 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
 
     const { model } = this.props;
-    const partitionFilter = partitionColumn
-      .filter()
-      .eq(
-        partitionColumn.type === 'java.lang.String' ||
-          partitionColumn.type === 'char'
-          ? model.dh.FilterValue.ofString(
-              model.displayString(
-                partition,
-                partitionColumn.type,
-                partitionColumn.name
-              )
-            )
-          : model.dh.FilterValue.ofNumber(partition)
+
+    let partitionFilterValue;
+    if (
+      partitionColumn.type === 'java.lang.String' ||
+      partitionColumn.type === 'char'
+    ) {
+      partitionFilterValue = model.dh.FilterValue.ofString(
+        model.displayString(
+          partition,
+          partitionColumn.type,
+          partitionColumn.name
+        )
       );
+    } else if (
+      partitionColumn.type === 'java.lang.Boolean' ||
+      partitionColumn.type === 'boolean'
+    ) {
+      partitionFilterValue = model.dh.FilterValue.ofBoolean(partition);
+    } else {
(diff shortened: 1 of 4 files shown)


georgecwan:
There is still an error that occurs when trying to partition by a boolean column:
```
heduler-Concurrent-2 | i.d.s.s.SessionService    | Internal Error '5c0a20aa-a105-4ad6-b89f-f1aab9995272' java.lang.IllegalArgumentException: Unknown type boolean for MatchFilter value auto-conversion
```
The code I ran is:
```

write(part, "/tmp/bool-test/boolCol=true/part.parquet")
write(part, "/tmp/bool-test/boolCol=false/part.parquet")

partition_table = read("/tmp/bool-test")
```
I suspect that the type of the boolCol being `boolean` instead of `java.lang.Boolean` might be related to the issue. I am not sure how to resolve this, some help would be appreciated.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -173,12 +178,16 @@ class IrisGridPartitionSelector<T> extends Component<
...
         </div>
         <div className="input-group">
           <input
             type="text"
-            value={partition}
+            value={
+              column.type === 'char' && partition.toString().length > 0
+                ? String.fromCharCode(parseInt(partition, 10))
+                : partition
+            }
```
mofojed, line 190: Might be able to use `IrisGridUtils.convertValueToText` here instead.

packages/iris-grid/src/PartitionSelectorSearch.tsx
```diff
@@ -275,19 +281,26 @@ class PartitionSelectorSearch<T> extends Component<
...
+        column.type === 'char' ||
+        column.type === 'java.lang.Boolean' ||
+        !Number.isNaN(Number(filterText))
+      ) {
+        const filter = this.tableUtils.makeQuickFilterFromComponent(
+          column,
+          column.type === 'java.lang.String' ? `~${filterText}` : filterText
         );
-      filters.push(filter);
+        if (filter) {
```
mofojed, line 300: If `filter == null`, we should log an error at the least; I think it even makes sense to just `throw` here (since the filter should _never_ by null, otherwise we screwed something up).

mofojed (changes requested):

mofojed:
@georgecwan For the different column type (`boolean` vs `java.lang.Boolean`), I'd check in #ddl-team-dhc if that is expected or not. It seems like `boolean` is always promoted to a `java.lang.Boolean` when you create a new table, but not in this case. E.g. if you do:
```
t = empty_table(100).update("x=(boolean)(i%2==0)")
print(t.columns)
```
You get:
```
[Column(name='x', data_type=java.lang.Boolean, component_type=None, column_type=NORMAL)]
```

packages/iris-grid/src/IrisGridPartitionSelector.tsx
georgecwan: I looked into it and `IrisGridUtils.convertValueToText` doesn't convert `partition` to a `char` because partition is always a `string` type and `convertValueToText` only performs the conversion for `number` types.

georgecwan:
Error with types `boolean` vs `java.lang.Boolean` will be resolved in https://github.com/deephaven/deephaven-core/pull/4547. I have tested my changes against that PR and tables partitioned on `boolean` columns will work on that version of core.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1921,10 +1921,39 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+
     const { model } = this.props;
-    const partitionFilter = partitionColumn
-      .filter()
-      .eq(model.dh.FilterValue.ofString(partition));
+
+    let partitionFilterValue;
+    if (
+      partitionColumn.type === 'java.lang.String' ||
+      partitionColumn.type === 'char'
```
mofojed, line 1933: Use `TableUtils.isTextType`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1921,10 +1921,39 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      partitionFilterValue = model.dh.FilterValue.ofString(
+        model.displayString(
+          partition,
+          partitionColumn.type,
+          partitionColumn.name
+        )
+      );
+    } else if (
+      partitionColumn.type === 'java.lang.Boolean' ||
+      partitionColumn.type === 'boolean'
```
mofojed, line 1944: Use `TableUtils.isBooleanType`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1921,10 +1921,39 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   }
 
   updatePartition(partition: string, partitionColumn: Column): void {
+    if (partitionColumn.type === 'char' && partition === '') {
```
mofojed, line 1924: Use `TableUtils.isCharType`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1921,10 +1921,39 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      partitionColumn.type === 'java.lang.Boolean' ||
+      partitionColumn.type === 'boolean'
+    ) {
+      partitionFilterValue = model.dh.FilterValue.ofBoolean(
+        typeof partition === 'string'
+          ? TableUtils.makeBooleanValue(partition)
+          : partition
+      );
+    } else {
+      partitionFilterValue = model.dh.FilterValue.ofNumber(partition);
```
mofojed, line 1952: Did use `TableUtils.makeQuickFilter` not work?

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -173,12 +179,17 @@ class IrisGridPartitionSelector<T> extends Component<
...
         <div className="status-message">
-          <span>Filtering &quot;{columnName}&quot; partition to</span>
+          <span>Filtering &quot;{column.name}&quot; partition to</span>
         </div>
         <div className="input-group">
           <input
             type="text"
-            value={partition}
+            value={
+              // IrisGridUtils.convertValueToText(partition, column.type)
```
mofojed, line 188: Remove unused code
```suggestion
```

packages/iris-grid/src/PartitionSelectorSearch.tsx
```diff
@@ -275,19 +281,29 @@ class PartitionSelectorSearch<T> extends Component<
...
-        .filter()
-        .invoke(
-          'matches',
-          dh.FilterValue.ofString(`(?s)(?i).*\\Q${filterText}\\E.*`)
+      if (
+        column.type === 'java.lang.String' ||
+        column.type === 'char' ||
+        column.type === 'java.lang.Boolean' ||
+        !Number.isNaN(Number(filterText))
+      ) {
```
mofojed, line 295: Use `TableUtils.isTextType` etc to check the column type...
Instead of having a check like this though, it would be better if we validated the input when it is changed; for example, the `input` field could change to `type="number"` with an appropriate `step` if it's a integer or a double type, and `handleTextChange` could filter it so that only numbers from the input are accepted. Then we shouldn't need a check here.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
georgecwan: It used to break the `char` and `boolean` cases but those problems have been resolved so I have changed back to using `TableUtils.makeQuickFilterFromComponent`

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -173,12 +179,16 @@ class IrisGridPartitionSelector<T> extends Component<
...
         <div className="status-message">
-          <span>Filtering &quot;{columnName}&quot; partition to</span>
+          <span>Filtering &quot;{column.name}&quot; partition to</span>
         </div>
         <div className="input-group">
           <input
             type="text"
-            value={partition}
+            value={
+              column.type === 'char' && partition.toString().length > 0
```
mofojed, line 188: TableTools.isCharType

mofojed (changes requested):

mofojed (approved):

PR #1547, opened 2023-09-25
fix: Formatting Rule Doesn't use default set by user

Zhou-Ziheng:
* Default Formatting Rule should change depending on the defaults set by the user
* Changing columnType should fill the input with the default string
~~* Pressing `enter` on empty formatting string input fields should fill the input with the default string~~
~~* Updated eslint config to handle lines ending in CRLF correctly~~

### DIFF
--- a/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
+++ b/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
@@ -212,7 +212,7 @@ export class ColumnSpecificSectionContent extends PureComponent<
         let resetKeys = {};
         if (key === 'columnType') {
           resetKeys = {
-            format: '',
+            format: this.makeDefaultFormatterItemByType(value as string),
           };
         }
         const newEntry = {
@@ -431,11 +431,6 @@ export class ColumnSpecificSectionContent extends PureComponent<
       this.handleFormatRuleChange(i, 'isNewRule', false);
     const onTypeChange = (e: ChangeEvent<HTMLSelectElement>): void => {
       this.handleFormatRuleChange(i, 'columnType', e.target.value);
-      this.handleFormatRuleChange(
-        i,
-        'format',
-        this.makeDefaultFormatterItemByType(e.target.value)
-      );
     };
     const ruleError = this.getRuleError(rule);
 
--- a/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
+++ b/packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
@@ -71,9 +71,6 @@ interface ColumnSpecificSectionContentState {
   showTimeZone: boolean;
   showTSeparator: boolean;
   timeZone: string;
-  defaultDateTimeFormat: string;
-  defaultDecimalFormatOptions: FormatOption;
-  defaultIntegerFormatOptions: FormatOption;
   truncateNumbersWithPound: boolean;
   timestampAtMenuOpen: Date;
 }
@@ -116,9 +113,6 @@ export class ColumnSpecificSectionContent extends PureComponent<
 
     const {
       formatter,
-      defaultDateTimeFormat,
-      defaultDecimalFormatOptions,
-      defaultIntegerFormatOptions,
       showTimeZone,
       showTSeparator,
       timeZone,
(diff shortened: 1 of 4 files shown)


packages/eslint-config/index.js
```diff
@@ -14,7 +14,12 @@ module.exports = {
     'react-refresh',
   ],
   rules: {
-    'prettier/prettier': ['error'],
+    'prettier/prettier': [
+      'error',
+      {
+        endOfLine: 'auto',
+      },
```
mofojed, line 21: We shouldn't be changing the `endOfLine` default, we want to keep it to `lf` instead of `crlf`. Might need to configure something on Windows so that line endings are handled correctly in git.

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -617,7 +617,7 @@ export class FormattingSectionContent extends PureComponent<
                 checked={truncateNumbersWithPound ?? null}
                 onChange={this.handleTruncateNumbersWithPoundChange}
               >
-                Truncate numbers with #
+                Show truncated numbers as ###
```
mofojed, line 620: A contextual help bubble could be useful here... I think the phrase "Mask truncated numbers with ###" may be better as well, @dsmmcken any opinion?

packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
```diff
@@ -552,7 +550,22 @@ export class ColumnSpecificSectionContent extends PureComponent<
         })}
         data-lpignore
         id={formatId}
-        placeholder={IntegerColumnFormatter.DEFAULT_FORMAT_STRING}
+        placeholder={
+          defaultFormatString ?? IntegerColumnFormatter.DEFAULT_FORMAT_STRING
+        }
+        onKeyDown={e => {
+          if (e.key === 'Enter' && value === '') {
```
mofojed, line 557: I'm not sure I like this - it's kind of hidden functionality that is difficult for the user to find. Granted, I don't know a better way to start with the default/edit it, other than copying pasting from above.
@dsmmcken Do you have any ideas for creating a new rule based off the default but edited slightly?

mofojed (changes requested):
Some questions for Don, and make sure we don't screw up line endings.

packages/code-studio/src/settings/FormattingSectionContent.tsx
dsmmcken: I am fine with this suggestion. Using the word show is more consistent wit the other labels, like "Show T seperator", and I like the multiple hashtags in the label. It's a bit long now is my only complaint.

dsmmcken: It only applies to numbers though

Zhou-Ziheng: The other option is to say `Show truncated numbers as #`

packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
```diff
@@ -221,7 +212,7 @@ export class ColumnSpecificSectionContent extends PureComponent<
         let resetKeys = {};
         if (key === 'columnType') {
           resetKeys = {
-            format: '',
+            format: this.makeDefaultFormatterItemByType(value as string),
```
mofojed, line 215: I wonder if there's a way using function overloading and/or generics to correct this so you don't need to cast it: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

Figuring that out isn't really necessary for this though.

packages/code-studio/src/settings/ColumnSpecificSectionContent.tsx
```diff
@@ -399,8 +429,9 @@ export class ColumnSpecificSectionContent extends PureComponent<
       this.handleFormatRuleChange(i, 'columnName', e.target.value);
     const onNameBlur = (): void =>
       this.handleFormatRuleChange(i, 'isNewRule', false);
-    const onTypeChange = (e: ChangeEvent<HTMLSelectElement>): void =>
+    const onTypeChange = (e: ChangeEvent<HTMLSelectElement>): void => {
       this.handleFormatRuleChange(i, 'columnType', e.target.value);
+    };
```
mofojed, line 434: Not sure why you added these here - `handleFormatRuleChange` doesn't return anything, so you can keep the shorthand. Or update the `onNameChange` and `onNameBlur` declarations to match for consistency.

mofojed (changes requested):
Looks good overall. Just a very minor cleanup.

mofojed (approved):

mofojed:
@Zhou-Ziheng good work. We need to port the same fix over to Enterprise as well (Settings menu isn't shared currently - that would be another task that would be sweet).

PR #1551, opened 2023-09-27
feat: Add copy/rename/delete options to notebook overflow menu

### LINKED ISSUE #1359: Copy/Rename/Delete are missing from notebook menu options
Enterprise notebooks expose Copy/Rename/Delete on the notebook overflow menu dropdown, these options are missing in community.

image

**Versions**
Engine Version: 0.26.0
Web UI Version: 0.40.4
Java Version: 17.0.7
Barrage Version: 0.5.0

georgecwan:
* Adds options to make a copy of, rename, or delete the file currently being viewed in the notebook panel
* Resolves #1359

#### Testing Instructions:
* Open any file create a new file
* Navigate to the overflow menu in the top right of the notebook panel
* The "copy file" option should create a copy of the currently open file with all of its contents copied over
    * The created copy should not appear in the file explorer until a save operation is done on it
* The "rename file" option should cause the file to appear in the file explorer with the same contents and specified name
    * If the file was saved under a different name before, the old file should be deleted
    * If the file was never saved before, this operation will be identical to a save
* The "delete file" option should close the file that the notebook panel is currently displaying.
    * If the file has been saved before, the saved file should be deleted and no longer in the file explorer
    * If the file was never saved before, this operation will be identical to closing the file and discarding any changes


mofojed (approved):
Great work! For future PRs I'd like details on what testing you did.

Issue #1553, opened 2023-09-27
fix: Broken e2e test - table-operations Drag


bmingles:
**Description**

One of our e2e tests is failing due to a mismatched screenshot. This is not currently reproducible locally.

```
1) [webkit] › table-operations.spec.ts:369:5 › custom column › Drag ──────────────────────────────

    Error: Screenshot comparison failed:

      4289 pixels (ratio 0.02 of all image pixels) are different.

    Call log:
      - expect.toHaveScreenshot with timeout 15000ms
      -   verifying given screenshot expectation
      - waiting for locator('.iris-grid-column')
      -   locator resolved to <div class="iris-grid-column">…</div>
      - taking element screenshot
      -   disabled all CSS animations
      -   waiting for element to be visible and stable
      -   element is visible and stable
      - 4289 pixels (ratio 0.02 of all image pixels) are different.
      - waiting 100ms before taking screenshot
      - waiting for locator('.iris-grid-column')
      -   locator resolved to <div class="iris-grid-column">…</div>
      - taking element screenshot
      -   disabled all CSS animations
      -   waiting for element to be visible and stable
      -   element is visible and stable
      - captured a stable screenshot
      - 4289 pixels (ratio 0.02 of all image pixels) are different.

    Expected: /home/runner/work/web-client-ui/web-client-ui/tests/table-operations.spec.ts-snapshots/custom-column-3-webkit-linux.png
    Received: /home/runner/work/web-client-ui/web-client-ui/test-results/table-operations-custom-column-webkit/custom-column-3-actual.png
        Diff: /home/runner/work/web-client-ui/web-client-ui/test-results/table-operations-custom-column-webkit/custom-column-3-diff.png

      444 |
      445 |     await waitForLoadingDone(page);
    > 446 |     await expect(page.locator('.iris-grid-column')).toHaveScreenshot();
          |                                                     ^
      447 |   });
      448 | });
      449 |

        at /home/runner/work/web-client-ui/web-client-ui/tests/table-operations.spec.ts:446:53
        at /home/runner/work/web-client-ui/web-client-ui/tests/table-operations.spec.ts:4[17](https://github.com/deephaven/web-client-ui/actions/runs/6331018335/job/17194694692#step:9:18):3

    attachment #1: custom-column-3-expected.png (image/png) ────────────────────────────────────────
    tests/table-operations.spec.ts-snapshots/custom-column-3-webkit-linux.png
[...]

PR #1554, opened 2023-09-27
fix: Disabled failing e2e test

bmingles:
Temporarily disabling failing e2e test until we can figure out the root cause.

#1553


mattrunyon:
This doesn't actually fix the linked ticket. I linked the existing closed tickets in the new ticket. We should have something open for the failing test still.

mattrunyon (approved):

PR #1556, opened 2023-09-29
feat: data bar gui

ethanalvizo:
Need to run your own instance of DHC. 
Use this branch: https://github.com/ethanalvizo/deephaven-core/tree/3549-databar-jsapi-setting

Run this in the console
```
from deephaven.column import int_col, double_col, string_col, long_col
students = new_table([
    string_col("Name", ["Andy", "Claire", "Jane", "Steven"]),
    int_col("MathGrade", [1, 2, 3, 4]),
    long_col("GPA", [-3, 4, 3, 2]),
])
```

Go to conditional formatting menu and change column to GPA

Note: this supercedes the previous data bar branch so if this does get merged later the other one can be closed


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1091,7 +1150,19 @@ class IrisGridTableModelTemplate<
 
     log.debug2('copyTotalsData', dataMap);
 
+    const dataBarColumnIndex = 2;
```
ethanalvizo, line 1153: this was the issue I was trying to describe earlier in the meeting. In the table that I created there were three columns (Name, MathGrade, GPA). If I want to access min or max from the map, I have to use the column index as the key to access it. I hard coded it to 2 since I know GPA will be the 3rd column in the example I generated but I need a way to dynamically find this

ethanalvizo:
One solution I thought of is access totals data from the `dataBarOptionsForCell` and using the model index param to get that min and max that I need. I'm getting multiple typescript errors saying a lot of the elements I chain through in order to get that value might be null. Is there a better way for me to access this value through all the layers?

image

PR #1557, opened 2023-10-03
fix: Handle deletion of unsaved copied file in NotebookPanel

georgecwan:
* Previously, duplicating a notebook file and then trying to delete it without saving it first will cause an error to occur
* Added a `fileExists` function to `FileUtils` to handle this check
* Added a `getUniqueCopyFileName` function to `FileUtils`
* Behaviour of copying a file from `NotebookPanel` changed from opening an unsaved copy of the file to duplicating the file in `FileStorage` first
* Related to #1359 

#### Testing Instructions:
* Open a file in the notebook panel
* Copy the opened file by using the 'Copy File' option in the overflow menu
* Delete the file by using the overflow menu (No error should occur)
* Repeat the same steps but copy the file by right clicking the file in the file explorer
* Delete the copied file again, this time it should also disappear from the file explorer


mofojed (changes requested):
There's still a case that can result in lost data:
1) Have a file named `foo.py`
2) In File Explorer, right-click the file and select "Copy File". A new file `foo-copy.py` is created.
3) Open foo-copy.py, make some edits, then save and close it.
4) Open the original `foo.py` notebook
5) From the notebook menu, choose "Copy File". It will open a new tab with the name `foo-copy.py`, effectively masking the file from the same name already on disk.
6) From the new tab with the name `foo-copy.py`, select `Delete File`. It will delete the `foo-copy.py` file from disk.

I think the fix for that is in `handleCopy` to loop when getting the copy file name until you know that file doesn't exist.

georgecwan:
That wouldn't fix the problem since the act of copying the file doesn't save it (this was the behaviour before my change) so trying to copy the same file twice without saving either of them first will still have the same problem. Is there a way to get all of files that `NotebookPanel` has open?

mofojed:
Right that's a good point; or you could copy it from the notebook panel, then copy it from the file explorer after.
Yes, there is an `openFileMap` defined in `ConsolePlugin` that keeps track of all the files that are open. You could use redux to set that for the current dashboard (similar to `setDashboardConsoleSettings`) and then use that when copying files...

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
```diff
@@ -166,17 +166,9 @@ export class FileExplorerPanel extends React.Component<
...
-        await fileStorage.info(newName);
-        return true;
-      } catch (error) {
-        return false;
-      }
-    };
     // await in loop is fine here, this isn't a parallel task
-    // eslint-disable-next-line no-await-in-loop
-    while (await checkNewName()) {
+    // eslint-disable-next-line no-await-in-loop, @typescript-eslint/strict-boolean-expressions
```
mofojed, line 170:
```suggestion
    // eslint-disable-next-line no-await-in-loop
```
Doesn't seem to be necessary to remove the strict-boolean-expressions warning.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -681,23 +681,40 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
-    const content = this.getNotebookValue();
     const { language } = settings;
     const { itemName } = fileMetadata;
-    const copyName = FileUtils.getCopyFileName(itemName);
+    let copyName = FileUtils.getCopyFileName(itemName);
+    // await in loop is fine here, this isn't a parallel task
+    // eslint-disable-next-line no-await-in-loop, @typescript-eslint/strict-boolean-expressions
+    while (await FileUtils.fileExists(fileStorage, copyName)) {
+      copyName = FileUtils.getCopyFileName(copyName);
+    }
```
mofojed, line 695: This whole block could be factored into a method in `FileUtils`, e.g.
```
async function getUniqueCopyFileName(fileStorage, originalName) { ... }
```
We do the exact same thing in FileExplorerPanel

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -681,23 +681,40 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+    const notebookSettings = {
+      value: null,
+      language,
+    };
+    glEventHub.emit(
+      NotebookEvent.SELECT_NOTEBOOK,
+      session,
+      language,
+      notebookSettings,
+      newFileMetadata
```
mofojed, line 708: We should pass `true` for the `shouldFocus` parameter here, so the newly copied panel gets focus (as it did before).

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -681,23 +681,40 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
-    const copyName = FileUtils.getCopyFileName(itemName);
+    let copyName = FileUtils.getCopyFileName(itemName);
+    // await in loop is fine here, this isn't a parallel task
+    // eslint-disable-next-line no-await-in-loop, @typescript-eslint/strict-boolean-expressions
+    while (await FileUtils.fileExists(fileStorage, copyName)) {
+      copyName = FileUtils.getCopyFileName(copyName);
+    }
     log.debug('handleCopy', fileMetadata, itemName, copyName);
-    this.createNotebook(copyName, language, content);
+    await fileStorage.copyFile(itemName, copyName);
```
mofojed, line 697: Ideally we'd be catching an error if one occurred here, and displaying a toast for it. However, we don't really have toasts wired up at all at the moment, and it's possible we just wire up a toast service for catching uncaught exceptions anyway so do not worry about it for now.

mofojed (commented):
Make sure we get @dsmmcken 's approval before merging, as this does change behaviour (since the file is created right on copy, instead of requiring an explicit save after). I think it's better overall (too many weird cases the other way).

packages/dashboard-core-plugins/src/panels/FileExplorerPanel.tsx
georgecwan: Without disabling the strict-boolean-expressions check, I get this warning:
image

Should I ignore it then?

mofojed: I think your TypeScript server be trippin', if you restart it (Ctrl+P, then type `> Restart TypeScript Server`), it'll fix it. If it's still causing issues just leave it.

mofojed (approved):

dsmmcken (approved):

PR #1558, opened 2023-10-04
feat: Read settings from props/server config when available

Zhou-Ziheng:
Updated the follow values to use ServerConfig when possible:
- defaultDateTimeFormat
- timeZone
- defaultDecimalFormatOptions
- defaultIntegerFormatOptions
- truncateNumbersWithPound
- defaultNotebookSettings.isMinimapEnabled

To test, first create a config file  (see https://deephaven.io/core/docs/how-to-guides/configuration/config-file/), then add
```- ./deephaven.prop:/opt/deephaven/config/deephaven.prop```
as a volumn for services.deephaven.columns

### CODE REVIEW

**Review by mofojed — CHANGES_REQUESTED** [review]

**Review by mofojed — COMMENTED** [review]
Can you include the steps you took to configure the server/test different default values being read?

File: packages/code-studio/src/settings/FormattingSectionContent.tsx [hunk 1]
```diff
@@ -201,199 +197,135 @@ export class FormattingSectionContent extends PureComponent<
...
+    };
+    this.setState(update);
+    if (
+      isValidFormat(
+        dh,
+        TableUtils.dataType.INT,
+        IntegerColumnFormatter.makeCustomFormat(event.target.value)
+      )
+    ) {
+      this.commitChanges(update);
```
mofojed, line 241: We should still debounce these changes, as they could have implications for the entire UI (e.g. if you have a table open and you're updating the formatting, it'll update every cell of every table open with every keystroke).
See my other comment in this file about debouncing for a suggestion.

packages/redux/src/store.ts
```diff
@@ -65,6 +65,15 @@ export interface WorkspaceData {
   settings: WorkspaceSettings;
 }
 
+export interface CustomzableWorkspaceData
```
mofojed, line 68: Typo
```suggestion
export interface CustomizableWorkspaceData
```

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -201,199 +197,135 @@ export class FormattingSectionContent extends PureComponent<
...
-      () => {
-        this.debouncedCommitChanges();
-      }
-    );
+    this.setState({
+      timeZone,
+    });
+    this.debouncedCommitChanges({
+      timeZone: undefined,
+    });
```
mofojed, line 289: Hm. So previously with `debouncedCommitChanges`, it would just get the current state and then set everything in settings... with this change, we're calling `debouncedCommitChanges` with a parameter now of the setting that has changed, and only sending the save for that. The only problem with that is that it's the same `commitChanges` method; so if you change the timezone and then immediately change decimal format afterwards, since the method is debounced it may drop the timezone update and only send the decimal format update (as the first call to `commitChanges` with the timezone param is dropped due to debounce).

To make sure we don't drop any updates, we basically need to keep a queue of all the updates and then apply them. So something like:

```ts
/** Updates that are pending */
pendingUpdates: Partial<WorkspaceSettings>[];

constructor(...) {
  ...
  this.pendingUpdates = []
}

commitChanges(): void {
    // Reduce the updates array to one partial workspace settings update
    // This will merge all the updates into one object
    // This is needed because the debounced function can be called multiple times
    // before the debounce timeout is reached
    const updates = this.pendingUpdates.reduce(
      (acc, update) => ({
        ...acc,
        ...update,
      }),
      {}
    );
    this.pendingUpdates = [];
    
    const { updateSettings } = this.props;
    updateSettings(updates);
}
```

mofojed (changes requested):

packages/redux/src/selectors.ts
```diff
@@ -41,38 +41,71 @@ export const getWorkspaceStorage = <State extends RootState>(
...
+  if (workspace == null) {
+    // this will only be null on initial render
+    return null as never;
+  }
+  const defaultInjectedWorkspace: Workspace = {
+    data: {
+      ...workspace.data,
+      settings: getDefaultWorkspaceSettings(store) ?? {},
+    },
+  };
```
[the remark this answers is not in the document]
Zhou-Ziheng: I wonder if we are fetching the entire workspace then manually destructure the settings object from it somewhere. My original intention was so that the Workspace object would look identical to before the change, to avoid creating unintended changes elsewhere

packages/redux/src/selectors.ts
```diff
@@ -41,38 +41,71 @@ export const getWorkspaceStorage = <State extends RootState>(
...
+
+  for (let i = 0; i < keys.length; i += 1) {
+    const key = keys[i];
+    const value = customizedWorkspaceSettings[key];
+    // only set pass in customized defaults if defined
+    if (value !== undefined) {
+      defaultInjectedWorkspace.data.settings[key] =
+        value as WorkspaceSettings[typeof key] as never;
+    }
+  }
```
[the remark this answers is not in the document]
Zhou-Ziheng: This was because sometimes the userSettings have values intentionally set to undefined (not just missing) to let user reset values to default etc. setting a value to undefined means the value is still keyed in userSettings so doing object spreading would make those values undefined as opposed to the default.

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -23,45 +23,39 @@ import {
...
-  defaultDecimalFormatOptions: FormatOption;
-  defaultIntegerFormatOptions: FormatOption;
+  defaultDateTimeFormat?: string;
+  showTimeZone?: boolean;
+  showTSeparator?: boolean;
+  timeZone?: string;
+  truncateNumbersWithPound?: boolean;
+  saveSettings: (settings: Partial<WorkspaceSettings>) => void;
+  defaultDecimalFormatOptions?: FormatOption;
+  defaultIntegerFormatOptions?: FormatOption;
```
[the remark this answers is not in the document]
Zhou-Ziheng: This was mostly wrestling with TypeScript. I had to change the Redux state type for `workspace.data.settings` to be optional (because it really is optional now), so the getWorkspace's return value is now typed as CustomizableWorkspace instead of Workspace, marking these parameters optional. I'll investigate for a workaround

Zhou-Ziheng:
This is the `deephaven.prop` I tested with:
```includefiles=dh-defaults.prop

internal.webClient.appInit.canCopy=false
internal.webClient.appInit.canDownloadCsv=false
integerFormat=##123

client.configuration.list=java.version,deephaven.version,barrage.version,internal.webClient.appInit.canCopy,internal.webClient.appInit.canDownloadCsv,integerFormat,```
the only thing of importance is the integerFormat

File: packages/code-studio/src/storage/LocalWorkspaceStorage.ts [hunk 7]
```diff
@@ -21,35 +22,93 @@ const log = Log.module('LocalWorkspaceStorage');
...
+        defaultFormatString: IntegerColumnFormatter.DEFAULT_FORMAT_STRING,
+      },
+      truncateNumbersWithPound: false,
+      defaultNotebookSettings: {
+        isMinimapEnabled: false,
+      },
+    };
+  }
+
+  static async makeWorkspaceData(
```
mofojed, line 60: We're still doing this incorrectly - this `LocalWorkspaceStorage.makeWorkspaceData` should _not_ take in the `serverConfigValues`. Just provide an empty object for `settings` here (e.g. no default overrides).
The `makeDefaultWorkspaceSettings` method above _should_ take the serverConfigValues, and populate the settings there if the settings if the server overrides the settings. Whatever the server provides are the "defaults".

mofojed (changes requested):
Still a few issues with functionality, I think I've pointed out all the problematic areas but here is a test case that I think should be comprehensive to illustrate what the issue is. This assumes starting from an incognito tab, so no previous workspace data

1. Set `integerFormat=##000` in props
2. Login as user and check in the Settings menu
  - Integer format option should show `##000` and not have the "Reset Integer Format" button visible, since we are using the default value
3. Shut down the server, set `integerFormat=##111` in props
4. From the same browser as step 2 (so workspace data is intact), login and check the settings tab
  - Integer format option should show `##111` and not have the "Reset Integer Format" visible
5. Change the option to `##999`. Reset button should become visible
6. Shut down the server, set `integerFormat=##222` in props
7. Login and check settings tab
  - Integer format should show `##999` as you set in step 5. Reset button should be visible
8. Click the Reset button - the integer format should change to `##222` and reset button should disappear.

You can also inspect what you have set in Redux using the Redux dev tools, and what's stored in LocalStorage using the browser dev tools. You want to make sure the default settings in redux are populated with the settings read from the server (anything not defined by the server should fallback to whatever previous defaults we have set right now), and workspace storage settings should only have the options the user has _changed_.

packages/code-studio/src/storage/LocalWorkspaceStorage.ts
Zhou-Ziheng: Yeah think I was trying to keep our default and put the server config defaults along with the user set values. What I should've done is override our defaults with server config defaults, and then keep the user set values separately.

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -114,7 +100,7 @@ export class FormattingSectionContent extends PureComponent<
       this.commitChanges.bind(this),
```
mofojed, line 100: Above - you no longer need the `defaults` set in the `defaultProps`, since it's passed in via redux

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -40,44 +39,31 @@ import {
...
+  defaults: WorkspaceSettings;
 }
 
 interface FormattingSectionContentState {
-  formatSettings: FormatterItem[];
   showTimeZone: boolean;
   showTSeparator: boolean;
   timeZone: string;
-  defaultDateTimeFormat: string;
+  defaultDateTimeFormat?: string;
```
mofojed, line 66: Unsure why this is optional, it shouldn't be as we always have a default provided.
```suggestion
  defaultDateTimeFormat: string;
```

packages/dashboard-core-plugins/src/linker/Linker.tsx
```diff
@@ -55,7 +55,7 @@ interface StateProps {
   dh: DhType;
   isolatedLinkerPanelId?: string;
   links: Link[];
-  timeZone: string;
+  timeZone?: string;
```
mofojed, line 58: Shouldn't be optional
```suggestion
  timeZone: string;
```

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -60,7 +60,7 @@ interface ConsolePanelProps extends DashboardPanelProps {
 
   sessionWrapper: SessionWrapper;
 
-  timeZone: string;
+  timeZone?: string;
```
mofojed, line 63:
```suggestion
  timeZone: string;
```

packages/redux/src/actions.ts
```diff
@@ -41,11 +41,22 @@ export const setApi: PayloadActionCreator<DhType> = api => ({
...
   payload: workspace,
 });
 
+export const setDefaultWorkspaceSettings: PayloadActionCreator<
+  WorkspaceSettings
+> = settings => ({
+  type: SET_DEFAULT_WORKSPACE_SETTINGS,
+  payload: {
+    ...settings,
+  },
```
mofojed, line 57: Pretty sure you don't need the spread here:
```suggestion
  payload: settings,
```

packages/redux/src/reducers/defaultWorkspaceSettings.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * The Server Configs Values retrieved from the Server
```
mofojed, line 2: A little pedantic, but at this layer we should just say these are the default settings. In our particular implementation, some of them are read in from the server, and some of them fallback to the hardcoded defaults we have in the Web UI. 
```suggestion
 * Default values for workspace settings. Used when the user has not modified a setting yet.
```

mofojed (approved):
Some minor cleanup, but looks great! Thanks

mofojed (approved):

PR #1560, opened 2023-10-05
feat: Monaco theming

### LINKED ISSUE #1542: Theming - Monaco
Update Monaco theming to use theme variables. Requires deriving CSS prop values and converting to .js based theme.

Epic: #1537

bmingles:
- Base color palette now uses hsl colors with a single `--dh-color-xxx-hue` variable defining each of the base colors (see not in testing section below)
- Updated Monaco theme to consume --dh-color variables
- Refined default dark theme a bit with more semantic variables mostly in support of Monaco.
- Utils for normalizing colors + deriving css variables

**Methodology for Resolving Css Variables and Normalizing Colors**
ThemeUtils.resolveCssVariablesInRecord(record, targetEl)
- Add tmp div to target element for resolving variables
- Extract distinct `var` expressions from values in given record
- For each distinct `var` expression, create a tmp css variable on the temp div
- Call getComputedStyle() on the tmp div
- For each value on the given record
  - Replace `var` expressions with resolved values. Values are resolved using `getPropertyValue()` on the result of `getComputedStyle()` in previous step.
  - Normalize any color values to 8 character hex
    - Convert to rgb/a - ColorUtils.asRgbOrRgbaString() - create a non-attached div element, set its background color then read the background color
    - Parse rgb/a and convert to hex with math

**Testing**
The MonacoTheme object now gets passed through a normalization function. You can see the before and after in the debug logs "Monaco theme:" and "Monaco theme derived:". The inputs are various `var(--dh-` expressions`, and the outputs should all be 8 character hex values.

> Note: The new base color palette is hsl based derived from original hex colors Don created in his theming work. A mapping of the original hex to the new hsl version can be found here:
> https://bmingles.github.io/deephaven-theming-spike/
> 

resolves #1542

BREAKING CHANGE: Theme variables have to be present on body to avoid Monaco init failing


packages/components/src/theme/theme_default_dark_semantic.css
```diff
@@ -0,0 +1,72 @@
+/* stylelint-disable alpha-value-notation */
+:root {
+  /* General */
+  --dh-color-black: var(--dh-color-gray-50);
```
bmingles, line 4: Move black and white to base color palette

packages/components/src/theme/theme_default_dark_semantic.css
```diff
@@ -0,0 +1,72 @@
...
+  --dh-color-accent: var(--dh-color-blue-700);
+  --dh-color-border: var(--dh-color-gray-500);
+  --dh-color-background: var(--dh-color-black);
+  --dh-color-foreground: var(--dh-color-white);
+  --dh-color-grid-background: var(--dh-color-background);
+  --dh-color-content-background: var(--dh-color-gray-100);
+
+  /* Text */
+  --dh-color-text: var(--dh-color-gray-800);
+  --dh-color-text-highlight: rgba(76 125 238 / 0.3);
```
bmingles, line 15: Base theme shouldn't have any hard coded colors. These should be derived from the base color palette. Also should make it easy to change the base color they are based on. e.g. A client may prefer a green based highlight over blue based.

packages/components/src/theme/theme_default_light.css
```diff
@@ -47,8 +47,8 @@
...
   --dh-color-black: var(--dh-color-gray-50);
   --dh-color-white: var(--dh-color-gray-75);
-  --dh-accent-color: var(--dh-color-blue-700);
-  --dh-background-color: var(--dh-color-white);
-  --dh-foreground-color: var(--dh-color-black);
-  --dh-grid-background-color: var(--dh-background-color);
+  --dh-color-accent: var(--dh-color-blue-700);
+  --dh-color-background: var(--dh-color-white);
+  --dh-color-foreground: var(--dh-color-black);
+  --dh-color-grid-background: var(--dh-color-background);
```
bmingles, line 53: This probably belongs in a grouping for Iris Grid related semantic colors

packages/console/src/monaco/MonacoTheme.module.scss
```diff
@@ -3,71 +3,72 @@
...
-  keyword: $blue;
-  storage: $red;
-  number: $purple;
-  operator: $red;
-  identifier: $gray-100;
-  namespace-identifier: $red;
-  identifier-js: $yellow;
-  comment: mix($gray-500, $gray-400, 40%);
+  //code ruless
+  string: var(--dh-color-visual-yellow);
```
bmingles, line 12: Need to introduce semantic base variables for things like this. MonacoTheme.module.scss shoudn't be hard tied to things like "yellow" since that may differ from theme to theme.

mofojed:
@bmingles I should add some more context to this quote on the [CONTRIBUTING.md](https://github.com/deephaven/web-client-ui/blob/main/CONTRIBUTING.md):
> NOTE: Do not use the ! notation for marking a breaking change - you must use the BREAKING CHANGE: footer and include details of the breakage/migration.
At the time I didn't think lerna was supporting this correctly, but upon re-reading [this comment](https://github.com/lerna/lerna/issues/2668#issuecomment-708632441) about the issue, we just need to set the changelog-preset: https://github.com/lerna/lerna/tree/main/libs/commands/version#--changelog-preset
I'll submit a PR to update that. That being said - you put `BREAKING_CHANGE` in your description, it should be `BREAKING CHANGE` (with a space, no underscore), so please update taht.

mofojed:
@bmingles scratch that, I tried making a PR to use the correct preset: https://github.com/deephaven/web-client-ui/pull/1565
But that just resulted in the `version` command hanging. I'm not sure why it's hanging. Rather than dive into it, I've filed a ticket for cleaning that up later: https://github.com/deephaven/web-client-ui/issues/1566
Just remove the exclamation mark from the PR title instead, the BREAKING CHANGE footer is sufficient.

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -119,6 +142,162 @@ export function getThemePreloadData(): ThemePreloadData | null {
...
+    result[key as keyof T] = resolveCssVariablesInString(
+      resolver,
+      value
+    ) as T[keyof T];
+  });
+
+  // Remove the temporary css variables
+  tmpPropEl.remove();
+
+  log.debug('Resolved css variables', performance.now() - perfStart, 'ms');
```
mofojed, line 265: Side note - perhaps we should add a `.time` method to our Logger, that wraps `console.time`: https://developer.mozilla.org/en-US/docs/Web/API/console/time
I would put that at a finer grain of detail than `debug2` even. Just noting it.

packages/utils/src/ColorUtils.ts
```diff
@@ -32,5 +48,93 @@ class ColorUtils {
...
+    }
+
+    const [r, g, b, a = 1] = tokens.map(Number);
+
+    return {
+      r,
+      g,
+      b,
+      a,
+    };
```
mofojed, line 108: I'd much prefer without the custom line breaks here
```suggestion
    return { r, g, b, a };
```

mofojed (approved):
Minor cleanup suggestion.

packages/utils/src/ColorUtils.ts
bmingles: ah yeah. I'll fix on next PR

PR #1563, opened 2023-10-10
fix: Change display of rollup key columns from null to empty string

### LINKED ISSUE #1483: Rollup grouping columns shouldn't show as 'null'
The key columns for a rollup in the UI shouldn't show `null` for a rolled-up key — it would be better to show completely blank cells or details of the key columns (e.g. a count of values/the themselves if they'd fit).

image

We could also allow right-clicking to show all the values (or all of the complete keys).

georgecwan:
* Previously, key columns containing non-numeric types would display `null` is the rolled up rows have different values for the same column
    * This behaviour is updated to show an empty cell instead of `null`
* Resolves #1483 

#### Testing Instructions:
1. Run:
  ```python
  
  size = 100
  scale = 1000
  
  random_values_with_null_NaN = empty_table(size).update([
      "MyString=(i%11==0? null : `a`+(int)(scale*(Math.random()*2-1)))"
  ])
```
2. Add the string and char columns and any other arbitrary columns as rollup columns in any order
3. The key column should only show `null` if it actually represents a null value in the column

### CODE REVIEW

File: packages/iris-grid/src/IrisGridTreeTableModel.ts [hunk 1]
```diff
@@ -33,6 +33,13 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
         const value = this.valueForCell(x, y);
         return this.displayString(value, column.constituentType, column.name);
       }
+      if (
+        row.hasChildren &&
+        row.depth <= x + 1 &&
+        this.table.groupedColumns.includes(column)
```
mofojed, line 39: This is unlikely to cause performance issues in this particular scenario since there's not too many cells that will realistically fall into this block, but right now this call will need to iterate through the `groupedColumns` array looking for `column` each time you get in here. If they had thousands of grouped columns (which, realistically is extremely unlikely, but we need to be prepared for anyway), this could cause performance issues.
We should be caching the `groupedColumns` array as a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) and using `.has` on that instead.
See `getCachedFilterableColumnSet` for an example, but just make something like `getCachedGroupedColumnSet` that returns the memoized column groups and call .has on that.

Also - to be double paranoid, I would also get the value of the cell `const value = this.valueForCell(x, y)` and only return the empty string if it is indeed `null`. And include a comment in this block about why we are doing this (can reference the issue number if you want), as this is a strange thing to do without more context.

mofojed (changes requested):

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -182,6 +191,15 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
       )
   );
 
+  getCachedGroupedColumnSet = memoize(
+    (groupedColumns: Column[]) =>
```
mofojed, line 195: Can be explicit about this being a `readonly` array:
```suggestion
    (groupedColumns: readonly Column[]) =>
```

mofojed (approved):

mofojed (approved):

PR #1564, opened 2023-10-10
feat: Widget plugins

### LINKED ISSUE #1167: Allow dashboard plugins to advertise what type of objects they support
As a user, I want plugins to be able to advertise what objects they support opening so that the UI can be smart about display those object types. Should be able to include a friendly name of the object and icon as well (for descriptions).

Currently, if there's an unknown object type exported (such as `PartitionedTable`), it appears as a button like other objects but just doesn't do anything: https://github.com/deephaven/web-client-ui/issues/1143

### LINKED ISSUE #1455: Add support for WidgetPlugins
Add a `WidgetPlugin`. This is a simplified version of basically everything from the js-plugins repo. The plugin author specifies a component, types supported, an optional icon, and then we register it to listen for panel open events matching the supported types. This should also wrap the component in our panel wrapper

mattrunyon:
Fixes #1455
Fixes #1167 by allowing widget plugins to advertise their supported types. Dashboard plugins don't make as much sense supporting specific types and all of our existing dashboard plugins that open widgets from the server could be converted to widget plugins.

I debated making `supportedTypes` have a type of `string | { type: string, icon: Icon }` (+ array version), but felt it complicated things unnecessarily for our current types/icons. We can always add it in the future, but for now each widget plugin could support multiple types, but have 1 icon for the console.

I converted the pandas plugin as a test of widget plugins. Confirmed that if I create a dataframe and apply a sort/filter those persist across page refreshes.

Example code to make a dataframe

```python
from deephaven.pandas import to_pandas
t = empty_table(40000).update(["Timestamp=epochMillisToInstant(i)","x=(double)i", "y=i%12==11?null:Math.sin(i)"])
df = to_pandas(t)
```

Still needed
- [x] Wrapping in panel component
- [x] Config option to not automatically wrap


packages/console/src/console-history/ConsoleHistory.tsx
```diff
@@ -13,43 +13,45 @@ interface ConsoleHistoryProps {
...
+        supportsType={supportsType}
+        iconForType={iconForType}
+      />
     );
+    historyElements.push(historyElement);
   }
+
+  return (
+    <div className="container-fluid console-history">{historyElements}</div>
+  );
```
mofojed, line 54: Not that this was you, and no need to make this change, but now I'd prefer to write this with a map:
```
  return (
    <div className="container-fluid console-history">
      {items.map((item, i) => (
        <ConsoleHistoryItem
          key={itemKey(i, item)}
          disabled={disabled}
          item={item}
          openObject={openObject}
          language={language}
          supportsType={supportsType}
          iconForType={iconForType}
        />
      ))}
    </div>
  );
```

packages/console/src/console-history/ConsoleHistoryItem.tsx
```diff
@@ -18,8 +18,10 @@ const log = Log.module('ConsoleHistoryItem');
 interface ConsoleHistoryItemProps {
   item: ConsoleHistoryActionItem;
   language: string;
-  openObject: (object: VariableDefinition) => void;
+  openObject(object: VariableDefinition): void;
   disabled?: boolean;
+  supportsType(type: string): boolean;
+  iconForType(type: string): ReactElement;
```
mofojed, line 24: I prefer the other format for TS function definitions, and what was there previously:
```
openObject: (object: VariableDefinition) => void;
```
Are we using the other format anywhere else? Figure we should be consistent, unsure about a linter rule for this.

packages/console/src/console-history/ConsoleHistoryItem.tsx
```diff
@@ -77,28 +79,30 @@ class ConsoleHistoryItem extends PureComponent<
...
-              disabled={btnDisabled}
-              icon={<ObjectIcon type={object.type} />}
-            >
-              {title}
-            </Button>
-          );
-          resultElements.push(element);
-        });
+        [...created, ...updated]
+          // .filter(object => supportsType(object.type))
```
mofojed, line 83: Guessing we'll enable this once all plugins are converted? Should put a TODO then

packages/dashboard-core-plugins/src/PandasPlugin.tsx
```diff
@@ -1,28 +1,22 @@
...
+    );
 
-  useDashboardPanel({
-    dashboardProps: props,
-    componentName: PandasPanel.COMPONENT,
-    component: PandasPanel,
-    supportedTypes: dh.VariableType.PANDAS,
-    hydrate,
-  });
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
mofojed, line 15: We should probably just disable this rule globally. We spread props all the time.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
```diff
@@ -0,0 +1,94 @@
...
+  DehydratedDashboardPanelProps,
+  PanelEvent,
+  PanelOpenEventDetail,
+  LayoutUtils,
+  useListener,
+} from '@deephaven/dashboard';
+import { usePlugins } from '@deephaven/app-utils';
+import { isWidgetPlugin } from '@deephaven/plugin';
+
+export function WidgetLoaderPlugin(
```
mofojed, line 16: JSDocs would be handy for this component. It opens the widget plugin that supports the object when it is opened.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const typeMap = new Map<string, ComponentType>();
+    plugins.forEach(plugin => {
+      if (!isWidgetPlugin(plugin)) {
+        return;
+      }
+
+      [plugin.supportedTypes].flat().forEach(supportedType => {
+        if (supportedType != null && supportedType !== '') {
+          typeMap.set(supportedType, plugin.component);
+        }
```
mofojed, line 30: Should we warn if there's already a type set for that type? Which plugin should take priority? Or should we allow multiple plugins to register and open panels in each of them (similar to JSON object viewer today)? We should be explicit about what we want the behaviour to be here, as right now it won't log anything and the last plugin registered will win (which may be fine).

I think I would probably stick with the behaviour you have, just add a comment/log about when it's overwriting an existing type. But you can have the final call.

Side note - should we allow patterns/regexes? I suppose doesn't have to be now, but the JSON object viewer plugin won't work with this. Which, again, is fine.

mofojed (changes requested):

packages/console/src/console-history/ConsoleHistoryItem.tsx
mattrunyon: I think I've picked the shorthand in other spots. Turns out there is a slight difference in that TS is more strict on the property version (what you prefer) due to how it handles variance.

Here's the linter rule. Want me to make a ticket to enable it?

https://typescript-eslint.io/rules/method-signature-style/

packages/dashboard-core-plugins/src/PandasPlugin.tsx
mattrunyon: Idk I think I agree with the rule of being explicit usually. We mostly use it when we're just blindly passing through props and there's no alternative.

I think TS checks spread props in JSX now, but not sure if it checks for excess props.

Here's the rule. Can also open a ticket to disable this if you think we should.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -140,13 +141,19 @@ export function DashboardLayout({
             CType.WrappedComponent.prototype.isReactComponent != null) ||
           (CType.prototype != null && CType.prototype.isReactComponent != null);
 
+        // For some unknown reason, ReactIs.isForwardRef returns false, but comparing $$typeof directly works
+        const isForwardRef =
+          !isWrappedComponent(CType) && CType.$$typeof === ReactIs.ForwardRef;
```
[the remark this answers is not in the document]
mattrunyon: Ahh good catch. I think I misread indent levels when looking at that code.

I'm going to make a util for detecting if we should forward ref because we will be wrapping components in 2 places. Might be able to make a util to just create the wrapped component actually.

packages/console/src/console-history/ConsoleHistoryItem.tsx
mattrunyon: Yep. Didn't want to break existing dashboard plugins and the more I think about it, dashboard plugins shouldn't have supported types. They're meant to be mounted in the react tree on any dashboard. If you had dashboard plugin supporting tables, you wouldn't be able to fetch the actual grid component to render the table as part of a panel with multiple components.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
mattrunyon: Ya I've been thinking about that. I'm leaning towards something that only lets one plugin handle a widget. What would we expect to happen in Deephaven UI if there were 2 plugins that claim to handle a grid and we need to pick one?

The JSON object viewer can still work as a dashboard plugin with this.

I figure keep it simple w/ just strings for now and if there's a compelling use case for regex later we can add it.

I added a warning if 2 plugins register to handle the same widget type and the last plugin registered will be the plugin that handles the widget.

packages/dashboard-core-plugins/src/PandasPlugin.tsx
mofojed: Eh let's not worry about it for now.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
mofojed: Okay sounds good.

packages/jsapi-types/src/dh.types.ts
```diff
@@ -82,6 +82,20 @@ export interface VariableDefinition<T extends string = string> {
   id?: string;
 }
 
+export interface JsWidget extends Evented {
+  getDataAsBase64: () => string;
+  getDataAsU8: () => string;
```
mofojed, line 87: Should be Uint8Array.
Either way we should be replacing this with the automatic typings, so let's not worry too much about the manual typings right now as long as it's working correctly.

packages/plugin/src/PluginTypes.ts
```diff
@@ -1,4 +1,9 @@
 import type { BaseThemeType } from '@deephaven/components';
+import { type JsWidget } from '@deephaven/jsapi-types';
```
mofojed, line 2: I thought it had to be `import type { JsWidget } from '@deephaven/jsapi-types';`. But I guess this works too? Neat. But also inconsistent with the line immediately before it.

packages/console/src/console-history/ConsoleHistoryItem.tsx
mattrunyon: #1578

packages/jsapi-types/src/dh.types.ts
mattrunyon: Good point. I don't even use the U8, but figured I'd add it in case we do use it before we get the automatic typings

packages/plugin/src/PluginTypes.ts
mattrunyon: Yep at some point this became valid syntax (looks like TS 4.5 added the inline types). It lets you combine type and non-type named imports on the same line. In some cases we have

```ts
import { a } from 'abc';
import type { b } from 'abc';
```

which can now be

```ts
import { a, type b } from 'abc';
```

There's a few related rules. Should I create 2 tickets for them?
1 for [import type specifier](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md) and 1 for [consistent imports and exports](https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/#enforcing-with-typescript-eslint)? The 2nd will likely require more touched files, but I think it's auto-fixable

mofojed: Sure

mofojed (approved):

mofojed (approved):

PR #1567, opened 2023-10-10
feat: web-client-ui changes required for deephaven.ui

mofojed:
- Add some more shared packages to the `remote-component.config.ts`
  - Now plugins can externalize these packages in their build process, and share the same instance
- Add Chart handling of model update
- Add some more externalized packages
- Add max-old-space-size flag to embed-chart
  - Noticed it was missing, and was causing a build issue with these changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "clean": "run-s clean:*",
     "types": "tsc --build",
     "watch:types": "tsc --build --watch --preserveWatchOutput",
-    "build": "run-s build:necessary types build:packages build:app build:embed-grid build:embed-chart",
+    "build": "run-s build:necessary types build:packages build:apps",
     "build:apps": "lerna run --scope=@deephaven/{code-studio,embed-chart,embed-grid} build",
     "build:app": "lerna run --scope=@deephaven/code-studio build",
     "build:embed-chart": "lerna run --scope=@deephaven/embed-chart build",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "clean": "run-s clean:*",
     "types": "tsc --build",
     "watch:types": "tsc --build --watch --preserveWatchOutput",
-    "build": "run-s build:necessary types build:packages build:apps",
+    "build": "run-s build:necessary types build:packages build:app build:embed-grid build:embed-chart",
     "build:apps": "lerna run --scope=@deephaven/{code-studio,embed-chart,embed-grid} build",
     "build:app": "lerna run --scope=@deephaven/code-studio build",
     "build:embed-chart": "lerna run --scope=@deephaven/embed-chart build",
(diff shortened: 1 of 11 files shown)


packages/jsapi-types/src/dh.types.ts
```diff
@@ -335,6 +336,17 @@ export interface Figure extends Evented {
   close(): void;
 }
 
+export interface Widget {
+  readonly EVENT_MESSAGE: string;
+
+  addEventListener: (
+    type: string,
+    listener: (event: unknown) => void
```
mattrunyon, line 344: Might want to make a separate interface for the message payload and also we should add another signature for the message event. There is also a close event which I believe is only fired on server sending a not ok status. We probably need another status/event to differentiate server hanging up vs server disconnecting due to something like network instability.

```
addEventListener: (
  type: 'message', // Or typeof EVENT_MESSAGE and set that to its value
  listener: (event: CustomEvent<Omit<Widget, 'addEventListener', 'sendMessage', 'EVENT_MESSAGE'>>) => void
)
```

packages/golden-layout/scss/goldenlayout-dark-theme.scss
```diff
@@ -70,7 +70,6 @@ body:not(.lm_dragging) .lm_header .lm_tab .lm_close_tab:hover {
 // Single Pane content (area in which final dragged content is contained)
 .lm_content {
   background: $content-bg;
-  overflow: visible;
```
mattrunyon, line 73: Looks like this is the only non-prettier change to this file. Why does this need to be removed?

mattrunyon (changes requested):

packages/golden-layout/scss/goldenlayout-dark-theme.scss
mofojed: Yea it's not really related to _this_ change so I'll back it out... basically two issues I have with this:
1. Specifically I had an overflowing panel, and that was causing the entire UI to get scroll bars
2. This is the `dark-theme.scss` file, and should really only contain stuff for a "dark" theme, whereas this is about the layout of things, which is relevant in the base theme. This kind of scss should just be in the base theme instead (if it's what we actually want), instead of being overridden in the dark theme.

Created a ticket to follow up on this.

packages/jsapi-types/src/dh.types.ts
mofojed: Yea, ultimately it extends the `HasEventHandling` interface... I don't really want to do too much here right now since I want to publish the automatic types and then use those, would like to see how that shakes out before re-writing this.

package.json
```diff
@@ -16,7 +16,6 @@
     "clean:cache": "run-s clean:cache:*",
     "clean:cache:eslint": "rimraf .eslintcache",
     "clean:cache:jest": "jest --clearCache",
-    "clean:cache:nx": "nx clear-cache",
```
mattrunyon, line 19: Odd that this is in this PR as it is already merge into main to remove this

mattrunyon (approved):

PR #1568, opened 2023-10-10
feat: Theming Iris Grid

bmingles:
- Created new base theme for Iris grid and mapped existing styles to it
- Added color palettes to styleguide
- IrisGrid now creates and resolves css variables when creating the default theme

**Testing**
- Easiest way to see the new color palettes is in the styleguide. 
- Iris Grid should look as it did before except for a few cases where the new color palette doesn't exactly align with the old one. aka. a color may have a subtle change in saturation but there shouldn't be any changes to the visual color.

Current mapping of Bootstrap gray palette to new `--dh-color-gray` vars based on Don's POC. The Bootstrap variables will get updated in a future PR to pull from the new vars, but this is how things line up right now:
image

BREAKING CHANGE: Enterprise will need ThemeProvider for the css variables to be available


mofojed (approved):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1337,16 +1337,25 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
-      rowFooterWidth: floatingRowCount > 0 ? theme.rowFooterWidth : 0,
-    }),
+      const rowFooterWidth =
+        floatingRowCount > 0
+          ? theme?.rowFooterWidth ?? defaultTheme.rowFooterWidth
+          : 0;
+
+      return {
+        ...defaultTheme,
+        ...theme,
```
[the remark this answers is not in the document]
bmingles: I had the same surprise

mofojed (changes requested):
Not sure why some of the grid tests are failing - looks like the colour isn't matching up. Double check that the correct colour is being used, if it is a new colour then update the snapshots (though I'm confused why `can open a simple table` passes but rollup/goto don't).

packages/iris-grid/src/IrisGridTheme.module.scss
```diff
@@ -12,92 +12,90 @@ $header-separator-color: $gray-900;
 $header-height: 30px;
 
 :export {
-  grid-bg: $gray-900;
+  grid-bg: var(--dh-color-grid-bg);
```
mofojed, line 15: Many of the variables above are no longer used (e.g. `$selection-color`).
Also the new selection color doesn't match the old color; I think that may be intentional? Confirm with @dsmmcken , but that's one reason the goto row tests are failing.

packages/components/src/theme/theme-dark/theme-dark-semantic-grid.css
```diff
@@ -1,3 +1,96 @@
...
+  /* Column */
+  --dh-color-grid-column-linker-hover-bg: var(
+    --dh-color-highlight-selected-hover
+  );
+  --dh-color-grid-column-grouped-divider: var(--dh-color-gray-50);
+
+  /* Tree */
+  --dh-color-grid-tree-line: var(--dh-color-gray-700);
+  --dh-color-grid-tree-marker: var(--dh-color-gray-800);
+  --dh-color-grid-tree-marker-hover: var(--dh-color-gray-900);
```
mofojed, line 76: These don't match what they were before:
```
  tree-line-color: $gray-500; 
  tree-marker-color: $gray-300;
  tree-marker-hover-color: $gray-100;
```

If that's intentional, then need to update snapshots. Seems odd we're going in the reverse on the gray scale though (the scale may not be the same anymore? Dunno).

mofojed (changes requested):

packages/components/src/theme/theme-dark/theme-dark-semantic-grid.css
bmingles: Spectrum numbers ascending 1-100 darkest to lightest. Here's how Don mapped the colors in his POC (left side old, right side new)
image

bmingles: I've updated the tree variables to be closer to the original for now. Seems to have fixed these tests

packages/iris-grid/src/IrisGridTheme.module.scss
bmingles: I deleted the unused vars. Will check with Don on the selection color diff.
image

mofojed (dismissed):

Issue #1477, opened 2023-09-01
isStuckToBottom behaviour when filtering


dsmmcken:
**Description**

Given a table and a scroll position at the top of the table, filtering to a result of less than a full viewport, then clearing the filter sets the table to isStuckToBottom (and further, only applies after a re-render). As a user this can be confusing and frustrating that you end up at the bottom of a table after the filter clears.

Although, this may be the desired behaviour if the user was already at the bottom. Open to heuristic suggestions for this behaviour. Maybe isStuckToBottom isn't triggered until on short viewports until it a tick actually goes past the end of the screen.

In comparison, if you are at the top, filter by a result that has a full viewport worth of results (example filter MyBoolean = true from below), then clear that filter you remain at the top. So behaviour is inconsistent.

**Steps to reproduce**

1. Create a simple table
```
from deephaven import empty_table

simple_static = empty_table(1000).update([
        "MyString=new String(`a`+(char) ((i%3)+97))",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i % 3)",
        "MyDouble=new Double(i+i/10)",
        "MyFloat=new Float(i+i/10)",
        "MyBoolean=new Boolean(i%2==0)",
        "MyChar= new Character((char) ((i%26)+97))",
        "MyShort=new Short(Integer.toString(i%32767))",
        "MyByte= new java.lang.Byte(Integer.toString(i%127))"])
```
2. From a scroll position at the top of the table, filter the table to 1 result (filter by MyInt = 0)
3. Clear the filter with a shortcut (ctrl/cmd+e)
4. Move the mouse anywhere over the table to trigger a re-render

**Expected results**

I feel like I would expect the table to be back at the top.

**Actual results**

The table is stuck to the bottom. (It doesn't move until after step 4, which is even weirder).

**Versions**

Engine Version: 0.27.0
Web UI Version: 0.46.0
Java Version: 17.0.7
Barrage Version: 0.6.0

mattrunyon:
The original fix for this made it impossible to get unstuck from the bottom. See #1615. Reverted in #1616

PR #1571, opened 2023-10-12
test: add e2e tests to context menu 

### LINKED ISSUE #1379: Context menu filter by e2e tests

ethanalvizo:
Closes #1379


packages/iris-grid/src/GotoRow.tsx
```diff
@@ -385,6 +385,7 @@ const GotoRow = forwardRef<GotoRowElement, GotoRowProps>(
 
                 <div>
                   <Button
+                    data-testid="increment-go-to"
```
mofojed, line 388: Instead of setting `data-testid`, set the `tooltip` and use that (it automatically sets the `aria-label` property if it's an icon only button). Also need to pass the icon as the `icon` prop instead of a child.:
```suggestion
<Button
  kind="ghost"
  icon={vsArrowUp}
  tooltip="Next match"
  disabled={gotoValue === ''}
  onClick={() => { onGotoValueSubmit(true): }}
  />
```

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -394,6 +395,7 @@ const GotoRow = forwardRef<GotoRowElement, GotoRowProps>(
                     <FontAwesomeIcon icon={vsArrowUp} />
                   </Button>
                   <Button
+                    data-testid="decrement-go-to"
```
mofojed, line 398: Same thing here.

mofojed (changes requested):
Not sure why the double context menu and advanced filter are failing... advanced filter looks like it may have gotten stuck to the bottom and then scrolled to the bottom of the grid somehow? Shouldn't happen if you're updating a filter, it should reset to the top...
And the double context menu, maybe just rebase off latest main?

ethanalvizo:
> Not sure why the double context menu and advanced filter are failing... advanced filter looks like it may have gotten stuck to the bottom and then scrolled to the bottom of the grid somehow? Shouldn't happen if you're updating a filter, it should reset to the top... And the double context menu, maybe just rebase off latest main?

Yea it was pretty funky. There were two errors coming from the tests

1. Iris grid scrolled down to the bottom every once in a while -> I created a function that scrolls up in the IrisGrid to circumvent this
2. Console input couldn't be found when a context menu was open -> I made sure to click somewhere else in the UI to close the context menu before ending the test

dsmmcken:
> Iris grid scrolled down to the bottom every once in a while -> I created a function that scrolls up in the IrisGrid to circumvent this

https://github.com/deephaven/web-client-ui/issues/1477 ?

packages/grid/src/Grid.tsx
```diff
@@ -1323,14 +1323,34 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+        eventType.key === 'PageDown';
+    }
     if (top != null) {
-      this.setState({ isStuckToBottom: isStickyBottom && top >= lastTop });
+      this.setState({
+        isStuckToBottom: isStickyBottom && top >= lastTop && isUserInputDown,
+      });
     }
     if (left != null) {
       this.setState({ isStuckToRight: isStickyRight && left >= lastLeft });
```
mofojed, line 1356: We should be mirroring the same logic for stuck to right. We don't have tables that grow by columns right now, but with pivot tables (coming at some point) it could happen.

packages/grid/src/Grid.tsx
```diff
@@ -1323,14 +1323,34 @@ class Grid extends PureComponent<GridProps, GridState> {
    * @param viewState New state properties to set.
    * @param forceUpdate Whether to force an update.
    */
-  setViewState(viewState: Partial<GridState>, forceUpdate = false): void {
+
+  setViewState(
+    viewState: Partial<GridState>,
+    forceUpdate = false,
+    eventType: WheelEvent | GridKeyboardEvent | null = null
```
mofojed, line 1330: Just name it `event`, since you're passing a whole event, not just the type.
Also would just mark it optional:
```
eventType?: WheelEvent | GridKeyboardEvent
```

packages/grid/src/Grid.tsx
```diff
@@ -1323,14 +1323,34 @@ class Grid extends PureComponent<GridProps, GridState> {
    * @param viewState New state properties to set.
    * @param forceUpdate Whether to force an update.
```
mofojed, line 1324: Update JS docs with the `event` parameter.

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -194,9 +194,12 @@ class SelectionKeyHandler extends KeyHandler {
 
         grid.moveCursorToPosition(top, cursorColumn, isShiftKey, false);
 
-        grid.setViewState({ top });
+        grid.setViewState({ top }, false, event);
       } else {
+        const { lastLeft, lastTop } = grid.metrics;
+
         grid.moveCursor(deltaColumn, deltaRow, isShiftKey);
+        grid.setViewState({ left: lastLeft, top: lastTop }, false, event); // need to call setViewState again since moveCursor calls a series of functions that sets view state
```
mofojed, line 202: `lastLeft` and `lastTop` is the furthest point that can be scrolled - with this change trying to navigate with the keyboard always jumps to the bottom.
We're probably just going to have to pass the event through `grid.moveCursor` and the other methods it has to go through, since we don't actually know where the cursor/viewport ends up until the state is set (that's done in `grid.moveViewToCell`).

tests/context-menu.spec.ts-snapshots/advanced-filters-2-chromium-linux.png
mofojed, line 1: There is a row hover in the snapshot here... I think that's fine, but just noting it. This row is highlighted only because it was underneath where the Done button was, and if that button moves, then this snapshot will fail.
It would be better to move the mouse to the top left of the grid or something before each snapshot to make sure we're not hovering over anything unexpectedly.

tests/context-menu.spec.ts-snapshots/advanced-filters-3-webkit-linux.png
mofojed, line 1: Yea you can see in this webkit snapshot vs the chromium/firefox one that there is a hover here, but not in the other two. I don't like the inconsistency, just move the mouse out of the way after clicking Done.

packages/grid/src/Grid.tsx
mofojed: You'll also need to handle a `GridMouseEvent` and wire up `GridVerticalScrollBarMouseHandler.onDrag` and `GridHorizontalScrollBarMouseHandler.onDrag` methods to call the appropriate `setViewState` options.

mofojed (changes requested):
Pull the changes out for the isStuckToBottom fix #1477 into a separate PR and merge that first.

packages/grid/src/Grid.tsx
ethanalvizo: I left this part out intentionally. I thought it would be weird for it to stick to the bottom if the user is dragging the scrollbar to a specific position.

mofojed: That makes sense.

packages/grid/src/Grid.tsx
ethanalvizo: for checking if user input is right I would only be checking for the right arrow key right?

tests/context-menu.spec.ts
```diff
@@ -1,4 +1,262 @@
...
+} from './utils';
+
+test.describe.configure({ mode: 'serial' });
+
+async function openAdvancedFilters(page: Page) {
+  await page
+    .locator('.iris-grid .grid-wrapper')
+    .click({ button: 'right', position: { x: 20, y: 20 } });
+  await page.getByRole('button', { name: 'Advanced Filters' }).click();
+}
```
mofojed, line 17: This method does open advanced filters, but it just seems to be doing just the first column of the first grid (e.g. you can't pass in a specific column or even a specific grid to open the advanced filter on). I'd rather be more specific with the name of the function (e.g. `openFirstAdvancedFilters`), or allow passing in the grid locator and an optional offset, e.g.:
```
/**
 * Open advanced filters for a grid.
 * @param gridLocator Locator to specify which grid to open the advanced filters on
 * @param xOffset Offset of the column to open advanced filters for. Defaults to the first column.
 */
async function openAdvancedFilters(gridLocator: Locator, xOffset = 20) 
  ...
}
```

tests/context-menu.spec.ts
```diff
@@ -1,4 +1,262 @@
...
+
+async function moveMouseAwayFromTable(page: Page) {
+  await page.mouse.move(0, 0);
+  await page.mouse.click(0, 0);
+}
```
mofojed, line 22: Similarly with this method - you're using it to move the mouse away from the table, but the actual function is just trying to reset the mouse position and focus. I'd probably do something like:
```
async function resetMousePosition(page: Page, resetFocus = true) {
  await page.mouse.move(0, 0);
  if (resetFocus) {
    await page.mouse.click(0, 0);
  }
}
```

tests/context-menu.spec.ts
```diff
@@ -1,4 +1,262 @@
...
+  // Model is loaded, need to make sure table data is also loaded
+  await waitForLoadingDone(page);
+
+  const tableOperationsMenu = page.locator(
+    'data-testid=btn-iris-grid-settings-button-table'
+  );
+  await tableOperationsMenu.click();
+
+  // Wait for Table Options menu to show
+  await expect(page.locator('.table-sidebar')).toHaveCount(1);
```
mofojed, line 52: This is exactly the same setup as in `table-operations.spec.ts`. We could refactor this into a utility method used by both, but I think in this case I would rather just move these table specific context menu tests to `table-operations.spec.ts` (since we're really testing the table operations here, we're just accessing them from the context menu).

mofojed (approved):
Looks good overall, some suggestions for possible cleanup/renaming of methods.

PR #1579, opened 2023-10-18
fix: stuck to bottom on filter clear

### LINKED ISSUE #1477: isStuckToBottom behaviour when filtering

ethanalvizo:
Closes #1477 

Previous comments are in #1571 PR. 

Note: Once this PR is merged #1571 can go in as well

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1181,11 +1181,13 @@ class Grid extends PureComponent<GridProps, GridState> {
    * @param deltaColumn Number of columns to move the cursor
    * @param deltaRow Number of rows to move the cursor
    * @param extendSelection True if the current selection should be extended, false to start a new selection
+   * @param event The event that triggered the update.
    */
   moveCursor(
     deltaColumn: number,
     deltaRow: number,
-    extendSelection: boolean
+    extendSelection: boolean,
+    event?: GridKeyboardEvent
   ): void {
     const { cursorRow, cursorColumn, selectionEndColumn, selectionEndRow } =
       this.state;
@@ -1193,14 +1195,14 @@ class Grid extends PureComponent<GridProps, GridState> {
     const row = extendSelection ? selectionEndRow : cursorRow;
     if (row === null || column === null) {
       const { left, top } = this.state;
-      this.moveCursorToPosition(left, top, extendSelection);
+      this.moveCursorToPosition(left, top, extendSelection, true, false, event);
     } else {
       const { model } = this.props;
       const { columnCount, rowCount } = model;
 
       const left = clamp(column + deltaColumn, 0, columnCount - 1);
       const top = clamp(row + deltaRow, 0, rowCount - 1);
-      this.moveCursorToPosition(left, top, extendSelection);
+      this.moveCursorToPosition(left, top, extendSelection, true, false, event);
     }
   }
 
@@ -1260,13 +1262,15 @@ class Grid extends PureComponent<GridProps, GridState> {
    * @param extendSelection Whether to extend the current selection (eg. holding Shift)
    * @param keepCursorInView Whether to move the viewport so that the cursor is in view
    * @param maximizePreviousRange With this and `extendSelection` true, it will maximize/add to the previous range only, ignoring where the selection was started
+   * @param event The event that triggered the update.
    */
   moveCursorToPosition(
     column: GridRangeIndex,
     row: GridRangeIndex,
     extendSelection = false,
     keepCursorInView = true,
-    maximizePreviousRange = false
+    maximizePreviousRange = false,
+    event?: GridKeyboardEvent
   ): void {
     if (!extendSelection) {
       this.beginSelection(column, row);
@@ -1275,7 +1279,7 @@ class Grid extends PureComponent<GridProps, GridState> {
     this.moveSelection(column, row, extendSelection, maximizePreviousRange);
 
     if (keepCursorInView) {
-      this.moveViewToCell(column, row);
+      this.moveViewToCell(column, row, event);
     }
   }
 
@@ -1284,8 +1288,13 @@ class Grid extends PureComponent<GridProps, GridState> {
    *
    * @param column The column index to bring into view
    * @param row The row index to bring into view
+   * @param event The event that triggered the update.
    */
-  moveViewToCell(column: GridRangeIndex, row: GridRangeIndex): void {
+  moveViewToCell(
+    column: GridRangeIndex,
+    row: GridRangeIndex,
+    event?: GridKeyboardEvent
+  ): void {
     if (!this.metrics) throw new Error('metrics not set');
 
     const { metricCalculator } = this;
@@ -1314,7 +1323,7 @@ class Grid extends PureComponent<GridProps, GridState> {
       }
     }
 
-    this.setViewState({ top, left, topOffset, leftOffset });
+    this.setViewState({ top, left, topOffset, leftOffset }, false, event);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -1314,26 +1323,52 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+    } else if (
+      event instanceof KeyboardEvent ||
+      (event && 'nativeEvent' in event) // used to catch the case that a synthetic react keyboard event is passed
+    ) {
+      isUserInputDown =
+        event.key === 'ArrowDown' ||
+        event.key === 'End' ||
+        event.key === 'PageDown';
+      isUserInputRight = event.key === 'ArrowRight';
+    }
```
mofojed, line 1362: I know I said to do these checks here in the last review - but this logic doesn't really belong here. `SelectionKeyHandler` has all the logic for which keys map to which direction, and arguably the logic from `handleWheel` for scrolling should be extracted to it's own mouse handler.

The reason I mention this though is because there are cases that aren't handled here - for example, if you press Shift+End, it will go to the far right instead of the bottom.

So instead of taking an event and passing it through, we really should just be passing in an option, e.g.
```
type StickyOptions = { shouldStickRight: boolean; shouldStickBottom: boolean };
setViewState(..., stickyOptions?: StickyOptions)
```

mofojed (changes requested):

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -279,7 +286,15 @@ class SelectionKeyHandler extends KeyHandler {
...
       false
     );
-    grid.setViewState({ top: viewportPosition });
+
+    const stickyOptions: StickyOptions = {
+      shouldStickBottom:
+        event.key === 'ArrowDown' ||
+        event.key === 'End' ||
+        event.key === 'PageDown',
+      shouldStickRight: event.key === 'ArrowRight',
```
mofojed, line 295: We shouldn't be doing these checks again - we're already in the `handlePageDown` method, which is triggered only on a keyboard event when navigating page down. Just do `stickOptions = { shouldStickBottom: true, shouldStickRight: false }`

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -177,6 +177,13 @@ class SelectionKeyHandler extends KeyHandler {
 
       const { theme } = grid.props;
       const { autoSelectRow = false, autoSelectColumn = false } = theme;
+      const stickyOptions: StickyOptions = {
+        shouldStickBottom:
+          event.key === 'ArrowDown' ||
+          event.key === 'End' ||
+          event.key === 'PageDown',
+        shouldStickRight: event.key === 'ArrowRight',
+      };
```
mofojed, line 186: Similar to my other comment - we shouldn't be checking the key itself again, just look at the `deltaColumn`/`deltaRow` to set the sticky options. Also need to pass it through in the above `grid.moveCursorToPosition` call I think.

packages/grid/src/Grid.tsx
```diff
@@ -1314,26 +1342,47 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+      });
     }
     if (left != null) {
-      this.setState({ isStuckToRight: isStickyRight && left >= lastLeft });
+      this.setState({
+        isStuckToRight:
+          isStickyRight &&
+          left >= lastLeft &&
+          (stickyOptions?.shouldStickRight ?? false),
+      });
```
mofojed, line 1385: Perhaps I named the sticky options incorrectly. I think we only want to update the options when sticky options are `true`, we shouldn't be updating `isStuckToRight` or `isStuckToBottom` at all if the values are false. So would be more like:

```
if (top != null && (stickOptions?.shouldStickToBottom ?? false)) {
  this.setState({ isStuckToBottom: isStickBottom && top >= lastTop });
}
```

As it is right now, you'll always wipe out the `isStuckToBottom`/`isStuckToRight` state whenever the cursor is moved, even if not a result of a user interaction.

mofojed (changes requested):

packages/grid/src/key-handlers/SelectionKeyHandler.ts
ethanalvizo: Ah yes I missed that one just the very first moveCursorToPosition the other ones didn't need it

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -169,21 +174,22 @@ class SelectionKeyHandler extends KeyHandler {
...
-          event.key === 'PageDown',
-        shouldStickRight: event.key === 'ArrowRight',
-      };
+      // const stickyOptions: StickyOptions = {
+      //   shouldStickBottom:
+      //     event.key === 'ArrowDown' ||
+      //     event.key === 'End' ||
+      //     event.key === 'PageDown',
+      //   shouldStickRight: event.key === 'ArrowRight',
+      // };
```
mofojed, line 192:
```suggestion
```

mofojed (changes requested):

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -169,14 +174,22 @@ class SelectionKeyHandler extends KeyHandler {
...
 
       const { theme } = grid.props;
       const { autoSelectRow = false, autoSelectColumn = false } = theme;
+      // const stickyOptions: StickyOptions = {
+      //   shouldStickBottom:
+      //     event.key === 'ArrowDown' ||
+      //     event.key === 'End' ||
+      //     event.key === 'PageDown',
+      //   shouldStickRight: event.key === 'ArrowRight',
+      // };
```
mofojed, line 192: Still need to delete this unused code

mofojed (changes requested):

packages/grid/src/key-handlers/SelectionKeyHandler.ts
ethanalvizo: Ah sorry about that I accidentally made that change on the e2e context menu branch rather than this one. Should be good to go now

mofojed (approved):

Issue #1143, opened 2022-06-14
UI for PartitionedTable


rcaudy:
As a user, it might be nice to display PartitionedTable objects in a friendly way. We could easily present a table grid of the non-constituent columns, for example, but that might be misleading to users without some visual cue that this isn't *just* a table, although it certainly is that.

@chipkent reports that we currently create a clickable icon, which may also be misleading.

mofojed:
Related to multiple partition selector: https://github.com/deephaven/web-client-ui/issues/1438

And different types: https://github.com/deephaven/web-client-ui/issues/1441

mofojed:
An example of a code snippet that produces a `PartitionedTable` (you must have `deephaven-plugin-plotly-express` installed):
```python
import deephaven.plot.express as dx
stocks = dx.data.stocks()
stocks_partitioned = stocks.partition_by(['sym', 'exchange'])
```

Just using the core deephaven libraries:
```python
from deephaven import empty_table
t = empty_table(10000).update(["x=(int)Math.floor(i/100)", "y=(int)Math.floor(i/10)", "z=i"])
pt = t.partition_by(["x", "y"])
```

The UI we're going to use is the [IrisGridPartitionSelector](https://github.com/deephaven/web-client-ui/blob/main/packages/iris-grid/src/IrisGridPartitionSelector.tsx), which will need support for multiple partitions added (#1438) and other types (#1441).

You can see the existing selector in action with parquet tables on #1414

dsmmcken:
UI spec:
![image](https://github.com/deephaven/web-client-ui/assets/1576283/d427b993-1d09-43a5-960f-4e6cd0848f36)

PR #1580, opened 2023-10-18
feat: Add support for multi-partition parquet:kv tables

### LINKED ISSUE #1438: Multi-Partitiondo not display multiple partition drop-downs
**Description**

Multi-Partition parquet:kv tables do not display multiple partition drop-downs

**Steps to reproduce**

1. place a .parquet file in a multi-partition kv directory. for example, "/path/to/somewhere/date_partition=2023-08-01/id_partition=id-001/my.parquet"
2. emulate multiple partitions via soft-link; from "/path/to/somewhere" directory, `ln -s date_partition=2023-08-01/id_partition=id-001 date_partition=2023-08-01/id_partition=id-002` and `ln -s date_partition=2023-08-01 date_partition=2023-08-02`. this will make it seem that there are 2 "date" partitions, each of which has 2 "id" partiions
3. import the table(s) in a query; `ParquetTools.readTable("/path/to/somewhere")
4. the "most recent" date-partion (2023-08-02) will be selected from the drop-down of available "date" partitions. note that there no is drop-down for the "id" partition

**Expected results**

4. a "date" partition and an "id" partition are visible as drop-downs. i guess this gets nasty if i want to select the "date=2023-08-01" and "ALL" of the "id" partitions ...

**Actual results**

3. there is no way to select a specific "id" 

**Versions**

 - Engine Version: ...
 - Web UI Version: ...
 - Java Version: ...
 - Barrage Version: ...
 - OS: ...
 - Browser: ...
 - Docker: ...

georgecwan:
Added feature to `IrisGridPartitionSelector` to display additional partition selector dropdowns when more than one partition column is present. The values displayed in the dropdown depend on the selected partitions of columns to the left of it.

The "Append Command" button and associated functions are removed in preparation for #1143. 

Closes #1438

#### Testing instructions:
1. Create and open a multi-partition table:
```python

write(part, "/tmp/pt-1438/x=0/y=0/part.parquet")
write(part, "/tmp/pt-1438/x=0/y=1/part.parquet")
write(part, "/tmp/pt-1438/x=1/y=0/part.parquet")
write(part, "/tmp/pt-1438/x=1/y=1/part.parquet")

partition_table = read("/tmp/pt-1438")
```
2. Use the dropdown to select any combination of partitions and check that it updates the table correctly.
    * The dropdown should only display values that are available after filtering for all the partitions left  of it. (E.g. A state/province dropdown should only contain the states/provinces corresponding to the selected country.)
3. Test filtering by typing directly in the dropdown and searching by using the search bar above the list of available partitions.
4. Retest using other data types for partition columns (dates, strings, chars, doubles). E.g:
```python

write(part, "/tmp/pt2-1438/date_partition=2023-08-01/id_partition=id-001/decimal=0.1/part.parquet")
write(part, "/tmp/pt2-1438/date_partition=2023-08-01/id_partition=id-002/decimal=0.15/part.parquet")
write(part, "/tmp/pt2-1438/date_partition=2023-08-02/id_partition=id-001/decimal=0.08/part.parquet")
write(part, "/tmp/pt2-1438/date_partition=2023-08-02/id_partition=id-002/decimal=0.08/part.parquet")
write(part, "/tmp/pt2-1438/date_partition=2023-08-02/id_partition=id-003/decimal=0.1/part.parquet")

partition_table = read("/tmp/pt2-1438")
```


packages/iris-grid/src/IrisGridPartitionSelector.test.tsx
mofojed, line 1: We should beef up the tests in this file. Add some tests to ensure the correct selectors appear for a couple of different combinations, that the "Append" button appears correctly if the `onAppend` prop is provided, etc. Let me know if you want help figuring out which test cases to write, or how to use codecov to determine where more coverage is needed.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -88,35 +105,58 @@ class IrisGridPartitionSelector<T> extends Component<
...
+    const { partitions, partitionTables } = this.state;
+    const selectedMenu = this.searchMenu[index];
+    if (selectedMenu) {
+      selectedMenu.closeMenu();
+    }
+
+    const newPartitions = [...partitions];
+    newPartitions[index] =
+      TableUtils.isCharType(columns[index].type) && partition.length > 0
+        ? partition.charCodeAt(0).toString()
```
mofojed, line 145: Why do we need to limit it to one char here? Why are we allowed to enter more than one char in this case at all?

mofojed (changes requested):
Need to worry about legacy format from dehydrated state, so update that type and handle it correctly in `hydrateIrisGridPanelState`.

packages/iris-grid/src/IrisGridPartitionSelector.test.tsx
dsmmcken: the new mockup removes the append button.

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -713,11 +713,23 @@ export class IrisGridPanel extends PureComponent<
...
+          filterArray.push(
+            `"${column.name}=${
+              TableUtils.isTextType(column.type) ? `\`${value}\`` : value
+            }"`
+          );
+        }
+        return filterArray;
+      }, [])
+      .join(', ');
+    const command = `${tableName} = ${tableName}.where(filters=[${filters}])`;
```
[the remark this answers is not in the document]
georgecwan: How do I get which language I should be using? Is there a function that tells me if Deephaven is in Python or Groovy mode?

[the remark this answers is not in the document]
georgecwan: Since the append function seems to be getting removed, should I still worry about making this work?

packages/iris-grid/src/IrisGridUtils.test.ts
```diff
@@ -658,8 +658,8 @@ describe('dehydration methods', () => {
       'dehydrateIrisGridPanelState',
```
mofojed, line 658: These tests right now are lacking hydration tests. We should add hydration tests for both v1 and v2 cases, to make sure we're getting the expected results. Some other cases to cover:
- No column selected (v1 and v2)
- One column selected (v1 and v2)
- Partial selection (v2 only)
- Multiple selected (v2 only)
... and any other you can think of.

mofojed (changes requested):
It looks good, but with the added hydration method and the imperative that we handle legacy data correctly, we need to add unit tests that cover those hydration cases.

packages/iris-grid/src/IrisGridTestUtils.ts
```diff
@@ -92,7 +92,7 @@ class IrisGridTestUtils {
   makeTable({
     columns = this.makeColumns(),
     size = 1000000000,
-    sort = [],
+    sort = [] as Sort[],
     layoutHints = {} as LayoutHints,
   } = {}): Table {
```
mofojed, line 97: This is fine as is, and consistent with what's there.

These days I would prefer to detail the type rather than requiring casting, e.g.
```
  makeTable({
    columns = this.makeColumns(),
    size = 1000000000,
    sort = [],
    layoutHints = {},
  }: {
    columns?: Column[];
    size?: number;
    sort?: readonly Sort[];
    layoutHints?: LayoutHints;
  } = {}): Table {
```

Or even if you define the type prior, and export it so it can be used in another file, e.g.
```
export type MakeTableOptions = {
  columns: Column[];
  size: number;
  sort: readonly Sort[];
  layoutHints: LayoutHints;
}

  makeTable({
    columns = this.makeColumns(),
    size = 1000000000,
    sort = [],
    layoutHints = {},
  }: Partial<MakeTableOptions> = {}): Table {
```

packages/iris-grid/src/IrisGridUtils.test.ts
```diff
@@ -680,3 +667,131 @@ describe('dehydration methods', () => {
...
+      },
+    ],
+  ])('%s invalid column error', (_label, panelState) => {
+    expect(() =>
+      IrisGridUtils.hydrateIrisGridPanelState(
+        irisGridTestUtils.makeModel(
+          irisGridTestUtils.makeTable({
+            columns: irisGridTestUtils.makeColumns(5, 'name_'),
+          })
+        ),
```
mofojed, line 698: This same model is used in all tests, you could declare it at the beginning of the describe.

packages/iris-grid/src/IrisGridUtils.test.ts
```diff
@@ -680,3 +667,131 @@ describe('dehydration methods', () => {
...
+      },
+    ],
+    [
+      'hydrateIrisGridPanelStateV1 one selected partition',
+      {
+        isSelectingPartition: true,
+        partition: '',
+        partitionColumn: 'name_0',
+        advancedSettings: [],
+      },
```
mofojed, line 730: These tests don't seem right. A selected partition should have a value for partition,. and the "unselected partition" test - are we testing the empty string? Or `isSelectingPartition`?

packages/iris-grid/src/IrisGridUtils.test.ts
```diff
@@ -680,3 +667,131 @@ describe('dehydration methods', () => {
     ).toBe(true);
   });
 });
+
+describe('hydration methods', () => {
```
mofojed, line 671: Should have cases for no partition column as well.

mofojed (changes requested):
Tests need a bit of cleanup

packages/iris-grid/src/IrisGridUtils.test.ts
georgecwan: I believe that the first test is testing for `isSelectingPartition` being `false` (in which case the value of `partition`) shouldn't matter and the second test is just testing for selecting a partition normally. `partition` should be `null` to signify a partition without a value so the second test is just testing for the empty string. I can use a non-empty string to make the test clearer.

mofojed (approved):

PR #1584, opened 2023-10-19
style: Enable method-signature-style TypeScript ESLint rule

### LINKED ISSUE #1578: Enable method-signature-style TypeScript ESLint rule
https://typescript-eslint.io/rules/method-signature-style/

Using the property option makes TS a little stronger in its function signature checks compared to the method option. We should enable this rule in `property` mode (which is the default).

georgecwan:
Added method-signature-style rule to ESLint rules with default `property` mode.

Resolves #1578


mattrunyon (changes requested):

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingItem.tsx
```diff
@@ -13,13 +13,19 @@ type VisibilityOrderingItemProps = {
...
-  onClick(name: string, event: React.MouseEvent): void;
-  onGroupDelete(group: ColumnHeaderGroup): void;
-  onGroupColorChange(group: ColumnHeaderGroup, color: string | undefined): void;
-  onGroupNameChange(group: ColumnHeaderGroup, name: string): void;
-  validateGroupName(name: string): string;
-  handleProps: Record<string, unknown>;
+  onVisibilityChange: (modelIndexes: number[], isVisible: boolean) => void;
+  onClick: (
+    name: string,
+    event: React.MouseEvent<HTMLElement, MouseEvent>
```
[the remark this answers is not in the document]
georgecwan: It gave me an error of `Element` not being convertible to `HTMLElement` before, I will leave it as `React.MouseEvent<HTMLElement>`.

packages/code-studio/src/styleguide/Dialog.tsx
```diff
@@ -54,7 +54,7 @@ class Dialog extends Component<unknown, DialogState> {
   }
 
   handleUpdateCheckboxMap(
-    checkBoxMap: Map<string, Map<string, boolean>>
+    checkBoxMap: Map<string, boolean | Map<string, boolean>>
```
mattrunyon, line 57: Not a big deal since it's just the styleguide, but looks like there's a type for this in this file. `HierarchicalCheckboxValueMap`. Just use that instead

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
```diff
@@ -140,7 +140,7 @@ class AdvancedFilterCreatorSelectValueList<T = unknown> extends PureComponent<
     );
     if (selectedIndex >= 0) {
       selectedValues.splice(selectedIndex, 1);
-    } else {
+    } else if (value != null) {
```
mattrunyon, line 143: This changes the logic of this function. Need to test what happened if the value was null before to ensure we shouldn't be doing something else in that case or typing `selectedValues` differently

packages/iris-grid/src/GotoRow.tsx
```diff
@@ -115,9 +115,9 @@ const GotoRow = forwardRef<GotoRowElement, GotoRowProps>(
     };
 
     const handleGotoValueKeySubmit = (
-      e: KeyboardEvent<HTMLInputElement>
+      e?: KeyboardEvent<HTMLInputElement>
```
mattrunyon, line 118: This doesn't seem right. I think this should be guaranteed to get the event and instead you should go down through where this is passed and make those non-optional params as well. Looks like it goes through `DateTimeInput` and `MaskedInput`

mattrunyon (changes requested):

packages/iris-grid/src/AdvancedFilterCreatorSelectValueList.tsx
georgecwan: It seems like it should be possible to push `null` since `makeAdvancedFilter` will be called on this array in `AdvancedFilterCreator.tsx` and it should be possible to filter on a `null` value. I changed the typing of `selectedValues` to `T | null` to get rid of the TS errors.

mattrunyon (approved):

mattrunyon:
Good work! Looks like this caught a few places where our types weren't quite correct and we could have potentially done something not safe in the future

PR #1587, opened 2023-10-19
feat: Composite plugin loading

### LINKED ISSUE #1586: Add support for plugin "suites" or composite plugins. Most likely just need to support an array of config objects as an option for plugin configs

mattrunyon:
Fixes #1586


mattrunyon:
Closing as it's not a current priority and this PR is old. Can always revive it if needed since it's a relatively small change

PR #1594, opened 2023-10-20
fix: Adjusted Monaco "white" colors

### LINKED ISSUE #1592: Monaco autocomplete colors are incorrect
`--dh-color-editor-suggest-fg` should be `--dh-color-white`. Right now we end up w/ black on gray text.

image

We should also probably add a variable to set `--vscode-editorSuggestWidget-selectedForeground` to `--dh-color-white` as well. It is currently set to `#ffffff` by monaco. This affects the color of the non-matching portion when there is keyboard selection on an autocomplete item.

Might want to set something for the blue colors used in selected/non-selected suggestions as well. Those are `--vscode-editorSuggestWidget-highlightForeground` and `--vscode-editorSuggestWidget-focusHighlightForeground`

**Versions**

 - Web UI Version: 0.50.0

bmingles:
This fixes some Monaco suggestion colors

image

fixes #1592

### DIFF
--- a/packages/components/src/theme/theme-dark/theme-dark-semantic-editor.css
+++ b/packages/components/src/theme/theme-dark/theme-dark-semantic-editor.css
@@ -58,6 +58,7 @@
   --dh-color-editor-suggest-selected-bg: var(--dh-color-highlight-selected);
   --dh-color-editor-suggest-selected-fg: var(--dh-color-white);
   --dh-color-editor-suggest-highlight-fg: var(--dh-color-accent-700);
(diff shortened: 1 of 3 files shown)


mattrunyon (approved):
Context for @dsmmcken the colors that can be modified are the blue letters in the highlighted selection (1st item), blue in the non-highlighted, white in highlighted, white in non-highlighted. This sets highlighted and non-highlighted to be the same for both colors

dsmmcken (approved):

PR #1596, opened 2023-10-23
fix: Remove @deephaven/app-utils from @deephaven/dashboard-core-plugins dependency list

### LINKED ISSUE #1593: Remove @deephaven/app-utils from @deephaven/dashboard-core-plugins dependencies
**Description**

This is an issue w/ DHE dev mode only (there is an easy workaround by modifying the Vite config there).

The issue is that `@deephaven/app-utils` was added as a dependency to `@deephaven/dashboard-core-plugins`. As a result, it became an indirect dependency in DHE when upgrading to 0.50.0 (dependency change was in 0.49.0).

In dev mode, Vite uses esbuild to pre-bundle and cache all dependencies. As part of its tree traversal (so it can do tree shaking), it hits the `import 'fira'` in `FontBootstrap` and esbuild throws that it can't handle font files like `.ttf`. This doesn't seem to be an issue when building since rollup is used for production builds.

Even though `@deephaven/app-utils` is not used by DHE, it is part of the import tree (and then would get shaken out). This issue is a downside of using barrel files to house all the imports for the package.

The things needed from `app-utils` can be split out into other packages. Specifically the `ConnectionContext` and `useConnection` can be put in `jsapi-utils` or `jsapi-components`. Then `app-utils` just creates the provider used in the app.

**Steps to reproduce**

Install DHC 0.50.0 packages in DHE and run `npm start`.

**Expected results**

Functional dev mode

**Actual results**

Broken dev mode complaining that esbuild doesn't know how to deal w/ font files.

Adding `optimizeDeps.exclude = ['@deephaven/app-utils']` seems to make dev mode functional again in DHE.

**Versions**

 - Web UI Version: 0.49.0

mattrunyon:
Fixes #1593

I don't think these breaking changes should affect enterprise since it doesn't use `@deephaven/app-utils`

BREAKING CHANGE:
- `usePlugins` and `PluginsContext` were moved from `@deephaven/app-utils` to `@deephaven/plugin`.
- `useLoadTablePlugin` was moved from `@deephaven/app-utils` to `@deephaven/dashboard-core-plugins`.
- `useConnection` and `ConnectionContext` were moved from `@deephaven/app-utils` to `@deephaven/jsapi-components`.
- `DeephavenPluginModuleMap` was removed from `@deephaven/redux`. Use `PluginModuleMap` from `@deephaven/plugin` instead.


packages/app-utils/tsconfig.json
```diff
@@ -4,7 +4,13 @@
     "rootDir": "src/",
     "outDir": "dist/"
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.js", "src/**/*.jsx"],
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
+    "src/**/*.js",
+    "src/**/*.jsx",
+    "../plugin/src/usePlugins.ts"
```
bmingles, line 12: Is this include intentional?

mofojed: Yea seems like this was unintentional

mofojed (changes requested):

packages/redux/src/actions.ts
```diff
@@ -3,6 +3,7 @@ import type { ThunkAction } from 'redux-thunk';
 import type { CommandHistoryStorage } from '@deephaven/console';
 import type { FileStorage } from '@deephaven/file-explorer';
 import type { dh as DhType } from '@deephaven/jsapi-types';
+import type { PluginModuleMap } from '@deephaven/plugin';
```
bmingles, line 6: Should `packages/redux` include @deephaven/plugin in package dependencies?

bmingles (changes requested):
Left 2 comments

packages/app-utils/tsconfig.json
mattrunyon: Ya VSCode's move file added this when I said to update import paths

packages/redux/src/actions.ts
mattrunyon: I added it. I'm not entirely sure how the types would be affected by a consumer if they only installed redux, but since it's in the `tsconfig.json` for the package, I added it to dependencies

bmingles (approved):
LGTM

mofojed (approved):

PR #1598, opened 2023-10-25
feat: Convert DashboardPlugins to WidgetPlugins

### LINKED ISSUE #1573: Convert DashboardPlugins to WidgetPlugins
Some of our dashboard plugins should actually be widget plugins. They don't need to have listeners mounted globally and we will need to be able to use them as widget plugins for deephaven ui.

Definitely can convert
- [ ] GridPlugin
- [ ] ChartPlugin
- [ ] PandasPlugin

Possibly can convert
- [ ] MarkdownPlugin (doesn't listen to PanelEvent.Open)
- [ ] FilterPlugin (uses other global events that might not be simple conversions)

mattrunyon:
Fixes #1573 

This fix should be adequate for furthering the Deephaven UI alpha. Rather than trying to refactor 2 1k+ line files (IrisGridPanel and ChartPanel), I opted to add some DOM detection to the `Panel` component to prevent rendering tab context menu/tooltips if it is nested within another panel.

For Deephaven UI, we will need to ensure the UI panel uses `Panel` at least and passes through `glContainer` and `glEventHub` to the child elements so grids and charts function roughly as normal.

### Testing

To test how something would look/function when wrapped, remove the `panelComponent` line from either of the plugin configs. This will wrap in a generic `WidgetPanel` which you can see in dev tools. The panel tab will not have any of the custom tooltip for grid or custom context menus.

Test by opening grids/figures, applying some filters/sorts and then reloading the page to ensure the state persisted. Also test adding links and ensuring those persisted on reload. Test that a chart built with chart builder works and survives a reload as well.

### Followup

We might be able to convert `MarkdownPlugin` and `FilterPlugin`, but I wanted to get the main components in first. Not sure if those would work as WidgetPlugins as currently constructed since they don't listen for `PanelEvent.OPEN`. `FilterPlugin` probably makes sense to split into a suite of plugins for each filter type which would need #1587

### DIFF
--- a/packages/dashboard-core-plugins/src/ChartPluginConfig.ts
+++ b/packages/dashboard-core-plugins/src/ChartPluginConfig.ts
@@ -1,10 +1,15 @@
-import { PluginType, DashboardPlugin } from '@deephaven/plugin';
+import { PluginType, type WidgetPlugin } from '@deephaven/plugin';
+import { vsGraph } from '@deephaven/icons';
 import ChartPlugin from './ChartPlugin';
 
-const ChartPluginConfig: DashboardPlugin = {
-  name: 'ChartPlugin',
-  type: PluginType.DASHBOARD_PLUGIN,
+const ChartPluginConfig: WidgetPlugin = {
+  name: 'ChartPanel',
+  title: 'Chart',
+  type: PluginType.WIDGET_PLUGIN,
   component: ChartPlugin,
+  panelComponent: ChartPlugin,
+  supportedTypes: 'Figure',
+  icon: vsGraph,
 };
 
 export default ChartPluginConfig;
--- a/packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
+++ b/packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
@@ -72,7 +72,6 @@ export function WrapWidgetPlugin(
             {...props}
           />
         )}
-        )
       </WidgetPanel>
     );
   }
--- a/packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
@@ -81,6 +81,7 @@ export interface ChartPanelFigureMetadata extends PanelMetadata {
    * @deprecated use `name` instead
    */
   figure?: string;
+  sourcePanelId: never;
 }
 
 export interface ChartPanelTableMetadata extends PanelMetadata {
@@ -118,7 +119,7 @@ export interface GLChartPanelState {
   table?: string;
   figure?: string;
 }
-export interface ChartPanelProps extends DashboardPanelProps {
+interface OwnProps extends DashboardPanelProps {
   metadata: ChartPanelMetadata;
   /** Function to build the ChartModel used by this ChartPanel. Can return a promise. */
   makeModel: () => Promise<ChartModel>;
@@ -128,16 +129,19 @@ export interface ChartPanelProps extends DashboardPanelProps {
   containerRef?: RefObject<HTMLDivElement>;
 
   panelState: GLChartPanelState;
-  settings: Partial<WorkspaceSettings>;
 }
 
-interface PropsFromRedux {
+interface StateProps {
   inputFilters: InputFilter[];
   links: Link[];
   isLinkerActive: boolean;
   source?: TableTemplate;
   sourcePanel?: PanelComponent;
   columnSelectionValidator?: ColumnSelectionValidator;
+  settings: Partial<WorkspaceSettings>;
+}
+
+interface DispatchProps {
   setActiveTool: (tool: string) => void;
   setDashboardIsolatedLinkerPanelId: (
     id: string,
@@ -181,14 +185,12 @@ function hasPanelState(
   return (panel as { panelState: IrisGridPanelState }).panelState != null;
 }
 
-type ChartPanelPropsWithRedux = ChartPanelProps &
-  PropsFromRedux &
+export type ChartPanelProps = OwnProps &
+  StateProps &
+  DispatchProps &
   React.RefAttributes<ChartPanel>;
 
-export class ChartPanel extends Component<
-  ChartPanelPropsWithRedux,
-  ChartPanelState
-> {
+export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
   static defaultProps = {
     columnSelectionValidator: null,
     isLinkerActive: false,
@@ -203,7 +205,7 @@ export class ChartPanel extends Component<
 
   static COMPONENT = 'ChartPanel';
 
-  constructor(props: ChartPanelPropsWithRedux) {
+  constructor(props: ChartPanelProps) {
     super(props);
 
     this.handleColumnSelected = this.handleColumnSelected.bind(this);
@@ -273,7 +275,7 @@ export class ChartPanel extends Component<
   }
 
   componentDidUpdate(
-    prevProps: ChartPanelPropsWithRedux,
+    prevProps: ChartPanelProps,
     prevState: ChartPanelState
   ): void {
     const { inputFilters, source } = this.props;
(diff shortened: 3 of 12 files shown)


packages/dashboard-core-plugins/src/ChartPluginConfig.ts
```diff
@@ -1,10 +1,15 @@
...
-const ChartPluginConfig: DashboardPlugin = {
-  name: 'ChartPlugin',
-  type: PluginType.DASHBOARD_PLUGIN,
+const ChartPluginConfig: WidgetPlugin = {
+  name: 'ChartPanel',
+  title: 'Chart',
+  type: PluginType.WIDGET_PLUGIN,
   component: ChartPlugin,
+  panelComponent: ChartPlugin,
+  supportedTypes: 'Figure',
```
mofojed, line 11: Hm... so we can't use the `dh.VariableType.FIGURE` here anymore... Just making a note of that.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
```diff
@@ -72,7 +72,6 @@ export function WrapWidgetPlugin(
             {...props}
           />
         )}
-        )
```
mofojed, line 75: How'd this not throw an error before?

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -122,9 +122,18 @@ export interface ChartPanelProps extends DashboardPanelProps {
...
+  localDashboardId: string;
+  Plotly?: typeof PlotlyType;
+  /** The panel container div */
+  containerRef?: RefObject<HTMLDivElement>;
+
+  panelState: GLChartPanelState;
+  settings: Partial<WorkspaceSettings>;
+}
+
+interface PropsFromRedux {
```
mofojed, line 134: We've done this paradigm before, but slightly differently. Take a look at `Linker.tsx`, where we define `StateProps`, `OwnProps`, and then the `connector` and join them together. I think overall that way is a little bit cleaner - you can actually use `OwnProps` in `mapStateToProps`, and you don't need to redefine all the types from redux, they will implicitly get picked up from the connector.

packages/dashboard-core-plugins/src/panels/Panel.tsx
```diff
@@ -151,6 +155,11 @@ class Panel extends PureComponent<PanelProps, PanelState> {
     if (prevProps.componentPanel == null && componentPanel != null) {
       glEventHub.emit(PanelEvent.MOUNT, componentPanel);
     }
+
+    this.setState({
```
mofojed, line 159: Shouldn't we check this in `componentDidMount` instead?

packages/dashboard-core-plugins/src/panels/Panel.tsx
```diff
@@ -151,6 +155,11 @@ class Panel extends PureComponent<PanelProps, PanelState> {
...
+      isWithinPanel:
+        this.ref.current?.parentElement?.closest('.iris-panel') != null,
```
mofojed, line 161: We may as well rename it to `dh-panel` at this point as well.

packages/dashboard-core-plugins/src/panels/Panel.tsx
```diff
@@ -335,21 +346,23 @@ class Panel extends PureComponent<PanelProps, PanelState> {
...
       >
         {children}
         <LoadingOverlay
           errorMessage={errorMessage}
           isLoaded={isLoaded}
           isLoading={isLoading}
         />
-        {glTab != null &&
+        {!isWithinPanel &&
+          glTab != null &&
```
mofojed, line 365: Right now we're defaulting to `isWithinPanel = true` on mount, and only setting it to false after the component is updated and we have a ref. I _believe_ that could leave a case where the portal here gets rendered, then we take it out immediately after on the next update. I'd flip the logic around so that we only render this after we verify that we are _not_ in another panel, just to be sure.

packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx
mattrunyon: This was actually displaying a parenthesis after the component if it was auto wrapped

packages/dashboard-core-plugins/src/panels/Panel.tsx
mattrunyon: Maybe. I wasn't sure if there would end up being a case where the ref or parent container could change, so I put it in update for max safety. Should be fine in `componentDidMount` though and it's not like it's an app breaking bug if we later find out it does need to be in `componentDidUpdate`

packages/dashboard-core-plugins/src/panels/Panel.tsx
mattrunyon: That's why I default to true. Initially we're not sure if it's in another panel, so assume we are. Portal should not be rendered on first render until we determine if we're in a wrapped panel and then a rerender would trigger

packages/dashboard-core-plugins/src/ChartPluginConfig.ts
mattrunyon: Yep. One downside to this, but also something that would be required for lazy loading from the manifest file for 3rd party plugins.

We could potentially provide a `createConfig` function option that gets the JS API or something, but not sure it adds much. Or do something where these are registered to use the JS API values instead, but also not sure that's really worth much

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
mattrunyon: You still need to do `OwnProps & StateProps` so the component knows, which is just a different name for what I did here. I changed to `OwnProps` and `StateProps` though.

A downside is if you need to use `OwnProps` from another file, it should be more specific. Minor thing, but `PandasPanel` needs to use `IrisGridOwnProps` because it doesn't not get/set the props from redux itself, but renders the connected component which does it. If you just use the `IrisGridProps` as `PandasProps` type, TS expects you to pass the redux props as well since it's not a connected component

mofojed: Within `IrisGridPanel.tsx`, just name it `OwnProps`; then the importer can rename it as needed, e.g. `import { type OwnProps as IrisGridOwnProps } from './IrisGridPanel'`. It's already namespaced within the file, the extra namespacing is a little redundant (though we do this in a ton of places).
The main advantage is not needing to specify `DispatchProps`, but you do need to move the `connect` and `mapStateToProps` up.

packages/dashboard-core-plugins/src/panels/Panel.tsx
mofojed: I feel like we're saying the same thing, but we both said it's defaulting to `true` which is incorrect - right now it's currently defaulting to `isWithinPanel: false` in the constructor. When it _should_ be `isWithinPanel: true`.

packages/dashboard-core-plugins/src/panels/Panel.tsx
```diff
@@ -118,10 +119,13 @@ class Panel extends PureComponent<PanelProps, PanelState> {
     this.handleTabClicked = this.handleTabClicked.bind(this);
     this.handleTab = this.handleTab.bind(this);
 
+    this.ref = React.createRef<HTMLDivElement>();
+
     const { glContainer } = this.props;
     this.state = {
       title: LayoutUtils.getTitleFromContainer(glContainer),
       showRenameDialog: false,
+      isWithinPanel: false,
```
mofojed, line 128:
```suggestion
      isWithinPanel: true,
```

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/Panel.tsx
mattrunyon: 🤦I should read my own code sometime instead of going off what I thought I wrote. Good catch

mofojed (approved):

PR #1600, opened 2023-10-27
fix: Theming - switched from ?inline to ?raw css imports

### LINKED ISSUE #1599: Theming ?inline css imports not consumable in Enterprise
**Description**

The ?inline css imports used in @deephaven/components for theming don't seem to be compatible with Enterprise

**Steps to reproduce**

Run Enterprise with dh v0.51.0

**Expected results**

CSS variable content should be inlined

**Actual results**

Theme css variable content is empty.

**Versions**
 - Web UI Version: 0.51.0

bmingles:
It seems that Enterprise does not handle css ?inline css imports in node_modules the same way it does in src. Namely, the raw content gets imported as an empty string instead of the raw content.

- Converted ?inline imports to ?raw. These seem to handle the raw content as expected even if in node_modules
- Added a sass minification step to @deephaven/components. Vite seems to minify ?inline imports but does not do so with raw
- Added a try / catch to MonacoUtils to avoid non-hex theme variables from crashing the entire app

fixes #1599

BREAKING CHANGE: Theme css imports were switched from `?inline` to `?raw`. Not likely that we have any consumers yet, but this would impact webpack config.

### DIFF
--- a/packages/components/src/theme/theme-light/index.ts
+++ b/packages/components/src/theme/theme-light/index.ts
@@ -1,5 +1,42 @@
-import themeLightPalette from './theme-light-palette.css?inline';
+import themeLightPalette from './theme-light-palette.css?raw';
 
+/**
+ * DH theme variables are imported via Vite `?raw` query which provides the
+ * text content of the variable files as a string. The exported theme is just a
+ * concatenation of the contents of all of these imports.
+ *
+ * Note that ?raw / ?inline imports are natively supported by Vite, but consumers
+ * of @deephaven/components using Webpack will need to add a rule to their module
+ * config.
+ * e.g.
+ * module: {
+ *  rules: [
+ *    {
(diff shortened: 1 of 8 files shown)


packages/components/src/theme/theme-light/index.ts
```diff
@@ -1,5 +1,42 @@
...
+ * }
+ * :root {
+ *   --dh-color-from-light-semantic-editor: #000;
+ * }
+ * :root {
+ *   --dh-color-from-light-semantic-grid: #000;
+ * }
+ * :root {
+ *   --dh-color-from-light-components: #000;
+ * }
```
mofojed, line 38: Not sure where this example came from?

mofojed (approved):

mattrunyon (approved):

PR #1603, opened 2023-10-30
feat: Theming - Bootstrap

bmingles:
- Updated bootstrap styles to pull from theme variables and added some mixin overrides

> NOTE I copy / pasted all of the original Bootstrap mixins (`packages/components/scss/bootstrap_override_*` in their own commit to make the diff easier for customizations I made. Should be able to see the diff here: https://github.com/bmingles/web-client-ui/compare/7ce4875ef065075c920f2e70b3462a77215b6d5e..bmingles:1540-theming-bootstrap

BREAKING CHANGE: Bootstrap color variables are now predominantly hsl based. SCSS will need to be updated accordingly. Theme providers are needed to load themes.


packages/components/scss/bootstrap_imports.scss
```diff
@@ -0,0 +1,61 @@
...
+
+@import 'bootstrap/scss/root';
+@import 'bootstrap/scss/reboot';
+@import 'bootstrap/scss/type';
+// @import 'bootstrap/scss/images';
+@import 'bootstrap/scss/code';
+@import 'bootstrap/scss/grid';
+// @import 'bootstrap/scss/tables';
+
+// NEED THIS THOUGH
```
bmingles, line 32: @dsmmcken I copy / pasted this verbatim from your PR. Do you remember any more context for this comment?

packages/components/scss/bootstrap_override_functions.scss
```diff
@@ -0,0 +1,36 @@
...
+    } @else {
+      @return $light;
+    }
+  }
+
+  // if the $color contains the string "yellow" or "orange" or "chatreuse" or "cyan" return $dark
+  // per the spectrum color guidelines. All other colors have white text.
+  // https://spectrum.adobe.com/page/using-color/#Solid-color-background-with-black-or-white
+  // Which for us means, informative and notice have dark text.
+  @if str-index($color, 'informative') {
```
bmingles, line 29: Note to myself, I need to update this since we now use our own --dh-color vars instead of Spectrum variables

packages/components/scss/bootstrap_imports.scss
dsmmcken: That was probably just a note to self, a can safely be deleted. Probably that forms was needed and for some  reason I didn't think it was.

packages/components/scss/BaseStyleSheet.scss
```diff
@@ -596,26 +597,26 @@ input[type='number']::-webkit-inner-spin-button {
 ::selection {
   //blink/webkit
   color: $foreground;
-  background: rgba($text-select-color, 0.99);
+  background: var(--dh-color-text-highlight);
```
mofojed, line 600: This doesn't seem to get mapped correctly in the derived monaco theme, and is appearing red. The value derived is `"hsla(222deg, 82.7%, 61.6%, 83%, 62%, 0.3)"` instead of the #rgba hex value.

image

packages/components/scss/BaseStyleSheet.scss
```diff
@@ -755,7 +757,7 @@ input[type='number']::-webkit-inner-spin-button {
       center;
     //make the dotted duplicate focus line on firefox go away
     &:-moz-focusring {
-      color: rgba(0, 0, 0, 0%);
+      color: transparent;
```
mofojed, line 760: Dropdown arrow in the select input is appearing black:
image

packages/code-studio/src/styleguide/colorUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+/** Return black or white contrast color */
+export function contrastColor(color: string): 'black' | 'white' {
+  const rgba = ColorUtils.parseRgba(ColorUtils.asRgbOrRgbaString(color) ?? '');
+  if (rgba == null || rgba.a < 0.5) {
+    return 'white';
+  }
+
+  const { r, g, b } = rgba;
+  const y = (299 * r + 587 * g + 114 * b) / 1000;
+  return y >= 128 ? 'black' : 'white';
```
mofojed, line 16: Use `ColorUtils.getBrightness` here:
```suggestion
  return ColorUtils.getBrightness([r, g, b]) >= 128 ? 'black' : 'white';
```

packages/components/scss/bootstrap_imports.scss
```diff
@@ -0,0 +1,60 @@
...
+// patch the mixins
+@import './bootstrap_override_mixins_background_variant.scss';
+@import './bootstrap_override_mixins_buttons.scss';
+@import './bootstrap_override_mixins_forms.scss';
+@import './bootstrap_override_mixins_text_emphasis.scss';
+
+@import 'bootstrap/scss/root';
+@import 'bootstrap/scss/reboot';
+@import 'bootstrap/scss/type';
+// @import 'bootstrap/scss/images';
```
mofojed, line 27: Maybe group the ones we're intentionally not importing together and just say "we're not importing these as they're not needed" or something?

packages/golden-layout/scss/goldenlayout-dark-theme.scss
```diff
@@ -362,7 +362,7 @@ body:not(.lm_dragging) .lm_header .lm_tab .lm_close_tab:hover {
 
 // If a specific Pane is maximized
 .lm_maximised {
-  border: 15px solid rgba($black, 0.85);
+  border: 15px solid background-opacity(80);
```
mofojed, line 365: Not sure exactly what's causing this, but maximise button is appearing black instead of gray:
image

mofojed (changes requested):
Two things I noticed while testing:
- The animation on login screen (`RandomAreaPlotAnimation`) doesn't display anymore, it's just a black screen.
- The _first_ time I ran this, the loading circle was black... I think because I switched from `main` to this branch? Can't seem to reproduce it with an incognito window or by re-opening the same thing, so probably not a big deal but thought I'd mention it.

packages/components/scss/BaseStyleSheet.scss
```diff
@@ -764,7 +766,7 @@ input[type='number']::-webkit-inner-spin-button {
     background-image: escape-svg($custom-select-indicator),
       linear-gradient(0deg, $gray-700, $gray-700);
     &:-moz-focusring {
-      color: rgba(0, 0, 0, 0%);
+      color: transparent;
       text-shadow: 0 0 0 $foreground !important;
     }
   }
```
mofojed, line 772: `.console-creator` isn't needed on Community, console-creator is only on Enterprise (and we may not even need this clause at all anymore)

packages/components/scss/BaseStyleSheet.scss
dsmmcken: I like how monaco yells at you if you set an invalid color by making it red. Real useful for theming actually.

packages/components/scss/BaseStyleSheet.scss
dsmmcken: This is gonna be slightly annoying to fix. My best suggestion is we need to change all usages of `<select>` to use `<Select>` (22 in core, 35 in enterprise). Then wrap `Select` with an extra element to add an icon that we can colorize with a css variable. Probably want to do this in it's own PR first.

packages/golden-layout/scss/goldenlayout-dark-theme.scss
dsmmcken: This will probably need to be changed to use masks for coloring the icons.

Can't style inline svg with a css variable, it was being colored at compile time.

packages/components/scss/BaseStyleSheet.scss
bmingles: Looks like I duplicated the S/L components here. Fixed

packages/components/scss/BaseStyleSheet.scss
bmingles: This is due to a limitation with inline SVGs not supporting injected CSS variables. I put in a static color value to address this temporarily. @dsmmcken is planning to change the implementation of our <select> to make it possible to style dynamically.

dsmmcken: @bmingles and I discussed a different plan since this is also an issue for input invalid and select invalid markers.

New plan is to make css variables that hold the the icon, and then use JS to string interpolate a resolved color into the variable on the client when it is being switched.

packages/components/scss/bootstrap_imports.scss
bmingles: Bootstrap import order is important which is why they were left in their original positions. That said, I can also add that detail to the comment telling devs to consult the original order if they need to add any back

packages/components/scss/BaseStyleSheet.scss
bmingles: I created https://github.com/deephaven/web-client-ui/issues/1635 to follow up on this one

packages/code-studio/src/styleguide/colorUtils.ts
bmingles: Addressed in [ColorUtils.getBrightness](https://github.com/deephaven/web-client-ui/pull/1651)

packages/code-studio/src/styleguide/colorUtils.ts
```diff
@@ -149,27 +176,67 @@ export function buildColorGroups(
...
   return groupData;
 }
+
+/**
+ * Normalize color to a hex string and drop alpha if it is 'ff'
+ * @param color
+ */
+export function normalizedOptionalAlpha(color: string): string {
+  return ColorUtils.normalizeCssColor(color).replace(
+    /^(#[a-f0-9]{6})ff$/,
```
mofojed, line 219: I was slightly concerned about casing here, but looks like normalize always returns it lowercase (converting to rgba and back).
We could specify the details in the type returned from `ColorUtils.normalizeCssColor` using [intrinsic types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype), e.g.
```ts
type HexColor = `#${Lowercase<string>}`
```

Though that still doesn't enforce length/hex characters only. May be too complicated for TypeScripts system to be complete, but at least with that type we guarantee it starts with a `#` and is lowercase only.

Doesn't need to be changed with this PR though, just something I noticed could be improved.

packages/code-studio/src/styleguide/colorUtils.ts
```diff
@@ -149,27 +176,67 @@ export function buildColorGroups(
...
+    '$1'
+  );
+}
+
+export function useContrastFgColorRef<
```
mofojed, line 224: I'd prefer this in some sort of hooks file rather than mixed with other utils - this hook requires React and can only be used in a render tree, whereas the other utils do not.

packages/components/scss/BaseStyleSheet.scss
```diff
@@ -16,6 +21,14 @@ html {
   min-height: 100%;
 }
 
+// override
+.text-black-50 {
+  color: background-opacity(50) !important;
```
mofojed, line 26: Wonder if this should actually use `dh-color-black` and `dh-color-white` instead... unsure where these are used, you probably know better than I.

packages/components/scss/bootstrap_imports.scss
```diff
@@ -0,0 +1,49 @@
+/*!
+ * Bootstrap v4.6.2 (https://getbootstrap.com/)
+ * Copyright 2011-2022 The Bootstrap Authors
+ * Copyright 2011-2022 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/main/LICENSE)
+ */
+
+/* 
+ * Deephaven manually imports only the necessary bootstrap files.
+ * See `packages/components/scss/bootstrap.scss` for the original list of imports.
```
mofojed, line 10: 404

packages/components/scss/bootstrap_overrides.scss
```diff
@@ -141,26 +167,31 @@ $btn-transition:
...
+$input-btn-focus-color: hsla(var(--dh-color-accent-hsl), 0.35);
 $input-btn-focus-box-shadow: 0 0 0 $input-btn-focus-width $input-btn-focus-color;
 
 //checkbox
 $custom-control-indicator-bg: $gray-600;
+$custom-control-indicator-active-bg: var(--dh-color-accent-1100);
 $custom-control-indicator-bg-size: 75% 75%;
 $custom-control-indicator-disabled-bg: $gray-800;
 $custom-control-indicator-checked-disabled-bg: $gray-800;
 $custom-control-label-disabled-color: $gray-400;
```
mofojed, line 187: Shouldn't these be variables?

packages/components/src/RandomAreaPlotAnimation.tsx
```diff
@@ -1,27 +1,26 @@
...
-const GRID_COLOR = ThemeExport['gray-800'];
-const BACKGROUND = ThemeExport['gray-900'];
 const PATTERN = { SIZE: 8, DOT_SIZE: 2, FILL_OPACITY: 0.25 };
 const PX_PER_SECOND = 18;
 const RESIZE_DEBOUNCE = 250;
 const INTERACTION_TIMEOUT = 60 * 1000;
 
 // Draw a background canvas, paint it with a fun chart looking animation
 const RandomAreaPlotAnimation = React.memo(() => {
+  const themeColors = useMemo(getLoginAnimationThemeColors, []);
```
mofojed, line 22: There's no way to change the theme on the login screen, but regardless - shouldn't this be dependent on the active theme anyway? In case we ever do add a changer to the login screen or we use this animation somewhere else or something.

packages/components/src/theme/theme-dark/theme-dark-components.css
```diff
@@ -1,13 +1,55 @@
...
 
+  /* Login */
+  --dh-color-login-form-bg: var(--dh-color-gray-400);
+  --dh-color-login-status-message: var(--dh-color-gray-600);
+
+  /* We should base this color on something from the palette, but for now just
+   * hard coding the original value */
+  --dh-color-login-animation-fg: #343e5d;
+  --dh-color-login-animation-bg: var(--dh-color-gray-75);
+  --dh-color-login-animation-grid: var(--dh-color-gray-300);
```
mofojed, line 18: Yeah I wonder what values we should actually be using here, unclear to me why this doesn't fit into our palette @dsmmcken 
Side note, I don't particularly like the name `login-animation` - yes, we use that animation only on the login screen, but there's nothing about the animation/component itself that is inherently tied to login; the animation could be used anywhere.

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -232,6 +234,28 @@ export function getExpressionRanges(value: string): [number, number][] {
...
+ */
+export function getLoginAnimationThemeColors(): LoginThemeColors {
+  const newThemeColors = resolveCssVariablesInRecord(
+    LOGIN_THEME_COLOR_VARIABLES
+  );
+
+  // Trim the alpha channel from the 8 digit hex colors
+
+  newThemeColors.animationBackground =
+    newThemeColors.animationBackground.substring(0, 7);
```
mofojed, line 248: Might make sense to bring the `normalizedOptionalAlpha` method and/or another method in `ColorUtils` for this.

mofojed (changes requested):

mofojed:
For the most part good, just the 404 link to bootstrap.scss and some minor cleanup suggestions to take a look at.

packages/code-studio/src/styleguide/colorUtils.ts
bmingles: The caveat here is that it can return the original value if it is not parseable, so not guaranteed to produce a hex value in all cases.

packages/code-studio/src/styleguide/Colors.tsx
```diff
@@ -66,25 +79,11 @@ function Colors(): React.ReactElement {
             in your .scss files when applicable for UI elements.
           </p>
           <p>
-            For refrence key colors are $primary for interactable elements,
+            For reference key colors are $primary for interactable elements,
             content-bg as background. White is same as gray-200. Gray-100 is
             used when white needs an active color. content-bg is between 800 and
             900.
```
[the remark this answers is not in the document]
dsmmcken: No, not really. We probably want to keep a translation section about sass -> css var equivelents, but the semantic variables section is now where we should draw from.

packages/components/scss/bootstrap_imports.scss
dsmmcken: the original import exists in the node_modules folder for bootstrap.

packages/code-studio/src/styleguide/Alerts.tsx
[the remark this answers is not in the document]
bmingles: yeah, Don told me to delete this so we don't have to include the bootstrap alerts styles which would involve more work to fix

packages/components/src/theme/theme-dark/theme-dark-components.css
dsmmcken: Needs to be composed with alpha now, and remove the animation from computing it's own alpha on primary-dark with a new constant.

Adjust the animation to have something like:
```js
// obviously, these need to be dynamically replaced with whatever cssvar -> js variable thing we do.
FOREGROUND_STROKE = hsla(    var(--dh-color-accent-hsl),    0.2  ); // something that looks like primary dark
FOREGROUND_FILL = hsla(    var(--dh-color-accent-hsl),    0.08  ); // new variable, with even more opacity
```
and the `FILL_OPACITY: 0.25` variable can be deleted.

packages/components/scss/bootstrap_imports.scss
bmingles: fixed

packages/components/src/theme/ThemeUtils.ts
bmingles: Added an `isAlphaOptional` flag to `normalizeCssColor` instead.

packages/code-studio/src/main/AppMainContainer.scss
```diff
@@ -182,7 +182,7 @@ $nav-space: 4px; // give a gap around some buttons for focus area that are in na
     flex-shrink: 0;
     background-image: linear-gradient(
       270deg,
-      rgba($background, 0) 0%,
+      hsla(var(--dh-color-bg-hsl), 0) 0%,
```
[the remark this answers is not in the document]
mofojed: The fourth parameter is defined as an `alpha-value`, but this is the second parameter (`s`), which is defined as a `percentage`:
> <legacy-hsl-syntax> = hsl( <hue> , [<percentage>](https://developer.mozilla.org/en-US/docs/Web/CSS/percentage) , [<percentage>](https://developer.mozilla.org/en-US/docs/Web/CSS/percentage) , [<alpha-value>](https://developer.mozilla.org/en-US/docs/Web/CSS/alpha-value)[?](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#question_mark) )

Pretty sure you're intentionally setting the saturation here as well, and not the alpha?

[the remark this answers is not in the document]
mofojed: Ahh okay my bad

packages/components/src/RandomAreaPlotAnimation.tsx
bmingles: I fixed this so that the animation now responds to theme changes. Added a sample to styleguide as well

packages/code-studio/src/styleguide/Colors.tsx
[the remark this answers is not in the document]
bmingles: @dsmmcken  and I discussed this in Slack. I've updated it accordingly

packages/code-studio/src/settings/SettingsMenu.scss
```diff
@@ -5,18 +5,18 @@ $settings-menu-width: 400px;
...
-$settings-menu-header-color: $foreground;
+$settings-menu-header-color: var(--dh-color-fg);
 
-$settings-menu-collapse-trigger-color: $foreground;
+$settings-menu-collapse-trigger-color: var(--dh-color-fg);
 
 $settings-menu-rule-bg: $gray-600;
 $settings-menu-rule-button-color: $gray-300;
 $settings-menu-rule-border: $gray-400;
 $settings-menu-rule-focused-border: $primary;
```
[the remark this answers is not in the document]
mofojed: Yup sounds good.

packages/components/src/theme/ThemeModel.ts
```diff
@@ -4,31 +4,106 @@ export type CssVariableStyleContent = `:root{${string}`;
...
+  | '--dh-color-accent-bg'
+  | '--dh-color-accent-hover-bg'
+  | '--dh-color-accent-down-bg'
+  | '--dh-color-accent-key-focus-bg'
+  | '--dh-color-login-form-bg'
+  | '--dh-color-login-status-message'
+  | '--dh-color-random-area-plot-animation-fg-fill'
+  | '--dh-color-random-area-plot-animation-fg-stroke'
+  | '--dh-color-random-area-plot-animation-bg'
+  | '--dh-color-random-area-plot-animation-grid'
```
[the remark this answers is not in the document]
bmingles: Sounds like @dsmmcken is saying the form colors should be specific to the form, so the issue here may be that the ThemeModel should not be responsible to specify these. I think these may not actually be required in the preload in DHC at all but will be needed for DHE, so I probably need to augment the preload to allow the app to specify additional vars. Then ThemeModel won't have to own these.

The other suggestions make sense. I'll change those as well.

packages/code-studio/src/styleguide/Swatch.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    const dhColorValue = getComputedStyle(ref.current).getPropertyValue(
+      dhColorVarName
+    );
+
+    setTooltip({
+      value: dhColorValue,
+      normalized: ColorUtils.normalizeCssColor(dhColorValue, true),
+    });
+  }, [ref]);
```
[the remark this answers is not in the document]
bmingles: I made this change with the exception that I am passing an elementRef to the new hook. This is necessary since `current` will be undefined until the effect runs.

packages/components/src/theme/ThemeModel.ts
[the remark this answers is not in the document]
bmingles: This turned out to be more involved than I want to include in this PR. I created https://github.com/deephaven/web-client-ui/issues/1679 to address it.

packages/components/scss/bootstrap_overrides.scss
bmingles: Yes, will follow up in separate PR

mofojed (changes requested):
One suggestion for cleanup

packages/components/src/RandomAreaPlotAnimation.tsx
```diff
@@ -5,23 +5,49 @@ import React, { useEffect, useState, useRef } from 'react';
...
+
 // Draw a background canvas, paint it with a fun chart looking animation
 const RandomAreaPlotAnimation = React.memo(() => {
+  const { activeThemes } = useTheme();
+
+  const [themeColors, setThemeColors] = useState<ThemeColors | null>(null);
+
+  useEffect(() => {
+    setThemeColors(getRandomAreaPlotAnimationThemeColors());
+  }, [activeThemes]);
```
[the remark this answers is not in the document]
bmingles: Most of the places that depend on resolving css vars need to go in an effect so that the latest data is available to the DOM right after a theme selection change. 

I'm not 100% sure, but I believe this happens because ThemeProvider renders style tags dynamically which would be in the same render as when useMemo gets calculated vs useEffect would fire after the DOM has actually updated the colors that are needed for the resolve.

For this particular example, it's easiest to see in the styleguide. When I tried useMemo and change the theme, the change trails behind till the next render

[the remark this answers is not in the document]
bmingles: I added a clarifying comment to the effect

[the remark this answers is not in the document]
mofojed: Okay - as we discussed in our meeting it would be nice if we pulled this logic up to the `ThemeProvider` so we don't need to do this pattern in so many places (here, `ChartThemeProvider`, `IrisGridThemeProvider`, etc).
So in `ThemeProvider`, the `value` it provides to the `ThemeContext.Provider` is only set _after_ a `useLayoutEffect` occurs in the `ThemeProvider`. Then any child can just do `useActiveThemes` (or maybe a `useThemeColors` type of hook?) and just get the active colours without needed to add their own `useLayoutEffect`
Can be a follow up ticket.

mofojed (approved):

mofojed (approved):

PR #1608, opened 2023-10-31
feat: Theming - Charts

### LINKED ISSUE #1572: Theming - Charts and Plots
Update charting and plotting to use theme variables.

Epic: https://github.com/deephaven/web-client-ui/issues/1537

bmingles:
Added ChartThemeProvider, added support for resolving css variables in chart theme, and updated chart theme to use css variables.

resolves #1572

BREAKING CHANGE: 
- ChartThemeProvider is now required to provide ChartTheme
- ChartModelFactory and ChartUtils now require chartTheme args


packages/components/src/theme/theme-dark/theme-dark-semantic-chart.css
```diff
@@ -0,0 +1,52 @@
...
+  --dh-color-chart-colorway-2: var(--dh-color-visual-green);
+  --dh-color-chart-colorway-3: var(--dh-color-visual-yellow);
+  --dh-color-chart-colorway-4: var(--dh-color-visual-purple);
+  --dh-color-chart-colorway-5: var(--dh-color-visual-orange);
+  --dh-color-chart-colorway-6: var(--dh-color-visual-red);
+  --dh-color-chart-colorway-7: var(--dh-color-visual-chartreuse);
+  --dh-color-chart-colorway-8: var(--dh-color-visual-fuchsia);
+  --dh-color-chart-colorway-9: var(--dh-color-visual-seafoam);
+  --dh-color-chart-colorway-10: var(--dh-color-visual-magenta);
+  --dh-color-chart-colorway-11: var(--dh-color-white);
```
mofojed, line 19: `colorway` should just be one variable - they could have less or more than 11 items in the colorway.

packages/chart/src/ChartThemeProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+ */
+export function ChartThemeProvider({
+  children,
+}: ChartThemeProviderProps): JSX.Element {
+  const { activeThemes } = useTheme();
+
+  const [chartTheme, setChartTheme] = useState<ChartTheme | null>(null);
+
+  // Running in an effect to ensure parent ThemeProvider has had a chance to add
+  // the <style> tags to the DOM that provide theme variables
```
mofojed, line 26: Not sure this comment is really necessary - if the `activeThemes` ever change, we need to re-initialize the chart theme, whether the theme provider has had a chance to add the style tags or not.
That being said, perhaps the `ThemeProvider` should gate until it has loaded?

packages/chart/src/ChartTheme.ts
```diff
@@ -1,20 +1,52 @@
...
+  area_color: string;
+  trend_color: string;
+  line_color: string;
+  error_band_line_color: string;
+  error_band_fill_color: string;
+  ohlc_increasing: string;
+  ohlc_decreasing: string;
+}
+
+export function initChartTheme(): Readonly<ChartTheme> {
```
mofojed, line 26: I'd probably name this `makeDefaultChartTheme()`, `getDefaultChartTheme()`, or `defaultChartTheme()`. The `init` makes me think it's initializing some sort of class or static property, when it's really just creating a new instance of the default theme.

packages/chart/src/useChartTheme.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import {
+  ChartThemeContext,
+  ChartThemeContextValue,
+} from './ChartThemeProvider';
+
+/**
+ * Hook to get the current `ChartThemeContextValue`.
+ */
+export function useChartTheme(): ChartThemeContextValue | null {
```
mofojed, line 11: Should we throw if there's no theme provided?

packages/console/src/HeapUsage.tsx
```diff
@@ -168,17 +170,17 @@ function HeapUsage({
                 colorway: ['#4878ea'],
                 xaxis: {
                   dtick: Math.round(monitorDuration / 6),
-                  gridcolor: ChartTheme.linecolor,
+                  gridcolor: chartTheme?.linecolor,
                   range: [lastTimestamp - monitorDuration, lastTimestamp],
-                  linecolor: ChartTheme.linecolor,
+                  linecolor: chartTheme?.linecolor,
```
mofojed, line 175: Right now we're just not setting these values if no chart theme is provided. I think either we should just throw if no theme is provided (in `useChartTheme`), or automatically provide the default theme if no provider is set.

packages/dashboard-core-plugins/src/ChartBuilderPlugin.tsx
```diff
@@ -44,8 +47,18 @@ export function ChartBuilderPlugin(
       table: Table;
     }) => {
       const { settings } = metadata;
-      const makeModel = (): Promise<ChartModel> =>
-        ChartModelFactory.makeModelFromSettings(dh, settings, table);
+      const makeModel = (): Promise<ChartModel> => {
+        if (chartTheme == null) {
+          throw new Error('Chart theme not available');
+        }
```
mofojed, line 53: Won't need this if `useChartTheme` throws.

packages/chart/src/ChartThemeProvider.tsx
bmingles: I actually tried that before, but it breaks Enterprise due to the way the component tree is structured.

Provider
  ThemeProvider
    MonacoBootstrap (depends on theming)
      ..
      App
        AppMainContainer (depends on theming)
        Login (can't block this from rendering)

bmingles: It's possible the `ChartThemeProvider` could be moved further down in the tree, but it would mean we can't bundle theming providers together such as the `ThemeBootstrap` does

packages/chart/src/useChartTheme.tsx
bmingles: I did this originally, but there were cases where the hook gets called before themes have loaded. I'll have to look at it again to remember the specifics

packages/chart/src/ChartThemeProvider.tsx
bmingles: Also to clarify, the comment is distinguishing that we have to load in useEffect to ensure any changes produced by setting the theme have completed. This would also apply if the theme is changed. This is in contrast to useMemo which will attempt to process the newly set theme before the style tags have actually been added and made the variables available. I can improve the comment if it helps, or let me know if there is something else I am missing.

packages/chart/src/useChartTheme.tsx
bmingles: I revisited this, and it seems to work to throw in `useChartTheme`. Must have been something else going on when I originally tested it.

packages/console/src/HeapUsage.tsx
bmingles: Changed to required

mofojed (approved):

PR #1613, opened 2023-11-02
fix: Plugins were re-registering on every re-render

mofojed:
- Forgot dependency array for `useEffect` hook


mattrunyon (approved):
Didn't realize the linter doesn't catch the case where you accidentally omit the dependency array

mofojed:
> Didn't realize the linter doesn't catch the case where you accidentally omit the dependency array

Yea same. I don't know if this behaviour changed in React or in our linter? Without the dependency array, it re-runs every render.

PR #1614, opened 2023-11-02
fix: Log figure errors, don't show infinite spinner

mofojed:
- Requires Core PR https://github.com/deephaven/deephaven-core/pull/4763
  - Also need to port that to Enterprise
- Actually look at the errors reported by the figure, and log them so we have a clearer idea of what the issue is when plots are reported as not working
- Also show the error message in the chart panel button bar. Clicking the button will display the full error message.
- Don't show the spinner infinitely if there are no series to load

### DIFF
--- a/packages/chart/src/Chart.tsx
+++ b/packages/chart/src/Chart.tsx
@@ -691,7 +691,7 @@ export class Chart extends Component<ChartProps, ChartState> {
           {shownError != null && (
             <>
               <div className="chart-error">{shownError}</div>
-              <CopyButton tooltip="Copy error" copy={shownError}>
+              <CopyButton tooltip="Copy Error" copy={shownError}>
                 Copy Error
               </CopyButton>
             </>
--- a/packages/chart/src/Chart.tsx
+++ b/packages/chart/src/Chart.tsx
@@ -692,7 +692,7 @@ export class Chart extends Component<ChartProps, ChartState> {
             <>
               <div className="chart-error">{shownError}</div>
               <CopyButton tooltip="Copy error" copy={shownError}>
-                Copy Versions
+                Copy Error
               </CopyButton>
             </>
           )}
--- a/packages/chart/src/Chart.tsx
+++ b/packages/chart/src/Chart.tsx
@@ -1,6 +1,7 @@
 import React, { Component, ReactElement, RefObject } from 'react';
 import deepEqual from 'deep-equal';
 import memoize from 'memoize-one';
+import { CopyButton, Popper } from '@deephaven/components';
 import {
   vsLoading,
   dhGraphLineDown,
@@ -57,10 +58,15 @@ interface ChartProps {
 
 interface ChartState {
   data: Partial<Data>[] | null;
+  /** An error specific to downsampling */
   downsamplingError: unknown;
   isDownsampleFinished: boolean;
   isDownsampleInProgress: boolean;
   isDownsamplingDisabled: boolean;
+
+  /** Any other kind of error */
+  error: unknown;
+  shownError: string | null;
   layout: Partial<Layout>;
   revision: number;
 }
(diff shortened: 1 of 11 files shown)


packages/chart/src/FigureChartModel.ts
```diff
@@ -265,6 +265,15 @@ class FigureChartModel extends ChartModel {
         ? this.dh.plot.DownsampleOptions.DISABLE
         : this.dh.plot.DownsampleOptions.DEFAULT
     );
+
+    if (this.figure.errors.length > 0) {
+      // We don't have a toast or anything we show for errors; just log the error for now
+      log.error('Errors in figure', this.figure.errors);
+    }
```
mattrunyon, line 272: Since this is just logged, will the panel be blank? Should we throw something to get caught and displayed by the error boundary?

mattrunyon (changes requested):
Need instructions on how to test this if I'm using the linked core PR

packages/chart/src/FigureChartModel.ts
mofojed: I think _ideally_ we'd display a toast, or some other type of error message; but also it's possible that we can still plot some of the other series that do not have errors, so I don't think we want to just throw and display the error at the error boundary.

mofojed:
@mattrunyon updated so you can see the full error message from the UI: 
image

packages/chart/src/FigureChartModel.ts
mofojed: It now displays a button in the button bar.

mofojed:
Use the linked Core PR with Groovy, and test with this snippet:
```
import static io.deephaven.csv.CsvTools.readCsv
import static io.deephaven.api.agg.Aggregation.AggAvg

cryptoTrades = readCsv("https://media.githubusercontent.com/media/deephaven/examples/main/CryptoCurrencyHistory/CSV/CryptoTrades_20210922.csv")

agg_list = [
    AggLast("Close = Price"),\
    AggFirst("Open = Price"),\
    AggMax("High = Price"),\
    AggMin("Low = Price"),
]

btcBin = cryptoTrades.where(FilterOr.of(Filter.from("Instrument = `BTC/USD`", "Instrument = `ETH/USD`")))
    .update("TimestampBin = lowerBin(Timestamp, HOUR)")
tOHLC = btcBin.aggBy(agg_list, "Instrument", "TimestampBin")

plotByOHLC = ohlcPlotBy("Instrument", tOHLC, "TimestampBin", "Open", "High", "Low", "Close", "Instrument").show()

btcBin2 = cryptoTrades.where("Instrument in `BTC/USD`, `ETH/USD`")
    .update("TimestampBin = lowerBin(Timestamp, HOUR)")
tOHLC2 = btcBin2.aggBy(agg_list, "Instrument", "TimestampBin")

plotByOHLC2 = ohlcPlotBy("Instrument", tOHLC2, "TimestampBin", "Open", "High", "Low", "Close", "Instrument").show()
```

I then commented out the changes in FigureWidgetTranslator part from that linked PR so that same snippet produces a plot with an error, and the error appears correctly.

packages/chart/src/Chart.tsx
```diff
@@ -617,6 +665,23 @@ export class Chart extends Component<ChartProps, ChartState> {
...
+          isShown={shownError != null}
+          onExited={this.handleErrorClose}
+          closeOnBlur
+          interactive
+        >
+          {shownError != null && (
+            <>
+              <div className="chart-error">{shownError}</div>
+              <CopyButton tooltip="Copy error" copy={shownError}>
+                Copy Versions
```
mattrunyon, line 680:
```suggestion
                Copy Error
```

packages/chart/src/FigureChartModel.ts
```diff
@@ -265,6 +265,12 @@ class FigureChartModel extends ChartModel {
         ? this.dh.plot.DownsampleOptions.DISABLE
         : this.dh.plot.DownsampleOptions.DEFAULT
     );
+
+    if (this.figure.errors.length > 0) {
+      // We don't have a toast or anything we show for errors; just log the error for now
```
mattrunyon, line 270: I think this will fail in DHE because `getErrors` is not marked as `@JsProperty` there, but it's added in the core PR this relies on. Either need a ticket/PR adding the decorator to the DHE JS API or optional chaining here

Also, you can remove the comment

mattrunyon (changes requested):

packages/chart/src/FigureChartModel.ts
mofojed: Yes, I have a branch already for that. I've created a ticket: https://deephaven.atlassian.net/browse/DH-15932

mattrunyon (approved):

PR #1620, opened 2023-11-03
feat: forward and back buttons for organize column search

### LINKED ISSUE #1529: Add forward and back buttons to organize column search
- Initial search should still select all items, and scroll into view the first item.
- Pressing left/right arrow should scroll into view the previous/next item, and select just that item. Show N of M for indicator
- Navigation should loop, pressing previous on first should go to last item and vice versa.

Note make sure padding is increased on the input when the buttons appear, so that it doesn't overlap the search text.

image

ethanalvizo:
Closes #1529


packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -1049,6 +1109,14 @@ class VisibilityOrderingBuilder extends Component<
       treeItems
     );
 
+    const queryParams = {
```
ethanalvizo, line 1112: i thought this would be nicer to pass in to SearchInput rather than creating two new optional props. Could revert to that if preferred though

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -168,6 +179,50 @@ class VisibilityOrderingBuilder extends Component<
...
+    });
+  }
+
+  /**
+   * Change the selected column to either the next or previous column
+   *
+   * @param direction The direction to move the selection
+   */
+
+  changeSelectedColumn(direction: 'forward' | 'back'): void {
```
ethanalvizo, line 195: i feel this could use a better name but not too sure what. This does change the selected column as the name implies but only from the previously queried list

dsmmcken (changes requested):
> The column search going back to the selected item, instead of picking up from your current selection seems counter intuitive to most find interfaces.  To wit, if I search for “log” in intellij; hit the button a few times to cycle; then select something else with my mouse and hit the button again it picks up the search from my cursor not the “n / m” I was on.

Arrows should go to the previous/next one before/after a manually selected index:

ex.
Search A, click BB, hit next get AC.
AA
AB
BB
AC
AD

dsmmcken:
My guess is a react re-render is interrupting the scroll.

ethanalvizo:
> My guess is a react re-render is interrupting the scroll.

fixed in the latest version now. It wasn't a re-render. My newest changes just broke the previous way I was passing in which item to focus on. All good now though!

dsmmcken (approved):
Approved for functionality, didn't code review.

PR #1621, opened 2023-11-03
fix: Fixing grid colors and grays

bmingles:
- Styleguide didn't have 850 actually mapped.
- dh-gray-100 is mapped to content-bg but doesn't have a Bootstrap gray-xxx counterpart, so I added a Gray-8XX for illustration while we are still mapping things
- Grid row was mapped to dh-gray-200 but should have been dh-gray-300
- I added calculations for the hsl for grays instead of the all zero hue based ones I had before. They are still not exact but closer. I'm looking into fixing the rounding error.

#1572


dsmmcken (changes requested):
--dh-color-grid-selection-outline-casing fix?

dsmmcken (approved):

Issue #1601, opened 2023-10-30
DateInput - onChange date value should not be optional


bmingles:
In `packages/components/src/DateInput.tsx` the `onChange` prop has it's date arg defined as optional. It doesn't appear that this is ever actually the case. We should update it so that consumers don't have to handle the undefined case just to appease TypeScript.

sakibian:
Can I work on this?

bmingles:
@sakibian Yes you may. Here are our contribution guidelines https://github.com/deephaven/web-client-ui/blob/fc543aab57189d5378ae1ebb280fc8024da9ce86/CONTRIBUTING.md

PR #1622, opened 2023-11-03
fix: Date argument non-optional for the onChange prop

sakibian:
Closes: #1601

### DIFF
--- a/packages/code-studio/src/styleguide/Inputs.tsx
+++ b/packages/code-studio/src/styleguide/Inputs.tsx
@@ -85,10 +85,6 @@ function Inputs(): React.ReactElement {
     []
   );
 
-  const handleChange = (date: string) => {
-    
-  };
-
   return (
     <div className="style-guide-inputs">
       <h2 className="ui-title">Inputs</h2>
@@ -334,7 +330,7 @@ function Inputs(): React.ReactElement {
             <TimeInput />
             <br />
             <h5>Date Input</h5>
-            <DateInput onChange={handleChange} />
+            <DateInput />
             <br />
             <h5>DateTime Input</h5>
             <DateTimeInput />
--- a/packages/code-studio/src/styleguide/Inputs.tsx
+++ b/packages/code-studio/src/styleguide/Inputs.tsx
@@ -85,6 +85,10 @@ function Inputs(): React.ReactElement {
     []
   );
 
+  const handleChange = (date: string) => {
+    console.log('Date changed:', date);
+  };
+
   return (
     <div className="style-guide-inputs">
       <h2 className="ui-title">Inputs</h2>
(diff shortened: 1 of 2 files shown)


sakibian:

packages/components/src/DateInput.tsx
```diff
@@ -12,7 +12,7 @@ const DATE_FORMAT = 'YYYY-MM-DD';
 
 type DateInputProps = {
   className?: string;
-  onChange?: (date?: string) => void;
+  onChange: (date: string) => void; // Remove the question mark to make date non-optional
```
mofojed, line 15: No need to add this comment here. 
```suggestion
  onChange: (date: string) => void;
```

packages/components/src/DateInput.tsx
```diff
@@ -23,7 +23,7 @@ const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
   (props: DateInputProps, ref) => {
     const {
       className = '',
-      onChange = () => undefined,
+      onChange = () => undefined, // Provide a default value for onChange
```
mofojed, line 26:
```suggestion
      onChange = () => undefined,
```

packages/components/src/DateInput.tsx
```diff
@@ -62,16 +62,6 @@ const DateInput = React.forwardRef<HTMLInputElement, DateInputProps>(
     );
   }
 );
-
 DateInput.displayName = 'DateInput';
 
-DateInput.defaultProps = {
```
mofojed, line 68: Good cleanup!

mofojed (changes requested):
Just remove the unnecessary comments. Thank you for the contribution!

mofojed (dismissed):
Thank you for the contribution!

packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -85,6 +85,10 @@ function Inputs(): React.ReactElement {
     []
   );
 
+  const handleChange = (date: string) => {
```
mofojed, line 88: We've still got linter issues here - our formatting is fairly strict, we include some extensions and VS Code settings for automatically formatting code on save: https://github.com/deephaven/web-client-ui/blob/main/.vscode/settings.json

For this case though, I'm going to recommend the following:
- Make the `onChange` prop on `DateInput` optional still (the ticket was specifically for the `date` parameter on the callback to be marked as required, we don't need to make `onChange` itself required as well)
- Don't add an `onChange` here in the styleguide.

That would make it consistent with our existing behaviour/other components (such as `DateTimeInput`).

mofojed (changes requested):

mofojed (approved):

PR #1625, opened 2023-11-06
feat: Add `LayoutManagerContext` and `useLayoutManager`

mofojed:
- Each dashboard provides the `LayoutManagerContext`


packages/dashboard/src/layout/useLayout.ts
```diff
@@ -0,0 +1,18 @@
+import { useContext } from 'react';
+import GoldenLayout from '@deephaven/golden-layout';
+import LayoutContext from './LayoutContext';
+
+/**
+ * Retrieve the current layout from the context
+ */
+function useLayout(): GoldenLayout {
```
mattrunyon, line 8: Calling this `layout` has always been a little confusing to me. I wonder if we should call it `useGoldenLayout` instead. I think there are some other spots where `layout` is the object that specifies the actual layout vs the instance of golden layout (often `glLayout` or something like that)

Or `useLayoutManager` which I think is what GoldenLayout calls the class

packages/dashboard/src/layout/useLayoutManager.ts
```diff
@@ -5,7 +5,7 @@ import LayoutContext from './LayoutContext';
 /**
  * Retrieve the current layout from the context
```
mattrunyon, line 6: Update the comment to layout manager

Also update the context to `LayoutManagerContext` to match

mattrunyon (changes requested):

packages/dashboard/src/layout/useLayoutManager.ts
```diff
@@ -0,0 +1,18 @@
+import { useContext } from 'react';
+import LayoutManager from '@deephaven/golden-layout';
+import LayoutManagerContext from './LayoutManagerContext';
+
+/**
+ * Retrieve the current LayoutManager from the context
+ */
+function useLayoutManager(): LayoutManager {
+  const layout = useContext(LayoutManagerContext);
```
mattrunyon, line 9: Might as well change this variable name while we're renaming it to `layoutManager`

packages/dashboard/src/layout/useLayoutManager.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (layout == null) {
+    throw new Error(
+      'Layout not available, did you add a LayoutManagerContext.Provider to the tree?'
```
mattrunyon, line 12:
```suggestion
      'LayoutManager not available, did you add a LayoutManagerContext.Provider to the tree?'
```

mattrunyon (changes requested):

mattrunyon (approved):

PR #1628, opened 2023-11-06
feat: Added test:debug script

### LINKED ISSUE #1627: Jest - Debugging Script
As a developer, I want to be able to debug unit tests. We should add a `test:debug` script to package.json to make this easier.

bmingles:
Added a `test:debug` npm script.

resolves #1627

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,12 +8,7 @@
     }
   ],
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "stylelint.validate": [
-    "css",
-    "html",
-    "sass",
-    "scss"
-  ],
+  "stylelint.validate": ["css", "html", "sass", "scss"],
   "launch": {
     "version": "0.2.0",
     "configurations": [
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,6 +94,26 @@ Note that log messages from other sources such as react prop types will still be
...
+- Create or modify the `.vscode/launch.json` config file by adding the following launch config:
+
+   ```jsonc
+   {
+      "type": "node",
+      "request": "attach",
+      "name": "Attach",
+      "port": 9229
+   }
+   ```
```
mattrunyon, line 109: You can add this to `.vscode/settings.json` and commit it. Then it's available to everyone w/o any extra setup. We have a launch chrome w/ debugger attached config in there already

mattrunyon (changes requested):
Nice. I think the version I had for debugging before didn't work in watch mode, so nice to have it (as well as have it documented)

README.md
bmingles: It looks like `launch.settings` is gitignored, so I don't think there's a good way to add it that gets committed.

package.json
```diff
@@ -41,6 +41,7 @@
     "start:embed-grid": "lerna run start --scope=@deephaven/embed-grid --stream",
     "pretest": "npm run build:necessary",
     "test": "jest --watch --changedSince origin/main",
+    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --watch --changedSince origin/main",
```
mattrunyon, line 44: I think I'd remove the `changedSince` flag for debug. I'd assume that we'll just specify the test we want when running this. You should be able to just do `npm run test:debug SomeFile`. Works without the double dash since it's not a flag you're passing

.vscode/settings.json
```diff
@@ -24,8 +19,35 @@
...
+        "name": "Debug Jest Tests",
+        "type": "node",
+        "request": "launch",
+        "runtimeArgs": [
+          "--inspect-brk",
+          "${workspaceRoot}/node_modules/.bin/jest",
+          "--runInBand",
+          "--watch",
+          "${input:pickFile}"
+        ],
```
mattrunyon, line 39: You can do this I believe so that it just runs the npm script and will stay in sync if that changes for some reason

```suggestion
        "runtimeExecutable": "npm",
        "runtimeArgs": ["run-script", "test:debug", "${input:pickFile}"],
```

mattrunyon (changes requested):
Debug w/ input for file is sweet. Just a few small things

package.json
bmingles: done

.vscode/settings.json
bmingles: Not sure why, but running npm here always auto breaks inside the jest module which is kind of annoying. Going to leave it as is for better UX. Should be rare the script would get changed anyway

mattrunyon (approved):

mattrunyon (changes requested):
Wait 1 more suggestion. Only run unit tests, not all tests. Add `--config jest.config.unit.cjs` to the command in both spots

mattrunyon (approved):

PR #1631, opened 2023-11-09
fix: Infinite loop with grid rendering

mofojed:
- Caused by #1626
- The children on IrisGrid were getting re-rendered as a result of IrisGrid.handleViewChanged getting called after updating the canvas in Grid.componentDidUpdate
- Another fix would be to memoize metrics and not emit a view change if they are exactly the same as previous metrics, but thought that was a bigger change (need to deep equals a bunch of maps and arrays in the metrics)
- Tested by opening up a table with React dev tools highlighting re-renders. Table did not re-render when not being interacted with.

### DIFF
--- a/packages/utils/src/ObjectUtils.tests.ts
+++ b/packages/utils/src/ObjectUtils.tests.ts
@@ -1,0 +1,66 @@
+import { getChangedKeys } from './ObjectUtils';
+
+describe('getChangedKeys', () => {
+  it('should get changed keys', () => {
+    const oldObject = {
+      foo: 'bar',
+      baz: 'qux',
+      quux: 'corge',
+    };
+    const newObject = {
+      foo: 'bar',
+      baz: 'quux',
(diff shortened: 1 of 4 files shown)


packages/utils/src/ObjectUtils.tests.ts
```diff
@@ -0,0 +1,66 @@
...
+      baz: 'qux',
+      quux: 'corge',
+    };
+    expect(getChangedKeys(oldObject, newObject)).toEqual([]);
+  });
+  it('should get changed keys when both objects are the same and empty', () => {
+    const oldObject = {};
+    const newObject = {};
+    expect(getChangedKeys(oldObject, newObject)).toEqual([]);
+  });
```
mattrunyon, line 65: This test is the same as 2 tests above it

mattrunyon (changes requested):

mattrunyon (approved):

PR #1637, opened 2023-11-10
refactor: replace html selects with Select component

dsmmcken:
Replaces all the places we used raw html select components with the Select component.

This was orignally needed for themeing so we could re-style the Select component, but we came up with an alternate solution.

### DIFF
--- a/packages/code-studio/src/styleguide/Inputs.tsx
+++ b/packages/code-studio/src/styleguide/Inputs.tsx
@@ -235,25 +235,38 @@ function Inputs(): React.ReactElement {
         <div className="col">
           <div className="form-group">
             <h5>Selection Menu</h5>
-            <select defaultValue="0" className="custom-select">
+            <Select
+              onChange={v => {
+                // no-op
+              }}
+              defaultValue="0"
+              className="custom-select"
+            >
               <option disabled value="0">
                 Custom Selection
               </option>
               <option value="1">One</option>
               <option value="2">Two</option>
               <option value="3">Three</option>
-            </select>
+            </Select>
           </div>
 
           <div className="form-group">
-            <select defaultValue="0" className="custom-select" disabled>
+            <Select
+              onChange={v => {
+                // no-op
+              }}
+              defaultValue="0"
+              className="custom-select"
+              disabled
+            >
               <option disabled value="0">
                 Custom Selection
               </option>
               <option value="1">One</option>
               <option value="2">Two</option>
               <option value="3">Three</option>
-            </select>
+            </Select>
           </div>
 
           <div className="form-group">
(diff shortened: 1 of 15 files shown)


packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -235,25 +235,38 @@ function Inputs(): React.ReactElement {
         <div className="col">
           <div className="form-group">
             <h5>Selection Menu</h5>
-            <select defaultValue="0" className="custom-select">
+            <Select
+              onChange={v => {
```
bmingles, line 239: I would break out a `noOp` function outside of the component as minor optimization to not cause re-renders

packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -235,25 +235,38 @@ function Inputs(): React.ReactElement {
...
               <option value="2">Two</option>
               <option value="3">Three</option>
-            </select>
+            </Select>
           </div>
 
           <div className="form-group">
-            <select defaultValue="0" className="custom-select" disabled>
+            <Select
+              onChange={v => {
```
bmingles, line 256: Same as other

packages/components/src/Select.tsx
```diff
@@ -1,50 +1,75 @@
...
+        onKeyDown={onKeyDown}
+        defaultValue={defaultValue}
+        value={value}
+        name={name}
+        title={title}
+        disabled={disabled}
+        data-testid={dataTestId}
+        aria-label={ariaLabel}
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {...rest}
```
bmingles, line 65: If we are going to spread props, any reason not to include others? Seems like most of them could be included in `rest`

dsmmcken: Like what?

packages/code-studio/src/styleguide/Inputs.tsx
dsmmcken: For the styleguide I think this is fine.

packages/components/src/Select.tsx
bmingles: Looks like this would work

```typescript
const {
  children,
  onChange,
  className,
  'data-testid': dataTestId,
  'aria-label': ariaLabel,
  ...rest
} = props;
```

```jsx
<select
  ref={ref}
  className={classNames('custom-select', className)}
  onChange={handleChange}
  data-testid={dataTestId}
  aria-label={ariaLabel}
  {...rest}
>
  {children}
</select>
```

bmingles (changes requested):
Added example of simplified props

bmingles (approved):
LGTM

Issue #1634, opened 2023-11-10
Theming - Inline svgs need to support dynamic css prop colors


bmingles:
We use inline svgs for certain icons. In some cases these are created via string concatenation with a Bootstrap color variable which is available at compile time. This does not work with dynamic CSS custom props which can change at runtime. We need to adjust this approach to support dynamic variables for theming.

- Golden layout icons (e.g. .lm_maximise)
- Form validation icons
- Select input arrows

It should be possible to address some of these using svg masks, but others will likely require a .js based option to dynamically update svgs when themes change (should be able to store the svgs in css props and update them, but need to test this)

PR #1639, opened 2023-11-13
feat: Styleguide regression tests

bmingles:
- Enabled styleguide in production builds
- Added a base set of e2e regression snapshot tests on the styleguide
- Changed router basename to be based on document.baseURI

#1634


packages/code-studio/vite.config.ts
```diff
@@ -19,13 +19,6 @@ export default defineConfig(({ mode }) => {
   // These are paths which should be proxied to the core server
   // https://vitejs.dev/config/server-options.html#server-proxy
   const proxy = {
-    // Proxy styleguide here instead of as a route in our app router
-    // That way, it is not included in the production build
```
mofojed, line 23: We need to update the `README.md` in `packages/code-studio` to fix the link to the styleguide; right now it just has http://localhost:4000/styleguide, should be http://localhost:4000/ide/styleguide

Also noting that this does not affect bundle size since you're using code splitting to only load the style guide when we go to that route (I think...), _however_ this also means that anybody can open up the styleguide. Which I guess is a good thing if we want people to develop their own themes.

packages/code-studio/src/main/AppRouter.tsx
```diff
@@ -7,16 +7,20 @@ import {
...
 
-const basename = document.location.pathname;
+const StyleGuideRoot = React.lazy(() => import('../styleguide/StyleGuideRoot'));
+
+const baseURI = new URL(document.baseURI).pathname;
+
 function AppRouter(): ReactElement {
   return (
-    <Router basename={basename}>
+    <Router basename={baseURI}>
```
mofojed, line 16: Reviewing the [ReactRouter documentation](https://v5.reactrouter.com/web/api/BrowserRouter/basename-string): 
> A properly formatted basename should have a leading slash, but no trailing slash.

Right now, all of our `BASE_URL`'s defined do have a trailing slash (`./`, `/ide/`). I don't think this manifests any issues, and some of our code that uses the `baseURI` expects a trailing slash there... so I wouldn't bother changing anything right now.

packages/code-studio/src/styleguide/Alerts.tsx
```diff
@@ -16,7 +18,7 @@ class Alerts extends Component {
       Alerts.renderAlert(brand)
     );
     return (
-      <div>
+      <div {...sampleSectionIdAndClasses('alerts')}>
```
mofojed, line 21: I probably would've thought to put this in a component instead that just takes props `name` and `className` - think it would be clearer. `Section`, `SectionView`, `SectionFlex` would cover `div` vs. `View` vs `Flex` cases. Not really worth refactoring though.

tests/styleguide.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        el.hasAttribute('disabled')
+      );
+
+      if (isDisabled) {
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+
+      // Focus
+      await button.focus({ timeout: 500 });
```
mofojed, line 66: We could probably set the `actionTimeout` to something like `500`, or use [page.setDefaultTimeout](https://playwright.dev/docs/api/class-page#page-set-default-timeout) to set a default and not have to specify it in all these tests.

tests/styleguide.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    for (let j = 0; j < buttonCount; j += 1) {
+      const button = buttons.nth(j);
+
+      const isDisabled = await button.evaluate(el =>
+        el.hasAttribute('disabled')
+      );
+
+      if (isDisabled) {
+        // eslint-disable-next-line no-continue
+        continue;
```
mofojed, line 62: No checks for a disabled button state? Why not have a snapshot for this case as well?

tests/styleguide.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      const [tagName, type, isDisabled] = await input.evaluate(el => [
+        el.tagName.toLowerCase(),
+        el.getAttribute('type'),
+        el.hasAttribute('disabled'),
+      ]);
+
+      if (isDisabled) {
+        // eslint-disable-next-line no-continue
+        continue;
+      }
```
mofojed, line 108: Ditto - why no tests for disabled state?

tests/styleguide.spec.ts-snapshots/alerts-chromium-linux.png
mofojed, line 1: Random side note - it's annoying there's no space after `Well done!`. We should probably include a space there.

tests/styleguide.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  await expect(sampleSections).toHaveCount(39);
+});
+
+test.afterEach(async () => {
+  await page.close();
+});
+
+// Iterate over all sample sections and take a screenshot of each one.
+test('UI regression test', async () => {
+  for (let i = 0; i < (await sampleSections.count()); i += 1) {
```
mofojed, line 24: Should only need to get the count once:
```suggestion
  const sectionCount = await sampleSections.count();
  for (let i = 0; i < sectionCount; i += 1) {
```

tests/styleguide.spec.ts-snapshots/alerts-firefox-linux.png
mofojed, line 1: Noticed the Menu button is visible in these screenshots... Wish there was some way to get just the section screenshot without the menu overlaid on top, but probably not important.

tests/utils.ts
```diff
@@ -22,6 +22,15 @@ export function generateVarName(prefix = 'v'): string {
...
 }
 
+/**
+ * Log browser type and version.
+ * @param browser
+ */
+export function logBrowserInfo(browser: Browser): void {
+  // eslint-disable-next-line no-console
+  console.log('Browser:', browser.browserType().name(), browser.version());
+}
```
mofojed, line 32: Honestly, we should probably just log the browser version, app version and other information when the app starts up. Would be useful for the support logs as well.

tests/styleguide.spec.ts-snapshots/theme-color-palette-chromium-linux.png
mofojed, line 1: Annoying that the menu button and arrow are appearing on top of this screenshot.

tests/styleguide.spec.ts-snapshots/spectrum-overlays-webkit-linux.png
mofojed, line 1: Menu and arrow up button do not appear on this screenshot...
So my concern with the menu/arrow showing up in some screenshots and not in others is that it is now dependent on where that section is; if `Overlays` is no longer near the bottom, then the menu button will probably start appearing in it; if `Wells` is no longer at the very bottom, then the little array won't appear in the screenshot.

Ideally the screenshots should not include the menu or arrow at all.

packages/code-studio/src/styleguide/Alerts.tsx
dsmmcken: Do something like:
`await page.locator('selector').evaluate(element => element.style.opacity= '0');`

packages/code-studio/src/main/AppRouter.tsx
bmingles: Oh, I missed that baseURI can have a trailing slash. I must have tested on a route that didn't

tests/utils.ts
bmingles: Updated

tests/styleguide.spec.ts-snapshots/spectrum-overlays-webkit-linux.png
bmingles: I used Don's approach and set opacity to 0 for the 2 navigation components

tests/styleguide.spec.ts-snapshots/alerts-firefox-linux.png
bmingles: This worked. Thanks

tests/styleguide.spec.ts
bmingles: Fixed

packages/code-studio/vite.config.ts
bmingles: FWIW, @dsmmcken is not a fan of surfacing the styleguide to users. I don't think it would be hard to conditionally show it based on an env var + vite config

packages/code-studio/src/main/AppRouter.tsx
mofojed: Sure, that'd be fine. May as well match what the docs expect.

Digging in the code, it doesn't _really_ matter, the `basename` prop just gets passed through to `createBrowserHistory`, which actually removes the trailing slash: https://github.com/remix-run/history/blob/6104a6a2e40ae17a47a297621afff9a6cb184bfc/modules/createBrowserHistory.js#L51
Side note their code for removing the trailing slash is just:
```
export function stripTrailingSlash(path) {
  return path.charAt(path.length - 1) === '/' ? path.slice(0, -1) : path;
}
```

tests/utils.ts
```diff
@@ -22,6 +31,25 @@ export function generateVarName(prefix = 'v'): string {
...
+  // eslint-disable-next-line no-restricted-syntax
+  for (const launcher of launchers) {
+    // eslint-disable-next-line no-await-in-loop
+    const browser = await launcher.launch();
+
+    // eslint-disable-next-line no-console
+    console.log('Browser:', browser.browserType().name(), browser.version());
+
+    browser.close();
+  }
```
mofojed, line 50: Take a look at the description of this rule: https://eslint.org/docs/latest/rules/no-await-in-loop
> However, performing an await as part of each operation is an indication that the program is not taking full advantage of the parallelization benefits of async/await.
> Usually, the code should be refactored to create all the promises at once, then get access to the results using Promise.all(). Otherwise, each successive operation will not start until the previous one has completed.

I think that applies in this case - we should be able to launch all the browsers in parallel and it should be faster (I think?).

In any case, I'd write this something like this instead of disabling the warnings:
```suggestion

  await Promise.all(
    launchers.map(async launcher => {
      const browser = await launcher.launch();
      // eslint-disable-next-line no-console
      console.log('Browser:', browser.browserType().name(), browser.version());

      return browser.close();
    })
  );
```

mofojed (approved):

tests/utils.ts
bmingles: I'll try this in my next PR to avoid having to wait for the e2e tests on this one

PR #1641, opened 2023-11-14
feat: forward and back button for organize column search

### LINKED ISSUE #1529: Add forward and back buttons to organize column search

ethanalvizo:
Closes #1529


packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -1067,8 +1227,9 @@ class VisibilityOrderingBuilder extends Component<
           <SearchInput
             className="visibility-search"
             value={searchFilter}
-            matchCount={searchFilter ? selectedColumns.size : undefined}
+            matchCount={searchFilter ? matchCount : undefined}
             onChange={this.handleSearchInputChange}
+            queryParams={queryParams}
```
ethanalvizo, line 1232: i thought passing in queryParams would be nicer than adding 2 optional params into SearchInput. Open to doing the other method though

packages/components/src/SearchInput.tsx
```diff
@@ -52,7 +79,47 @@ class SearchInput extends PureComponent<SearchInputProps> {
...
+              queryParams.changeQueriedColumnIndex('forward');
+            }}
+            icon={vsArrowRight}
+            tooltip="Next match"
+          />
+        </>
+      );
+    } else {
+      matchCountSection = matchCount > 0 && (
+        <span className="match_count">{matchCount}</span>
```
mofojed, line 119: Class names shouldn't have an underscore in them like this. HTML attributes and CSS properties should be [kebab case](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case), not [snake case](https://developer.mozilla.org/en-US/docs/Glossary/Snake_case).

Also not sure why you changed the name - should just keep the name `search-match`, and update the existing CSS definition for it (right now it's defined in `SearchInput.scss` but not used anymore).

mofojed (changes requested):

dsmmcken (changes requested):
Would you murder me if I ask for ~one~ two more things?

1. When in the find field and there are results, enter and shift+enter should act as keyboard shortcuts for next/previous match, and should be exposed as the shortcut on the tooltip as well.

    Same shortcut as vs code:
    image

2. Also the left button should say "Previous match (Shift+Enter)" and the right "Next match (Enter)". They are currently say the same thing.
  image

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -1201,9 +1201,9 @@ class VisibilityOrderingBuilder extends Component<
       treeItems
     );
 
-    const queryParams = {
-      queriedColumnIndex,
-      changeQueriedColumnIndex: (direction: 'forward' | 'back') =>
+    const cursor = {
+      index: queriedColumnIndex as number,
```
mofojed, line 1205: You shouldn't need this cast, and in fact it's pointing out an error - you could be passing `index: undefined` here, which is not what `SearchInput` is expecting.
Instead, you should just check if it is not set and just return `undefined` for that instead of the parameter, e.g.:
```
    const cursor =
      queriedColumnIndex != null
        ? {
            index: queriedColumnIndex,
            next: (direction: 'forward' | 'back') =>
              this.changeSelectedColumn(direction),
          }
        : undefined;
```

mofojed (changes requested):

dsmmcken (changes requested):
Searching then doing previous/next starts at the end instead of the start.

1. search a column
2. hit next

expected:
Item selected is 1 of N

actual:
Item selected is Nth of N (which would be expected if I pressed previous)

dsmmcken (changes requested):
Holding down the Enter key after searching will accumulate changes and then lock the UI while it applies all of them at once, rather than on key repeat.

@mofojed for advice.
Maybe because contextactions? Could use key events directly if that gives better control.

packages/components/src/SearchInput.tsx
```diff
@@ -1,9 +1,14 @@
...
-import { vsSearch } from '@deephaven/icons';
+import { vsArrowLeft, vsArrowRight, vsSearch } from '@deephaven/icons';
 import classNames from 'classnames';
+import Button from './Button';
 import './SearchInput.scss';
 
+interface QueryParams {
+  queriedColumnIndex: number | undefined;
+  changeQueriedColumnIndex: (direction: 'forward' | 'back') => void;
+}
```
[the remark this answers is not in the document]
ethanalvizo: I commented on this below

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
ethanalvizo: I unresolved one of your previous comments. I didn't think it through properly but queriedColumnIndex should be able to be undefined. When the user searches initially there is no index yet. But cursor should still be passed in since we need the next function for the forward and back buttons

dsmmcken:
> Holding down the Enter key after searching will accumulate changes and then lock the UI while it applies all of them at once, rather than on key repeat.
> 
> @mofojed for advice. Maybe because contextactions? Could use key events directly if that gives better control.

This is probably the issue #1650, poor drag performance. Can be ignored for this ticket, no changes needed here.

dsmmcken (changes requested):
dimisses

Find gets dismissed if you search, click an item, click back in search, click back on an item.

dsmmcken (changes requested):
scroll_into_view

Scroll into view isn't appropriately taking into account groups.

Have a table with a lot of columns, create two groups, one at the top with a bunch of columns, and a second group a few pages off screen. Search for the group name and attempt to have it scrolled into view. I think it ends up higher by the size of the first group.

dsmmcken:
image_360

I somehow managed to get this once, but completely unable to reproduce. 

All my other issues are resolved.

Approved for functionality.

mofojed (approved):

dsmmcken (approved):

PR #1644, opened 2023-11-15
fix: Changes for Deephaven UI embedding widget plugins

mattrunyon:
These changes are needed so Deephaven UI can embed widget plugins. This also creates a difference between the widget component and the widget panel component.

Will open a companion UI plugin PR to utilize these in Deephaven UI.

Tested the panels still function as intended for grid, chart, and dataframe.


packages/dashboard-core-plugins/src/GridWidgetPlugin.tsx
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    init();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [dh, fetch]);
+
+  // eslint-disable-next-line react/jsx-props-no-spreading
```
mofojed, line 36: You're not using spreading anyway
```suggestion
```

packages/dashboard-core-plugins/src/panels/PandasPanel.tsx
```diff
@@ -11,6 +17,42 @@ import ConnectedIrisGridPanel, {
 } from './IrisGridPanel';
 import './PandasPanel.scss';
 
+export function PandasReloadButton({
```
mofojed, line 20: Should be in a separate file.

mofojed (changes requested):

mofojed (approved):

mofojed (approved):

PR #1648, opened 2023-11-16
fix: Default to `Skip` operation instead of `Sum` operation

### LINKED ISSUE #1355: Aggregate Columns ignores columns selection
**Description**

image

**Steps to reproduce**

Using the following table:
```

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",

complex_ticking = simple_ticking.update([
        "MyBigDecimal= new java.math.BigDecimal(i+i/10)",
        "MyBigInteger= new java.math.BigInteger(Integer.toString(i))"
])                          
```
1. select from table options sidebar "Aggregate Columns"
2. Select sum from drop down and click add aggregation
3. From the sum button click the pencil to edit aggregations
4. De-select a number of columns

**Expected results**

Sum is only computed for the selected columns

**Actual results**

Sum is selected for all columns

**Versions**
Vermillion and 

Engine Version: 0.26.0
Web UI Version: 0.40.4
Java Version: 17.0.7
Barrage Version: 0.5.0

mofojed:
- Previously we were unable to remove columns from an aggregate selection
- Default to `Skip` operation unless the user has actually applied one or more operations for a column
- Requires https://github.com/deephaven/deephaven-core/pull/4780
- Tested using the steps in the description of #1355
- Fixes #1355

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -854,7 +854,7 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.startListening(model);
   }
 
-  componentDidUpdate(prevProps: IrisGridProps): void {
+  componentDidUpdate(prevProps: IrisGridProps, prevState: IrisGridState): void {
     const {
       inputFilters,
       isSelectingColumn,
@@ -922,6 +922,18 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
           return { loadingScrimProgress };
         });
       });
+    }
+
+    const { openOptions } = this.state;
+    if (openOptions !== prevState.openOptions) {
+      const isAggEditType = (option: OptionItem): boolean =>
+        option.type === OptionType.AGGREGATION_EDIT;
+      if (
+        !openOptions.some(isAggEditType) &&
+        prevState.openOptions.some(isAggEditType)
+      ) {
+        this.removeEmptyAggregations();
+      }
     }
 
     this.sendStateChange();
@@ -3206,26 +3218,16 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     log.debug('handleAggregationChange', aggregation);
 
     this.startLoading(`Aggregating ${aggregation.operation}...`);
-    this.setState(({ aggregationSettings }) => {
-      let newAggregations = [...aggregationSettings.aggregations];
-      let aggregationIndex = newAggregations.findIndex(
-        a => a.operation === aggregation.operation
-      );
-      if (aggregationIndex === -1) {
-        aggregationIndex = newAggregations.length;
-      }
-      newAggregations[aggregationIndex] = aggregation;
-      newAggregations = newAggregations.filter(
-        a => a.selected.length > 0 || a.invert
-      );
-      return {
-        selectedAggregation: aggregation,
-        aggregationSettings: {
-          ...aggregationSettings,
-          aggregations: newAggregations,
-        },
-      };
-    });
+
+    this.setState(({ aggregationSettings }) => ({
+      selectedAggregation: aggregation,
+      aggregationSettings: {
+        ...aggregationSettings,
+        aggregations: aggregationSettings.aggregations.map(a =>
+          a.operation === aggregation.operation ? aggregation : a
+        ),
+      },
+    }));
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1287,7 +1287,12 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       const operationMap = this.getOperationMap(columns, aggregations);
       const operationOrder = this.getOperationOrder(aggregations);
 
-      return { operationMap, operationOrder, showOnTop };
+      return {
+        operationMap,
+        operationOrder,
+        showOnTop,
+        defaultOperation: AggregationOperation.SKIP,
```
mattrunyon, line 1294: How does enterprise handle this? Does it just ignore it for now?

mofojed: Handles it as it should. The `SKIP` operation was already implemented on Enterprise side, and it was also already looking for `defaultOperation` in the totals table config: https://github.com/deephaven-ent/iris/blob/304a3267f77cff5afedc96cf004f4352b59dba68/web/client-api/src/main/java/com/illumon/iris/web/client/api/JsTotalsTableConfig.java#L69

Tested against dev-grizzly with these changes pulled in using `start-community`.

mattrunyon (changes requested):
If I create aggregations (created a sum using the table in the linked ticket) and select none of the columns, a server error is logged to the console (doesn't happen before this change). Here's the top of the stack

```
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: io.deephaven.proto.backplane.grpc.AggregateRequest must have at least one aggregations (5)
```

There's a 2nd issue where if you have no columns selected in the aggregations menu, it also still shows the aggregation after you leave the edit menu. Not sure if that's intentional, but it is that way in main so probably just file as a separate ticket

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1275,12 +1275,17 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
-        (aggregations?.length ?? 0) === 0
-      ) {
+      if ((config?.columns?.length ?? 0) > 0) {
+        // If we've got rollups, then aggregations are applied as part of that...
+        return null;
+      }
+
+      // Filter out aggregations without any columns actually selected
+      const aggregations = aggregationSettings.aggregations.filter(
+        agg => agg.selected.length > 0 || agg.invert
```
mattrunyon, line 1285: If you somehow got into a state with all "selected", but `invert = true` then it's actually none selected. Not sure if that's possible though

I think `invert` makes logic around this more complex than it needs to be, but not worth changing in this PR. Like why track selected + invert instead of just selected. The invert seems to only get set on toggle all and reset. Which for reset it sets selection to none + invert. Should just set selection to all in that case

I'm guessing it's because of a "select all" case if columns or custom columns are added. Just feels like there should be `selectAll` instead or something

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3196,15 +3206,26 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
-      },
-    }));
+    this.setState(({ aggregationSettings }) => {
+      let newAggregations = [...aggregationSettings.aggregations];
+      let aggregationIndex = newAggregations.findIndex(
+        a => a.operation === aggregation.operation
+      );
+      if (aggregationIndex === -1) {
+        aggregationIndex = newAggregations.length;
+      }
```
mattrunyon, line 3216: Were we just not adding aggregations in this case before?

mattrunyon (changes requested):
Looks like it's functioning correctly. Just a few minor things

mattrunyon:
Also just found out BigInt and BigDecimal show up as possible (on sum at least), but don't actually aggregate. They can cause the same error as the error with none selected before if only a BigInt or BigDecimal is selected

packages/iris-grid/src/IrisGrid.tsx
mofojed: We do `invert` so it scales up to large lists. If you have hundreds of columns and only deselect 1, rather than storing the hundreds of columns it just sends the one and the `invert` bit.
Not so important here, but in places like the Advanced Filter Creator it is important, as you could be deselecting values from a list of potentially millions.
In this case `invert = true` with all "selected", we flip the inversion bit (`AggregationEdit#handleSelect`). Could wrap this logic in some sort of utility method or something though to make it clearer, or resolve it to the list of aggs and just look at that. Both which I'm not inclined to do in this PR.

packages/iris-grid/src/IrisGrid.tsx
mofojed: This case never happened before, because the aggregation always existed and we just replaced it. Now we're removing it if the aggregation is empty.

mofojed:
@mattrunyon The BigInteger issue looks to be a JS API issue, I've logged a ticket for it: https://github.com/deephaven/deephaven-core/issues/4877

packages/iris-grid/src/IrisGrid.tsx
mofojed: There is a minor difference in behaviour, where if you add a few aggregations, then edit the top one in the list, deselect everything, and then start selecting stuff again, it will now be at the end of the list.
I'm just going to change it to remove it only after the user exits the AggregationEdit screen.

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: Makes sense. Although the actual logic looks like invert is only false if `toggleAll` is used and thus  you can still end up with `n-1` columns in your aggregations list instead of what should be the max of `n/2` if the inversion is flipped any time it would be cheaper to track the smaller portion.

Either way not important to this change

mattrunyon (approved):

PR #1651, opened 2023-11-17
feat: Theming - Inline svgs

bmingles:


dsmmcken:
would I then have to specify an INLINE_THEME.TS file to replace the icons or change the color if I, as a third party theme write was trying to create one?

I imagined it would have just been like variable as part of theme:

--icon-whatever: url("svg path blah blah blah fill="var(--some-color)"...")


and then before you insert the variables/on theme change you resolve it to the actual color

packages/code-studio/src/styleguide/colorUtils.ts
```diff
@@ -0,0 +1,105 @@
...
+      };
+    })
+    .flat();
+}
+
+/** Group color data based on capture group value */
+export function buildColorGroups(
+  styleText: string,
+  captureGroupI: number,
+  reassignVarGroups: Record<string, string> = {},
```
mofojed, line 62: Thinking `reassignVarGroups` should be declared in `colorUtils.ts`, and then used as the default value here as the last parameter (since we're always passing in the same value anyway).

mofojed (changes requested):
Just a little bit of cleanup. Looks good overall.

mofojed (approved):

PR #1655, opened 2023-11-22
fix: Fix button snapshots

bmingles:
Realized there was a race condition where button snapshots were not still running. Added detection to ensure button tests run + re-ran snapshots


tests/styleguide.spec.ts
```diff
@@ -46,40 +47,51 @@ test.afterEach(async () => {
...
     // Isolate the section
-    await page.goto(`/ide/styleguide?isolateSection=true#${id}`);
-
-    // Need to reload since we are calling in a loop and only the hash is changing
-    await page.reload();
+    const sectionUrl = `/ide/styleguide?isolateSection=true#${id}`;
+    await page.goto(sectionUrl);
 
     const section = page.locator(`#${id}`);
+    await section.waitFor();
```
bmingles, line 87: This is needed to ensure button tests run

dsmmcken: you should be awaiting most of these page actions, then this is not needed.

Just `await page.locator`

https://playwright.dev/docs/locators

bmingles: @dsmmcken have you seen any examples of awaiting the `page.locator()` call? I've only seen ones awaiting methods off of that, and it doesn't seem to return a promise

bmingles: For reference, these are the docs I followed for this particular case:
https://playwright.dev/docs/api/class-locator#locator-wait-for

dsmmcken:
```
 const grid = await page.locator('.iris-grid-panel .iris-grid');
 ```
hmm, I did this in another test and it is working? 🤷  Maybe that isn't actually awaiting the locator, I don't know. But it runs.

**Review by mofojed — DISMISSED** [review]

File: tests/styleguide.spec.ts [hunk 2]
```diff
@@ -1,34 +1,35 @@
 /* eslint-disable no-await-in-loop */
-import { expect, Page, test } from '@playwright/test';
+import { expect, test } from '@playwright/test';
 
-let page: Page;
+const EXPECTED_SAMPLE_SECTION_COUNT = 40;
+const EXPECTED_BUTTON_SECTION_COUNT = 4;
+
+// let page: Page;
```
mofojed, line 7:
```suggestion
```

PR #1657, opened 2023-11-24
feat: View cell contents in context menu

### LINKED ISSUE #1605: "View full contents" for text cells should be available from context menu
When a String column is too narrow to display its full contents, there is a button that lets you open a dialog that shows the full contents:

image

This is helpful because the "Cell Contents" dialog lets you select and copy just a portion of the text:

image

However, if the column happens to already be wide enough to display the full cell contents, there is no way to get to the "Cell Contents" dialog, since the button disappears:

image

And it is not available in the context menu:

image

dsmmcken:
Resolves #1605


mofojed (approved):
Looks good.

The signature for `getCellActions` could be improved (`modelColumn` can't be `null`/`undefined`, it's not even using `grid`, and it gets the `modelRow` again even though it's only called from a spot where we already know `modelRow` is not `null`), but none of that is part of your change.

PR #1658, opened 2023-11-24
feat: theme fontawesome icon size wrapped in spectrum icons

dsmmcken:
Spectrum icons are 18px. However, our icons look best at 16px as they were designed on a 16px grid. This overrides the default spectrum icon size when wrapping a font-awesome icon inside a spectrum icon wrapper.

Made overrides not a module, so it doesn't add hashes to the classes in overrides, as I have a feeling this file will be used to target specific classes and that wouldn't be desirable.

Note: please suggest somewhere for that variable I added to live if you think that location sucks


packages/components/scss/BaseStyleSheet.scss
```diff
@@ -3,6 +3,11 @@
 @import './custom.scss';
 @import 'bootstrap/scss/bootstrap';
 
+:root {
+  // TODO: where's a better place to put this?
+  --dh-svg-inline-icon-size: 16px;
```
bmingles, line 8: This seems fine to me for now. Possible we could move it to the module containing the inline SVG variables once my PR is done, but there's not really a better spot for now

bmingles (changes requested):
Changes look good. There are failing unit tests + e2e tests

PR #1659, opened 2023-11-27
test: set pull_policy to always for e2e:docker tests

dsmmcken:
Always pull the latest edge server, as unless you manally pull this regularly your tests can fail.

### DIFF
--- a/tests/docker-scripts/docker-compose.yml
+++ b/tests/docker-scripts/docker-compose.yml
@@ -3,6 +3,7 @@ version: '3.8'
 services:
   dhc-server:
     image: ghcr.io/deephaven/server:${DHC_VERSION:-edge}
+    pull_policy: always
     expose:
       - 10000
     environment:


tests/docker-scripts/docker-compose.yml
```diff
@@ -3,6 +3,7 @@ version: '3.8'
 services:
   dhc-server:
     image: ghcr.io/deephaven/server:${DHC_VERSION:-edge}
+    pull_policy: always
```
mofojed, line 6: Just reading the docs: https://docs.docker.com/compose/compose-file/05-services/#pull_policy
> always: Compose always pulls the image from the registry.
I just want to verify - does it _always_ redownload the image? Or does it just check to see if you've already got the latest image in the cache and just use that? I want to ensure it's still pulling from the cache if you already have the latest edge.

bmingles (approved):
LGTM. Verified locally

mofojed (approved):

PR #1661, opened 2023-11-29
feat: Theme Selector

### LINKED ISSUE #1660: Theming - Theme selector
We need to add a theme selector to the settings menu. 

- Should show all themes available from default or plugins
- Should change theme realtime when selection changes
- For now we can hide it if only 1 theme is available. We can then easily enable during dev by adding plugins and keep it hidden from users while we still work through remaining theme support

See Don's POC for starting point
https://github.com/deephaven/web-client-ui/pull/1460/files#diff-d0402466d6aa89364526f1efba471ff5eb597b40ed81273323a5dc6c73137a89

bmingles:
- Added a theme selector to settings menu when there are multiple themes available
- IrisGrid and Monaco now respond dynamically to selected theme change
- Misc moving of theme utils
- Added vite config for local plugin dev

resolves #1660


tsconfig.eslint.json
```diff
@@ -5,6 +5,8 @@
     "packages/*/src/**/**.tsx",
     "packages/*/src/**/**.js",
     "packages/*/src/**/**.jsx",
+    /* Vite config including any local overrides (e.g. vite.config.local.ts) */
```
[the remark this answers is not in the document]
bmingles: Yeah, vscode supports `jsonc / JSON with Comments`. tsconfig seems to be configured as this by default. Sadly package.json is not

README.md
```diff
@@ -52,6 +52,45 @@ If your DHC address is different from the default `http://localhost:10000`, edit
...
+The plugins repo supports [serving plugins locally](https://github.com/deephaven/deephaven-plugins/blob/main/README.md#serve-plugins). DHC can be configured to proxy `js-plugins`requests to the local dev server by setting `VITE_JS_PLUGINS_DEV_PORT` in `packages/code-studio/.env.development.local`.
+
+e.g. To point to the default dev port:
+
+```
+VITE_JS_PLUGINS_DEV_PORT=4100
+```
+
+## Local Vite Config
+If you'd like to override the vite config for local dev, you can define a `packages/code-studio/vite.config.local.ts` file that extends from `vite.config.local`. This file is excluded via `.gitignore` which makes it easy to keep local overrides in tact.
```
mofojed, line 65:
```suggestion
If you'd like to override the vite config for local dev, you can define a `packages/code-studio/vite.config.local.ts` file that extends from `vite.config.ts`. This file is excluded via `.gitignore` which makes it easy to keep local overrides in tact.
```

packages/code-studio/src/styleguide/ThemeColors.tsx
```diff
@@ -16,16 +16,29 @@ import {
...
+  const [swatchDataGroups, setSwatchDataGroups] = useState(
+    buildSwatchDataGroups
+  );
+
+  const { selectedThemeKey } = useTheme();
+
+  useEffect(() => {
+    // Rebuild swatches when themes change
+    setSwatchDataGroups(buildSwatchDataGroups());
+  }, [selectedThemeKey]);
```
mofojed, line 40: I think you can just use `useMemo` here.
```suggestion
  const swatchDataGroups = useMemo(buildSwatchDataGroups, [selectedThemeKey]);
```

mofojed (changes requested):
Some minor cleanup, looks good overall

packages/plugin/src/PluginUtils.tsx
```diff
@@ -32,3 +42,37 @@ export function getIconForPlugin(plugin: PluginModule): React.ReactElement {
...
+/**
+ * Extract theme data from theme plugins in the given plugin map.
+ * @param pluginMap
+ */
+export function getThemeDataFromPlugins(
+  pluginMap: PluginModuleMap
+): ThemeData[] {
+  const themePluginEntries = [...pluginMap.entries()].filter(
+    (entry): entry is [string, ThemePlugin] => isThemePlugin(entry[1])
+  );
```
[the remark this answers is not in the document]
bmingles: The keys aren't used here, but they do get consumed in the next block when `themePluginEntries` is iterated, so the key + value needs to be present

packages/code-studio/src/styleguide/ThemeColors.tsx
bmingles: Cool. eslint complained about unused selectedThemeKey when I tried:

```
useMemo(() => buildSwatchDataGroups(), [selectedThemeKey])
```

Looks like your version doesn't have this issue, presumably because eslint can't see inside the closure

mofojed (approved):

PR #1663, opened 2023-11-29
feat: Create UI to Display Partitioned Tables

### LINKED ISSUE #1143: UI for PartitionedTable

georgecwan:
- `PartitionedTable` objects can now be opened and displayed with a new UI
  - Supports switching between partitions and viewing the merged or key table
- Partition aware parquet tables will also share this new UI
  - The new UI will no longer allow users to enter invalid partitions
- Closes #1143
- Depends on the following changes to core:
  - https://github.com/deephaven/deephaven-core/pull/4789
  - https://github.com/deephaven/deephaven-core/pull/4931
  - https://github.com/deephaven/deephaven-core/pull/4940

#### PartitionedTable
1. Run the code and open the table `pt`
    ```py
  
    _t = empty_table(100).update(["verylongcolumn=(int)Math.floor(i/5)", "veryveryverylongcolumn=i"])
    pt = _t.partition_by(["verylongcolumn", "veryveryverylongcolumn"])
    ```
2. Check that the features specified in the [spec](https://user-images.githubusercontent.com/1576283/268390627-d427b993-1d09-43a5-960f-4e6cd0848f36.png) are present:
    - Resizing the panel horizontally wraps the dropdown without wrapping the '>'
    - Hovering over the 'Key' and 'Merge' buttons displays the correct labels
    - The initial partition should be the first valid partition available when all columns are sorted in descending order
    - Options in the dropdown are displayed in descending order
3. Clicking the 'Key' and 'Merge' tables should correctly display the respective table and the button should visually indicate if one of them is being displayed. All the dropdowns should show empty values while one of the buttons is active.
    - While one of the toggle buttons is active, only the leftmost dropdown should be enabled
4. After clearing the dropdowns by clicking either the 'Key' or 'Merge' button, selecting any value on any dropdown should automatically set the remaining dropdowns and display a valid partition.
5. Dropdowns should only contain values that are valid with respect to the selected values of all the dropdowns left of it
6. Changing the value of a dropdown should try to preserve dropdowns to the right of it. If this is not possible, the values of the dropdowns right of it should be changed so that a valid partition can be displayed.

#### Parquet Tables
1. Run the following code and verify that all tables display and function correctly for every data type
    ```py
  
    write(part, "/tmp/pt-test/intCol=0/part.parquet")
    write(part, "/tmp/pt-test/intCol=1/part.parquet")
    int_partition = read("/tmp/pt-test")
    
    write(part, "/tmp/string-test/stringCol=hello/part.parquet")
    write(part, "/tmp/string-test/stringCol=world/part.parquet")
    string_partition = read("/tmp/string-test")
    
    write(part, "/tmp/double-test/doubleCol=1.5/part.parquet")
    write(part, "/tmp/double-test/doubleCol=2.5/part.parquet")
    double_partition = read("/tmp/double-test")
    
    write(part, "/tmp/char-test/charCol=a/part.parquet")
    write(part, "/tmp/char-test/charCol=b/part.parquet")
    char_partition = read("/tmp/char-test")
    
    write(part, "/tmp/long-test/longCol=2147483648/part.parquet")
    write(part, "/tmp/long-test/longCol=2147483650/part.parquet")
    long_partition = read("/tmp/long-test")
    
    write(part, "/tmp/bool-test/boolCol=true/part.parquet")
    write(part, "/tmp/bool-test/boolCol=false/part.parquet")
    bool_partition = read("/tmp/bool-test")

    write(part, "/tmp/multi_test/x=0/y=0/part.parquet")
    write(part, "/tmp/multi_test/x=0/y=1/part.parquet")
    write(part, "/tmp/multi_test/x=1/y=0/part.parquet")
    write(part, "/tmp/multi_test/x=1/y=1/part.parquet")
    write(part, "/tmp/multi_test/x=1/y=2/part.parquet")
    multi_partition = read("/tmp/multi_test")
    ```


packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,442 @@
...
+
+const log = Log.module('IrisGridPartitionedTableModel');
+
+export function isIrisGridPartitionedTableModel(
+  model: IrisGridModel
+): model is IrisGridPartitionedTableModel {
+  return (
+    (model as IrisGridPartitionedTableModel).partitionedTable !== undefined
+  );
+}
```
mofojed, line 46: One thing that I think will help clean this up a bit... 
- Define the interface for a partitioned grid model in a file `PartitionedGridModel.ts`. See `ExpandableGridModel` as an example (though you still want it in the `iris-grid` package).
- Have `IrisGridPartitionedTableModel` and `IrisGridProxyModel` add the `implements PartitionedGridModel` to the class definition
  - That way if there's ever new methods added to PartitionedGridModel, they'll both throw an error
  - Side note - IrisGridTreeTableModel and IrisGridProxyModel should also `implements ExpandableGridModel` as well

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -452,6 +461,14 @@ class IrisGridProxyModel extends IrisGridModel {
...
   }
 
+  get partitionColumns(): readonly Column[] {
+    return this.originalModel.partitionColumns;
+  }
+
+  set partitionColumns(columns: readonly Column[]) {
+    this.originalModel.partitionColumns = columns;
+  }
+
```
mofojed, line 471: You shouldn't be able to set partition columns, should only have a getter.
```suggestion
```

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,442 @@
...
+
+  get partitionColumns(): Column[] {
+    // TODO
+    return this.partitionedTable.keyColumns;
+  }
+
+  set partitionColumns(columns: Column[]) {
+    // Do nothing, partition columns of PartitionedTable can't be modified
+  }
+
```
mofojed, line 416: Shouldn't have a setter here, this is a read-only property
```suggestion
```

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,442 @@
...
+  dataBarOptionsForCell(
+    column: number,
+    row: number,
+    theme: IrisGridThemeType
+  ): DataBarOptions {
+    return this.model.dataBarOptionsForCell(column, row, theme);
+  }
+
+  get partitionColumns(): Column[] {
```
mofojed, line 408: Enforce returning a `readonly` array
```suggestion
  get partitionColumns(): readonly Column[] {
```

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -476,6 +493,74 @@ class IrisGridProxyModel extends IrisGridModel {
     this.model.filter = filter;
   }
 
+  get partition(): readonly unknown[] {
+    return this.originalModel.partition;
+  }
+
+  set partition(partition: readonly unknown[]) {
+    log.debug('set partition', partition);
```
mofojed, line 501: You should add a check in here to see if our original model is a partitioned table model, and if not just throw. E.g:
```suggestion
    if (!isIrisGridPartitionedTableModel(this.originalModel)) {
      throw new Error('Partitions are not available');
    }
    log.debug('set partition', partition);
```

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -476,6 +493,74 @@ class IrisGridProxyModel extends IrisGridModel {
...
+    this.originalModel.partition = partition;
```
mofojed, line 502: I don't really like this, where you're mutating the original underlying model to get the next model... instead should just be creating the next model here and calling `setNextModel`.

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -476,6 +493,74 @@ class IrisGridProxyModel extends IrisGridModel {
...
+    if (isIrisGridPartitionedTableModel(this.originalModel)) {
+      if (partition.length === 0) {
+        this.setNextModel(
+          this.originalModel.partitionedTable.getMergedTable().then(table => {
+            const newModel = makeModel(this.dh, table, this.formatter);
+            (this.originalModel as IrisGridPartitionedTableModel).model =
```
mofojed, line 508: We shouldn't be using `IrisGridPartitionedTableModel` as a proxy like this - we've already got the `IrisGridProxyModel` as a proxy... and now we're mixing the responsibility between the two here.
`IrisGridPartitionedTableModel` should just be the top level partitioned table model I think, don't need to set a child model in there. Just do `setNextModel` here.

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,442 @@
...
+  get isValuesTableAvailable(): boolean {
+    return this.model.isValuesTableAvailable;
+  }
+
+  get isChartBuilderAvailable(): boolean {
+    return this.model.isChartBuilderAvailable;
+  }
+
+  get isRollupAvailable(): boolean {
+    return this.model.isRollupAvailable;
```
mofojed, line 274: Hmm we've got a problem right now if you try and set a rollup or set a select distinct on a partitioned table then it throws an error. I think you should just return `false` in these cases.

mofojed (changes requested):

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -178,139 +149,218 @@ class IrisGridPartitionSelector<T> extends Component<
...
     return (
       <div className="iris-grid-partition-selector">
+        <div className="table-name">
+          <span>Partitioned Table </span>
+        </div>
+        <div className="partition-button-group">
+          <Button
+            type="button"
+            className="btn btn-sm"
+            onClick={this.handleKeyTableClick}
```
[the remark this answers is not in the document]
georgecwan: I was under the impression to use these types of inline buttons from @dsmmcken. Don could I have some confirmation?

packages/iris-grid/src/IrisGridProxyModel.ts
georgecwan: I also need to be able to set the partition for `IrisGridTableModel` for the partitioned parquet tables

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -178,139 +149,218 @@ class IrisGridPartitionSelector<T> extends Component<
...
+            kind="inline"
+            tooltip="View keys as table"
+            icon={<FontAwesomeIcon icon={vsKey} />}
```
dsmmcken, line 346: this can just accept vsKey directly

packages/iris-grid/src/IrisGridPartitionSelector.tsx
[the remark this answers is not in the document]
dsmmcken: I suggested inline, spectrum isn't done being themed yet. inline + active is a toggle button.

I don't think you need to specify a type, or the additional first "btn" class, as that is also added automatically.

packages/iris-grid/src/IrisGridProxyModel.ts
mofojed: Ahh we should have a flag for checking if the model is partitionable then. `isPartitionRequired` I think is the right flag to check? I don't think there's a case where it's partitionable but a partition is not required is there?

georgecwan: The original `isPartitionRequired` is actually equal to `table.isUncoalesced` (for parquet tables), which returns `false` after a partition is set for the first time. Is there some other way I can check whether a partition is required that I can use to replace the existing implementation of `isPartitionRequired`?

dsmmcken (changes requested):
Boolean and Char columns don't seem to work in the UI as parition keys, even though they are supported by the engine:
image

georgecwan:
> Boolean and Char columns don't seem to work in the UI as parition keys, even though they are supported by the engine: image

Currently breaks when partition columns are not in order

dsmmcken (changes requested):
Something awful is happening with the performance here. I can only replicate it with the parition viewer, but it might be something in core. I don't know.

Steps to reproduce:
1. run this, time_table, creating new a parition on each tick
```

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",

p = simple_ticking.partition_by("MyString")
```
2. Wait like a minute or two then take a performance snapshot. You'll get a million little nano-second locks too the point of it breaking everything.
image

georgecwan:
When partitioning a table by a column of `Long`'s, users would be unable to change partitions due to `LongWrapper` not supporting an `equals` function. PR for this bug has been opened at https://github.com/deephaven/deephaven-core/pull/4940

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.partitionKeys = partition;
+  }
+
+  get partitionKeysTable(): Promise<Table> {
+    const sorts = this.partitionColumns.map(column => column.sort().desc());
+    return this.partitionedTable.getKeyTable().then(table => {
+      table.applySort(sorts);
+      return table;
+    });
+  }
```
mofojed, line 85: We shouldn't be doing an expensive operation like this/creating a table for a property accessor. Property accessors should be either just quick reads or memoized results, so that accessing the property multiple times is consistent. Right now, this will create a table every time you read this property, which is not what you're expecting; e.g if you did something like:

```
if (model.partitionKeysTable) {
  setNextModel(createModel(model.partitionKeysTable))
}
```
You'd have a table created in the `if` expression that would never be closed/cleaned up properly. And actually, in `IrisGrid.tsx` you do exactly that, leaking a new table on each re-render:
```
           {model.partitionKeysTable && (  // Creating one key table here
                <IrisGridPartitionSelector
                  dh={model.dh}
                  tablePromise={model.partitionKeysTable} // Creating another key table here that is getting passed into IrisGridPartitionSelector
```
Instead, this should be a method so it's explicitly clear you're creating a new object. I would mimick `valuesTable` - basically just remove the `get `

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2403,23 +2313,19 @@ export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+  handlePartitionMerge(): void {
+    const { model } = this.props;
+    model.partition = [];
   }
 
-  handlePartitionDone(): void {
-    this.setState({ isSelectingPartition: false });
+  handlePartitionKeyTable(): void {
+    const { model } = this.props;
+    model.openPartitionKeysTable();
```
mofojed, line 2328: Much like setting the partition above, this is problematic because we're not saving/restoring this state; we want the table to be in the same state as when we closed the dashboard.
I think a good model to follow here is the `rollupConfig` model, as it's doing something similar. In my general review comment I'll give some overarching guidelines.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -307,12 +334,21 @@ abstract class IrisGridModel<
...
   /**
+   * @deprecated Replaced with isPartitionRequired()
    * @returns True if this model requires a filter to be set
    */
   get isFilterRequired(): boolean {
     return false;
   }
 
+  /**
+   * Replaces isPartitionRequired()
```
mofojed, line 345:
```suggestion
```

mofojed (changes requested):
Not quite there yet - some performance issues with leaking tables, and we are not persisting the state when saving/restoring. Here are my suggestions:
- Right now you have added the following methods to `IrisGridModel`: `get partition()`, `set partition(partition)`, `get partitionColumns()`, `get partitionKeysTable`, and `openPartitionKeysTable`. Instead, create a type `PartitionConfig` and an interface `PartitionedGridModel` with the following methods/property accessors (similar to how `RollupConfig` works):
```typescript
export type PartitionConfig {
  /** The partition values to set on the model */
  partitions: unknown[];

  /** What data to display - the keys table, the merged table, or the selected partition */
  mode: 'keys' | 'merged' | 'partition';
}

export interface PartitionGridModel {
  get partitionConfig(): PartitionConfig;

  set partitionConfig(partitionConfig: PartitionConfig);

  /**
   * Retrieve the columns this model is partitioned on
   * @returns All columns to partition on
   */
  get partitionColumns(): readonly Column[];

  /** Whether a partition is required to be set on this model */
  get isPartitionRequired(): boolean;

  /** Whether partitions are available on this model */
  get isPartitionAvailable(): boolean;

  /** Get a keys table for the partitions */
  partitionKeysTable(): Promise<Table>
```

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -1,316 +1,394 @@
...
     };
   }
 
   async componentDidMount(): Promise<void> {
-    const { columns, table } = this.props;
-    const { partitions } = this.state;
+    const { columns, partitionConfig, tablePromise } = this.props;
 
-    const partitionTables = await Promise.all(
+    const table = await tablePromise;
```
mofojed, line 72: As I commented in IrisGrid, pass in the `model` instead of a `tablePromise`, and load the table from the model here.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -1,316 +1,394 @@
...
+    if (!deepEqual(prevConfig, partitionConfig)) {
+      this.updateConfig(partitionConfig);
+    }
   }
 
   componentWillUnmount(): void {
-    const { partitionTables } = this.state;
-    partitionTables?.forEach(table => table.close());
+    this.table?.close();
+    this.partitionTables?.forEach(table => table.close());
```
mofojed, line 92: One case this _doesn't_ handle is if you unmount immediately after mount, before the tables are loading. In those cases, those tables will never get closed and we will leak them.
What you can do is make use of the `Pending` class to keep track of promises to cancel when the component unmounts. It's used in `IrisGrid` as an example, but in the constructor you'd have a `this.pending = Pending()`, and then say in `componentDidMount` where you have:
```
    this.partitionTables = await Promise.all(
      columns.map(async (_, i) => table.selectDistinct(columns.slice(0, i + 1)))
    );
```
You'd do:
```
    try {
      this.partitionTables = await Promise.all(
        // Add the inner promise to the pending queue, and add a cleanup `t => t.close()` for if it gets canceled.
        columns.map(async (_, i) => 
          pending.add(table.selectDistinct(columns.slice(0, i + 1))), t => t.close())
      );
    } catch (e) {
      if (!PromiseUtils.isCanceled(e)) {
        log.error('Unable to get partition tables', e);
      }
    }
    this.updateConfig(partitionConfig);
```

You'll need similar handling for the `awaits` in the `updateConfig` below.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -1,316 +1,394 @@
...
+    const partitionFilters = [...this.partitionTables[index].filter];
+    const validPartitions = [...partitionConfig.partitions];
+    let lastValidPartition = null;
 
-    const partitionFilters = [];
-    for (let i = 0; i < columns.length - 1; i += 1) {
-      const partition = partitions[i];
+    // Update partition filters
+    for (let i = index; i < partitionConfig.partitions.length; i += 1) {
+      /* eslint-disable  no-await-in-loop */
```
mofojed, line 218: Add a comment about how you need to do the await in a loop because each subsequent partition list is based on the results from the previous one (cascading).
We also need to check that user didn't do something else that invalidates this work. Easiest way to do that is to set a loading state, disable the controls while in that state (since it's loading), await on updateConfig when you call it, and then unset the loading state when `updateConfig` has finished.

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,105 @@
...
+  async initializePartitionModel(): Promise<IrisGridModel> {
+    const initTable = await this.partitionedTable.getTable(
+      this.config.partitions
+    );
+    const tableCopy = await initTable.copy();
+    return Promise.resolve(
+      new IrisGridTableModel(this.dh, tableCopy, this.formatter)
+    );
+  }
+}
```
mofojed, line 103: We shouldn't need this function. Model should already be initialized in IrisGridPanel

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,105 @@
...
+import EmptyIrisGridModel from './EmptyIrisGridModel';
+import { PartitionConfig, PartitionedGridModel } from './PartitionedGridModel';
+
+export function isIrisGridPartitionedTableModel(
+  model: IrisGridModel
+): model is IrisGridPartitionedTableModel {
+  return (
+    (model as IrisGridPartitionedTableModel).partitionedTable !== undefined
+  );
+}
```
mofojed, line 21: We shouldn't need this function, we already have `isPartitionedGridModel` method.

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,105 @@
...
+  set partitionConfig(rollupConfig: PartitionConfig) {
+    this.config = rollupConfig;
+  }
+
+  partitionKeysTable(): Promise<Table> {
+    const sorts = this.partitionColumns.map(column => column.sort().desc());
+    return this.partitionedTable.getKeyTable().then(table => {
+      table.applySort(sorts);
+      return table;
+    });
```
mofojed, line 91: We shouldn't apply sorts here. That should be done in `IrisGridPartitionSelector`.

mofojed (changes requested):

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
georgecwan: This function is still used in `set partitionConfig` inside `IrisGridProxyModel` since how the next model is set depends on the which model type it is

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -476,6 +497,61 @@ class IrisGridProxyModel extends IrisGridModel {
...
+        this.setNextModel(
+          this.originalModel.partitionedTable
+            .getMergedTable()
+            .then(table => makeModel(this.dh, table, this.formatter))
+        );
+        return;
+      }
+      const tablePromise = this.originalModel.partitionedTable
+        .getTable(partitionConfig.partitions)
+        .then(table => table.copy());
```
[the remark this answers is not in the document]
georgecwan: Without this `.copy()` the constituent table is closed when I swap to a different partition so if I swap back afterwards I can no longer display the table.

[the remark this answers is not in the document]
mofojed: Ahh that's an annoying thing with the table returned from the `JsPartitionedTable`... it memoizes the results, but then it doesn't close any of those tables when you close the `PartitionedTable`. Makes it hard to use... I'll need to verify it's leaking something, probably mark this as a TODO to follow up on.

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
mofojed: Shouldn't need to use that there

packages/iris-grid/src/IrisGridPartitionSelector.test.tsx
```diff
@@ -1,72 +1,53 @@
...
 import IrisGridPartitionSelector from './IrisGridPartitionSelector';
 import IrisGridTestUtils from './IrisGridTestUtils';
+import { PartitionConfig, PartitionedGridModel } from './PartitionedGridModel';
+
+const irisGridTestUtils = new IrisGridTestUtils(dh);
+
+function makeModel(
+  columns = irisGridTestUtils.makeColumns()
+): PartitionedGridModel {
+  const model = {
```
bmingles, line 14: Not worth changing, but `createMockProxy` could make this a bit cleaner

packages/iris-grid/src/IrisGridPartitionedTableModel.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  get columns(): readonly Column[] {
+    return this.partitionedTable.columns;
+  }
+
+  get partitionColumns(): readonly Column[] {
+    return this.partitionedTable.keyColumns;
+  }
+
+  partitionKeysTable(): Promise<Table> {
```
bmingles, line 61: 2 of these are not marked async, but `partitionTable` is. Should probably make consistent.

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -181,10 +193,101 @@ class IrisGridTableModel extends IrisGridTableModelTemplate<Table, UIRow> {
     );
   }
 
+  get partitionColumns(): readonly Column[] {
+    return this.getCachedPartitionColumns(this.columns);
+  }
+
+  partitionKeysTable(): Promise<Table> {
```
bmingles, line 200: `async` ?

bmingles (changes requested):
Left 2 minor comments about missing `async`

bmingles (dismissed):
LGTM

mofojed (dismissed):

dsmmcken (changes requested):
image

String column not working. Boolean column not working.

And I get this error occasionally:
image

Tried testing using dx.stocks and mostdatatypes queries.

mofojed:
@dsmmcken should be fixed. Also, you'll need to make sure you're on the latest `main` that includes both https://github.com/deephaven/deephaven-core/pull/5033 and https://github.com/deephaven/deephaven-core/pull/5009

mofojed:
Don's found two errors that are still occurring, both of them are because it is a ticking table starting with no partition. Using this snippet you can reproduce both issues:
```python

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",

p = simple_ticking.partition_by(["MyString", "MyInt", "MyBoolean"])
```
1. Run the snippet above. Wait a couple of ticks, then run the snippet again. Sometimes you get stuck with a "Rebuilding Filters..." message in the bottom corner.
  - In the logs there is a "Table does not have any data" message. The issue is we're trying to load the first partition key, but there are no partition keys. The fix is to instead wait for an "updated" event to get the first key.
2. Run the snippet above and let it go a few ticks. Then change partitions, wait a second, then run the snippet above again. You'll get an error message saying `layoutHints` is `null`:
image
```
IrisGrid.tsx:2885 [IrisGrid] request failed: TypeError: Cannot read properties of null (reading 'layoutHints')
    at get layoutHints [as layoutHints] (IrisGridTableModel.ts:124:23)
    at new IrisGridTableModelTemplate (IrisGridTableModelTemplate.ts:251:12)
    at new IrisGridTableModel (IrisGridTableModel.ts:57:5)
    at makeModel (IrisGridProxyModel.ts:65:10)
    at IrisGridProxyModel.ts:551:26
```
  - The JsPartitionedTable.getTable API can return `null` instead of rejecting. I have a PR to fix that: https://github.com/deephaven/deephaven-core/pull/5050
  - The main issue is that we re-open the table with the previously selected key, which is no longer valid (since the table started from scratch). We could handle that error more gracefully.

Workaround for both issues is to re-open the table.

dsmmcken (approved):
Looks good, all working now. Changed a couple color variables to theme better.

PR #1664, opened 2023-11-30
fix: add right margin to <Button kind='inline'/> using icons

dsmmcken:
btn-inline did not have a rule to add margin to svgs in the css, like other kinds. This should do it in a backwards compatible way, and ensure all button kinds have the correct margin. We have a bunch of css for other kinds that would be no longer necessary, but not removing in case we use it in a non-standard way elsewhere.

```jsx
// previously missing margin between icon and text
<Button kind="inline" icon="dhTruck">Test</Button>
```

edit: was somewhat more complicated then expected, because children doesn't mean visible children
```jsx
// should handle this case without adding margin
<Button kind="inline" icon="dhTruck"><DropdownMenu/></Button>
```


mattrunyon (changes requested):
Looks good. Just wanted to add a little more context to the comment for future us

PR #1666, opened 2023-12-04
fix: adjust filter bar colour

dsmmcken:
- add full accent-100-1400 hsl colours
- use accent colours with transparency to more closely match original grid design for filter bar colour
- use those variables in the filter input field, with a hacky stacked solid colour linear gradient, such that transperncy has the colour underneath as canvas would
- --dh-color-grid-row-hover-bg was too dark, bumped up one level (no longer the same as elsewhere, but grid stripes are lighter so it needs more contrast

### DIFF
--- a/packages/components/src/theme/theme-dark/theme-dark-semantic.css
+++ b/packages/components/src/theme/theme-dark/theme-dark-semantic.css
@@ -6,6 +6,21 @@
   --dh-color-foreground-hsl: var(--dh-color-white-hsl);
 
   --dh-color-accent-hsl: var(--dh-color-blue-600-hsl);
+  --dh-color-accent-100-hsl: var(--dh-color-blue-100-hsl);
+  --dh-color-accent-200-hsl: var(--dh-color-blue-200-hsl);
+  --dh-color-accent-300-hsl: var(--dh-color-blue-300-hsl);
+  --dh-color-accent-400-hsl: var(--dh-color-blue-400-hsl);
+  --dh-color-accent-500-hsl: var(--dh-color-blue-500-hsl);
+  --dh-color-accent-600-hsl: var(--dh-color-blue-600-hsl);
+  --dh-color-accent-700-hsl: var(--dh-color-blue-700-hsl);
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/FilterInputField.scss
```diff
@@ -56,13 +56,29 @@
...
 
     &.iris-grid-has-filter {
-      // same as filter-bar-expanded-active-bg
-      background: scale-color($primary, $saturation: -75%, $lightness: -60%);
+      // This is a hack to get the filter bar inputs to match the canvas rendered equivalent
+      // --dh-color-filter-bar-expanded-active-bg relies on transparency of being drawn overtop of
+      // the grid background. Since we can't do that here, we need to mimic the effect by stacking
+      // backgrounds using a single color linear gradients as a hack. Performance is not an issue
+      // since it only applies to the selected filter bar cell.
+      background: linear-gradient(
```
bmingles, line 64: I added a similar `solid-gradient` helper for this stacking hack as part of the inline SVGs work. Not sure if it can be re-purposed for this as well.

dsmmcken: I feel like the less scss we use, the easier it will be to switch off it someday.

packages/components/src/theme/theme-dark/theme-dark-semantic.css
```diff
@@ -6,6 +6,21 @@
   --dh-color-foreground-hsl: var(--dh-color-white-hsl);
 
   --dh-color-accent-hsl: var(--dh-color-blue-600-hsl);
+  --dh-color-accent-100-hsl: var(--dh-color-blue-100-hsl);
```
bmingles, line 9: If we are going to keep the non-hsl `--dh-color-accent-xxx` versions, we should probably update them to derive from these new variables

dsmmcken: sure

bmingles (approved):
LGTM

PR #1673, opened 2023-12-06
ci: Set gitHub reporter for Playwright tests

mofojed:
- Add the `github` reporter to add annotations directly in our code when playwright tests fail
- See an example summary report from a previous run: https://github.com/deephaven/web-client-ui/actions/runs/7121141960?pr=1673
- The annotations show up in the Summary, along with the HTML report that you can download:
image
- Annotations also show up in the Files Changed tab, annotating the test case explicitly that failed:
image


bmingles (approved):
Very nice

PR #1682, opened 2023-12-11
feat: Theming - Moved ThemeProvider updates into effect

### LINKED ISSUE #1669: Theming - Spectrum variable mapping
Map remaining React spectrum theme variables to `--dh-color-xxx` ones.

bmingles:
Moved ThemeProvider updates into effect
resolves #1669

### DIFF
--- a/packages/chart/src/ChartThemeProvider.tsx
+++ b/packages/chart/src/ChartThemeProvider.tsx
@@ -20,10 +20,7 @@ export function ChartThemeProvider({
 }: ChartThemeProviderProps): JSX.Element {
   const { activeThemes } = useTheme();
 
-  const chartTheme = useMemo(
-    () => (activeThemes == null ? null : defaultChartTheme()),
-    [activeThemes]
-  );
+  const chartTheme = useMemo(defaultChartTheme, [activeThemes]);
 
   return (
     <ChartThemeContext.Provider value={chartTheme}>
--- a/packages/chart/src/ChartThemeProvider.tsx
+++ b/packages/chart/src/ChartThemeProvider.tsx
@@ -1,4 +1,4 @@
-import { createContext, ReactNode, useEffect, useState } from 'react';
+import { createContext, ReactNode, useMemo } from 'react';
 import { useTheme } from '@deephaven/components';
 import defaultChartTheme, { ChartTheme } from './ChartTheme';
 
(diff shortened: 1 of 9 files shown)


packages/components/src/theme/ThemeProvider.tsx
```diff
@@ -88,17 +90,16 @@ export function ThemeProvider({
...
-    }),
-    [activeThemes, selectedThemeKey, themes]
-  );
+    });
+  }, [activeThemes, selectedThemeKey, themes]);
 
-  return (
+  return value == null ? null : (
     <ThemeContext.Provider value={value}>
       {activeThemes == null ? null : (
```
mofojed, line 104: I think we need to render this style block before the `useEffect` and `value` get set, no? Should be something like:
```
    <>
      {activeThemes == null ? null : (
        <>
          {activeThemes.map(theme => (
            <style data-theme-key={theme.themeKey} key={theme.themeKey}>
              {theme.styleContent}
            </style>
          ))}
        </>
      )}
      {value == null ? null : (
        <ThemeContext.Provider value={value}>
          <SpectrumThemeProvider>{children}</SpectrumThemeProvider>
        </ThemeContext.Provider>
      )}
    </>
```

packages/chart/src/ChartThemeProvider.tsx
```diff
@@ -20,20 +20,10 @@ export function ChartThemeProvider({
...
-  // available by the time the effect runs. Therefore, it is important to derive
-  // the chart theme in an effect instead of deriving in a `useMemo` to ensure
-  // we have the latest CSS variables.
-  useEffect(() => {
-    if (activeThemes != null) {
-      setChartTheme(defaultChartTheme());
-    }
-  }, [activeThemes]);
+  const chartTheme = useMemo(
+    () => (activeThemes == null ? null : defaultChartTheme()),
```
mofojed, line 24: This doesn't seem right...
Looking at `ChartModelFactory.makeModel`, it's always expecting a non-null theme. Why are we allowing the theme to be null here? Shouldn't it always call `defaultChartTheme()` and possibly even throw if it's `null`?
Same comment under IrisGridThemeProvider.

mofojed (changes requested):

packages/chart/src/ChartThemeProvider.tsx
bmingles: The `activeThemes` needs to remain null until plugins load to differentiate pending load vs an empty array that would designate plugins loaded, but no custom themes. We could use the default chart theme / grid theme, but seemed like unnecessary parsing when we know loading isn't complete yet. The `IrisGridThemeContext` and `ChartThemeContext` support null for the theme, but the consumers can opt to throw or provided a default if theme is null. `useChartTheme` throws if null, while IrisGrid handles the default if no theme is present.

mofojed: Ah right, this is just in the bootstrapping process that it's `null`.

Wondering if we should just change `PluginsBootstrap` to not render children until plugins are actually loaded; we need it for auth and for theming, which are the next things that get loaded ... then we wouldn't have to worry about having `null` plugins here. What do you think?
The initialization order in `AppBootstrap.tsx` could be something like:
```
<ClientBootstrap ...>
  <PluginsBootstrap ...>
    <ThemeBootstrap ...>
```

bmingles: I think that would work in DHC, but IIR, I think it bites us in DHE due to Login component being further down the tree.

mofojed: We're not even using `PluginsBootstrap` or `PluginsContext` in DHE... and yeah on DHE we can't load worker plugins until user is actually authenticated. We will need to wire up `PluginsContext` there at some point... regardless, why would we ever want to return `null` here? I think that's breaking the contract of what this bootstrap component should do, and may make it confusing to debug later.

I'd gate `PluginsBootstrap`, and these as well so there's never `null`s. On Enterprise, it'll need to have it's own plugin bootstrapping step.

PR #1688, opened 2023-12-12
fix: Made selector return types generic

### LINKED ISSUE #1687: Redux selector return types should be generic
**Description**

The `getSettings` and `getWorkspace` return types should be derived from the generic `State extends RootState` arg to support DHE consumption.

**Versions**
0.56

bmingles:
fixes #1687

### DIFF
--- a/packages/utils/src/TypeUtils.ts
+++ b/packages/utils/src/TypeUtils.ts
@@ -28,6 +28,13 @@ export type OnlyOneProp<T> = {
 }[keyof T];
 
 /**
+ * Remove `Partial` wrapper from a type. Note that this is slightly different
+ * than `Required` because it will preserve optional properties on the original
+ * target type.
+ */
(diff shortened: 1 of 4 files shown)


packages/utils/src/TypeUtils.ts
```diff
@@ -27,6 +27,13 @@ export type OnlyOneProp<T> = {
   [P in keyof T]: { [ONEPROP in P]: T[ONEPROP] };
 }[keyof T];
 
+/**
+ * Remove `Partial` wrapper from a type. Note that this is slightly different
+ * than `Required` because it will preserve optional properties on the original
+ * target type.
+ */
+export type UndoPartial<T> = T extends Partial<infer U> ? U : never;
```
mofojed, line 35: Ohh neat

mofojed (approved):

Issue #1585, opened 2023-10-19
Shortcut to copy value from an individual cell with alt/opt + click


dsmmcken:
As a user, I would like a quicker way to copy a cells value. 

Copying and individual values from a cell is such a common task that it would be good to have a dedicated shortcut for doing so. Since tables in Deephaven typically use row selection mode, copying an individual cell requires using the context menu.

My proposed shortcut is an alt (opt on mac) + click of a cell to quickly copy it's value. To make it more discoverable, I suggest we switch to a custom copy cell cursor when hovering a cell with a copyable value in it and the user presses or holds alt. Further we can display the shortcut text in the context menu next to the copy cell value.

Copy cell cursor:
![image](https://github.com/deephaven/web-client-ui/assets/1576283/9182b275-9d56-483d-a224-0d94e533e63e)

SVG file:
![Copy-Cell-Cursor](https://github.com/deephaven/web-client-ui/assets/1576283/70984851-52e7-441a-979e-4912a628eb84)

Context menu:
![image](https://github.com/deephaven/web-client-ui/assets/1576283/36479da5-377b-404a-bcdd-2e99ad7931f5)

PR #1694, opened 2023-12-14
feat: Add alt+click shortcut to copy cell and column headers

georgecwan:
Adds a alt/opt+click shortcut to copy the value of a cell or column header.

Resolves deephaven/web-client-ui#1585.


packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -632,6 +633,9 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
       actions.push({
         title: 'Copy Cell',
         group: IrisGridContextMenuHandler.GROUP_COPY,
+        shortcutText: ContextActionUtils.isMacPlatform()
```
mofojed, line 636: I was debating if we should add support for Shortcut to handle mouse clicks... would be kind of interesting in that they could be configurable, but it's more of a headache than we need right now.

packages/components/src/context-actions/ContextActionUtils.ts
```diff
@@ -17,6 +17,7 @@ export interface ContextAction {
   icon?: IconDefinition | React.ReactElement;
   iconColor?: string;
   shortcut?: Shortcut;
+  shortcutText?: string;
```
mofojed, line 20: While we're in here, should add some JSDocs for the new property (ideally we'd have JSDocs for every prop, but adding just for the new property is good for now). Just explaining this will display text for a shortcut in the menu, but does not actually wire up the shortcut.

packages/grid/src/Grid.tsx
```diff
@@ -1732,6 +1733,24 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+    for (let i = 0; i < keyHandlers.length; i += 1) {
+      const keyHandler = keyHandlers[i];
+      const result = keyHandler.onUp(event, this);
+      if (result !== false) {
+        const options = result as EventHandlerResultOptions;
+        if (options?.stopPropagation ?? true) event.stopPropagation();
+        if (options?.preventDefault ?? true) event.preventDefault();
+        break;
+      }
+    }
```
mofojed, line 1751: We could refactor this into another method `notifyKeyboardHandlers` similar to `notifyMouseHandlers` since both `handleKeyUp` and `handleKeyDown` have the same content other than the function name.

Also - looks like you didn't actually end up using this at all? Which is fine, may as well leave it in since it's already done, but ideally we'd have it as it's own PR.

packages/iris-grid/src/mousehandlers/IrisGridCopyCellMouseHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+    event: GridMouseEvent
+  ): EventHandlerResult {
+    if (
+      event.altKey &&
+      !ContextActionUtils.isModifierKeyDown(event) &&
+      !event.shiftKey
+    ) {
+      this.cursor = null;
+      if (gridPoint.columnHeaderDepth !== undefined) {
+        this.irisGrid.copyColumnHeader(gridPoint.column);
```
mofojed, line 33: We should be passing the `columnHeaderDepth` to the `copyColumnHeader` here, and copying the one at the depth you've selected. Right now with a column group created it always copies to lowest header.
Group columns under "Organize Columns" in the menu to try it out.

packages/iris-grid/src/mousehandlers/IrisGridCopyCellMouseHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return false;
+  }
+
+  onMove(
+    gridPoint: GridPoint,
+    _grid: Grid,
+    event: GridMouseEvent
+  ): EventHandlerResult {
+    if (
```
mofojed, line 47: We should be checking if the user `canCopy` here (on the `irisGrid.props`. @dsmmcken should we display a different cursor if copy is not available (ie. user doesn't have permissions)?

packages/iris-grid/src/mousehandlers/IrisGridCopyCellMouseHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+      event.altKey &&
+      !ContextActionUtils.isModifierKeyDown(event) &&
+      !event.shiftKey
+    ) {
+      this.cursor = null;
+      if (gridPoint.columnHeaderDepth !== undefined) {
+        this.irisGrid.copyColumnHeader(gridPoint.column);
+      } else {
+        this.irisGrid.copyCell(gridPoint.column, gridPoint.row);
```
mofojed, line 35: Instead of calling `copyCell` directly, we should be calling `copyRanges` with the cell as a range. That way a notification bar will be displayed indicating a copy has occurred (or if user does not have permission to copy).
Same with the `copyColumnHeader` - it will need to go through that same flow.

mofojed (changes requested):
Good start! Couple of things we can improve.

mofojed:
@georgecwan you can test with copy permissions disabled by setting the permission flag on startup, e.g.:
```
START_OPTS="-Dinternal.webClient.appInit.canCopy=false" ./gradlew server-jetty-app:run -Pdebug -Ppsk=iris
```

packages/iris-grid/src/mousehandlers/IrisGridCopyCellMouseHandler.ts
dsmmcken: Nah, no separate icon, just hide this feature.

mofojed (changes requested):
Something is strange with the snapshot function in general. When I'm trying to copy 100,000 rows, it's only copying 50,000. Table I'm using is:
```
t = empty_table(100000).update(["X=i", "Y=i*i"])
```
I don't think related to your change though, seems to be in the JS API. Interesting. Logged it with ticket https://github.com/deephaven/web-client-ui/issues/1703 for further investigation.

Other than the minor `handleKeyDown` cleanup on Grid, should also write a unit test in `IrisGridCopyHandler.test.tsx` for the copy header cases.

mofojed (approved):

PR #1698, opened 2023-12-19
feat: NavTabList component

mattrunyon:
This is ported from DHE and generalized a bit so it can be used in DHC and DHE. Also cleaned up some of the CSS, but could still use some improvements for theming.

Test in the styleguide Navigations section. The 1st tab in both lists is non-closable. Add custom context actions for the styleguide which selects the tab to the left/right of the tab you opened the context menu on.


packages/components/src/navigation/NavTab.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+interface NavTabProps {
+  tab: NavTabItem;
+  onSelect: (key: string) => void;
+  onClose?: (key: string) => void;
+  isActive: boolean;
+  activeRef: React.RefObject<HTMLDivElement>;
+  index: number;
+  isDraggable: boolean;
+  contextActions?: ResolvableContextAction[];
```
mofojed, line 18: Should take `ResolvableContextAction | ResolvableContextAction[]` (like `ContextActions actions` prop does

packages/components/src/navigation/NavTabList.tsx
```diff
@@ -0,0 +1,455 @@
...
+      <NavTab
+        tab={tab}
+        key={key}
+        index={index}
+        isActive={isActive}
+        activeRef={activeTabRef}
+        onSelect={onSelect}
+        onClose={onClose}
+        isDraggable={isReorderAllowed}
+        contextActions={tabContextActionMap.get(key)}
```
mofojed, line 378: Instead of making all the context actions ahead of time, you should be passing a `ResolvableContextAction`, which can just be a function to get the context actions for that tab item, and then they generate on demand when the item is right-clicked.
And no need to join them, context actions already handles that. So you can pass:
```
contextActions={[() => makeBaseContextItems(tab, tabs, onClose), () => (makeContextItems?.(tab) ?? [])}
```
Or you can just do that in the `tabContextActionMap` above, so it'd be:
```
      const contextActions = [
        () => makeBaseContextItems(tab, tabs, onClose),
        () => (makeContextItems?.(tab) ?? []),
      ];
```

packages/components/src/navigation/NavTabList.tsx
```diff
@@ -0,0 +1,455 @@
...
+
+  /**
+   * Context items to add to the tab in addition to the default items.
+   * The default items are Close, Close to the Right, and Close All.
+   * The default items have a group value of 20.
+   *
+   * @param tab The tab to make context items for
+   * @returns Additional context items for the tab
+   */
+  makeContextItems?: (tab: T) => ResolvableContextAction[];
```
mofojed, line 53: I'd rather name this `contextActionsForTab` or `makeContextActions`. Should also be `ResolvableContextAction | ResolvableContextAction[]`

mofojed (changes requested):
Functionality looks good, context menu should be generated on click though

packages/components/src/navigation/NavTabList.tsx
mattrunyon: Ya good point. They're generated ahead of time on DHE and I just kept that behavior

packages/components/src/navigation/NavTabList.tsx
mattrunyon: The (small) benefit to just returning the array is it simplifies the merge code within this component. It's an infrequently used component, and the only actual difference in someone writing this function in DHE is adding square brackets to their return

mofojed (dismissed):

packages/components/src/navigation/NavTabList.scss
```diff
@@ -0,0 +1,166 @@
...
+$tab-button-separator-color: $gray-600;
+
+$tab-control-btn-offset: -8px;
+
+.nav-container {
+  display: flex;
+  flex-shrink: 0;
+
+  .nav-tabs {
+    --dragging-ant-color: $gray-300;
```
dsmmcken, line 27: unused?

packages/components/src/navigation/NavTabList.scss
```diff
@@ -0,0 +1,166 @@
...
+    flex-wrap: nowrap;
+    overflow-x: hidden;
+    position: relative;
+
+    &.dragging {
+      @include ants-base($gray-300, $background);
+    }
+
+    .btn-nav-tab {
+      // border: $tab-drag-border-width solid transparent;
```
dsmmcken, line 41: meant to be commented or deleted?

dsmmcken (changes requested):
Yes, that's the correct dh-color variable.

packages/components/src/navigation/NavTabList.scss
mattrunyon: Ya I think I initially tried to pass this to the `ants-base` mixin, but it can't handle css variables properly

packages/components/src/navigation/NavTabList.scss
mattrunyon: I think deleted. Looks like the base style sheet includes this border for `.btn-link:not(.btn-link-icon)` already, so not sure if we need this here as it's redundant

mofojed (approved):

dsmmcken (approved):

PR #1702, opened 2023-12-21
fix: column preview types not accurate

### LINKED ISSUE #1597: UI calls column types "string" when they have no native arrow js-api implementation
**Description**

In the UI, columns whose types should be (for example) `ZoneId` or `LocalDateTime` get called `String` types when hovering over the column names. These columns should be called their correct types, with an additional message that mentions the string preview.

**Steps to reproduce**

Run the following code:
```

t = empty_table(10).update(['tz1 = parseTimeZone("GMT")', 'tz2 = parseTimeZone("America/New_York")'])
t = t.update(["ldate1 = '2022-03-28'", 'ldate2 = parseLocalDate("2075-08-08")'])
t = t.update(["ltime1 = '10:15:30.46'", 'ltime2 = parseLocalTime("12:01:01.4567")'])
```

**Expected results**

When I hover my mouse over any of these columns, I expect to see their types displayed as `ZoneId`, `LocalDate`, etc.

**Actual results**

The types get labelled and displayed as strings, but the metadata table confirms that they are actually of the correct type:

Screenshot 2023-10-24 at 10 30 05 AM
Screenshot 2023-10-24 at 10 30 17 AM

**Versions**

 - Engine Version: 0.29.0
 - Java Version: 17
 - OS: MacOS
 - Browser: Safari
 - Docker: NA - built & run with gradle

ethanalvizo:
Closes #1597

### DIFF
--- a/packages/iris-grid/src/ColumnStatistics.tsx
+++ b/packages/iris-grid/src/ColumnStatistics.tsx
@@ -171,7 +171,10 @@ class ColumnStatistics extends Component<
       model.isColumnStatisticsAvailable;
     const statisticElements = [];
     const columnType = column.type.substring(column.type.lastIndexOf('.') + 1);
-    const description = column.description === null ? null : column.description;
+    const description =
+      column.description === null
+        ? null
+        : column.description.substring(column.description.lastIndexOf(' ') + 1);
     if (statistics != null) {
       for (let i = 0; i < statistics.length; i += 1) {
         const { operation, className, value, type } = statistics[i];
@@ -201,7 +204,7 @@ class ColumnStatistics extends Component<
       <div className="column-statistics">
         <div className="column-statistics-title">
           {column.name}
-          <span className="column-statistics-type">&nbsp;({columnType})</span>
+          <span className="column-statistics-type">&nbsp;({description})</span>
           <CopyButton
             className="column-statistics-copy"
             tooltip="Copy column name"
@@ -209,7 +212,9 @@ class ColumnStatistics extends Component<
           />
         </div>
         {description != null && (
-          <div className="column-statistics-description">{description}</div>
+          <div className="column-statistics-description">
+            Previewing as {columnType}
+          </div>
         )}
         {columnIndex != null && !model.isColumnSortable(columnIndex) && (
           <div className="column-statistics-status">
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/ColumnStatistics.tsx
```diff
@@ -170,8 +170,14 @@ class ColumnStatistics extends Component<
       statistics == null &&
       model.isColumnStatisticsAvailable;
     const statisticElements = [];
-    const columnType = column.type.substring(column.type.lastIndexOf('.') + 1);
-    const description = column.description === null ? null : column.description;
+    let columnType = column.type.substring(column.type.lastIndexOf('.') + 1);
+    let description;
+    if (column.description != null && column.description !== undefined) {
```
mofojed, line 175: This check is redundant. `!= null` will mean that `!== undefined` is always true. You'd only need to strictly check against `undefined` if you strictly checked against null, e.g. `!== null` instead of just `!= null`.
When checking against both null and undefined, doing a loose check `!= null` is fine.

mofojed (changes requested):
So there's a problem I didn't realize when assigning this ticket - we don't actually _know_ it's a preview column, the column description is used on non-preview column types as well. For example, on rollup tables:
```
from deephaven import read_csv, agg

insurance = read_csv("https://media.githubusercontent.com/media/deephaven/examples/main/Insurance/csv/insurance.csv")

agg_list = [agg.avg(cols=["bmi", "expenses"])]
by_list = ["region", "age"]

test_rollup = insurance.rollup(aggs=[], by=by_list, include_constituents=True)
insurance_rollup = insurance.rollup(aggs=agg_list, by=by_list, include_constituents=True)
```
If you hover over `expenses`, you'll see there's an actual description:
image

So we don't want to swap the description in that case.

I think we need to push back on this ticket as, and either have a specific field detailing the preview type (rather than stuffing it in the description). Then we can treat it a little more differently (and also disable stuff like filtering, rather than trying to filter it as a string).

Right now the description is getting set in the API for column previews: https://github.com/deephaven/deephaven-core/blob/01b6cf49dad3e855067f0cb411a858c89b875d2d/engine/table/src/main/java/io/deephaven/engine/table/impl/preview/ColumnPreviewManager.java#L121

We have tickets on the Core side for this: https://github.com/deephaven/deephaven-core/issues/3358
and https://github.com/deephaven/deephaven-core/issues/2102
So I don't think we can complete this ticket as desired until addressing those two tickets first.

ethanalvizo:
Waiting on core tickets mentioned above to go through first

PR #1706, opened 2024-01-02
feat: Action button tooltips

### LINKED ISSUE #1705: feat: Action button tooltips
The `ActionButtonDialogTrigger` `ConfirmActionButton` and `IconActionButton` components should support tooltips. This is specifically needed to support DH-16121, but it's a good general feature to have for action buttons.

bmingles:
Added tooltip prop to action button components

resolves #1705


packages/components/src/dialogs/ActionButtonDialogTrigger.tsx
```diff
@@ -1,15 +1,18 @@
+/* eslint-disable camelcase */
```
mofojed, line 1: Unnecessary. In cases where we do need to go against eslint rules, should opt for disabling a single line only wherever possible.
```suggestion
```

mofojed (changes requested):

mofojed (approved):

PR #1713, opened 2024-01-08
fix: Moved logos so they show in production build

### LINKED ISSUE #1712: Logos missing in production build
**Description**

Themed logos don't work in production deployment

**Steps to reproduce**

- `npm run build`

**Expected results**

Logo should be present after build

**Actual results**

- `packages/components/theme/logos` folder is not copied to `dist` where they are expected by `Logo.css`
- `npm run preview` shows missing logo**Additional details and attachments**

**Versions**

0.31.0-SNAPSHOT

bmingles:
Moved logos so they show in production build

fixes #1712

**Test Plan**
`npm run build`
`npm run preview`

Open DH, logo should show up for both dark and light themes


mofojed (changes requested):
This won't work when you actually publish the package. The outputted result in dist/theme/Logo.css is:
```
```

bmingles:
@mofojed I fixed the files config. Tested via docker compose build of deephaven-core pointed to alpha

mofojed (approved):
In the future, we may want to consider doing an `assets` folder at the top level, and then having css/scss/svg be subfolders or something.

PR #1714, opened 2024-01-08
feat: Multiple dashboards

### LINKED ISSUE #1683: Support multiple dashboards
For Deephaven UI, we need multiple dashboard support so a user can create a dashboard which opens in a separate tab within the app. This should work like the enterprise tabs at the top of the app where dashboards are closeable and movable (except main dashboard)

Will need a way to emit a dashboard open event as well as saving the open dashboards and their layouts

mattrunyon:
Fixes #1683

Test with the dh.ui plugin from https://github.com/deephaven/deephaven-plugins/pull/176

Example dashboard code

```
from deephaven import ui, time_table
from deephaven.ui import use_memo, use_state

def use_wave_input():
    """
    Demonstrating a custom hook.
    Creates an input panel that controls the amplitude, frequency, and phase for a wave
    """
    amplitude, set_amplitude = use_state(1.0)
    frequency, set_frequency = use_state(1.0)
    phase, set_phase = use_state(1.0)

    input_panel = ui.flex(
        ui.slider(
            label="Amplitude",
            default_value=amplitude,
            min_value=-100.0,
            max_value=100.0,
            on_change=set_amplitude,
            step=0.1,
        ),
        ui.slider(
            label="Frequency",
            default_value=frequency,
            min_value=-100.0,
            max_value=100.0,
            on_change=set_frequency,
            step=0.1,
        ),
        ui.slider(
            label="Phase",
            default_value=phase,
            min_value=-100.0,
            max_value=100.0,
            on_change=set_phase,
            step=0.1,
        ),
        direction="column",
    )

    return amplitude, frequency, phase, input_panel

@ui.component
def multiwave():
    amplitude, frequency, phase, wave_input = use_wave_input()

    tt = use_memo(lambda: time_table("PT1s").update("x=i"), [])
    t = use_memo(
        lambda: tt.update(
            [
                f"y_sin={amplitude}*Math.sin({frequency}*x+{phase})",
                f"y_cos={amplitude}*Math.cos({frequency}*x+{phase})",
                f"y_tan={amplitude}*Math.tan({frequency}*x+{phase})",
            ]
        ),
        [amplitude, frequency, phase],
    )
    p_sin = use_memo(
        lambda: Figure().plot_xy(series_name="Sine", t=t, x="x", y="y_sin").show(), [t]
    )
    p_cos = use_memo(
        lambda: Figure().plot_xy(series_name="Cosine", t=t, x="x", y="y_cos").show(),
        [t],
    )
    p_tan = use_memo(
        lambda: Figure().plot_xy(series_name="Tangent", t=t, x="x", y="y_tan").show(),
        [t],
    )

    return [
        ui.column(
            ui.row(
                ui.stack(
                    ui.panel(wave_input, title="Wave Input"),
                    ui.panel(t, title="Wave Table"),
                    activeItemIndex=0
                ),
                height=25
            ),
            ui.row(
                ui.stack(ui.panel(p_sin, title="Sine"), width=50),
                ui.stack(ui.panel(p_cos, title="Cosine"), width=30),
                ui.stack(ui.panel(p_tan, title="Tangent"))
            )
        )
    ]

mw = ui.dashboard(multiwave())
```


packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -415,21 +432,42 @@ export class AppMainContainer extends Component<
...
   }
 
   handleGoldenLayoutChange(goldenLayout: GoldenLayout): void {
     this.goldenLayout = goldenLayout;
-  }
-
-  handleLayoutConfigChange(layoutConfig?: DashboardLayoutConfig): void {
-    const { updateWorkspaceData } = this.props;
-    updateWorkspaceData({ layoutConfig });
+    this.goldenLayout.eventHub.on(
```
mofojed, line 445: As mentioned in other review, let's have this defined as `emitDashboardOpenEvent` and `listenDashboardOpenEvent` methods that can enforce the typing.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -689,6 +683,15 @@ export class AppMainContainer extends Component<
     });
   }
 
+  openDashboard(widget: VariableDefinition): void {
+    const { connection } = this.props;
+    this.emitLayoutEvent('dashboardOpen', {
```
mofojed, line 688: Is this event getting used anywhere? I think this method unused and can be removed

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -706,11 +709,73 @@ export class AppMainContainer extends Component<
...
+      const [t] = newTabs.splice(from, 1);
+      newTabs.splice(to, 0, t);
+      return { tabs: newTabs };
+    });
+  }
+
+  handleTabClose(tabId: string): void {
+    const { setDashboardData } = this.props;
+    // TODO: Figure out how to remove the dashboard data
+    // without updates after this recreating some dashboard data
```
mofojed, line 732: Not sure I follow this TODO - what's wrong with this?

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -706,11 +709,73 @@ export class AppMainContainer extends Component<
...
+    id: string;
+    layoutConfig: ItemConfigType[];
+  }[] {
+    const { tabs } = this.state;
+    const { allDashboardData, workspace } = this.props;
     const { data: workspaceData } = workspace;
     const { layoutConfig } = workspaceData;
+    // TODO: Move this to read dashboard data
+    // const { dashboardData } = this.props;
+    // return (dashboardData.layoutConfig ?? []) as ItemConfigType[];
```
mofojed, line 763: Guessing this is waiting on @vbabich 's piece?

packages/code-studio/src/main/EmptyDashboard.tsx
```diff
@@ -1,16 +1,9 @@
-import React, { MouseEvent } from 'react';
+import React from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Button } from '@deephaven/components';
 import { vsArrowUp } from '@deephaven/icons';
 import './EmptyDashboard.scss';
 
-export interface EmptyDashboardProps {
-  onAutoFillClick?: (event: MouseEvent) => void;
```
mofojed, line 8: We're just getting rid of the autofill behaviour entirely now? Why?

packages/dashboard/src/LazyDashboard.tsx
```diff
@@ -0,0 +1,52 @@
...
+    [id, dispatch]
+  );
+
+  if (!isLoaded && isActive) {
+    setIsLoaded(true);
+  }
+
+  if (!isLoaded) {
+    return <LoadingOverlay />;
+  }
```
mofojed, line 37: Don't really need separate if blocks for `isLoaded` here, but not a big deal:
```suggestion
  if (!isLoaded) {
    if (isActive) {
      setIsLoaded(true);
    }
    return <LoadingOverlay />;
  }
```

I tend to shy away from calling setState functions from within the render as well, though it does seem to be valid: https://react.dev/reference/react/useState#storing-information-from-previous-renders
My instinct would have been to write this as an effect, e.g.

```
useEffect(() => {
  if (isActive) {
    setIsLoaded(true);
  }
}, [isActive]);
```

mofojed (changes requested):

packages/code-studio/src/main/AppMainContainer.tsx
mattrunyon: So right now when you close a dashboard, we null out the dashboard data, but then other parts of the app emit stuff like the `closed` array which recreates dashboard data.

If we track dashboards via a different property then that can be fine. But since we have no way of the user managing dashboards, we need to clean them up at some point. Otherwise the dashboard data grows unbounded with unreachable dashboards

packages/dashboard/src/LazyDashboard.tsx
mattrunyon: My instinct is that too, but at some point I read through this article on the new react docs and this is actually in the examples for when not to use an effect.

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

packages/code-studio/src/main/EmptyDashboard.tsx
mattrunyon: I forgot to commit stashed changes. Fixed

packages/dashboard/src/LazyDashboard.tsx
mofojed: Neat

packages/code-studio/src/main/AppDashboards.tsx
```diff
@@ -0,0 +1,95 @@
...
+export function AppDashboards({
+  dashboards,
+  activeDashboard,
+  onGoldenLayoutChange,
+  plugins,
+  onAutoFillClick,
+}: AppDashboardsProps): JSX.Element {
+  const connection = useConnection();
+
+  const hydratePanel = useCallback(
```
mofojed, line 35: How will this work from Enterprise? We need to be able to pass in the hydration function and pull the query from the metadata for Enterprise.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -415,21 +434,40 @@ export class AppMainContainer extends Component<
...
   }
 
-  handleDataChange(data: AppDashboardData): void {
+  handleDataChange(data: DashboardData): void {
     const { updateWorkspaceData } = this.props;
 
     // Only save the data that is serializable/we want to persist to the workspace
-    const { closed, filterSets, links } = data;
-    updateWorkspaceData({ closed, filterSets, links });
+    const { closed } = data;
```
mofojed, line 441: I actually don't know why we're saving `closed` with the workspace data as well - that's something that should just be saved with the dashboard (for re-opening closed markdown panels, though it'd be useful for re-opening _any_ panel as well).

packages/code-studio/src/main/AppMainContainer.tsx
mofojed: See my other comment about remove `closed` array from the workspace data... does this TODO still need to be followed up on? File a separate ticket for it?

mofojed (changes requested):

packages/code-studio/src/main/AppDashboards.tsx
mattrunyon: I don't think this hydration specifically will affect Enterprise. This is just moving the hydration function from `AppMainContainer` to `AppDashboards` since `AppMainContainer` doesn't need to know anything about this once the dashboards component is split out

Enterprise has its own hydration which would be part of the `Dashboard` already

As far as Enterprise changes go, it will need to add `dashboardPluginData` and also listen to the event for creating a dashboard like DHC does. There might be something w/ the plugin and plugin data that we need to modify to work w/ enterprise PQs since right now the plugin is responsible for fetching its  widget

mattrunyon: I've thought about extending `WidgetPlugin` in some manner so that it doesn't have to mount a panel. Right now it's kind of hacky IMO because the dashboard widget from dh.ui is loaded as part of a dashboard plugin. Instead it could be a widget plugin that mounts, but doesn't render a panel.

Or we need some sort of context/hook that lets the plugin get its own fetch function or create one with just the widget name and type. In DHE the hook could wrap any other info like PQ data into the fetch function it provides

packages/code-studio/src/main/AppMainContainer.tsx
mattrunyon: I'm just going to undo this for now as out of scope for this PR. Should include it with persistence

packages/code-studio/src/main/AppMainContainer.tsx
mattrunyon: I've removed the todo and the call to `setDashboardData`. Will leave this for a separate persistence PR

mattrunyon:
I've fixed the console focus issue. The problem was if the console was focused when a dashboard was opened, it didn't resize monaco when it became visible again. In enterprise we're doing some extra stuff that calls `updateDimensions` on the console from the parent container when the tab is focused. I added a resize observer instead which fires when the display changes so it doesn't need the parent container to tell it to resize

mofojed (dismissed):

packages/code-studio/src/main/AppDashboards.tsx
mofojed: A plugin that mounts but doesn't have a panel sounds kinda neat. Need to give it more thought.

mofojed (approved):

PR #1717, opened 2024-01-08
fix: GoToRow timestamp fails when selected row is out of view

### LINKED ISSUE #1561: Unable to pick a Timestamp column in the GoToRow dialog if the selected row is out of the viewport
**Description**

Attempting to select a Timestamp column throws an error in the browser log and fails to update the column dropdown in the GoToRow dialog if the selected row in the table is out of the viewport.

**Steps to reproduce**

1. Open a large table with a Timestamp column, i.e. example A3 from the Real-time table ops notebook:
```

from datetime import datetime, timedelta
import random
import string

def random_character() -> str:
    return random.choice(string.ascii_uppercase)

def create_random_table(time_interval, start_time=None):
    """
    Creates a Deephaven table containing rows of random integers from 1 to 99, random
    uppercase characters, and timestamps.

    Parameters:
        time_interval (str||int): Duration string or int representation of the time interval between rows.
        start_time (str||Instant): Optional string or DateTime representation of the start time.
    Returns:
        A Deephaven Table containing the random data.
    """
    table = None
    if start_time is None:
        table = time_table(time_interval)
    else:
        table = time_table(period=time_interval, start_time=start_time)

    return table.update(formulas=["Number = randomInt(1, 100)", "Character = random_character()", "Boolean = randomBool()"])

daily_data_0 = create_random_table(time_interval="PT1M", start_time=datetime.now() - timedelta(hours=240, seconds=2))
```
2. Select the first row by clicking on a non-timestamp column, scroll the table to the bottom
3. Press Cmd+G to open the GoToRow dialog
4. Select Timestamp in the column selection dropdown

**Expected results**
The timestamp column is selected

**Actual results**
Timestamp not selected,
browser log shows an exception: 
```
Uncaught TypeError: Cannot read properties of undefined (reading ‘asDate’)
    at PYt (IrisGridUtils.js:29:16)
    at Ns.convertValueToText (IrisGridUtils.js:801:46)
    at lm.handleGotoValueSelectedColumnNameChanged (IrisGrid.js:2907:32)
    at onChange (GotoRow.js:270:7)
    at Object.YXe (react-dom.production.min.js:52:317)
    at QXe (react-dom.production.min.js:52:471)
    at JXe (react-dom.production.min.js:53:35)
    at dfe (react-dom.production.min.js:100:68)
    at Kke (react-dom.production.min.js:101:380)
    at react-dom.production.min.js:113:65
    at Tke (react-dom.production.min.js:292:189)
    at XXe (react-dom.production.min.js:50:57)
    at Zke (react-dom.production.min.js:105:469)
    at cte (react-dom.production.min.js:75:265)
    at ute (react-dom.production.min.js:74:124)
    at t.unstable_runWithPriority (scheduler.production.min.js:18:343)
    at Nx (react-dom.production.min.js:122:325)
    at Cke (react-dom.production.min.js:292:48)
    at hYe (react-dom.production.min.js:73:352)
```

If on step 2 I click on the Timestamp column, the GoToRow dialog won't open at all with the same error in the browser log.

**Versions**

Engine Version: 0.28.0
Web UI Version: 0.47.0
Java Version: 17.0.8.1
Barrage Version: 0.6.0

wusteven815:
- Fixes #1561 
- Added the optional chaining operator to prevent an error from being thrown

### DIFF
--- a/packages/iris-grid/src/IrisGridUtils.ts
+++ b/packages/iris-grid/src/IrisGridUtils.ts
@@ -188,7 +188,7 @@ function isValidIndex(x: number, array: readonly unknown[]): boolean {
 }
 
 function isDateWrapper(value: unknown): value is DateWrapper {
-  return (value as DateWrapper).asDate != null;
+  return (value as DateWrapper)?.asDate != null;
 }
 
 class IrisGridUtils {


mofojed (approved):
Great work!

Issue #1701, opened 2023-12-20
Reject promise immediately if variable is not in first message of field updates


mofojed:
Right now there is a 10s timeout when we're fetching an object: https://github.com/deephaven/web-client-ui/blob/73e0b658edffc7ef89b3b786f3fe30c0e64c96f9/packages/jsapi-utils/src/ConnectionUtils.ts#L37
However, if the object isn't in the first message, we should be able to reject immediately. The underlying call though implies that all known objects should be returned in the first message. From https://github.com/deephaven/deephaven-core/blob/6d203ab9dfbe1fbe203be161d340b8ee0c2e29ed/proto/proto-backplane-grpc/src/main/proto/deephaven/proto/application.proto#L27
>  The first received message contains all fields that are currently available on the worker. None of these fields will be RemovedFields.

mofojed:
Good first issue for @wusteven815

PR #1718, opened 2024-01-08
feat: Reject promise immediately if var not found

wusteven815:
- Implements #1701 
- `fetchVariableDefinition` now throws an `Error` when a variable is not found (instead of throwing `TimeoutError`)
- `TimeoutError` still used and will be thrown on an actual timeout
- Tests added to check for the correct error

### DIFF
--- a/packages/jsapi-utils/src/ConnectionUtils.ts
+++ b/packages/jsapi-utils/src/ConnectionUtils.ts
@@ -34,10 +34,12 @@ export function fetchVariableDefinition(
      */
     function handleFieldUpdates(changes: VariableChanges): void {
       const definition = changes.created.find(def => def.title === name);
+      clearTimeout(timeoutId);
+      removeListener?.();
       if (definition != null) {
-        clearTimeout(timeoutId);
-        removeListener?.();
         resolve(definition);
+      } else {
+        reject(new TimeoutError(`Variable ${name} not found`));
       }
     }
 
--- a/packages/jsapi-utils/src/ConnectionUtils.ts
+++ b/packages/jsapi-utils/src/ConnectionUtils.ts
@@ -25,7 +25,7 @@ export function fetchVariableDefinition(
 
     const timeoutId = setTimeout(() => {
       removeListener?.();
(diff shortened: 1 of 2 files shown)


packages/jsapi-utils/src/ConnectionUtils.ts
```diff
@@ -34,10 +34,12 @@ export function fetchVariableDefinition(
...
     function handleFieldUpdates(changes: VariableChanges): void {
       const definition = changes.created.find(def => def.title === name);
+      clearTimeout(timeoutId);
+      removeListener?.();
       if (definition != null) {
-        clearTimeout(timeoutId);
-        removeListener?.();
         resolve(definition);
+      } else {
+        reject(new TimeoutError(`Variable ${name} not found`));
```
mofojed, line 42: So in this case, we do _not_ want a `TimeoutError`, since this isn't being triggered from a `Timeout`. Just reject with `new Error(...)` instead:
```suggestion
        reject(new Error(`Variable ${name} not found`));
```

mofojed (changes requested):

packages/jsapi-utils/src/ConnectionUtils.ts
mofojed: We could modify the TimeoutError above to be different as well, e.g. `TimeoutError(\`Timeout looking for variable ${name}\`)`

packages/jsapi-utils/src/ConnectionUtils.ts
```diff
@@ -25,7 +25,7 @@ export function fetchVariableDefinition(
 
     const timeoutId = setTimeout(() => {
       removeListener?.();
-      reject(new TimeoutError(`Variable ${name} not found`));
+      reject(new TimeoutError(`Timeout looking for Variable ${name}`));
```
mofojed, line 28:
```suggestion
      reject(new TimeoutError(`Timeout looking for variable ${name}`));
```

mofojed (changes requested):
Looks good! Just a minor correction to the error message.

mofojed (approved):

PR #1721, opened 2024-01-10
feat: Default Plotly map colors

bmingles:
- Added theme variables for Plotly map colors. Wired them up to the default layout
- Removed unused chart theme variables
- Fixed scroll issue in styleguide

Testing
Here's a Python snippet that can be used to see the map colors
```python
import random 

def update(fig):
    pass

sourceh = time_table("PT1S").update(formulas=[
    "X = (float) random.uniform(-90, 90)", 
    "Y = (float) random.uniform(-180, 180)", 
    "Z = (float)random.gauss(3, 3)",
    "l1 = i % 20",
    "l2 = i % 30",
    ])

figs = dx.scatter_geo(
    sourceh, 
    lat="X", 
    lon="Y", 
    by="l1", 
    size="Z",
    color_discrete_sequence=["salmon", "lemonchiffon"],
    projection="natural earth",
    unsafe_update_figure=update
)
```


packages/chart/src/ChartTheme.module.scss
```diff
@@ -9,13 +9,15 @@
...
-  error-band-fill-color: var(--dh-color-chart-error-band-fill);
   ohlc-increasing: var(--dh-color-chart-ohlc-increase);
   ohlc-decreasing: var(--dh-color-chart-ohlc-decrease);
+
+  /* Geo */
+  coastline-color: var(-dh-color-plotly-geo-coastline);
+  land-color: var(--dh-color-plotly-geo-land);
+  ocean-color: var(--dh-color-plotly-geo-ocean);
+  lake-color: var(--dh-color-plotly-geo-lake);
+  river-color: var(--dh-color-plotly-geo-river);
```
mattrunyon, line 22: I think these should be `--dh-color-chart-geo-*` to match the name pattern of the other variables

packages/chart/src/ChartUtils.ts
```diff
@@ -1874,25 +1874,56 @@ class ChartUtils {
...
+      // gridcolor,
+      // linecolor,
+      // zerolinecolor,
+
+      // Used by `getPlotlyErrorBars()`
+      // error_band_line_color,
+
+      /* Used by `addStylingToSeriesData()` */
+      // ohlc_increasing,
+      // ohlc_decreasing,
```
mattrunyon, line 1904: Why are these commented out here?

packages/chart/src/ChartUtils.ts
```diff
@@ -1874,25 +1874,56 @@ class ChartUtils {
 
   makeDefaultLayout(theme: ChartTheme): Partial<Layout> {
     const { dh } = this;
+
```
mattrunyon, line 1877: Can you add a jsdoc to this method and include a link to https://plotly.com/javascript/reference/layout/ in it? It doesn't include everything (they're spread under different pages in the sidebar), but gives the starting point at least

mattrunyon (changes requested):

mattrunyon:
@dsmmcken Should we remove the frame? We can control `framecolor`, `framewidth`, and `showframe`

image

packages/chart/src/ChartUtils.ts
```diff
@@ -1908,8 +1929,21 @@ class ChartUtils {
         yaxis: this.makeLayoutAxis(dh.plot.AxisType.Y, theme),
         zaxis: this.makeLayoutAxis(dh.plot.AxisType.Z, theme),
       },
+      geo: {
+        showcoastlines: true,
```
mattrunyon, line 1933: Need to add `geo.bgcolor`. I'd say just set it to `plot_bgcolor` instead of making a separate variable for it

dsmmcken:
In your example, I would expect the area inside the red line to have the slightly lighter bg color, and everything outside to have the content-bg color.

Then hiding the frame is fine.

bmingles:
I set `bgcolor` to `paper_bgcolor` instead of `plot_bgcolor` so that it gets the outer color and keeps the frame boundary as the delineation of the 2 colors

mattrunyon (approved):

Issue #1679, opened 2023-12-08
Theming - Preload variable improvements


bmingles:
Certain UI elements are visible before theming plugins can finish loading. e.g. in DHC, the background color and loading spinner. In DHE everything in the login form. To improve the user experience, we are storing certain "preload" variables in localStorage and then loading them on next page load independent of the plugins. Preload variable names + default values are currently stored in `ThemeModel` `DEFAULT_PRELOAD_DATA_VARIABLES`.

There's a couple of things that don't quite feel right with the current implementation.

Variables for the login page + animation technically shouldn't be needed for DHC preload, but they are needed for DHE. 
```
| '--dh-color-login-form-bg'
| '--dh-color-login-status-message'
| '--dh-color-login-logo-bg'
| '--dh-color-login-footer-fg'
| '--dh-color-random-area-plot-animation-fg-fill'
| '--dh-color-random-area-plot-animation-fg-stroke'
| '--dh-color-random-area-plot-animation-bg'
| '--dh-color-random-area-plot-animation-grid';
```
It would be nice to have DHE be responsible for these, but this will require a way for apps to provide preload variables that they need

Also, @mofojed has called out that it seems odd for component specific variables to be declared in the preload as opposed to more generic semantic variables. We should think through this one to see if there is a better approach.

PR #1723, opened 2024-01-10
feat: Improved preload variable handling

### LINKED ISSUE #1695: Theming - merge default preload data into cached preload data
We "preload" a few css theming variables that are needed before plugins have a chance to load. `ThemeUtils.preloadTheme()` first attempts to load the variables from the localStorage cache. If this fails, or if no cached variables exist, it falls back to `calculatePreloadStyleContent()` where default values for our variables are provided. This means that when variables exist in the cache, `calculatePreloadStyleContent()` won't be called before the initial page load. This is by design under normal circumstances, but in cases where we may update the list of preload variables, it means the user can experience a 1 time page load that is missing colors. This will resolve itself after first login but does introduce the possibility of a user experiencing missing colors on the login page immediately following a DHC update.

To see the issue, logout of DHE, run the following in dev tools, and refresh the page
```
localStorage.setItem('deephaven.themeCache', '{"themeKey":"default-dark","preloadStyleContent":":root{--dh-color-input-bg:#fff}"}')
```

Note that DHC is less impacted by this issue since plugins load before login, but a user may still se the loading spinner or background color be off on initial page load.

One fix that could mitigate this would be to always call `calculatePreloadStyleContent()` and then override variables that are in the cache. This would ensure we always get new variables. It would not handle a scenario where we change a default variable value, but that's probably ok since things will resolve after login.

bmingles:
- Preload now always applies default preload values before applying cached ones. This ensures that cached preload values don't prevent new defaults from being applied if / when we add new variables to the preload list
- Default preload variables can now be passed in to ThemeUtils + ThemeProvider. This will allow DHE to specify additional variables if needed

resolves #1695 and part of #1679

### DIFF
--- a/packages/code-studio/src/styleguide/SpectrumComparison.tsx
+++ b/packages/code-studio/src/styleguide/SpectrumComparison.tsx
@@ -24,6 +24,7 @@ import {
   RadioItem,
   Select,
 } from '@deephaven/components';
+import { EMPTY_FUNCTION } from '@deephaven/utils';
 import {
   SAMPLE_SECTION_E2E_IGNORE,
   SPECTRUM_COMPARISON_SAMPLES_ID,
@@ -43,10 +44,6 @@ const options = [
   { title: 'Two', value: '2' },
   { title: 'Three', value: '3' },
 ];
-
-function noop(): void {
-  // no-op
-}
 
 export function SpectrumComparison(): JSX.Element {
   const [isChecked, setIsChecked] = useState(false);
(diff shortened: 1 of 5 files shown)


packages/code-studio/src/styleguide/SpectrumComparison.tsx
```diff
@@ -44,6 +44,10 @@ const options = [
   { title: 'Three', value: '3' },
 ];
 
+function noop(): void {
+  // no-op
+}
+
```
mofojed, line 50: You can use the `EMPTY_FUNCTION` constant instead from `@deephaven/utils`.

mofojed (approved):

packages/code-studio/src/styleguide/SpectrumComparison.tsx
bmingles: ah, I went ahead and updated

mofojed (approved):

Issue #1685, opened 2023-12-11
Open file blocks logout


bmingles:
**Description**

It seems that having an open file stops logout from working.

**Steps to reproduce**

- Run DHC such that it requires PSK for login
- Login
- Open a file
- Try logging out. It should fail
- Close the file
- Try logging out again. It should succeed

**Expected results**

Logout should always work

**Actual results**

Logout only succeeds when no files are open

**Versions**

Engine Version: 0.31.0-SNAPSHOT
Web UI Version: 0.56.0
Java Version: 21.0.1
Barrage Version: 0.6.0

mofojed:
Would be nice to take another look at this. The RefreshTokenBootstrap has been removed.

bmingles:
@mofojed  This still seems to be reproducible on `ghcr.io/deephaven/server-slim:edge`

PR #1726, opened 2024-01-11
fix: Handle undefined DashboardData props

### LINKED ISSUE #1684: ConsolePanel should not render if there is no session
**Description**

The `ConsolePanel` component has a connect prop to `sessionWrapper`. If this prop is undefined / null, the component crashes when attempting to destructure the prop with an error "Cannot destructure property 'config' of 'sessionWrapper'". This component should not be rendered at all if `sessionWrapper` is undefined.

**Steps to reproduce**

- Login
- Open a file (this seems to be necessary to fail the logout. See https://github.com/deephaven/web-client-ui/issues/1685 for related issue)
- Logout

**Expected results**

There should be no console errors. Console should not be rendered if no session is available

**Actual results**

On login there is a console error "Cannot destructure property 'config' of 'sessionWrapper'". On logout, the same error occurs.

image

**Versions**

Web UI Version: 0.56.0
Java Version: 21.0.1
Barrage Version: 0.6.0

bmingles:
- Handle undefined props on DashboardData. Don't render console panel if no session
- Fixed a set state on unmounted HeapUsage bug
- Removed chatty async interval debug logging

fixes #1684

**Testing**
- Run server with PSK enabled
- Login and open any file
- Logout should no longer show "ConsolePanel.tsx:372 Uncaught TypeError: Cannot destructure property 'config' of 'sessionWrapper' as it is undefined." error in debug console (Note that login will still fail until #1685 is done)

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -123,7 +123,7 @@ interface AppMainContainerProps {
   match: {
     params: { notebookPath: string };
   };
-  connection: IdeConnection;
+  connection?: IdeConnection;
   session?: IdeSession;
   sessionConfig?: SessionConfig;
   setActiveTool: (tool: string) => void;
@@ -255,6 +255,10 @@ export class AppMainContainer extends Component<
 
   initWidgets(): void {
     const { connection } = this.props;
+    if (connection == null) {
+      return;
+    }
+
     if (connection.subscribeToFieldUpdates == null) {
       log.warn(
         'subscribeToFieldUpdates not supported, not initializing widgets'
@@ -614,6 +618,10 @@ export class AppMainContainer extends Component<
 
   startListeningForDisconnect(): void {
     const { connection } = this.props;
+    if (connection == null) {
+      return;
+    }
+
     connection.addEventListener(
       dh.IdeConnection.EVENT_DISCONNECT,
       this.handleDisconnect
@@ -630,6 +638,10 @@ export class AppMainContainer extends Component<
 
   stopListeningForDisconnect(): void {
     const { connection } = this.props;
+    if (connection == null) {
+      return;
+    }
+
     connection.removeEventListener(
       dh.IdeConnection.EVENT_DISCONNECT,
       this.handleDisconnect
@@ -650,6 +662,7 @@ export class AppMainContainer extends Component<
   ): DehydratedDashboardPanelProps & { fetch?: () => Promise<unknown> } {
     const { connection } = this.props;
     const { metadata } = props;
+
     if (
       metadata?.type != null &&
       (metadata?.id != null || metadata?.name != null)
@@ -666,12 +679,14 @@ export class AppMainContainer extends Component<
               name: metadata.name,
               title: metadata.name,
             };
+
       return {
-        fetch: () => connection.getObject(widget),
+        fetch: async () => connection?.getObject(widget),
         ...props,
         localDashboardId: id,
       };
     }
+
     return DashboardUtils.hydrate(props, id);
   }
 
@@ -684,7 +699,7 @@ export class AppMainContainer extends Component<
     const { connection } = this.props;
     this.emitLayoutEvent(PanelEvent.OPEN, {
       dragEvent,
-      fetch: () => connection.getObject(widget),
+      fetch: async () => connection?.getObject(widget),
       widget,
     });
   }
--- a/packages/console/src/HeapUsage.tsx
+++ b/packages/console/src/HeapUsage.tsx
@@ -4,7 +4,7 @@ import { Tooltip } from '@deephaven/components';
 import type { QueryConnectable } from '@deephaven/jsapi-types';
 import { Plot, useChartTheme } from '@deephaven/chart';
 import Log from '@deephaven/log';
-import { useAsyncInterval } from '@deephaven/react-hooks';
+import { useAsyncInterval, useIsMountedRef } from '@deephaven/react-hooks';
 import './HeapUsage.scss';
 
 const log = Log.module('HeapUsage');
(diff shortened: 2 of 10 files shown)


packages/console/src/HeapUsage.tsx
```diff
@@ -41,9 +41,16 @@ function HeapUsage({
...
 
+  const isMountedRef = useIsMountedRef();
+
   const setUsageUpdateInterval = useCallback(async () => {
     try {
       const newUsage = await connection.getWorkerHeapInfo();
+
+      if (!isMountedRef.current) {
+        return;
+      }
```
mofojed, line 52: Technically this may be incorrect. E.g.
1. First pass, `monitorDuration` is `5000`, and a call to `connection.getWorkerHeapInfo()` is triggered. It is async and hasn't returned yet
2. `monitorDuration` changes to `10000`, triggering another call to `connection.getWorkerHeapInfo()`. This call returns before the call from step 1
3. Call from step 1 returns, now the data is out of date.

In practice that could never happen in this case, since the first call to `connection.getWorkerHeapInfo()` should always return before a subsequent call, but in another async situation, using the `isMountedRef` may not handle all cases.

I typically go with the `isCancelled` pattern, as that handles cases where the dependencies change, not just unmounting. An example from the React docs (looks like they name it `ignore` instead of `isCancelled` now): https://react.dev/reference/react/useEffect#fetching-data-with-effects

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -368,6 +375,11 @@ export class ConsolePanel extends PureComponent<
       timeZone,
       unzip,
     } = this.props;
+
+    if (sessionWrapper == null) {
+      return null;
```
mofojed, line 380: We shouldn't display _nothing_ I don't think we should be displaying an error message. When do we expect this to occur?

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
bmingles: This happens when logging out and in some cases when logging in

packages/console/src/HeapUsage.tsx
bmingles: I think `monitorDuration` is how long it holds onto the history and doesn't impact the interval (unless I am missing something)

I think your appoint does apply to the interval though `isOpen ? hoverUpdateInterval : defaultUpdateInterval`. So ideally we should be cancelling if either the component unmounts `or` if the interval changes?

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
mofojed: It can also happen if remote consoles are disabled. You can do that by running with the flag `deephaven.console.disable=true`, e.g.:
```
START_OPTS="-Ddeephaven.console.disable=true" ./gradlew server-jetty-app:run -Ppsk=iris
```
In this case we're displaying a blank console panel now:
image

Ideally in that scenario, we should display an error message. So I think we should display an appropriate error message here. It is still an edge case though; by default, if remote console is enabled, we won't even add a console panel to the layout. But if they already have a layout saved locally with the ConsolePanel, or they import a layout with ConsolePanel, they'll run into this error.

On Logout we immediately transition back to the login screen, so no concern about the error message appearing when logging out.

bmingles: @dsmmcken any suggestions on error message here?

dsmmcken: Is this two seperate error states?

1. "Console option is disabled."
2. "Unable to load console, no available session."

bmingles: Yes, although I'm not sure the user will ever actually see #2

bmingles: Is there a standard icon for this?

mofojed: Use a `LoadingOverlay` with `isLoading={null}` and the error message set, e.g.
```
      <LoadingOverlay
        isLoading={false}
        errorMessage="Console option is disabled."
      />
```

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -123,7 +123,7 @@ interface AppMainContainerProps {
   match: {
     params: { notebookPath: string };
   };
-  connection: IdeConnection;
+  connection?: IdeConnection;
```
mofojed, line 126: Is this true? I can see `session` being `undefined`, but in what case is `connection` ever `undefined`? This shouldn't be a case we have to handle.

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -368,6 +376,16 @@ export class ConsolePanel extends PureComponent<
       timeZone,
       unzip,
     } = this.props;
+
+    if (sessionWrapper == null) {
+      return (
+        <LoadingOverlay
+          isLoading={false}
+          errorMessage="Console option is disabled."
```
mofojed, line 384:
```suggestion
          errorMessage="Console is disabled."
```

mofojed (changes requested):
`getDashboardSessionWrapper` currently only returns `SessionWrapper` - that needs to be updated to allow `undefined` as well.

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
bmingles: fixed

packages/code-studio/src/main/AppMainContainer.tsx
mofojed: Ah true. Should update that type in redux as well then.

bmingles: I think it is already up to date? `dashboardData: Record<string, DashboardData>;` can also be an empty record.

mofojed: Ah whoops I was looking at old code, looks good.

mofojed (approved):

PR #1729, opened 2024-01-16
fix: loading spinner finishes before all series load

### LINKED ISSUE #1654: Chart plotBy does not wait for all series to load before displaying
**Description**

When using a plotBy, the chart will remove the loading spinner and fire its load finished event as soon as the 1st series data arrives.

This is because for a regular series/chart, the figure object from the server includes the series to plot so we can populate `seriesToProcess`.

For a multi-series (plotBy), the figure from the server does not include any of the actual series. So the `seriesToProcess` set is always empty if we have a `plotBy`. 

I think instead of populating `seriesToProcess` in `initAllSeries`, it should be part of `addSeries`. Then also remove from that set if the series is inactive in `initAllSeries`. A `plotBy` sends series added events which trigger `addSeries`. Then we would need to not fire updated events until the load is finished. Right now we fire updated events regardless of if we're waiting on other series.

**Steps to reproduce**

1. Create a chart with a plotBy in Python. This example takes advantage of downsampling being slow when there are nulls to show the series loading in at different times.
```python

t = empty_table(80000).update(["Timestamp=epochMillisToInstant(i)","x=(double)i", "y=i%12==11?null: i%5==0 ? i*2 : i", "series= i%5==0 ? `Double` : `Single`"])
f = Figure().plot_xy(series_name="Hello", t=t, x="Timestamp", y="y", by=["series"]).show()
```

**Expected results**

Loading spinner goes away when both series have loaded.

**Actual results**

One series displays, then 1-2 seconds later, the other series displays.

This was discovered because I was using the load finished event to detect when the data has arrived for the docs snapshot tool. This doesn't work with plotBy and recently an unresolved promise started appearing because of this case.

ethanalvizo:
Closes #1654

### DIFF
--- a/packages/chart/src/FigureChartModel.ts
+++ b/packages/chart/src/FigureChartModel.ts
@@ -157,6 +157,9 @@ class FigureChartModel extends ChartModel {
     const { charts } = this.figure;
     const axisTypeMap = ChartUtils.getAxisTypeMap(this.figure);
     const activeSeriesNames: string[] = [];
+
+    this.seriesToProcess = new Set();
+
     for (let i = 0; i < charts.length; i += 1) {
       const chart = charts[i];
 
@@ -216,7 +219,6 @@ class FigureChartModel extends ChartModel {
       const seriesName = inactiveSeriesNames[i];
       this.seriesDataMap.delete(seriesName);
     }
-    this.seriesToProcess = new Set([...this.seriesDataMap.keys()]);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/chart/src/FigureChartModel.ts
```diff
@@ -157,6 +157,9 @@ class FigureChartModel extends ChartModel {
     const { charts } = this.figure;
     const axisTypeMap = ChartUtils.getAxisTypeMap(this.figure);
     const activeSeriesNames: string[] = [];
```
mofojed, line 159: Side note - this could be optimized by using a `Set<string>` instead of a `string[]`.

mofojed (approved):

PR #1731, opened 2024-01-16
fix: re-colorize command codeblocks when theme changes

dsmmcken:
Fixes an issues where the command history doesn't get re-colorized when the theme changes.

Steps to reproduce:
Run a command, then change theme (depends on which theme you started on, but history will look wrong):
image

### DIFF
--- a/packages/console/src/common/Code.tsx
+++ b/packages/console/src/common/Code.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, ReactNode, useCallback } from 'react';
+import React, { useEffect, useState, ReactNode, ReactElement } from 'react';
 import * as monaco from 'monaco-editor';
 import { useTheme } from '@deephaven/components';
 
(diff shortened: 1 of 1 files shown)


packages/console/src/common/Code.tsx
```diff
@@ -1,51 +1,39 @@
...
-          {children}
-        </div>
-      </div>
-    );
-  }
+  }, [children, language]);
+
+  useEffect(() => {
+    if (activeThemes != null) colorize();
+  }, [activeThemes, colorize]);
```
mofojed, line 27: There's technically a race condition here - if you colorize a snippet, but before that returns the component changes and you colorize again, but the second colorize completes before the first one, it'll get set to the wrong value. Or it'll try to set it after an unmount. Should instead set an `isCanceled` flag and a `useEffect` straight up, e.g.

```typescript
  useEffect(() => {
    let isCanceled = false;
    async function colorize() {
      if (containerRef.current == null || children == null) {
        return;
      }
      const result = await monaco.editor.colorize(
        children.toString(),
        language,
        {}
      );
      if (isCanceled) {
        return;
      }
      containerRef.current.innerHTML = result;
    }
    colorize();
    return () => {
      isCanceled = true;
    };
  }, [activeThemes, children, language]);
```

Also - we should just use the `dangerouslySetInnerHTML` prop instead of using a container ref, e.g.
```

function Code({ children, language }: CodeProps): JSX.Element {
  const [colorizedHtml, setColorizedHtml] = useState<string | null>(null);
  const { activeThemes } = useTheme();

  useEffect(() => {
    let isCanceled = false;
    async function colorize() {
      if (children != null) {
        const result = await monaco.editor.colorize(
          children.toString(),
          language,
          {}
        );
        if (!isCanceled) {
          setColorizedHtml(result);
        }
      }
    }
    colorize();
    return () => {
      isCanceled = true;
    };
  }, [activeThemes, children, language]);

  return (
    <div
      // Add pointerEvents: 'none' has huge benefits on performance with Hit Test testing on large colorized elements.
      // You can still select the text event with this set
      style={{ pointerEvents: 'none' }}
      // eslint-disable-next-line react/no-danger
      dangerouslySetInnerHTML={
        colorizedHtml != null ? { __html: colorizedHtml } : undefined
      }
    />
  );
}
```

mofojed (changes requested):

packages/console/src/common/Code.tsx
dsmmcken: totally forgot `dangerouslySetInnerHTML` exists.

mofojed (approved):

PR #1733, opened 2024-01-17
fix: hcm caret shouldn't be allowed to shrink

dsmmcken:
Caret shouldn't shrink if component is truncating. Component only used in enterprise right now, and there's a duplicate change in deephaven-ent/iris#1274

### DIFF
--- a/packages/components/src/HierarchicalCheckboxMenu.scss
+++ b/packages/components/src/HierarchicalCheckboxMenu.scss
@@ -18,6 +18,7 @@
     @include caret-icon;
 
     margin-right: 4px;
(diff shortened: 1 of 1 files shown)


bmingles:
Is there any way to see this in the UI?

bmingles (approved):
I confirmed this works even when removing the hack one here in DHE:
https://github.com/deephaven-ent/iris/pull/1274/commits/95109dc081d2c08ca9688fe2de56e67b1fd49e29#diff-68b9bf7ddb27c0286469a4bb26223b15844511f56dbe2a37f7f5724d02e8505cR25-R27

Issue #1720, opened 2024-01-10
Add remark-math plugin to MarkdownEditor


mofojed:
As a Deephaven user, I want support for math syntax in markdown panels so that I can write math equations.

For example, writing the following in a markdown panel:
```
$$ x = {-b \pm \sqrt{b^2-4ac} \over 2a}$$

$(a+b)^2$

$$\eqalign{
(a+b)^2 &= (a+b)(a+b) \\
&= a^2 + ab + ba + b^2 \\
&= a^2 + 2ab + b^2
}$$
```
Should produce a mathematical output:

$$ x = {-b \pm \sqrt{b^2-4ac} \over 2a}$$

$(a+b)^2$

$$\eqalign{
(a+b)^2 &= (a+b)(a+b) \\
&= a^2 + ab + ba + b^2 \\
&= a^2 + 2ab + b^2
}$$

mofojed:
We need to add the remark-math plugin. Details are at https://github.com/remarkjs/react-markdown#use-remark-and-rehype-plugins-math
Should add this to both MarkdownNotebook and MarkdownEditor (they should be using the same plugins).

mattrunyon:
Katex does not support the `eqalign` environment https://katex.org/docs/support_table.html#e

GitHub uses Mathjax https://github.blog/2022-05-19-math-support-in-markdown/

PR #1734, opened 2024-01-18
feat: add LaTeX to Markdown

wusteven815:
- Adds #1720 
  - Updated `react-markdown` to `8.0.7`
  - Updated `remark-gfm` to `4.0.0`
  - Installed `rehype-katex 6.0.3`
  - Installed `remark-math 5.1.1`
  - Some of these are not the latest versions to match compatibility (with React 17)
- Updated Jest config for pure ESM libraries (thanks to @mattrunyon)


wusteven815:
A few notes:
- `import 'katex/dist/katex.min.css';` is needed, but ESLint is not happy (`'katex' should be listed in the project's dependencies. Run 'npm i -S katex' to add it (import/no-extraneous-dependencies)`)
  - Not sure if there will be any weird behaviours from being in a monorepo
- The following equation (from the issue) does not work
```
$$\eqalign{
(a+b)^2 &= (a+b)(a+b) \\
&= a^2 + ab + ba + b^2 \\
&= a^2 + 2ab + b^2
}$$
```

mofojed:
Check how big the katex.css file is, Don is concerned about loading of the component.

wusteven815:
> Check how big the katex.css file is, Don is concerned about size of loading it up front.

It's 23KB

mofojed:
@dsmmcken at 23kb I wouldn't be concerned about the extra weight

mattrunyon:
For the linting issue, add `katex` to the `package.json` of the package that needs it, then run `npm install` from the root. Might need to restart VSCode or hit `ctrl+shift+p` (cmd on mac) and select `Restart ESLint Server` for eslint to recognize the update locally

You can also do `npm install katex -w @deephaven/dashboard-core-plugins` from the root to install it to the package

packages/dashboard-core-plugins/src/panels/WidgetPanelTooltip.tsx
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import { CopyButton } from '@deephaven/components';
 import { GLPropTypes, LayoutUtils } from '@deephaven/dashboard';
 import './WidgetPanelTooltip.scss';
-import { ReactElement } from 'react-markdown';
 import type { Container } from '@deephaven/golden-layout';
+import { ReactElement } from 'react-markdown/lib/react-markdown';
```
mattrunyon, line 7: I don't know why this was imported from `react-markdown` to begin with, but it should just be imported from react. Or just use `React.ReactElement` in place of this type

In general, importing from paths that aren't the package root should be avoided unless the package says to do it (they can set up import paths for different import specifiers) or there is no other way to do it (like the CSS probably)

When you import from just the package name, the package's `package.json` is consulted for what file the user should be given. When you append paths, `package.json` is ignored and the import is completely dependent on the file structure of the package. These imports are much more brittle to any sort of rename

packages/dashboard-core-plugins/src/panels/MarkdownNotebook.tsx
```diff
@@ -16,8 +18,9 @@ import {
   ReactMarkdownProps,
   TransformImage,
   TransformLink,
-} from 'react-markdown/src/ast-to-react';
+} from 'react-markdown/lib/ast-to-react';
```
mattrunyon, line 21: I know this isn't code you added, but you should be able to remove the `/lib/ast-to-react` imports by adding `import { Components, Options } from 'react-markdown'` and then one of those should have these types available as either `A.TransformImage` or `A['TransformImage']` I believe. Give it a shot and let me know if you run into any issues with it

mattrunyon (changes requested):

mattrunyon:
Replace `jest.config.base.cjs` contents with this to get the tests to work

```js
const path = require('path');

// List of node_modules that need to be transformed from ESM to CJS for jest to work
const nodeModulesToTransform = [
  // monaco
  'monaco-editor',
  // plotly.js dependencies
  'd3-interpolate',
  'd3-color',
  // react-markdown and its dependencies
  'react-markdown',
  'vfile',
  'vfile-message',
  'unist-util.*',
  'unified',
  'bail',
  'is-plain-obj',
  'trough',
  'remark.*',
  'mdast-util.*',
  'micromark.*',
  'decode-named-character-reference',
  'trim-lines',
  'property-information',
  'hast-util.*',
  '.*separated-tokens',
  'ccount',
  'devlop',
  'escape-string-regexp',
  'markdown-table',
  'zwitch',
  'longest-streak',
  'rehype.*',
  'web-namespaces',
  'hastscript',
];

module.exports = {
  transform: {
    '.(ts|tsx|js|jsx)': [
      'babel-jest',
      {
        rootMode: 'upward',
        plugins: ['@deephaven/babel-preset/mockCssImportPlugin'],
      },
    ],
  },
  // Makes Jest transform some node_modules when needed. Usually because they are pure ESM and Jest needs CJS
  // By default, Jest ignores transforming node_modules
  // When switching to transform all of node_modules, it caused a babel error
  transformIgnorePatterns: [
    `node_modules/(?!(${nodeModulesToTransform.join('|')})/)`,
  ],
  moduleNameMapper: {
    '\\.(css|less|scss|sass)$': 'identity-obj-proxy',
    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
      path.join(__dirname, './__mocks__/fileMock.js'),
    '^fira$': 'identity-obj-proxy',
    '^monaco-editor$': path.join(
      __dirname,
      'node_modules',
      'monaco-editor/esm/vs/editor/editor.api.js'
    ),
    // Handle monaco worker files
    '\\.worker.*$': 'identity-obj-proxy',
    // All packages except icons use src code
    '^@deephaven/(?!icons)(.*)$': path.join(__dirname, './packages/$1/src'),
  },
  testEnvironment: 'jsdom',
  setupFilesAfterEnv: [path.join(__dirname, './jest.setup.ts')],
};

```

packages/dashboard-core-plugins/src/panels/MarkdownPanel.tsx
```diff
@@ -17,8 +23,12 @@ import { RootState } from '@deephaven/redux';
 import Panel from './Panel';
 import MarkdownContainer from '../controls/markdown/MarkdownContainer';
 import MarkdownStartPage from '../controls/markdown/MarkdownStartPage';
-import MarkdownEditor from '../controls/markdown/MarkdownEditor';
 import './MarkdownPanel.scss';
+import type MarkdownEditorType from '../controls/markdown/MarkdownEditor';
+
+const MarkdownEditor = lazy(
+  () => import('../controls/markdown/MarkdownEditor')
+);
```
mofojed, line 31: You'll need to wrap the usage of `MarkdownEditor` with a `Suspense` in the `render` method so it has an element to fallback to, e.g.
```
            <Suspense fallback={<LoadingOverlay />}>
              <MarkdownEditor
                ...
              />
            </Suspense>
```
That will show a loading spinner while it's loading, and fix the unit test that is broken. Similar thing will be needed in `NotebookPanel`.

https://react.dev/reference/react/lazy#usage

mofojed (changes requested):

mattrunyon (commented):
Due to our Vite config, the lazy loading doesn't actually offload much (only about 2.5kB). The vendor bundle increases from 792kB gzipped to 1417kB gzipped

I'll see if we can fix this. There was some reason we manually specified chunks and I think it was to split out monaco and plotly. I believe there was also some memory usage issue w/ the source maps and everything in 1 giant chunk

mattrunyon (approved):

mofojed (approved):

PR #1735, opened 2024-01-18
fix: show copy cursor in grid on key down and not just mouse move 

dsmmcken:
- Adds copy cell key handler to set cursor on key down. Grid was only showing copy cursor on mouse move, not just key down. Unfortunately we still require focus to get the key down, and I don't think it's worth hacking around that.
- Moved cursor styles to linker and irisgrid css respectively, so that it is shared to DHE. AppMainContainer.css isn't used in DHE.
- Moved cursors and logos in to asset folder (folders named "assets" are bundled by Vite
- Tweaked copy icon svg (aliasing is improved)
- default cursor to "copy" in iris grid so that it shows in embed widget and DHE

BREAKING CHANGE:  linker and iris grid custom cursor styling and assets are now provided by components directly. DHE css and svg files containing linker cursors should be removed/de-duplicated.


mofojed:
Test on Enterprise with start-community

dsmmcken:
Confirmed it works with start-community

mofojed (approved):

Issue #1233, opened 2023-04-18
Right-click filtering should filter column to all values in selection, not just the row clicked


rbasralian:
When multiple rows are selected, right-clicking and using "filter by value" should filter based on **all** selected values, not only the row that was right-clicked.

This is how it looks now — I can only filter to one row, even though I selected three:

![image](https://user-images.githubusercontent.com/9516585/232843951-37325833-da8a-49a5-a560-01f8393f5bb5.png)

In the Swing UI, the dialog changes to "Filter by Value**s**", and allows you to filter based on multiple rows at once:

![image](https://user-images.githubusercontent.com/9516585/232844371-282c6909-ffd4-44b4-abe5-e114996a6e8c.png)

This is extremely useful when drilling into big tables (e.g. selecting a handful of stock ticker symbols to look at, or selecting a handful of ID numbers from some state log). It also makes it easy to filter to a subset of data so you can look at summary statistics from the column headers.

dsmmcken:
Changes we want to make:

- [ ] Right-clicking on a row that isn’t selected should immediately change the selection to that row and present the menu options for that row
- [ ] Right-clicking inside an existing selection should present the filter by value menu based on the selection size. If there's only one selected row, it should function as it does currently. If there are multiple selected rows, it should present a "Filter by Values" (plural) menu. The preview at the top should show several values together to show that it is filtering by multiple values (although that may not be possible if the values are too long to show more than one). The UI will need to take a snapshot of the selection to get any values part of an offscreen selection. We should limit the snapshot to the first 1000 values or something as to not kill the client if the user has a whole table selected. For input tables where you might have multiple individual cells selected in different columns, use only the selected cells from the column that was clicked.


Display the number of rows selected below the values when there are multiple rows selected as shown:
![image](https://user-images.githubusercontent.com/1576283/233735293-fe597602-ad7d-49c8-b6ce-8425994ed1de.png)

wusteven815:
A few questions regarding the filter's behaviour:
- What should happen when null is part of the selection range? For example, for strings, should it use the null menu or the string menu? If string menu, what would happen to with the "contains", "starts with", and "ends with" actions? (And similarly with numbers and those comparisons)
- What should the logic be with number comparisons? For example, if 1, 2, 3 are selected, what would the "greater than" comparison mean (>1 or >3)?

dsmmcken:
[...]

PR #1736, opened 2024-01-19
feat: multiselect values

wusteven815:
- Feature: #1233 
  - Moved cell filter actions to its own `async` function that gets a snapshot to supports multiple values
  - A maximum of 1000 (`MAX_MULTISELECT_ROWS`) rows will be snapshotted for unique values. The reason why its rows and not unique values is to prevent the case of a very large amount of rows with a very small amount of unique values.
  - Add tests for above by running E2E tests on all the filters
- Changed right-click behaviour such that if right-clicked outside of the current selection, the new selection will be the row the cursor is on
  - Add tests for above, add function to right-click in Grid tests
- Add assert to check if an array/map is non-empty, and if an number is not NaN
  - Add tests for above


dsmmcken:
I'll want to review the right click behaviour changes when available.

dsmmcken (dismissed):
> equals/not equals should be ORs and contains/starts with/ends with should be ANDs
1. I know I said this in the ticket. I tried it and I was clearly wrong. It does not make sense for strings. It should really be just ORs for all _string_ comparisons types.
2. If you right click outside of the selection it should actually be changing the selection to the row you right-clicked on first.
> Right-clicking on a row that isn’t selected should immediately change the selection to that row and present the menu options for that row
like this:
selection

I see what you did with the greater than/less than for numbers. I like that.

It's a little confusing at first if your set contains infinity, but I guess that's reasonable as it can't have greater than or less. Wonder if there's some sort of user feedback we should be giving though. Thoughts?

wusteven815:
> It's a little confusing at first if your set contains infinity, but I guess that's reasonable as it can't have greater than or less. Wonder if there's some sort of user feedback we should be giving though. Thoughts?

How about if positive infinity is present, then > and >= are disabled, and the similarly for positive infinity?

mofojed:
> > It's a little confusing at first if your set contains infinity, but I guess that's reasonable as it can't have greater than or less. Wonder if there's some sort of user feedback we should be giving though. Thoughts?
> 
> How about if positive infinity is present, then > and >= are disabled, and the similarly for positive infinity?

`infinity` is similar to `null`, as the only options that make sense are `is equal to` and `is not equal to`. Doing `<infinity` is the same as saying `!=infinity`, and doing `<=infinity` doesn't make any sense. Converse is true for negative infinity. So if `infinity` and/or `-infinity` are present, only show the equals/not equals options.

dsmmcken:
> So if infinity and/or -infinity are present, only show the equals/not equals options.

Yes, that is what is presently happening. However, I am saying It can be confusing. For example, say you select 500 rows. On screen you see 1, 2, 3, 4 etc. but somewhere in that selection set is -infinity. You don't know it's there you only know that for some reason you can only select equals/not equals and it's unclear as to why.

Maybe we should treat it like null and exclude it from the set if selection an option other than equals not equals?

dsmmcken:
Tooltips in a context menu seems like something one shouldn't do.

I am gonna recommend dropping it from the filter for cases other than equals/not equals, as that would probably behave inline with a users intent of selecting less than/greater than.

packages/grid/src/Grid.tsx
```diff
@@ -355,6 +355,7 @@ class Grid extends PureComponent<GridProps, GridState> {
     this.handleMouseUp = this.handleMouseUp.bind(this);
     this.handleResize = this.handleResize.bind(this);
     this.handleWheel = this.handleWheel.bind(this);
+    this.getSelectedRanges = this.getSelectedRanges.bind(this);
```
mofojed, line 358: Binding this here is not strictly necessary, as `getSelectedRanges` isn't passed in anywhere. Should just be able to call `grid.getSelectedRanges` directly.

packages/grid/src/mouse-handlers/GridSelectionMouseHandler.ts
```diff
@@ -218,6 +218,41 @@ class GridSelectionMouseHandler extends GridMouseHandler {
...
+      ) {
+        isInRange = true;
+        break;
+      }
+    }
+
+    // only change the selected range if the selected cell is not in the selected range
+    if (!isInRange && gridPoint.row !== null) {
+      this.startPoint = undefined;
+      this.dragBounds = undefined;
```
mofojed, line 246: It looks like `this.dragBounds` is unused, you can just remove that private property entirely.

packages/grid/src/mouse-handlers/GridSelectionMouseHandler.ts
```diff
@@ -218,6 +218,41 @@ class GridSelectionMouseHandler extends GridMouseHandler {
...
+      this.stopTimer();
+      grid.clearSelectedRanges();
+      grid.moveSelection(gridPoint.column, gridPoint.row);
+      grid.commitSelection();
```
mofojed, line 250: I think you can just call `moveCursorToPosition` instead:
```suggestion
      grid.moveCursorToPosition(gridPoint.column, gridPoint.row);
```

mofojed (changes requested):

packages/utils/src/TextUtils.ts
```diff
@@ -58,6 +58,32 @@ class TextUtils {
...
+   *
+   * Examples:
+   *  - A || B || C
+   *  - !=A && !=B && !=C
+   * @param items List of terms (e.g. `['A', 'B', 'C']`)
+   * @param binaryOp Operator to combine terms
+   * @param toStr Gunction to turn list of items to string (also for prefixes and suffixes)
+   * @returns
+   */
+  static makeLogicalNormalForm<T>(
```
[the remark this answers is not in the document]
wusteven815: Changed to use map + joins

packages/grid/src/Grid.tsx
wusteven815: I'm getting `TypeError: Cannot read properties of undefined (reading 'state')` when I'm not binding it. I believe it's because the function is reading `selectedRanges` from `this.state`.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -531,70 +531,65 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+    //   - NOT first MAX_MULTISELECT_ROWS unique values (prevent case where there are a small
+    //     amount of values, but a large amount of rows with those values)
+    if (GridRange.containsCell(selectedRanges, sourceColumn, sourceRow)) {
+      let rowCount = GridRange.rowCount(selectedRanges);
+      while (rowCount > MAX_MULTISELECT_ROWS) {
+        const lastRow = selectedRanges.pop();
+        // should never occur, sanity check
+        if (lastRow === undefined) {
+          throw Error('Selected ranges should not be empty');
         }
```
mofojed, line 555: You can write this using an `assertNotNull` instead to make it a one-liner (need to add the import as well).
```suggestion
        assertNotNull(lastRow, 'Selected ranges should not be empty');
```

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -531,70 +531,65 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
-        // gone past source cell and not in previous selected ranges
-        if (sourceRow < selectedRange.startRow && !sourceInSelected) break;
-        // have to truncate the range
-        if (selectedRange.endRow - selectedRange.startRow > valuesLeft) {
-          slicedSelectedRanges.push(
+        const lastRowSize = GridRange.rowCount([lastRow]);
+        // should never occur, sanity check
+        if (Number.isNaN(lastRowSize)) {
+          throw Error('Selected ranges should not be unbounded');
+        }
```
mofojed, line 560: Optimization - we could create an `assertNotNaN` function in `Asserts.ts` to make this a one-liner.

mofojed (dismissed):
Looks good. A couple suggestions for optimizations/refactors.

mofojed (approved):
Excellent work!

PR #1737, opened 2024-01-19
feat: Add pluginDataMap to redux, add useDashboardPluginData hook

vbabich:
Add `useDashboardPluginData` hook for accessing plugin data in redux

### DIFF
--- a/packages/dashboard/src/redux/hooks.ts
+++ b/packages/dashboard/src/redux/hooks.ts
@@ -1,0 +1,19 @@
+import { useDispatch, useSelector } from 'react-redux';
+import { PluginData, RootState } from '@deephaven/redux';
+import { getPluginDataForDashboard } from './selectors';
+import { setDashboardPluginData } from './actions';
+
+export function useDashboardPluginData(
+  dashboardId: string,
+  pluginId: string
+): [PluginData, (data: PluginData) => void] {
+  const dispatch = useDispatch();
+  return [
+    useSelector((store: RootState) =>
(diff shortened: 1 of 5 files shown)


packages/dashboard/src/redux/hooks.ts
```diff
@@ -0,0 +1,19 @@
...
+  dashboardId: string,
+  pluginId: string
+): [PluginData, (data: PluginData) => void] {
+  const dispatch = useDispatch();
+  return [
+    useSelector((store: RootState) =>
+      getPluginDataForDashboard(store, dashboardId, pluginId)
+    ),
+    data => dispatch(setDashboardPluginData(dashboardId, pluginId, data)),
+  ];
```
mofojed, line 16: I typically would explicitly put the `useSelector` at the top level. Also we should use `useCallback` for the data callback, and we need JS docs on the parameters passed in and what the return values are. e.g.
```
const dispatch = useDispatch();
const data = useSelector( ... );
const setData = useCallback(newData => dispatch(...), [dashboardId, pluginId, dispatch]);
return [ data, setData ];
```

mofojed (changes requested):
Need to clean up the hook used in a bit.

Side note, ideally we'd have the plugins wrapped in a context such that they don't need to self-report their `pluginId` - we could have a situation here where a plugin is reading/writing another plugins data, and/or doesn't use it's _actual_ `pluginId`. But that seems more like a convenience thing (security-wise we're already storing the data in redux which is accessible by all).

mofojed (approved):

PR #1739, opened 2024-01-19
feat: added shortcut for copying version info and added browser/os to info

dsmmcken:
- Adds a shortcut for copying version info to clipboard to make it easier when making tickets.
- Adds a new package for parsing UA strings
- Adds browser/os to version info for copying

### DIFF
--- a/packages/code-studio/src/settings/SettingsUtils.tsx
+++ b/packages/code-studio/src/settings/SettingsUtils.tsx
@@ -8,6 +8,8 @@ import {
 } from '@deephaven/jsapi-utils';
 import type { dh as DhType } from '@deephaven/jsapi-types';
 import Log from '@deephaven/log';
+import type { ServerConfigValues } from '@deephaven/redux';
+import Bowser from 'bowser';
 
 const log = Log.module('SettingsUtils');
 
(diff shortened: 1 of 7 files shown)


mattrunyon (changes requested):

packages/code-studio/src/settings/SettingsUtils.tsx
```diff
@@ -43,18 +43,18 @@ export function getFormattedVersionInfo(
   serverConfigValues: ServerConfigValues
 ): Record<string, string> {
   const ua = Bowser.parse(window.navigator.userAgent);
-  const browser = `${(ua.browser.name ?? '') || 'Unknown'} ${
-    // use only the major version and minor version, rest usually empty 120.1.0.0 -> 120.1
+  const browser = `${ua.browser.name ?? ''} ${
+    // use only the major version and minor version, rest is usually empty 120.1.0.0 -> 120.1
     Number(parseFloat(ua.browser.version ?? '')) || ''
```
mattrunyon, line 48:
```suggestion
    Number(parseFloat(ua.browser.version ?? ''))
```

mattrunyon (changes requested):

packages/code-studio/src/settings/SettingsUtils.tsx
dsmmcken: No, the outer Number() can return NaN if the inner parseFloat returns NaN when version is undefined. And the inner ?? is needed to keep typescript happy as parseFloat is supposed to take strings, not undefined.

mattrunyon (approved):

packages/code-studio/src/settings/SettingsUtils.tsx
mattrunyon: Ahh yes my favorite number NaN

PR #1743, opened 2024-01-23
fix: keep manually entered value in GoToRow when changing to same column type

### LINKED ISSUE #1562: Keep manually entered value in the GoToRow search when changing columns if the column type is the same
Don't re-populate the GoToRow search when changing columns if it contains a manually entered value and the column type doesn't change.

wusteven815:
- Fixes #1562 
  - Tracks if the value was manually entered and changes appropriately
  - Adds test with a wait to account for the debounce

**Testing Methodology**
- Generated a relevant table with 2 integer and 1 string columns
- GoToRow has a column selector and a field to input the value, the possible configurations are
  - If the column selector has changed types
  - If the field was changed
- GoToRow should leave the field unmodified if the column selector has changed types and the field was changed, otherwise keep the original behaviour
  - One special case where this is the case AND GoToRow has been toggled, going back to the original behaviour
- Helper function `setColumnAndExpectInputValue` to make the changes and do the test


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -444,6 +444,7 @@ export interface IrisGridState {
 
   gotoValueSelectedColumnName: ColumnName;
   gotoValueSelectedFilter: FilterTypeValue;
+  goToValueManuallyChanged: boolean;
```
mofojed, line 447: Name should be `gotoValueManuallyChanged` to be consistent (`goto` vs. `goTo`)

mofojed (changes requested):
Include details in the PR description about how you tested this (e.g. what snippet of code did you use, what scenarios did you test).
Also we should probably reset the `gotoValueManuallyChanged` flag when `isGotoShown` is changed (e.g. when the user hides/shows the goto row).

mofojed (approved):
Great work!

PR #1744, opened 2024-01-23
fix: allow logout on open file

### LINKED ISSUE #1685: Open file blocks logout

wusteven815:
- Fixes #1685 
  - Moved `storeCookieToken` from `AuthPluginPsk.tsx` to `AuthPlugin.ts`, so it can be exported
  - Logout button now uses that to remove the cookies

### DIFF
--- a/packages/app-utils/src/components/AppBootstrap.tsx
+++ b/packages/app-utils/src/components/AppBootstrap.tsx
@@ -1,8 +1,10 @@
 import React, { useCallback, useMemo, useState } from 'react';
 import '@deephaven/components/scss/BaseStyleSheet.scss';
+import { storeCookieToken } from '@deephaven/auth-plugins';
 import { ClientBootstrap } from '@deephaven/jsapi-bootstrap';
 import {
   RefreshTokenBootstrap,
+  storeRefreshToken,
   useBroadcastLoginListener,
 } from '@deephaven/jsapi-components';
 import { type Plugin } from '@deephaven/plugin';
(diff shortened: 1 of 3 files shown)


mattrunyon (changes requested):
I think this only needs to be 1 line in `AuthPluginPsk` in `onLogout` add `storeRefreshToken(null)`.

Since we use plugins for auth, any clearing of stored data on logout should be handled by the plugin. We emit an event on a BroadcastChannel on logout so that any 3rd party plugin can respond to a logout request

With any of the plugin architecture, the mindset we use typically is "could I achieve this if I were a 3rd party writing this plugin". What that means is basically no app specific code (anything in app-utils or code-studio) to handle these things as a 3rd party writing a plugin cannot modify any code except their own plugin code. If it can't be achieved, then that's a shortcoming of the APIs we have for plugins

As an example, here's a plugin for KeyCloak which handles many different types of authentication https://github.com/deephaven/deephaven-plugins/blob/main/plugins/auth-keycloak/src/js/src/AuthPluginKeycloak.tsx#L67

wusteven815:
> I think this only needs to be 1 line in `AuthPluginPsk` in `onLogout` add `storeRefreshToken(null)`.

When I was debugging this, I noticed that of the 3 listeners (`AppBootstrap`, `AuthPluginPsk`, `RefreshTokenBootstrap`), only `AppBootstrap`'s `onLogout` was being run, and that was why I added those lines to it. Is the real issue that the other listeners are not running, instead of `AppBootstrap` not running the logout functions?

mattrunyon:
If they aren't being run, then that's the real issue. I added a line to the refresh token `onLogout` and it seemed to run when I logged out for me

PR #1745, opened 2024-01-23
feat: adjust display of  theme palette in styleguide

dsmmcken:
Note: contains no color or app changes.

- display styleguide as theme palette in a grid (shows in e2e)
- re-order blue to be after cyan in code
- group content-subdued-bg with content-bg (triggers a e2e change)


bmingles (approved):
LGTM. I looked at styleguide and new color layout looks good. Looked at DH very high level. Looks as before as expected.

PR #1749, opened 2024-01-25
fix: sorting frozen columns

### LINKED ISSUE #1645: Freezing a column then sorting columns in the organize column results in wrong column frozen
**Description**

Freezing a column then sorting columns in the organize column results in wrong column frozen. After the user is also unable to unfreeze the wrongly frozen column.

**Steps to reproduce**

```

        "MyString=new String(`a`+i)",
        "MyInt=new Integer(i)",
        "MyLong=new Long(i)",
        "MyFloat=new Float(i+i/10)",
```

1. Right click the "MyFloat" column header, and select "Freeze column" from the context menu.
2. Open the table overflow menu, and select organize columns, then press the sort ascending button
3. (optional): attempt to unfreeze the frozen column (now MyBoolean)

**Expected results**
MyFloat remains frozen, and the other columns are sorted.

**Actual results**

MyBoolean becomes the frozen column, and no columns are shown as frozen as the organize columns list. Right clicking the "MyBoolean" column doesn't give an option to unfreeze it, only freeze which then freezes a completely different column.

freeze_column

**Versions**

Engine Version: 0.31.0
Web UI Version: 0.54.0
Java Version: 21.0.1
Barrage Version: 0.6.0

wusteven815:
- Fixes #1645 
  - Moves are now generated for frozen columns (like front/end for initial moved columns) and only used when sorting
  - The range being sorted is truncated to remove frozen/front/end columns
- Added tests to make sure the sort is correct

### DIFF
--- a/packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
+++ b/packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
@@ -583,7 +583,8 @@ class VisibilityOrderingBuilder extends PureComponent<
   getSortMoves(
     itemsParam: readonly IrisGridTreeItem[],
     option: keyof typeof VisibilityOrderingBuilder.SORTING_OPTIONS,
-    movedColumns: readonly MoveOperation[]
+    movedColumns: readonly MoveOperation[],
+    frozenColumns: readonly string[]
   ): MoveOperation[] {
     const items = [...itemsParam];
     // Sort all the movable columns
@@ -592,6 +593,11 @@ class VisibilityOrderingBuilder extends PureComponent<
     items.sort((a, b) => {
       const aName = a.id.toUpperCase();
       const bName = b.id.toUpperCase();
+      if (frozenColumns.includes(a.id) && frozenColumns.includes(b.id)) {
+        return TextUtils.sort(aName, bName, isAscending);
+      }
+      if (frozenColumns.includes(a.id)) return -1;
+      if (frozenColumns.includes(b.id)) return 1;
       return TextUtils.sort(aName, bName, isAscending);
     });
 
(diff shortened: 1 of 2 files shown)


mofojed (commented):
I just want to confirm with @dsmmcken - do we want the frozen columns to be sorted as well when you click sort? My instinct is no, we want the frozen columns to stay in the same order they were frozen in.

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -635,11 +635,47 @@ class VisibilityOrderingBuilder extends PureComponent<
...
+    for (let i = 0; i < model.frozenColumns.length; i += 1) {
+      const frozenColumn = model.frozenColumns[i];
+      // get index of frozenColumn
+      let frozenIndex = 0;
+      for (let j = 0; j < model.columns.length; j += 1) {
+        if (model.columns[j].name === frozenColumn) {
+          frozenIndex = j;
+          break;
+        }
+      }
```
mofojed, line 654: Can use `getColumnIndexByName`:
```
const frozenIndex = model.getColumnIndexByName(frozenColumn.name);
```

packages/iris-grid/src/sidebar/visibility-ordering-builder/VisibilityOrderingBuilder.tsx
```diff
@@ -635,11 +635,47 @@ class VisibilityOrderingBuilder extends PureComponent<
...
+        }
+        // placed in front, increase index
+        if (to < frozenIndex) frozenIndex += 1;
+      }
+      if (frozenIndex !== i) {
+        initialAndFrozenMovedColumns.push({
+          from: frozenIndex,
+          to: i,
+        });
+      }
```
mofojed, line 672: You can use `GridUtils.getVisibleIndex(frozenIndex, initialAndFrozenMovedColumns)` instead of re-creating the algorithm.

mofojed (changes requested):

mofojed (approved):

PR #1750, opened 2024-01-26
feat: DH-16336: usePickerWithSelectedValues - boolean flags should be calculated based on trimmed search text

### LINKED ISSUE #1747: feat: DH-16336: usePickerWithSelectedValues - boolean flags should be calculated based on trimmed search text
In support of DH-16336, the `usePickerWithSelectedValues` hook needs to be able to handle non-trimmed search text but to base its return boolean flags on the trimmed value (`hasSearchTextWithZeroResults`, `searchTextExists`, `searchTextIsInSelectedValues`). It might be best to configure this with an optional input arg or something so that the current behavior can also be supported.

bmingles:
Supports DH-16336:
- Added an `isDebouncing` prop to `useDebouncedValue`
- Added support for search text trimming to `usePickerWithSelectedValues`
- Introduced a validation delay to account for viewport filtering to improve UX and avoid flicker of validation message

resolves #1747

Note that there's not any place in DHC that uses the `usePickerWithSelectedValues` hook yet. I've been testing locally as part of DH-16336.

BREAKING CHANGE: `usePickerWithSelectedValues` now takes an object as an argument instead of positional args

### DIFF
--- a/packages/react-hooks/src/useDebouncedValue.ts
+++ b/packages/react-hooks/src/useDebouncedValue.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 /**
  * Debounces a value.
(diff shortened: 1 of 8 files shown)


packages/react-hooks/src/useDebouncedValue.ts
```diff
@@ -1,25 +1,44 @@
...
+  value: T,
+  debounceMs: number
+): { isDebouncing: boolean; value: T } {
+  const [isDebouncing, setIsDebouncing] = useState(true);
   const [debouncedValue, setDebouncedValue] = useState<T>(value);
+
+  // Set isDebouncing to true immediately whenever the value changes. Using
+  // `useMemo` instead of `useEffect` so that state is never out of sync whenever
+  // value and / or debounceMs have changed.
+  useMemo(() => {
```
bmingles, line 21: An alternative to this would be 

```
  const [previousValue, setPreviousValue] = useState(value);
  const [previousDebounceMs, setPreviousDebounceMs] = useState(debounceMs);

  if (value !== previousValue || debounceMs !== previousDebounceMs) {
    setIsDebouncing(true);
    setPreviousValue(value);
    setPreviousDebounceMs(debounceMs);
  }
  ```
  
  This actually aligns with an example in the React docs whereas not sure if setting state inside of `useMemo` is recommended.

bmingles: Yeah, I didn't know this pattern until today.

mofojed (approved):

PR #1756, opened 2024-01-30
feat: allow themes to use any srgb color for definitions

dsmmcken:
Themes were previously limited to HSL raw value colors such that transparency could be created using HSLA. This made theme creation somewhat annoying and limiting. This change adopts color-mix (requires chrome 111, firefox 113 -- both older than 6 months) to create any necessary transparency.

- replaces hsl raw values with hex (allows skipping color normalization)
- replace hsla usage with color-mix
- fix miss-named legend-color
- created utils that use color-mix for easier specification
- move utils.scss to be auto-imported with custom.scss
- re-wrote color normilzation and resolving to resolve color-mix based colors
- changed iris grid theme context to not accept a paritial

Style changes:
- style console input to match other inputs, add hover state to border
- fix input border hover color to use correct variable
- darken water color in dark theme maps]
- adjust disabled button color in light theme]
- fix selection background and selection hover color to be more subtle. (No idea why I made it so opaque, it used to be 0.09 pre-spectrum)

BREAKING CHANGE:
- IrisGridThemeContext no longer accepts a paritial theme. By  guaranteeing the provider is a full theme we can resolve the CSS variables and normailze the colors only once per theme load globally, rather than having to do it once per grid.
- Themes must be defined using valid srgb CSS colors, and not hsl raw component values


packages/components/src/theme/ThemeUtils.ts
```diff
@@ -294,45 +294,48 @@ export function replaceSVGFillColor(
...
   const tmpPropEl = document.createElement('div');
-  targetElement.appendChild(tmpPropEl);
-
-  const varExpressions = [...extractDistinctCssVariableExpressions(record)];
-
-  // Set temporary css variables for resolving var expressions
-  varExpressions.forEach((varExpression, i) => {
-    const tmpPropKey = `--${TMP_CSS_PROP_PREFIX}-${i}`;
-    tmpPropEl.style.setProperty(tmpPropKey, varExpression);
+  tmpPropEl.style.display = 'none';
```
bmingles, line 303: What does this do?

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -294,45 +294,48 @@ export function replaceSVGFillColor(
...
-    const tmpPropKey = `--${TMP_CSS_PROP_PREFIX}-${varExpressions.indexOf(
-      varExpression
-    )}`;
-
-    const resolved = computedStyle.getPropertyValue(tmpPropKey);
-
-    return ColorUtils.normalizeCssColor(resolved, isAlphaOptional);
-  };
+  // append only once to avoid multiple re-layouts
+  targetElement.appendChild(tmpPropEl);
```
bmingles, line 316: Does this have to be added to the DOM?

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -294,45 +294,48 @@ export function replaceSVGFillColor(
...
-  isAlphaOptional = false
+  isAlphaOptional = true
 ): T {
   const perfStart = performance.now();
 
   // Add a temporary div to attach temp css variables to
   const tmpPropEl = document.createElement('div');
-  targetElement.appendChild(tmpPropEl);
-
-  const varExpressions = [...extractDistinctCssVariableExpressions(record)];
```
bmingles, line 305: Is this still able to resolve strings containing multiple expressions?
`var(--bbb1) var(--bbb2)`

dsmmcken: Yes.

```js
const el = document.createElement('div');
el.style.setProperty("--a", "#aaa");
el.style.setProperty("--b", "#bbb");
el.style.setProperty("--c", "var(--a) var(--b)");
document.body.appendChild(el);
window.getComputedStyle(el).getPropertyValue("--c");

// '#aaa #bbb'
```

packages/components/src/theme/ThemeUtils.ts
```diff
@@ -300,39 +300,48 @@ export function resolveCssVariablesInRecord<T extends Record<string, string>>(
...
+    if (!value.includes(CSS_VAR_EXPRESSION_PREFIX)) {
+      (result as Record<string, string>)[key] = value;
+      return;
+    }
+    let resolved = tempPropElComputedStyle.getPropertyValue(
+      `--${TMP_CSS_PROP_PREFIX}-${i}`
+    );
+    if (
+      !/^#[0-9A-F]{6}[0-9a-f]{0,2}$/i.test(resolved) &&
+      // skip if resolved is already hex
```
bmingles, line 331: Shouldn't these 2 comments go above the lines they annotate instead of below?

packages/utils/src/ColorUtils.ts
```diff
@@ -89,30 +102,42 @@ class ColorUtils {
...
       return null;
     }
 
-    const [r, g, b, a = 1] = tokens.map(Number);
+    if (rgbOrRgbaString.startsWith('color(srgb')) {
+      return {
+        r: Math.round(Number(r) * 255),
+        g: Math.round(Number(g) * 255),
+        b: Math.round(Number(b) * 255),
+        a: Number(a ?? 1),
```
bmingles, line 130: Does this need to handle the > 1 scenerio like the one below?

tests/styleguide.spec.ts-snapshots/inputs-focus-col3-row6-text-firefox-linux.png
bmingles, line 1: I wonder why only firefox changed on this one

packages/utils/src/ColorUtils.ts
dsmmcken: no, all values passed into the color(srgb) format are 0-1. Technically neither do, as we've tested css.supports elsewhere, but since this is a util I figured more safety around it wouldn't be a bad thing.

tests/styleguide.spec.ts-snapshots/inputs-focus-col3-row6-text-firefox-linux.png
dsmmcken: I have no idea, I thought that was weird too. The others actually already matched but this one didn't. Thresholds I guess?

packages/utils/src/ColorUtils.ts
dsmmcken: I mean if we really wanted to be particular, we would clamp everything to 0-255 only as well. I think you had a test with a non-alpha value, so I did this to fix it.

bmingles (approved):
Changes look good. Verified that local build + preview is working as well

PR #1758, opened 2024-01-30
refactor: Remove JSAPI databar code

mattrunyon:
Removing databar specific JSAPI code since this API will be moved to dh.ui at some point

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModelTemplate.ts
+++ b/packages/iris-grid/src/IrisGridTableModelTemplate.ts
@@ -3,7 +3,6 @@
 import memoize from 'memoize-one';
 import throttle from 'lodash.throttle';
 import {
-  CellRenderType,
   EditOperation,
   GridRange,
   GridUtils,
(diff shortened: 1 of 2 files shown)


dsmmcken:
> Removing databar specific code since this feature will be moved to dh.ui at some point

We've discussed the pyhon api being on ul.table... the JS would always be needed regardless of where the api originates from.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
mofojed, line 1: Only changes we need is for `dataBarOptionsForCell` to throw, and remove `formatDataBar` from the `Format` type in dh.types.ts.

mofojed (changes requested):
Just rip the API part out of databars, not the whole UI part (though it will be unused until we do wire it up to API).

Building into deephaven.ui we'll need to extend the model with our data bar options (so where `UITable.tsx` calls `IrisGridModelFactory.makeModel`, we'll need some way to add options in there or extend the model returned or have our own factory function or something).

Side note, in the future having plugins for cell renderers might be neat (customer I met with today was asking about them) but not our top priority.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
mattrunyon: `dataBarOptionsForCell` throws from `IrisGridModel`, so just removing it from the table model template. `renderTypeForCell` returns `text` from `GridModel`, so removing that from the table model template as well

mofojed (approved):

Issue #1728, opened 2024-01-16
Theming - Chart component should be responsible for its theme


bmingles:
We currently build a chart model that includes theme colors. We should separate theming from chart model creation and have the Chart apply its theme whenever the ChartThemeContext changes.

PR #1762, opened 2024-01-30
feat: Added dashboard-core-plugins to remote-component list

bmingles:
Including `dashboard-core-plugins` to remote-component list so we can externalize it in plugins repo when developing js plugins.

supports #1728


mattrunyon (approved):
Removed the linked issue since it doesn't actually fix the issue. Don't want it to close incorrectly

Issue #1671, opened 2023-12-06
Optimize e2e tests using application mode


mofojed:
We can use application mode to seed our deephaven-core container when running e2e tests with all the widgets/objects we need, and we don't need to paste code into the console and run it. This should speed up the tests quite a bit, and also allow for parallelization.

mofojed:
Created a draft PR starting off this work where it's launching with application mode: https://github.com/deephaven/web-client-ui/pull/1670
Need to extract the rest of the tests into files that are launched with application mode.

PR #1670, opened 2023-12-06
WIP Add files in app mode for e2e testing

mofojed:
- We should be able to just open panels from the panel menu instead
- Saves all the time typing
- We could run tests in parallel!
- Uses Application Mode: https://deephaven.io/core/docs/reference/app-mode/application-mode-config/

To test, start up deephaven-core with the application.dir setting set, e.g. (need to replace the path): 
```
START_OPTS="-DAuthHandlers=io.deephaven.auth.AnonymousAuthenticationHandler -Ddeephaven.application.dir=/path/to/web-client-ui/tests/docker-scripts/data/app.d" ./gradlew server-jetty-app:run
```
Then run `npm run e2e` to test.

PR #1766, opened 2024-02-01
test: use app mode for E2E tests

wusteven815:
- Implements #1671
- Continuation of #1670
- Point scripts to use the application mode to preload scripts
- Change most scripts to run in parallel mode (exceptions are those designed to run in serial and/or contain console commands)
- Add test utility to open table/plot and changed tests to use this
- Update some snapshots since tables are not longer opened by a command
- Add a util to wait for context menus, as multiselect filtering is async and might not load the Filter by Value button in time

### A few notes regarding parallelism:
- Core/web becomes the bottleneck as some tests may fail because it takes too long for the table/command/page to load
- As such, flakiness can arise from timeouts, whether that be from waiting for the page to load or a command to finish executing


playwright.config.ts
```diff
@@ -21,16 +21,16 @@ const config: PlaywrightTestConfig = {
...
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Global setup file for initializing before all other tests */
   globalSetup: require.resolve('./tests/globalSetup.ts'),
   /* Retry on CI only */
   retries: process.env.CI ? 1 : 0,
-  /* Only have one worker since we don't want tests running in parallel, trampling over each other on the backend */
-  workers: 1,
+  /* Default to 50% of cores, don't want too many as core or web will become bottleneck */
+  workers: undefined,
```
mofojed, line 33: GitHub actions by default has 4 cores: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
Did you try defaulting to 100% in CI to see how it does?

packages/console/src/ConsoleMenu.tsx
```diff
@@ -188,6 +188,7 @@ class ConsoleMenu extends PureComponent<ConsoleMenuProps, ConsoleMenuState> {
     return (
       <>
         <Button
+          id="table-actions"
```
mofojed, line 191: We should not add an `id` to these elements, especially for tests. IDs are bad because they must be unique to that elements subtree and also do not provide any accessibility. It would be better to use a `data-testid` instead for it to be accessible with tests, _but_ we can actually do better. Take a look at the Locators page: https://playwright.dev/docs/locators
First preference is to `getByRole`, then `getByTest`, and `getByLabel`. We should be able to use `getByRole` here because we are already providing a tooltip on the button. When you provide a tooltip on an `iconOnly` button, we automatically add it as the `aria-label`: https://github.com/deephaven/web-client-ui/blob/e90b627fed2c76b81ee96006abef14942388f3cc/packages/components/src/Button.tsx#L137

So in this case, we shouldn't need to add an ID. You should be able to just do `getByRole('button', { name: 'Tables' })` and this should work.
Ditto with the Widgets button below.

tests/console.spec.ts
```diff
@@ -30,8 +33,8 @@ test.describe('console input tests', () => {
     ).toHaveCount(1);
   });
 
-  test('object button is created when creating a table', async () => {
-    const tableName = generateVarName('t');
+  test('object button is created when creating a table', async ({ browserName }) => {
+    const tableName = `browserName${generateVarName('t')}`;
```
mofojed, line 37: You're not actually using `browserName` here, you're using the string literal "browserName" and the variable `browserName` is unused. Instead pass the `browserName` into the prefix to `generateVarName`:
```suggestion
    const tableName = generateVarName(`${browserName}_t`);
```

tests/console.spec.ts
```diff
@@ -17,8 +20,8 @@ test.afterAll(async () => {
 });
 
 test.describe('console input tests', () => {
-  test('print commands get logged', async () => {
-    const message = `Hello ${shortid()}!`;
+  test('print commands get logged', async ({ browserName }) => {
+    const message = `Hello ${browserName} ${shortid()}!`;
```
mofojed, line 24: Not strictly necessary, but good idea for debugging purposes.

tests/utils.ts
```diff
@@ -17,6 +17,69 @@ export enum TableTypes {
...
+  | 'multiselect_number'
+  | 'multiselect_string';
+
+/**
+ * Opens a table loaded from application mode and returns the grid location
+ * @param page
+ * @param type Either 'table' or 'plot'
+ * @param name Name of the table or plot
+ */
+export async function openTableOrPlot(
```
mofojed, line 44: I'd rename this function to just `openObject`, and that takes the type option (`table` or `plot`).

tests/utils.ts
```diff
@@ -17,6 +17,69 @@ export enum TableTypes {
...
+
+  await expect(
+    page.locator('.iris-grid .iris-grid-loading-status')
+  ).toHaveCount(0);
+
+  // get grid and return that
+  // const grid = await page.locator('.iris-grid-panel .iris-grid');
+  // const gridLocation = await grid.boundingBox();
+  // expect(gridLocation).not.toBeNull();
+  // return gridLocation;
```
mofojed, line 80: Remove unused code, and also this part about checking for `iris-grid` should really only be checked for tables loaded. I'd have a separate method `openTable` that calls this `openObject` method, e.g.

```
export async function openTable(page: Page, name: TableNames, waitForLoadFinished = true): Promise<void> {
  await openObject(page, 'table', name);

  if (waitForLoadFinished) {
    await expect(
    ).toHaveCount(0);
  }
}
```

And a similar `openPlot` function as well.

tests/utils.ts
```diff
@@ -17,6 +17,69 @@ export enum TableTypes {
   AllTypes,
 }
 
+type TablePlotNames =
```
mofojed, line 20: Split this into `TableNames`, `PlotNames`, and then `type ObjectNames = TableNames | PlotNames`

tests/docker-scripts/data/app.d/common_figures.py
mofojed, line 1: We should run these Python files through Python Black formatter. You can install the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) or just run them through an [online formatter](https://black.vercel.app/).

tests/styleguide.spec.ts-snapshots/spectrum-buttons-webkit-linux.png
mofojed, line 1: I wonder how we hit this change... it seems to be a font kerning change. Something to keep in mind when looking deeper into the style guide e2e tests.

tests/utils.ts
```diff
@@ -268,6 +330,7 @@ export async function openTableOption(
 
 export default {
   generateVarName,
+  openTable: openTableOrPlot,
```
mofojed, line 333: Export `openTable`, `openPlot` functions.

mofojed (changes requested):
Looks great overall! Some small things to clean up.

packages/console/src/ConsoleMenu.tsx
wusteven815: I noticed that the condition to set the tooltip as the `aria-label` isn't met since `iconOnly = false`.

https://github.com/deephaven/web-client-ui/blob/e90b627fed2c76b81ee96006abef14942388f3cc/packages/components/src/Button.tsx#L126-L127

mofojed: Ahhh right because we have the dropdown menu as a child. Whoops.
Add an `aria-label` explicitly then instead of `id`.

mofojed (changes requested):

tests/utils.ts
```diff
@@ -106,7 +106,7 @@ export async function openPlot(
   openObject(page, 'plot', name);
 
   if (waitForLoadFinished) {
-    await expect(page.locator('.iris-grid-panel')).toHaveCount(1);
+    await expect(page.locator('.chart-panel-container')).toHaveCount(1);
```
mofojed, line 109: I think we should instead be waiting for the loading spinner to disappear, similar to what we do for grids. But also for grids, we should wait for the panel/loading spinner to appear, then wait for loading to complete. What we had before:
```
  // Expect the panel to be open with a loading spinner first
  await expect(page.locator('.iris-chart-panel')).toHaveCount(1);

  // Wait until it's done loading
  await expect(page.locator('.iris-chart-panel .loading-spinner')).toHaveCount(
    0
  );
```
So I'd probably do:
```
  // Expect the panel to be open with a loading spinner first
  await expect(page.locator('.chart-panel-container .loading-spinner')).toHaveCount(1);

  // Wait until it's done loading
  await expect(page.locator('.chart-panel-container .loading-spinner')).toHaveCount(0);
```
Similarly what we had for tables before:
```

  // Wait for the panel to show
  await expect(page.locator('.iris-grid-panel')).toHaveCount(1);

  // Wait until it's done loading
  await expect(page.locator('.iris-grid-panel .loading-spinner')).toHaveCount(
    0
```

mofojed (changes requested):

tests/utils.ts
wusteven815: I've noticed that expecting the loading spinner to appear first can result in some flakiness if the object finishes loading before that part of the code can be executed.

mofojed (approved):

mofojed:
Great work!

PR #1768, opened 2024-02-01
fix: apply theme accent color scale and other small tweaks

dsmmcken:
- accent-100-1400 wasn't actually applied to spectrum, only actually matters in a few components (like checkboxes in lists), as most used other derived variables
- enable stylelint on css files
- change item list selected border color to match iris-grid selected border color (e2e item-list-selected-border change)
- manually override spectrum selected border color in lists and tables to not be hardcoded as blue, instead use accent
- adjusted inactive bg color in light theme
- fix mitre gaps in file list and command history list now that border is more visible
  fixes small mitre gaps:
  image_360


bmingles (approved):
Checked the mitre gaps and looks good. Verified that outlined button is now themed properly in styleguide

PR #1772, opened 2024-02-05
feat: Chart responsible for its own theme

### LINKED ISSUE #1728: Theming - Chart component should be responsible for its theme

bmingles:
This PR 
- Move theme colors down to `layout.template.layout` so that they are used as defaults. Some of the theme colors had to be set in `layout.template.data` for specific trace types
- Chart component now responds to ChartTheme context changes (this should enable chart theme switching, but it depends on some follow up plugins repo work to complete it).

**Testing**
For now, we should just make sure charts behave as they did before. Not all charts will update in response to a theme change, but re-running queries should show the theme change. I have tested this and have not seen any changes to existing behavior introduced by this PR

Once the following plugins repo PRs have merged, the chart should fully respond to theme changes:
https://github.com/deephaven/deephaven-plugins/pull/243
https://github.com/deephaven/deephaven-plugins/pull/251

resolves #1728

BREAKING CHANGE: 
- Renamed `ColorUtils.getColorwayFromTheme` to `normalizeColorway`
- Removed `chartTheme` arg from functions in `ChartUtils`, `ChartModelFactory` and `FigureChartModel` in @deephaven/chart


packages/chart/src/ChartUtils.ts
```diff
@@ -968,29 +966,8 @@ class ChartUtils {
       // The default histfunc in plotly is 'count', but the data passed up from the API provides explicit x/y values and bins
       // Since it's converted to bar, just set the widths of each bar
       seriesData.width = [];
-
-      if (seriesData.marker.line !== undefined) {
-        Object.assign(seriesData.marker.line, {
-          color: theme.paper_bgcolor,
-          width: 1,
-        });
```
mattrunyon, line 976: Just to confirm, this is an intentional change to the styling for histograms? It's not in the default template

packages/chart/src/ChartUtils.ts
```diff
@@ -1877,12 +1855,67 @@ class ChartUtils {
     return axis;
   }
 
+  /**
+   * Creates a plotly layout template object based on a given theme.
+   * See https://plotly.com/javascript/reference/layout/#layout-template
+   * @param theme The theme to use for the layout template
+   */
+  makeDefaultTemplate(theme: Partial<ChartTheme>): Template {
```
mattrunyon, line 1863: Add an `@returns`

Also, why is the param a partial? Shouldn't it just expect the full theme?

packages/chart/src/ChartUtils.ts
```diff
@@ -1877,12 +1855,67 @@ class ChartUtils {
...
+      title_color,
+    } = theme;
+
+    return {
+      data: {
+        bar: [
+          {
+            marker: {
+              line: { color: 'transparent' },
+            },
```
mattrunyon, line 1878: This is the only part I didn't see in the old series styling. Is it intentional?

packages/chart/src/ChartUtils.ts
```diff
@@ -1877,12 +1855,67 @@ class ChartUtils {
...
+    };
+  }
+
   /**
    * Creates a plotly layout object based on a given theme.
    * See https://plotly.com/javascript/reference/layout/
    * @param theme The theme to use for the layout
    */
-  makeDefaultLayout(theme: ChartTheme): Partial<Layout> {
+  makeDefaultLayout(theme: Partial<ChartTheme>): Partial<Layout> {
```
mattrunyon, line 1918: No need for partial on the param if the default template just expects the full theme

packages/chart/src/FigureChartModel.ts
```diff
@@ -72,17 +72,13 @@ class FigureChartModel extends ChartModel {
...
-      layout: this.chartUtils.makeDefaultLayout(theme),
-    };
     this.layout = {
       grid: {
         rows: figure.rows,
         columns: figure.cols,
         pattern: 'independent',
       },
-      template,
+      template: this.chartUtils.makeDefaultTemplate(theme),
```
mattrunyon, line 81: It would be ideal if the model didn't need the theme at all, but looks like it does for some dynamic axis stuff still.

You should be able to remove the template at least because it's overridden by `Chart` anyway

mattrunyon: I did a bit of testing and you can remove theme from the `updateFigureAxes` call. The only spot where it ends up using it is to set the axis if one doesn't exist, but plotly applies from the template in that case it seems.

Just to test, get a subplotting example from [here](https://deephaven.io/core/docs/how-to-guides/plotting/subplots/) and then change something in `makeLayoutAxis`. I made the y axis `zerolinecolor` red and it properly applied to both series even w/o the assignment in `updateLayoutAxes`

That then leaves just the error bars in the model which I bet can remove the theme as well since you added it to the default template. I couldn't find a good example on the docs to test this though

packages/chart/src/plotly/plotly-extends.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Fix some issues with plotly types.
+ */
+
+declare module 'plotly.js' {
```
mattrunyon, line 5: Can this go in the root `@types` folder? I'd try putting it there to see if you get the same results. It is meant to serve as a central types override

mattrunyon (changes requested):
Looks like the padding on the styleguide snapshots got a little smaller for the charts. I'm guessing it's fine, but wanted to make sure it was a known change

packages/chart/src/ChartUtils.ts
```diff
@@ -511,24 +511,26 @@ class ChartUtils {
...
-    let colorway: string[] = [];
-    if (theme.colorway) {
-      if (Array.isArray(theme.colorway)) {
-        colorway = theme.colorway;
-      } else if (typeof theme.colorway === 'string') {
-        colorway = theme.colorway.split(' ');
-      } else {
-        log.warn(`Unable to handle colorway property: ${theme.colorway}`);
-      }
+  static normalizeColorway(colorway?: string | string[]): string[] | undefined {
```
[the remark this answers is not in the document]
bmingles: ah good catch

packages/chart/src/ChartUtils.ts
bmingles: Yes, @dsmmcken specified that histogram bars should not have any border

packages/chart/src/ChartUtils.ts
```diff
@@ -968,29 +966,8 @@ class ChartUtils {
...
-      }
-    } else if (plotStyle === dh.plot.SeriesPlotStyle.OHLC) {
-      (seriesData as Partial<OhlcData>).increasing = {
-        line: { color: theme.ohlc_increasing },
-      };
-      (seriesData as Partial<OhlcData>).decreasing = {
-        line: { color: theme.ohlc_decreasing },
-      };
     } else if (plotStyle === dh.plot.SeriesPlotStyle.PIE) {
       seriesData.textinfo = 'label+percent';
```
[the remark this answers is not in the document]
bmingles: Possibly? Since there are some things that will have to stay here, I'm not too worried about it.

packages/chart/src/plotly/plotly-extends.d.ts
bmingles: It doesn't seem to merge with the existing Plotly types when I do this

packages/chart/src/ChartUtils.ts
bmingles: Yes, this is part of removing borders from histograms and gets rid of the default white border.

packages/chart/src/ChartUtils.ts
bmingles: I fixed the partial. Do we need a `@returns` here? I don't see that we consistently do this, and I don't see it adding any clarity to this particular util.

bmingles:
> Looks like the padding on the styleguide snapshots got a little smaller for the charts. I'm guessing it's fine, but wanted to make sure it was a known change

This is a known change. The styleguide was overriding some of the default layout. Now that we have moved to template.layout, the defaults properly stay in place.

packages/chart/src/ChartUtils.ts
```diff
@@ -1326,9 +1318,6 @@ class ChartUtils {
             axisProperty,
             figureAxisIndex
           );
-          if (layout[axisLayoutProperty] == null) {
-            layout[axisLayoutProperty] = this.makeLayoutAxis(axisType, theme);
-          }
```
mattrunyon, line 1356: I wonder if this should set the axis to an empty object instead by default. Then the template part would still be applied by the template, but any other axis specifics (like range I think) should get applied to the axis. The code after calls `updateLayoutAxis` only if the axis exists. In subplots, the axis wouldn't exist (but it does seem to get created somewhere, I just don't know where)

```ts
layout: {
  xaxis: ...,
  xaxis2: { range: ... },
  template: {
    layout: {
      xaxis: { styleStuff }
    }
  }
}
```

mattrunyon (changes requested):

packages/chart/src/ChartUtils.ts
```diff
@@ -1261,7 +1255,6 @@ class ChartUtils {
    * @param layoutParam The layout object to update
    * @param chartAxes The chart axes to update the layout with
    * @param figureAxes All figure axes to update the layout with
-   * @param theme Theme used for displaying the plot
```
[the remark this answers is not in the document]
bmingles: I took your suggestion to create the empty object, so comment should still be good.

packages/chart/src/ChartUtils.ts
bmingles: Good call. Doing this allowed me to restore most of the full tests I had deleted before which feels better to me.

mattrunyon (approved):

PR #1773, opened 2024-02-05
feat: always show close button on the active panel in a stack

dsmmcken:
Based on user feedback, we've made close buttons a little more prominent. Mainly the active tab in a stack now always shows it's close button. This mirrors behaviour in other IDEs like vs code. The intent is this will make it slightly less likely to accidently close tabs.

Summary:

close_button_change

Changes:
- show close button on active panels
- removed transition so close button appears instantly on hover of non-active panel
- added hover background to close button to make it stand-out slightly more
- adjusted overall tab size and spacing to be a little tighter
- reduced minimum tab size slightly to 10ch to fit more tabs in tab bar


mattrunyon (approved):
Did you add a bit more shadow to the tabs to give a little more depth to them?

dsmmcken:
> Did you add a bit more shadow to the tabs to give a little more depth to them?

No. Are you basing that on the diff of tests/golden-layout.spec.ts-snapshots/tests-golden-layout-operations-golden-layout-can-drag-tab-to-left-edge-2 images? 

Shadows were changed in a previous PR, but I guess playwright decided the shadow change alone was within tolerance to not regenerate. This PR actually changes the tab width, which is enough of a diff.

mattrunyon:
Yep just looking at the diffs I noticed a more prominent shadow

PR #1774, opened 2024-02-06
test: new styleguide tests

wusteven815:
- Refactored the style guide tests to use the same page instead of loading a new one per test
  - Moved the section isolation feature to client side instead of has a search param


bmingles:
It looks like this suffers from one of the issues I had when I first isolated the sections. The easiest way to see it is to introduce extra color variables into `packages/components/src/theme/theme-dark/theme-dark-palette.css`

image

and then run `npm run e2e:update-ci-snapshots` again

This "should" only impact the `Theme Color Palette` section, but it causes diffs in many other sections (presumably due to vertical shift). I see 185 files change.

PR #1776, opened 2024-02-06
feat: new error panel

### LINKED ISSUE #1520: Downsampling is automatically applied to charts that don't support it
Newly-created plots with too many points to actually plot appear to have downsampling enabled even when it's not supported (i.e. when the X axis is not a datetime). For example, the following query:

```groovy
t = emptyTable(250_000).updateView("RowIdx = ii").update("Val = Math.random()")

plot = figure().plot("Plot", t, "RowIdx", "Val").show()
```

Creates this plot (notice the 'downsampling failed' error icon):

image

There is nothing indicating that downsampling should be used or would be required; the only thing you see is a blank chart and that downsampling failed.

When disabling downsampling, the UI isn't always responsive — the panel doesn't 'spin' or anything, so it's hard to tell that it's working. For example, when I click to disable downsampling on that plot, this is still all I see ~3 seconds later:

image

It does eventually draw (although for larger tables w/ millions of rows, it appears to hang indefinitely):

image

When downsampling or plotting fails, it would be helpful to display the number of source table rows — makes it easier as a user to understand what you're telling DH to do/what you're doing wrong.

ethanalvizo:
Closes #1520 

Needs an accompanying DH-Core ticket to update the error messages and catch the new case of slow performance with too many points


packages/components/src/ChartErrorOverlay.scss
```diff
@@ -0,0 +1,17 @@
...
+    .btn {
+      margin-top: $spacer-3;
+      margin-right: $spacer-3;
+    }
+    .info-message,
+    .waiting-filter-list {
+      text-align: left;
+    }
+  }
+}
```
mofojed, line 17: Should rebase and use CSS variables. We recently switched to using CSS variables and had a lunch 'n' learn on this topic: https://github.com/deephaven-examples/lunch-n-learn/tree/main/css-variables (or https://github.com/deephaven-examples/lunch-n-learn/tree/main/css-variables#stop-using-scss-color-variables)

Variables are here: https://github.com/deephaven/web-client-ui/blob/main/packages/components/src/theme/theme-light/theme-light-semantic.css

(Doesn't look like you're using the SCSS variables declared anyway).

packages/components/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+  errorMessage,
+  clearError,
+  'data-testid': dataTestId,
+}: ChartErrorOverlayProps): ReactElement {
+  const messageTestId =
+    dataTestId != null ? `${dataTestId}-message` : undefined;
+
+  const slowPerformanceMessage = errorMessage.includes(
+    'Plot contains more than'
+  ); // TODO: core side changes to detect slow performance and show this message
```
mofojed, line 22: Add the ticket number to this.

packages/components/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const unableToDownsampleMessage =
+    errorMessage === 'Too many items to disable downsampling';
+
+  const undismissableError = unableToDownsampleMessage;
```
mofojed, line 27: Not sure why you redeclared this variable?

packages/components/src/LoadingOverlay.tsx
```diff
@@ -52,14 +50,13 @@ function LoadingOverlay({
...
             </div>
             {errorMessage != null && (
-              <div className="message-text" data-testid={messageTestId}>
-                {errorMessage}
-              </div>
+              <ChartErrorOverlay
+                errorMessage={errorMessage}
+                clearError={clearError}
+                data-testid={dataTestId}
+              />
```
mofojed, line 59: `LoadingOverlay` is built as a generic component; `ChartErrorOverlay` is specific to charting. `LoadingOverlay` should not depend on `ChartErrorOverlay`.

packages/components/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <div className="chart-panel-overlay chart-error-overlay">
+      <div
+        className={classNames(
+          'chart-panel-overlay-content chart-error-overlay-content'
+        )}
+      >
```
mofojed, line 35: All of these class names are specific to the charting plugin, which is in the `dashboard-core-plugins` package. The `components` package (which this is currently in) should be generic components and should not depend on `dashboard-core-plugins`. This file should probably be in `dashboard-core-plugins` or `chart`

packages/dashboard-core-plugins/src/panels/Panel.tsx
```diff
@@ -355,6 +358,7 @@ class Panel extends PureComponent<PanelProps, PanelState> {
           errorMessage={errorMessage}
           isLoaded={isLoaded}
           isLoading={isLoading}
+          clearError={clearError ?? undefined}
```
mofojed, line 361: Shouldn't need the `?? undefined` here...

packages/dashboard-core-plugins/src/panels/WidgetPanel.tsx
```diff
@@ -243,6 +246,7 @@ class WidgetPanel extends PureComponent<WidgetPanelProps, WidgetPanelState> {
         isClonable={isClonable}
         isRenamable={isRenamable}
         additionalActions={additionalActions}
+        clearError={clearError ?? undefined}
```
mofojed, line 249: Shouldn't need the `?? undefined` here

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -684,10 +685,17 @@ export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
     });
   }
 
-  handleError(): void {
+  handleError(error: Error): void {
     // Don't want to set an error state, because the user can fix a chart error within the chart itself.
     // We're not loading anymore either so stop showing the spinner so the user can actually click those buttons.
-    this.setState({ isLoading: false });
+    this.setState({ isLoading: false, error: error.message });
+  }
```
mofojed, line 692: Take a look at the comment there:

> Don't want to set an error state, because the user can fix a chart error within the chart itself.

With this change, you're wiring up an error dismissal mechanism through `Panel` for this feature. The errors presented in `Panel` are for when there's an issue displaying the panel or your connection (see in `WidgetPanel`) - they're not really errors you can dismiss, as you won't be able to do anything.
Instead, I'd make a `ChartErrorOverlay` in the `chart` package for the overlay, then in `Chart.tsx` display the error overlay when appropriate, and still emitting the `onError`. No need to wire this through `WidgetPanel` or `Panel`, that would require some more architecture.

mofojed (changes requested):

Issue #1646, opened 2023-11-15
Add formatting option to disable empty/null rendering in strings


dsmmcken:
Some users have expressed a preference to not show null strings/empty strings, instead show a blank cell. Add a user options to disable that formatting feature.

Make sure they get a prop to be set as well.

![image](https://github.com/deephaven/web-client-ui/assets/1576283/b0422342-9cf5-4da7-8301-f0bc5511aa3b)

PR #1778, opened 2024-02-07
feat: toggle empty/null rendering

wusteven815:
- Adds #1646 
  - Add a new state in Redux for if empty and null should be rendered
  - Add a check in `IrisGridTableModelTemplate` to check for if the setting is enabled


mofojed (approved):
Looks good to me. I also tested to make sure that the props were read correctly when specified in `deephaven.prop`. By default the file is in `~/.config/deephaven/deephaven.prop`, and I specified the `showEmptyStrings` and `showNullStrings` props by adding the following:
```
showEmptyStrings=true
showNullStrings=true

client.configuration.list=java.version,deephaven.version,barrage.version,showEmptyStrings,showNullStrings
```
User was still able to override the setting, which is expected.

PR #1786, opened 2024-02-09
test: update styleguide methodology

wusteven815:
- Related to optimization from #1671
- Updated `styleguide.spec.ts` to decrease the amount of `page.goto`
  - Troublesome tests still use the original methodology
  - Other tests are combined to the same page
- Added a way to do section isolation by client side, which the normal tests use


packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -71,24 +74,34 @@ function StyleGuide(): React.ReactElement {
...
-        <Flex
-          {...stickyProps}
-          UNSAFE_className={HIDE_FROM_E2E_TESTS_CLASS}
-          top="calc(100vh - 40px)"
-          marginTop={-32}
-          marginEnd={hasMultipleThemes ? -234 : 0}
-        >
-          <GotoTopButton />
-        </Flex>
+        {isTestMode && (
```
bmingles, line 94: This seems to hide the GotoTopButton in the styleguide during normal usage

packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -31,19 +31,22 @@ import { GoldenLayout } from './GoldenLayout';
...
-  justifyContent: 'end',
-  zIndex: 1,
-  UNSAFE_style: {
-    float: 'right',
-  },
-} as const;
-
 function StyleGuide(): React.ReactElement {
-  const isolateSection = window.location.search.includes('isolateSection=true');
+  const isTestMode = window.location.search.includes('testMode=true');
```
bmingles, line 35: I would make this case insensitive with something like `search.toLowerCase().includes('testmode=true')`

packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -71,24 +76,34 @@ function StyleGuide(): React.ReactElement {
...
-          top="calc(100vh - 40px)"
-          marginTop={-32}
-          marginEnd={hasMultipleThemes ? -234 : 0}
-        >
-          <GotoTopButton />
-        </Flex>
+        {!isTestMode && (
+          <Flex
+            {...stickyProps}
+            UNSAFE_className={HIDE_FROM_E2E_TESTS_CLASS}
```
bmingles, line 99: Out of curiosity, did the HIDE_FROM_E2E_TESTS_CLASS not hide this?

packages/code-studio/src/styleguide/StyleGuide.test.tsx
```diff
@@ -24,3 +24,22 @@ describe('<StyleGuide /> mounts', () => {
...
+  test('default does not activate the isolate input box', () => {
+    const { container } = render(
+      <ApiContext.Provider value={dh}>
+        <ThemeProvider themes={[]}>
+          <StyleGuide />
+        </ThemeProvider>
+      </ApiContext.Provider>
+    );
+
+    expect(container.querySelector('input[placeholder="Isolate"]')).toBeNull();
```
bmingles, line 43: I like that you are testing this. I would recommend using an `id` instead of depending on placeholder text since this is a special input that should only ever have 1 instance.

bmingles: Actually see suggestion regarding hashchange event handler below

packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -71,24 +76,34 @@ function StyleGuide(): React.ReactElement {
...
+          // UNSAFE_className={HIDE_FROM_E2E_TESTS_CLASS}
           marginTop={-56}
           top={20}
           gap={10}
           alignItems="end"
         >
           {hasMultipleThemes ? <ThemePicker /> : null}
           <SamplesMenu />
+          {isTestMode && (
+            <input
```
bmingles, line 88: Since we already use hash routing to identify which section is selected, we could probably simplify this by listening to the hash change event instead of requiring this input.

```typescript
 useEffect(() => {
    const hashChangeHandler = () => {
      setTargetSection(window.location.hash.replace('#', ''));
    };

    window.addEventListener('hashchange', hashChangeHandler);

    return () => {
      window.removeEventListener('hashchange', hashChangeHandler);
    };
  }, []);
  ```
  I think the unit test could then just set the `window.location.hash` and it "should" just work. Setting the hash alone shouldn't reload the page

File: tests/styleguide.spec.ts [hunk 6]
```diff
@@ -1,151 +1,178 @@
...
-test('UI regression test - Styleguide section count', async ({ page }) => {
-  await page.goto('/ide/styleguide');
+test.describe.configure({ mode: 'serial' });
+
+let page: Page;
+
+async function isolateAndGetSection(sectionId: string) {
+  const isolator = page.getByPlaceholder('Isolate');
+  await expect(isolator).toHaveCount(1);
+  await isolator.fill(sectionId);
```
bmingles, line 64: If you go with the "hashchange" event approach suggested in Styleguide.tsx, you would need to figure out how to set `window.location.hash = sectionId`

tests/styleguide.spec.ts
```diff
@@ -1,151 +1,178 @@
...
+      await expect(sampleSection).toHaveScreenshot(`${id}.png`);
+    });
   });
 });
 
-buttonSectionIds.forEach((id, i) => {
-  test(`Buttons regression test - ${id}`, async ({ page }) => {
-    // Isolate the section
-    await page.goto(`/ide/styleguide?isolateSection=true#${id}`);
+test.describe('Buttons regression test', () => {
```
bmingles, line 119: Does a failing test in the forEach cause remaining tests to be skipped, or will all still run?

bmingles (changes requested):
I left some comments / questions

tests/styleguide.spec.ts
```diff
@@ -1,151 +1,176 @@
...
+});
 
-  const sampleSections = await page.locator(
-    '.sample-section:not(.sample-section-e2e-ignore)'
-  );
+let page: Page;
 
-  await expect(sampleSections).toHaveCount(sampleSectionIds.length);
-});
+async function isolateAndGetSection(sectionId: string) {
```
bmingles, line 63: I like how you encapsulated all this

packages/code-studio/src/styleguide/StyleGuide.tsx
wusteven815: Seems like I somehow commented it out, brought it back and it looks fine.

tests/styleguide.spec.ts
wusteven815: It does cause it to be skipped over since it is in serial and Playwright doesn't have a feature to continue after a failed test in serial mode (even with soft assertions). However, I've made it so that the test will only be serial for CI. When running locally, it will be in parallel and continue with all the tests.

PR #1792, opened 2024-02-09
test: adds a test to make sure light theme and dark theme contain all the same variables

### LINKED ISSUE #1782: Create test to confirm Default Light and Default Dark contain exactly the same variables
Light theme and Dark theme are base themes that third parties can extend to create custom themes. 

Theme variables must be present in both themes to ensure good defaults are set for each base.

Write a test to verify the defined variable names match (ignoring value of that variable) between Default Light and Dark, and fail if they don't.

dsmmcken:
Ensure base default dark and default light theme always contains the same set of variables, as they are independant base themes.

Resolves #1782

### DIFF
--- a/packages/components/src/theme/ThemeUtils.test.ts
+++ b/packages/components/src/theme/ThemeUtils.test.ts
@@ -1,3 +1,6 @@
+import fs from 'fs';
+import path from 'path';
+
 import { ColorUtils, TestUtils } from '@deephaven/utils';
 import {
   DEFAULT_DARK_THEME_KEY,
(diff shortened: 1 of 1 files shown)


packages/components/src/theme/ThemeUtils.test.ts
```diff
@@ -333,6 +336,47 @@ describe('getDefaultBaseThemes', () => {
...
+        const filePath = path.join(directory, file);
+        const fileContent = fs.readFileSync(filePath, 'utf8');
+        // strip comments as they may contain css variable names
+        const fileContentsNoComments = fileContent.replace(
+          /\/\*[\s\S]*?\*\//g,
+          ''
+        );
+        const lines = fileContentsNoComments.split('\n');
+        // extract css variables from each line, before the colon
+        const matches = lines.map(line => line.match(/(--[a-zA-Z-]+):/)?.[1]);
```
bmingles, line 362: It doesn't look like we use this eslint rule in our codebase, but in the past I was led to believe that `RegEx.prototype.exec` may be preferred over `String.prototype.match`. Up to you whether you care on this one.

https://typescript-eslint.io/rules/prefer-regexp-exec/

packages/components/src/theme/ThemeUtils.test.ts
```diff
@@ -333,6 +336,47 @@ describe('getDefaultBaseThemes', () => {
...
+   * This test is to ensure that the css variables in theme-dark and theme-light are in sync,
+   * to prevent the case where a css variable is added to one theme and not the other.
+   */
+  it('light and dark themes should contain the same css variables', () => {
+    const themeDarkDir = './packages/components/src/theme/theme-dark';
+    const themeLightDir = './packages/components/src/theme/theme-light';
+
+    function extractCssVariablesFromFiles(directory: string): Set<string>[] {
+      const files = fs.readdirSync(directory);
+      const fileVariables = [] as Set<string>[];
```
bmingles, line 350: I would set the type on the left side in this case as it is slightly stricter. `const fileVariables: Set<string>[] = []`

packages/components/src/theme/ThemeUtils.test.ts
```diff
@@ -333,6 +336,47 @@ describe('getDefaultBaseThemes', () => {
...
+          /\/\*[\s\S]*?\*\//g,
+          ''
+        );
+        const lines = fileContentsNoComments.split('\n');
+        // extract css variables from each line, before the colon
+        const matches = lines.map(line => line.match(/(--[a-zA-Z-]+):/)?.[1]);
+        matches?.forEach(match => {
+          if (match != null) return variables.add(match as string);
+        });
+        fileVariables.push(new Set([...variables].sort()));
```
bmingles, line 366: I'm not sure the sorting is necessary here unless I'm misreading. I think jest will treat this as true
```
expect(new Set('ABC')).toEqual(new Set('BCA'));
```

bmingles (changes requested):
A few minor comments

packages/components/src/theme/ThemeUtils.test.ts
dsmmcken: It's not necessary for the comparison no. But it seemed to make the results more predicatble for humans.

bmingles (approved):
LGTM

PR #1802, opened 2024-02-14
feat: Lazy loading and code splitting

mattrunyon:
This adds lazy loading for `chart` and `iris-grid` and enables it for `MarkdownComponent` as well.

Replaced the exports for `Chart` and `IrisGrid` with lazy loading wrappers. This should mean anywhere else we use the components in the app should be automatically lazy loaded.

`iris-grid` doesn't seem to really lazy load anything. Might be that most/all of its dependencies are already used in other spots in the app, so they all load anyway.

`chart` I confirmed lazy loads by watching the network tab on page load with no charts open. When I open a chart or view the memory usage graph, the network tab shows 2 chunks loading in including plotly.

`MarkdownComponent` I confirmed the same was as `chart` but with opening a markdown panel. There is a big chunk lazy loaded which includes mathjax.

We could remove the `iris-grid` stuff for now if we want, or keep it and maybe future changes will enable splitting. Outside of monaco I don't think it's a very heavy import


dsmmcken:
Good work.

> monaco will basically always be required because of the console

An app mode dashboard can disable the console, and is actually probably the expected pattern. Iris grid sidebar does use Monaco for custom columns, and that would benefit from lazy loading in monaco.

mofojed:
> monaco will basically always be required because of the console.

That's not true - in App mode, you can specify a default layout without a console at all. We would need console to lazy load it as well too though, and would need to initialize Monaco correctly still (AppRoot calls MonacoUtils.init right now)

packages/chart/src/plotly/Plot.tsx
```diff
@@ -0,0 +1,15 @@
+import { LoadingOverlay } from '@deephaven/components';
+import { lazy, Suspense } from 'react';
+
+const PlotBase = lazy(() => import('./PlotBase.js'));
```
mofojed, line 4: No need for .js.

I'm inexplicably sad you didn't name this `LazyPlot` and use the same naming used for `LazyChart`/`Chart`.

Side note - why _does_ `HeapUsage` use `Plot` directly instead of `Chart`? Were we just being lazy there/didn't want to make a model to pass to `Chart`?

packages/chart/src/LazyChart.tsx
```diff
@@ -0,0 +1,15 @@
+import { LoadingOverlay } from '@deephaven/components';
+import { lazy, Suspense } from 'react';
+
+const Chart = lazy(() => import('./Chart.js'));
```
mofojed, line 4: I don't think you need the `.js` here.
```suggestion
const Chart = lazy(() => import('./Chart'));
```

mofojed (changes requested):

mofojed:
Seems to have broken the Style guide?

packages/chart/src/plotly/Plot.tsx
mattrunyon: I can rename it to match the other naming.

That would be my guess is we didn't want to make a model. Or there's other things `Chart` does that we didn't want in `HeapUsage`. Not sure

packages/chart/src/LazyChart.tsx
mattrunyon: I'll check if our babel add extensions step adds it to dynamic imports. Adding the `.js` is proper ESM

mattrunyon: Our babel plugin does not add extensions to dynamic imports, so I'm going to leave the `.js` so it's valid ESM

mattrunyon:
Added @bmingles to confirm he's ok with the CSS fix for the chunks loading out of order in production

mattrunyon:
Also, I didn't think about app mode not necessarily requiring a monaco load. That said, I won't worry about splitting out monaco in this PR as I mostly wanted mathjax and plotly split

packages/chart/src/index.ts
```diff
@@ -1,12 +1,12 @@
...
+export { default as Chart } from './LazyChart';
 export { default as ChartModelFactory } from './ChartModelFactory';
 export { default as ChartModel } from './ChartModel';
 export { default as ChartUtils } from './ChartUtils';
 export * from './ChartUtils';
 export * from './DownsamplingError';
 export { default as FigureChartModel } from './FigureChartModel';
 export { default as MockChartModel } from './MockChartModel';
-export { default as Plot } from './plotly/Plot';
+export { default as Plot } from './plotly/LazyPlot';
```
bmingles, line 9: Are there any breaking change implications of changing default imports to lazy ones? The only one I can think of is does this now require a Suspense boundary in places that wouldn't have required one before?

bmingles: Oh nvm, looks like you baked the Suspense boundary into the exported default.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -451,7 +452,7 @@ export interface IrisGridState {
   columnHeaderGroups: readonly ColumnHeaderGroup[];
 }
 
-export class IrisGrid extends Component<IrisGridProps, IrisGridState> {
+class IrisGrid extends Component<IrisGridProps, IrisGridState> {
```
bmingles, line 455: Why remove the named export?

packages/iris-grid/src/index.ts
```diff
@@ -8,6 +8,7 @@ export * from './CommonTypes';
 export { default as ColumnHeaderGroup } from './ColumnHeaderGroup';
 export * from './PartitionedGridModel';
 export * from './IrisGrid';
+export { default as IrisGridType } from './IrisGrid';
```
bmingles, line 11: Isn't this the lazy component? Why the `Type` suffix?

packages/code-studio/vite.config.ts
```diff
@@ -101,32 +101,6 @@ export default defineConfig(({ mode }) => {
       emptyOutDir: true,
       sourcemap: true,
       target: 'esnext',
-      rollupOptions: {
-        output: {
-          manualChunks: id => {
```
bmingles, line 106: Curious what changed so that we don't need this anymore?

packages/chart/src/index.ts
mattrunyon: Yep it should be non-breaking since `Suspense` and `lazy` has existed since React 16 and we specify/use 17

packages/components/src/theme/theme-spectrum/theme-spectrum-alias.module.css
```diff
@@ -1,6 +1,13 @@
...
 /* stylelint-disable alpha-value-notation */
-.dh-spectrum-alias {
+
+/**
+ * The higher-alias-specificity class is so we always take precedence over spectrum's definitions 
+ * It must be different from any other CSS module specificity classes for spectrum's theme
+ * Otherwise one of the classes will not get applied since the JS part receives the same key,
+ * but the CSS module will generate a different class name suffix. This ends up with spectrum picking only 1.
+ */
+.dh-spectrum-alias.higher-alias-specificity {
```
bmingles, line 10: I would prefix the specificity class with `.dh-` for consistency. Alternative would be to just double the original class `.dh-spectrum-alias.dh-spectrum-alias` and then you don't need a new name. IMO makes it a little more clear you are just doubling the specificity not targetting a different thing.

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: I'm pretty sure it wouldn't conflict w/ the index export since we explicitly list `IrisGrid` as an export, but I wanted to make sure it wouldn't conflict since we do `export * from './IrisGrid'`

bmingles: ok

packages/iris-grid/src/index.ts
mattrunyon: It's just for one spot where we assign a ref and call into `IrisGrid` methods. It is a component (I might be able to export the typeof instead), but it's intended to be used as just a type in that one spot

bmingles: Gotcha. Yeah if you can export as a type, this makes sense. If you can't, I would suggest renaming it to `IrisGridBase` or something of your liking that doesn't imply it is just a type.

packages/components/src/theme/theme-spectrum/theme-spectrum-alias.module.css
mattrunyon: Neat. Didn't know that was valid in CSS to increase specificity. I'll change and make sure the CSS modules handles it properly (doesn't assign 2 separate suffixes for some reason)

packages/code-studio/vite.config.ts
mattrunyon: I think part of the reason for chunking like this initially was we wanted to split out monaco and plotly, but it didn't actually do any lazy loading. I'm not sure if the manual chunks works properly if I specify just those that we want

We added mathjax to parse markdown latex math syntax which was a heavy package, so that is part of the driver for splitting the chunks. I can check if things properly lazy load (mathjax and plotly) w/ the manual chunks and just adding a mathjax chunk

bmingles: I honestly don't have any understanding of what the original config was for or whether it matters, just wanted to document reasoning that it used to be required but now is not.

mattrunyon: Found out Vite still splits chunks out properly w/ lazy loading even if I use manual chunks for the node_modules stuff. This allowed me to write an e2e test verifying that plotly and mathjax chunks don't load. Previously, the Vite chunks could be random. Like the mathjax chunk ended up in 1 file called `svg-abc123.js`

This might be slightly less efficient for chunking since the smaller dependencies like rehype-math get included in the general vendor chunk than the auto chunking, but I like that it can be tested more easily

packages/iris-grid/src/index.ts
mattrunyon: I exported as type and verified that `new IrisGridBase()` results in a TS error because it was exported w/ the type modifier

It's not intended to be used as its own component (the `IrisGrid` lazy wrapper should just work) and is just because in the panel we call some class methods directly using the ref

mattrunyon:
Added an e2e test which works because of the manual chunks for the large packages. Tested the e2e test locally by changing the `LazyPlot` export in the chart index to just export from `Plot` which causes plotly to not lazy load and the test to fail

packages/components/src/theme/theme-spectrum/index.ts
```diff
@@ -12,7 +12,9 @@ import './theme-spectrum-overrides.css';
  * e.g.
  * {
  *   'dh-spectrum-palette': '_dh-spectrum-palette_abr16_1',
+ *   'higher-palette-specificity': '_higher-palette-specificity_18mbe_1',
```
bmingles, line 15: Should probably revert the comment block too

tests/lazy-loading.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { test, expect } from '@playwright/test';
+
+test('does not load heavy dependencies by default', async ({ page }) => {
```
bmingles, line 3: Is there a way to verify the requests do actually happen once they are required?

packages/components/src/theme/theme-spectrum/theme-spectrum-alias.module.css
```diff
@@ -1,6 +1,10 @@
 /* stylelint-disable custom-property-empty-line-before */
 /* stylelint-disable alpha-value-notation */
-.dh-spectrum-alias {
+
+/**
+ * Intentionally using the classname twice so we have higher specificity than spectrum's definitions
+ */
```
mofojed, line 6: Why are we doing this and why is it a part of this change? Should add to the comment why we need to do this (chunks loading out of order).

Though that being said - chunks loading out of order implies there's something off with the packaging. Why is this happening now?

mattrunyon: I'll update the comment.

I didn't dig too far into why the chunks were out of order (we don't directly import spectrum's css nor can we because it's spread across every component). We could also have an import order wrong somewhere in our code. Either way, just increasing the specificity is safer in the long run I think (although it did fail e2e tests because of the import order)

tests/lazy-loading.spec.ts
mattrunyon: Added a check for this and included rough size checks to ensure the chunk is not just some 1kB wrapper. Although the size checks don't seem to work in safari (size listed as 0), but the request still completes

tests/lazy-loading.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  expect(requests.some(req => req.url().includes('assets/plotly'))).toBe(false);
+
+  await openPlot(page, 'simple_plot');
+
+  const plotlyRequest = requests.find(req =>
+    req.url().includes('assets/plotly')
+  );
+  expect(plotlyRequest).toBeDefined();
+  await expectResponseSize(plotlyRequest, 300 * 1000); // 300kB
```
bmingles, line 40: Is this dependable, or can the size change as we upgrade things in the future?

tests/lazy-loading.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Checks the size of the response body of a request
+ * If the response body size is 0, the request is not checked
+ * This seems to happen for Safari sometimes
+ *
+ * @param request The request object. Playwright provides size on the request, not the response
+ * @param size The minimum size in bytes
+ */
+async function expectResponseSize(request: Request | undefined, size: number) {
```
bmingles, line 12: I missed this before. It would be good to name this something that indicates it is the minimum response size it checks for `expecteMinimumResponseSize` or something like that.

tests/lazy-loading.spec.ts
bmingles: Ah, I misinterpreted what `expectResponseSize` actually does. I left a suggestion on renaming that for more clarity.

mofojed (approved):

PR #1806, opened 2024-02-15
fix: Update plotly to v2.29, resolve some rendering issues

mofojed:
- v2.28 had some fixes for M1 macs, including for the scattergl issue we've fallen back to scatter for in the past
- Did not resolve the range breaks issue, so that still requires `scatter` type
- See https://github.com/plotly/plotly.js/issues/3522 for more details
- Needed to uninstall plotly, then reinstall it - until I did, line plots were not appearing correctly in `scattergl` mode
  - Did `npm uninstall react-plotly.js --workspace packages/chart` and `npm uninstall react-plotly.js --workspace packages/chart` followed by `npm install plotly.js --workspace packages/chart` and `npm install react-plotly.js --workspace packages/chart`
- Tested on Linux desktop and M1 mac, ensured plots were appearing correctly. Will verify e2e tests as well
- ScatterGL seems to render things _slightly_ differently, so snapshots needed to be updated for a couple plots


mattrunyon (commented):
Looks like the e2e tests failed

mofojed:
I'm not sure why it's not rendering lines in scattergl. When I try on codepen with the same data/layout, it all seems to work fine. Perhaps something wrong with the plotly modules we're including??

mattrunyon (approved):

Issue #1746, opened 2024-01-23
Persist multiple dashboards


mattrunyon:
#1683 is to add support for displaying multiple dashboards. This issue will be for persisting multiple dashboards across page reloads. I've added a few TODOs with this issue number to spots where data should be stored/updated for dashboards, but it might not be comprehensive.

- Cleanup what is stored in workspaceData and migrate things like `layoutConfig` and `links` to `dashboardData`
- Persist `dashboardData` from redux (likely with a `LocalDashboardStorage` similar to `LocalWorkspaceStorage`)
- Properly rehydrate dh.ui dashboards

PR #1810, opened 2024-02-15
fix: Load default dashboard data from workspace data

mofojed:
- Until we have support for #1746 , load the default dashboard data from workspace
  - Needed for testing the widget hydration PR
- Tested with my changes for widget hydration
  - Ensured deephaven.ui widgets re-opened upon refresh
  - Ensured links remained between existing tables

### DIFF
--- a/packages/dashboard/src/DashboardEvents.ts
+++ b/packages/dashboard/src/DashboardEvents.ts
@@ -2,22 +2,30 @@ import type { EventHub } from '@deephaven/golden-layout';
 
 export const CREATE_DASHBOARD = 'CREATE_DASHBOARD';
 
-export interface CreateDashboardPayload {
+export interface CreateDashboardPayload<T = unknown> {
   pluginId: string;
   title: string;
-  data: unknown;
+  data: T;
 }
 
-export function listenForCreateDashboard(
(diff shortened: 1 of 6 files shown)


packages/dashboard/src/DashboardEvents.ts
```diff
@@ -2,22 +2,30 @@ import type { EventHub } from '@deephaven/golden-layout';
 
 export const CREATE_DASHBOARD = 'CREATE_DASHBOARD';
 
-export interface CreateDashboardPayload {
+export interface CreateDashboardPayload<T = unknown> {
   pluginId: string;
   title: string;
-  data: unknown;
+  data: T;
```
mattrunyon, line 8: Why the generic? The data comes from the plugin and we only really care it's serializable

mofojed: So the listen/emit events can add typing for payload. For the emit function which will come from the plugin, can add typing so we know we're not passing a malformed payload.

mattrunyon (commented):
Looks good. I think we should wait until the plugin side of this change is also ready in case there are any more generics/type changes that are needed

Issue #1815, opened 2024-02-20
Remove page.goto from e2e openObject


mattrunyon:
As a developer, I want  to open multiple objects in an e2e test without page reload so that I can test cases with multiple objects or where an object load is delayed until another check.

The `openObject` util function starts with `page.goto('')` which I think should not be part of the util. The test should worry about navigating pages

PR #1819, opened 2024-02-20
test: improve openObject

### LINKED ISSUE #1816: e2e openTable and openChart utils do not check the panel exists
It seems e2e tests are running fine, but the `openTable` and `openChart` utils could introduce flakiness because they don't wait to check if the panel actually exists, just that there is no loading spinner.

The issue is there's no loading spinner to begin with, so if the panel hasn't even begun loading, the check will pass saying it's loaded.

One idea is to check the number of panels of that type before opening the object, then wait for 1 more panel of that type to exist, then wait for no loader on that n-th panel specifically. There's an issue with this if you were to open the same panel twice though because the util is using the console panels menu instead of the app panels menu.

So 2 fixes really
1. Verify the panel exists before saying it's loaded
2. Use the app panels menu instead of the console panels menu so we can open 2 copies of the same object. This is so we can call `openTable` on the same table twice without failing the loading check

wusteven815:
- Adds #1815 
  - Move all `page.goto` out of the `openObject` util method and into each test
  - Add util function for to go to page and wait for load
- Fixes #1816 
  - Use app panel menu instead of console panel menu
  - Add check that a new panel is created in `openTable` and `openPlot`

### DIFF
--- a/tests/context-menu.spec.ts
+++ b/tests/context-menu.spec.ts
@@ -14,6 +14,8 @@ async function moveMouseAwayFromTable(page: Page) {
 }
 
 test.beforeEach(async ({ page }) => {
+  await page.goto('');
+  await expect(page.locator('.loading-spinner')).toHaveCount(0);
   await openTable(page, 'all_types');
 
   const tableOperationsMenu = page.locator(
--- a/tests/context-menu.spec.ts
+++ b/tests/context-menu.spec.ts
@@ -1,5 +1,10 @@
 import { test, expect, Page } from '@playwright/test';
-import { waitForLoadingDone, openTable, expectContextMenus } from './utils';
+import {
+  waitForLoadingDone,
+  openTable,
+  expectContextMenus,
+  gotoPage,
+} from './utils';
 
 async function openAdvancedFilters(page: Page) {
   await page
@@ -14,8 +19,7 @@ async function moveMouseAwayFromTable(page: Page) {
 }
 
 test.beforeEach(async ({ page }) => {
-  await page.goto('');
-  await expect(page.locator('.loading-spinner')).toHaveCount(0);
+  await gotoPage(page, '');
   await openTable(page, 'all_types');
 
   const tableOperationsMenu = page.locator(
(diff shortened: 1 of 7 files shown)


tests/context-menu.spec.ts
```diff
@@ -14,6 +14,8 @@ async function moveMouseAwayFromTable(page: Page) {
 }
 
 test.beforeEach(async ({ page }) => {
+  await page.goto('');
+  await expect(page.locator('.loading-spinner')).toHaveCount(0);
```
mofojed, line 18: Since we're doing the exact same thing in a dozen spots now, we should put this in a utility method we can call. In utils.ts, something like:
```
async function gotoPage(url: string, options?: ...): Promise<void> {
  await page.goto(url, options);
  await expect(...)
}
```

mofojed (changes requested):

wusteven815:
Would it be better to combine `opentable` and `openPlot` together (including their panel number change) to a generic `openPanel`, like in

https://github.com/deephaven/deephaven-plugins/blob/e5ded24d75559e704197fb9fcf22b4a489590350/tests/utils.ts#L31-L33

tests/utils.ts
```diff
@@ -105,10 +108,11 @@ export async function openPlot(
...
-  await openObject(page, 'plot', name);
+  const panelCount = await page.locator('.iris-chart-panel').count();
+  await openPanel(page, name);
 
   if (waitForLoadFinished) {
-    // Wait until it's done loading
+    await expect(page.locator('.iris-chart-panel')).toHaveCount(panelCount + 1);
     await expect(
       page.locator('.chart-panel-container .loading-spinner')
     ).toHaveCount(0);
```
mattrunyon, line 118: I'd add a check after the loading as well that checks the internal chart wrapper exists. This caused some inconsistency w/ the lazy loading PR.

See https://github.com/deephaven/web-client-ui/pull/1802/files#diff-bccd87d19079cf7b35274281848f7e5a32a070e5101bff9c5b67b58fb82fc16d

mattrunyon (dismissed):

mofojed (changes requested):
Should match the `panelCount` checks, otherwise looks ready to merge.

wusteven815:
Succeeded by #1837

PR #1825, opened 2024-02-22
fix: spectrum textfield validation icon position with set content-box

dsmmcken:
Moves a DHE fix to DHC.

fixes text field spectrum validation icon position: 
image

BREAKING CHANGE: the duplicate `spectrum-Textfield-validationIcon` css in DHE should be removed

### DIFF
--- a/packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
+++ b/packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
@@ -9,6 +9,11 @@ label[class^='spectrum-'] {
   margin-bottom: 0;
 }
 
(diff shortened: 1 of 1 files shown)


dsmmcken:
feels like a bit of an abuse of breaking change, but not sure how else to flag that the DHE code can be removed when this is merged to DHE.

bmingles (changes requested):
Looks like e2e tests need to be updated

dsmmcken:
Flaky test failure, re-ran and it passed.

bmingles (approved):
LGTM. Verified that the icon is positioned correctly in the styleguide y setting `validationState="invalid"`

PR #1833, opened 2024-02-26
feat: exposes editor-line-number-active-fg theme variable

dsmmcken:
Fixes issue with active line number looking bad in light theme.

- Exposes editor-line-number-active-fg as editor semanatic variable
- Sets to blue/black in light theme to match vs code light theme approach


bmingles (approved):
Confirmed line numbers + active line numbers look good in light + dark theme.

PR #1842, opened 2024-02-28
refactor: Switch to generated jsapi types

### LINKED ISSUE #1770: Consume generated jsapi types

mattrunyon:
Fixes #1770 

Here's the list of changes that weren't just updating the type import

ChartUtils
- [ ] dh.plot.AxisType.Z does not exist any more. Changed to null for Z for now since we don't do any specific checks for Z axis in the template creation
- [ ] addTickSpacing add a null check to gapBetweenMajorTicks

Console
- [ ] Changed openObject to take VariableDescriptor instead of VariableDefinition. Removed title from openObject calls since it's not part of the descriptor

MonacoProviders
- [ ] monacoToLspPosition added a pick to the return type

TableSaver
- [ ] cast readonly to non-readonly column array for snapshot call

Integer/Decimal/DateTimeColumnFormatter
- [ ] Add format.formatString null check in isValid

IrisGridUtils
- [ ] Removed readonly on columnHeaderGroups line 138
- [ ] Changed DehydratedSort.direction to just string

IrisGrid
- [ ] Casting api sort direction (just string) to TableUtils SortDirection (enum)
- [ ] Changed getOperationMap return type to not readonly (JSAPI doesn't mark as readonly)
- [ ] Removed readonly from CommonTypes OperationMap
- [ ] Changed CommonTypes UITotalsTableConfig to not extend dh.TotalsTableConfig because of readonly issues

IrisGridContextMenuHandler
- [ ] Casted a few ofNumber calls to value as number since we mark it unknown
- [ ] dateFilterActions changed value to type Date and convert to DateWrapper before number filter value
- [ ] Throw if value is not Date at line ~500

FormatContextMenuUtils
- [ ] Line 91 add formatString ?? undefined to coerce string | null | undefined to just string | undefined

IrisGridTreeTableModel
- [ ] Removed formatRowColor from virtualColumns in constructor. It's marked as static in the types
- [ ] Cast as unknown as DisplayColumn because it's not a real column

IrisGridTableModelTemplate
- [ ] Added ?? null to line 1318 because TreeTable setViewport returns void. Table setViewport returns TableViewportSubscription

ServerConfigBootstrap
- [ ] Added cast to getServerConfigValues to make Map constructor happy

AuthConfigBootstrap
- [ ] Added cast to getAuthConfigValues to make Map constructor happy

ConsolePanel, Console, useObjectFetcher
- [ ] Changed some VariableDefinition uses to dh.ide.VariableDescriptor & { title?: string }

DashboardLayout
- [ ] Removed title from hydrateDefault when metadata.id is null (it's redundant as name is the field we want and title is not part of the descriptor)

FileStorage
- [ ] Changed FileType to just string to match JS API

IrisGridPanelTypes
- [ ] Added Partial to isLegacyIrisGridPanelMetadata

Added a blank line to logo.css because of stylelint? Might have accidentally bumped stylelint


packages/app-utils/src/components/AuthBootstrap.tsx
```diff
@@ -48,7 +48,10 @@ export function AuthBootstrap({ children }: AuthBootstrapProps): JSX.Element {
       let isCanceled = false;
       async function loadAuthConfigValues(): Promise<void> {
         try {
-          const newAuthConfigValues = await client.getAuthConfigValues();
+          const newAuthConfigValues = (await client.getAuthConfigValues()) as [
+            string,
+            string,
+          ][];
```
mofojed, line 54: This kind of sucks, but I don't have any clever solutions. Would need a type predicate to map the `string[][]` to `[string, string][]` by checking each element I guess? `isMapConstructorArray` or something.
I also assume that GWT can't express this typing, otherwise that would be the best option to fix it in the jsapi-types source (though can have this temporarily).

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -30,7 +30,7 @@ export function ConnectionBootstrap({
   const api = useApi();
   const client = useClient();
   const [error, setError] = useState<unknown>();
-  const [connection, setConnection] = useState<IdeConnection>();
+  const [connection, setConnection] = useState<dh.IdeConnection>();
```
mofojed, line 33: A fun PR to watch about importing classes from namespaced packages: https://github.com/Microsoft/TypeScript/issues/13135

packages/chart/src/ChartTestUtils.ts
```diff
@@ -22,9 +11,9 @@ class ChartTestUtils {
 
   static DEFAULT_SERIES_NAME = 'MySeries';
 
-  private dh: DhType;
+  private dh: typeof DhType;
```
mofojed, line 14: Interesting you need to `typeof` a `namespace`.

packages/chart/src/ChartTestUtils.ts
```diff
@@ -37,25 +26,25 @@ class ChartTestUtils {
...
-  } = {}): Axis {
+  } = {}): DhType.plot.Axis {
     const { dh } = this;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     return new (dh as any).Axis({
       label,
       type: type ?? dh.plot.AxisType.X,
       position: position ?? dh.plot.AxisPosition.BOTTOM,
-      formatType: formatType ?? dh.Axis.FORMAT_TYPE_NUMBER,
+      formatType: formatType ?? dh.plot.AxisFormatType.NUMBER,
```
mofojed, line 41: Guess that's always been broken, but never gets used.

packages/chart/src/ChartUtils.ts
```diff
@@ -1963,7 +1962,7 @@ class ChartUtils {
       scene: {
         xaxis: this.makeLayoutAxis(dh.plot.AxisType.X, theme),
         yaxis: this.makeLayoutAxis(dh.plot.AxisType.Y, theme),
-        zaxis: this.makeLayoutAxis(dh.plot.AxisType.Z, theme),
+        zaxis: this.makeLayoutAxis(null, theme),
```
mofojed, line 1965: Ah, on Enterprise it did used to have this constant. I guess we still need to set this for theming purposes of 3d plots in plotly-express? Wondering if we can just remove it and `makeLayoutAxis` doesn't need to take `null` then.

packages/file-explorer/src/FileStorage.ts
```diff
@@ -11,7 +11,7 @@ export interface FileMetadata {
   basename: string;
 }
 
-export type FileType = 'file' | 'directory';
+export type FileType = string;
```
mofojed, line 14: Instead of this change, we should throw in `GrpcFileStorage` if the type does not match a known type `'file' | 'directory'`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4399,7 +4389,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
                     model,
                     column,
                     advancedFilterOptions,
-                    sortDirection,
+                    sortDirection as SortDirection,
```
mofojed, line 4392: Guessing GWT can't express `Literal` types for sort either... rather than casting should we have a method to convert it and throw if it can't be converted? Would be more "correct".

Side note, not sure how this isn't an error currently, since `null` shouldn't match `SortDirection | undefined`

mofojed (changes requested):
Not done yet but posting some minor feedback so far. Couple of things to change.

Still need to review the rest.

packages/app-utils/src/components/AuthBootstrap.tsx
mattrunyon: I did note this in a non-critical improvements ticket https://github.com/deephaven/deephaven-core/issues/5082

Ya we could add a util to actually check it since this is just assuming it's an array of string pairs

packages/chart/src/ChartUtils.ts
mattrunyon: We use just the X or Y in `makeLayoutAxis` to add a few params to X or Y axis specifically

That said, w/ the theme changes I think that could be removed entirely and in the defaults here we could instead do something like this for the 1-2 props we add to the X and Y axis. I'd say do that as a followup though to keep this PR just related to types

```ts
xaxix: {
  ...this.makeLayoutAxis(theme),
  showgrid: true
}
```

packages/jsapi-utils/src/TableUtils.ts
```diff
@@ -1776,7 +1778,9 @@ export class TableUtils {
       );
     }
 
-    return dh.FilterValue.ofNumber(TableUtils.removeCommas(value));
+    return dh.FilterValue.ofNumber(
+      TableUtils.removeCommas(value) as unknown as number
+    );
```
mofojed, line 1783: Gross. Do we have a follow up ticket to improve JS API? I feel like I commented on this at the time on the JS API review as well as `ofNumber` needing to take `string`, but unsure why that wasn't added to the Union type.

packages/jsapi-utils/src/ViewportDataUtils.ts
```diff
@@ -74,11 +74,11 @@ export function createOnTableUpdatedHandler<T>(
  */
 export function defaultRowDeserializer<T>(
   row: ViewportRow,
-  columns: Column[]
+  columns: dh.Column[]
 ): T {
   return columns.reduce((result, col) => {
     // eslint-disable-next-line no-param-reassign
-    result[col.name as keyof T] = row.get(col);
+    result[col.name as keyof T] = row.get(col) as T[keyof T];
```
mofojed, line 81: We do have the `ValueOf` utility type for this case, e.g.
```
    result[col.name as keyof T] = row.get(col) as ValueOf<T>;
```
Doesn't really matter.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: Ya I think there were some issues with it/not a high priority. I'll add a typeguard and check for this.

`SortDirection` is `"ASC" | "DESC" | "REVERSE" | null``, so that's why it doesn't fail

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -579,7 +561,7 @@ class IrisGridTableModelTemplate<
         assertNotNull(theme.nullStringColor);
         return theme.nullStringColor;
       }
-      if (format && format.color) {
+      if (format?.color != null) {
```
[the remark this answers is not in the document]
mattrunyon: Good catch

packages/jsapi-utils/src/TableUtils.ts
mattrunyon: https://github.com/deephaven/deephaven-core/issues/5082

mofojed (approved):

PR #1843, opened 2024-02-29
test: fix open table flakiness

### LINKED ISSUE #1845: openTable causes flaky tests with lazy loading
Tests related to right-clicking are sometimes flaky because they run before lazy loading

wusteven815:
- Fixes #1845 
  - Add an expect for loading spinner to disappear

The expects will be in the following order:
1. Table panel is loaded (new `.iris-grid-panel` appears)
2. Grid is lazy loaded (`.iris-grid-panel .loading-spinner` is gone)
3. Grid content is loaded (`.iris-grid .iris-grid-loading-status` is gone)

### DIFF
--- a/tests/utils.ts
+++ b/tests/utils.ts
@@ -92,6 +92,7 @@ export async function openTable(
     await expect(
       page.locator('.iris-grid .iris-grid-loading-status')
     ).toHaveCount(0);
+    await expect(page.locator('.iris-grid-panel .iris-grid')).toHaveCount(0);
   }
 }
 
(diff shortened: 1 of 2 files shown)


mofojed (dismissed):

tests/utils.ts
```diff
@@ -89,6 +89,9 @@ export async function openTable(
 
   if (waitForLoadFinished) {
     await expect(page.locator('.iris-grid-panel')).toHaveCount(panelCount + 1);
+    await expect(page.locator('.iris-grid-panel .loading-spinner')).toHaveCount(
+      0
+    );
     await expect(
       page.locator('.iris-grid .iris-grid-loading-status')
     ).toHaveCount(0);
```
mattrunyon, line 97: You still need to make a positive assertion here that content is available of some format. Just the spinners being gone could flake out if there is nothing rendered as these assertions are always true even if there's no grid on the page

mattrunyon (dismissed):

PR #1847, opened 2024-02-29
feat: Added section support to Picker

bmingles:
- Picker component now supports <Section> children
- Re-exported Spectrum Item + Section components

Supports deephaven/deephaven-plugins/issues/292


packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -71,10 +77,25 @@ export function Picker({
     [tooltip]
   );
 
+  const renderItem = useCallback(
+    ({ content, textValue }: NormalizedPickerItem) => (
+      <Item textValue={textValue === '' ? 'Empty' : textValue}>
```
mofojed, line 82: Was this `'Empty'` string part of the spec? Seems odd to have, as I'm not seeing anything else to visually distinguish this now empty string from a value of `Empty`.
In addition, we should be passing the `key` into `Item` here - that seems like a bug that we're not.

packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -89,14 +110,21 @@ export function Picker({
...
-      {({ content, textValue }) => (
-        <Item textValue={textValue === '' ? 'Empty' : textValue}>
-          <PickerItemContent>{content}</PickerItemContent>
-          {tooltipOptions == null || content === '' ? null : (
-            <Tooltip options={tooltipOptions}>{content}</Tooltip>
-          )}
-        </Item>
-      )}
+      {itemOrSection => {
+        if ('items' in itemOrSection) {
```
mofojed, line 114: Should write a type predicate for this, e.g. `isPickerSectionItem`

mofojed (changes requested):

packages/components/src/spectrum/picker/Picker.tsx
bmingles: `textValue` is an accessibility thing that isn't visible to the user. It shows warnings if set to empty string. 

`Item` is not actually rendered, and the `key` gets auto determined from the items in the `items` prop. That said, it shouldn't hurt to be explicit, so I can add.

bmingles: @mofojed  More details on how the `textValue` prop is used. It seems to get used as part of `<select><option>` elements that back the Picker. Specifically it seems to provide the content of the `<option>` elements, while the `value` attribute comes from the `key`

image

packages/components/src/spectrum/picker/Picker.tsx
bmingles: done

packages/components/src/spectrum/picker/Picker.tsx
mofojed: Ah my concern was that this would fire with `Empty` on the change event. But if it's firing the event key that's okay. Disregard my comment.

mofojed (approved):

PR #1850, opened 2024-03-05
feat: new chart error panel

### LINKED ISSUE #1520: Downsampling is automatically applied to charts that don't support it

ethanalvizo:
Closes #1520


packages/chart/src/Chart.tsx
```diff
@@ -704,23 +708,22 @@ class Chart extends Component<ChartProps, ChartState> {
...
-          {shownError != null && (
-            <>
-              <div className="chart-error">{shownError}</div>
-              <CopyButton tooltip="Copy Error" copy={shownError}>
-                Copy Error
-              </CopyButton>
-            </>
-          )}
-        </Popper>
+        {downsamplingError != null && (
```
ethanalvizo, line 711: wasn't quite sure what the difference is between downsampling errors and other types of errors. Do we not want them to show in this window as well?

Currently I have them trigger their own overlay warning

ethanalvizo:
image


New error popup^

packages/chart/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,59 @@
...
+              Dismiss
+            </Button>
+          )}
+          {onConfirm && !undismissableError && (
+            <Button onClick={onConfirm} kind="primary">
+              Continue
+            </Button>
+          )}
+        </div>
+        {/* )} */}
```
mofojed, line 53:
```suggestion
```

packages/chart/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,59 @@
...
+    errorMessage === 'Too many items to disable downsampling';
+
+  return (
+    <div className="chart-panel-overlay chart-error-overlay">
+      <div className="chart-panel-overlay-content chart-error-overlay-content">
+        <div className="info-message" data-testid={messageTestId}>
+          {errorMessage}
+          <CopyButton copy={errorMessage} style={{ margin: '0' }} />
+        </div>
+        {/* {!undismissableError && ( */}
```
mofojed, line 35: Remove undismissable commented out code
```suggestion
```

packages/chart/src/ChartErrorOverlay.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onCancel,
+  'data-testid': dataTestId,
+}: ChartErrorOverlayProps): ReactElement {
+  const messageTestId =
+    dataTestId != null ? `${dataTestId}-message` : undefined;
+
+  // TODO: DHC #5220 to detect slow performance and show this message
+
+  const undismissableError =
+    errorMessage === 'Too many items to disable downsampling';
```
mofojed, line 26: Remove this `undismissableError` part.
```suggestion
```

mofojed (changes requested):

mofojed (approved):

mofojed (approved):

PR #1855, opened 2024-03-06
feat: Picker - Item description support

bmingles:
- Normalized keys for item / section now remain undefined if not provided. Same for item `textValue`
- Proper rendering of items with descriptions
- Fixed tooltips for items with descriptions. Also changed default placement to `right`
- isElementOfType util
- Added SpectrumThemeProvider to Popper

Supports deephaven/deephaven-plugins/issues/293

### DIFF
--- a/packages/code-studio/src/styleguide/SamplesMenu.tsx
+++ b/packages/code-studio/src/styleguide/SamplesMenu.tsx
@@ -52,7 +52,7 @@ export function SamplesMenu(): JSX.Element {
     const categories: LinkCategory[] = [currentCategory];
 
     const spectrumComponentsSamples = document.querySelector(
-      `#sample-section-${SPECTRUM_COMPONENT_SAMPLES_ID}`
+      `#${SPECTRUM_COMPONENT_SAMPLES_ID}`
     );
 
     const spectrumComparisonSamples = document.querySelector(
--- a/packages/code-studio/src/styleguide/SamplesMenu.tsx
+++ b/packages/code-studio/src/styleguide/SamplesMenu.tsx
@@ -56,7 +56,7 @@ export function SamplesMenu(): JSX.Element {
     );
 
     const spectrumComparisonSamples = document.querySelector(
-      `#${SPECTRUM_COMPARISON_SAMPLES_ID}`
+      `#sample-section-${SPECTRUM_COMPARISON_SAMPLES_ID}`
     );
 
     document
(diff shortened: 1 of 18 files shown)


packages/components/src/spectrum/picker/PickerItemContent.tsx
```diff
@@ -12,25 +14,57 @@ export interface PickerItemContentProps {
...
+            // content. The index suffix is an extra precation for when 2 <Text>
+            // nodes have the same value.
+            key: item.key ?? `${item.props.children}_${i}`,
+            UNSAFE_className: cl(
+              item.props.UNSAFE_className,
+              stylesCommon.spectrumEllipsis
+            ),
+          })
+        : item
+    );
```
mofojed, line 59: Ah, when we were talking about this, I didn't realize the children we were talking about were _inside_ the picker items - i.e. there's no need for a key.

You should be able to get around this warning by just using `React.Children` instead:
```
    content = React.Children.map(content, (element, i) =>
      isElementOfType(element, Text)
        ? cloneElement(element, {
            ...element.props,
            UNSAFE_className: cl(
              element.props.UNSAFE_className,
              stylesCommon.spectrumEllipsis
            ),
          })
        : element
    );
```

packages/components/src/spectrum/picker/PickerUtils.ts
```diff
@@ -120,11 +120,11 @@ export function isNormalizedPickerSection(
  * @param itemOrSection The picker item or section
  * @returns A `PickerItemKey` for the picker item
  */
-function normalizeItemKey(item: PickerItem): PickerItemKey;
-function normalizeItemKey(section: PickerSection): Key;
+function normalizeItemKey(item: PickerItem): PickerItemKey | undefined;
```
mofojed, line 123: I'd update the doc `@returns` as to when it returns `undefined`, as this is unexpected from the current docs/name.

mofojed (changes requested):

mofojed:
Need to update e2e snapshots as well.

packages/components/src/spectrum/picker/PickerItemContent.tsx
bmingles: Ah nice, glad you knew about this one. It felt gross the way I implemented it.

packages/code-studio/src/styleguide/SamplesMenu.tsx
```diff
@@ -56,7 +56,7 @@ export function SamplesMenu(): JSX.Element {
     );
 
     const spectrumComparisonSamples = document.querySelector(
-      `#${SPECTRUM_COMPARISON_SAMPLES_ID}`
+      `#sample-section-${SPECTRUM_COMPARISON_SAMPLES_ID}`
```
bmingles, line 59: This was a bug causing duplicate ids and the menu would always navigate to the first occurrence

mofojed (approved):

PR #1857, opened 2024-03-07
feat: Add support to pass in mouseHandlers into IrisGrid

mofojed:
- Move the doubleClick behaviour from IrisGridDataSelectMouseHandler into GridSelectionMouseHandler, so it's baked into Grid itself (double-clicking a grid in the styleguide was exhibiting erroneous behaviour)
- Allow passing in mouseHandlers to IrisGrid. Will be necessary to support https://github.com/deephaven/deephaven-plugins/issues/320
- Tested using my branch for `on_row_press` support on deephaven.ui: https://github.com/deephaven/deephaven-plugins/pull/346
- Verified Linker still worked as expected

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -282,7 +282,7 @@ export interface IrisGridProps {
   onCreateChart: (settings: ChartBuilderSettings, model: IrisGridModel) => void;
   onColumnSelected: (column: DhType.Column) => void;
   onError: (error: unknown) => void;
-  onDataSelected: (index: ModelIndex, map: Record<ColumnName, unknown>) => void;
+  onDataSelected: (index: ModelIndex, map: RowDataMap) => void;
   onStateChange: (irisGridState: IrisGridState, gridState: GridState) => void;
   onAdvancedSettingsChange: AdvancedSettingsMenuCallback;
 
@@ -341,6 +341,10 @@ export interface IrisGridProps {
   canToggleSearch: boolean;
 
   columnHeaderGroups?: readonly ColumnHeaderGroup[];
+
+  // Optional key and mouse handlers
+  keyHandlers: readonly KeyHandler[];
+  mouseHandlers: readonly GridMouseHandler[];
 }
 
 export interface IrisGridState {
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/mousehandlers/IrisGridDataSelectMouseHandler.ts
```diff
@@ -12,7 +12,7 @@ import type IrisGrid from '../IrisGrid';
  */
 class IrisGridDataSelectMouseHandler extends GridMouseHandler {
   constructor(irisGrid: IrisGrid) {
-    super();
+    super(825);
```
bmingles, line 15: This could probably use a comment as to what `825` is

bmingles: ah, I see it is `order` now in the base class.

bmingles (changes requested):
Left a question about event handlers

packages/iris-grid/src/mousehandlers/IrisGridDataSelectMouseHandler.ts
mofojed: Yea, we could (and probably should) have constants for a few different levels of this... https://github.com/deephaven/web-client-ui/issues/1859

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1394,6 +1403,26 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+
+  getKeyHandlers(): readonly KeyHandler[] {
+    const { keyHandlers } = this.props;
+    return this.getCachedKeyHandlers(keyHandlers);
+  }
+
+  getCachedMouseHandlers = memoize(
+    (
+      mouseHandlers: readonly GridMouseHandler[]
+    ): readonly GridMouseHandler[] => [...mouseHandlers, ...this.mouseHandlers]
```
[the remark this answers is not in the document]
mofojed: `this.mouseHandlers` and `this.keyHandlers` are not getting by initial props, that's a local value that is initialized in the constructor.
Then the `getCachedMouseHandlers` combines the default/code IrisGrid mouseHandlers with the custom provided `mouseHandlers` provided from the props.
We use a similar pattern in `Grid.tsx` as well.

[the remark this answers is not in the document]
bmingles: ah, misread. Looks good then

bmingles (dismissed):
LGTM

bmingles (approved):
LGTM

Issue #1863, opened 2024-03-10
Cannot use `@deephaven/components` in React v18


mofojed:
In a React v18 project, if you try and install `@deephaven/components`, you get an error:
```sh
✗ npm install @deephaven/components
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: demo-day@0.0.0
npm ERR! Found: react@18.2.0
npm ERR! node_modules/react
npm ERR!   react@"^18.2.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^17.x" from @deephaven/components@0.68.0
npm ERR! node_modules/@deephaven/components
npm ERR!   @deephaven/components@"*" from the root project
```

We should allow support for more versions of React as a peer dependency, like other component libraries such as Spectrum do: https://github.com/adobe/react-spectrum/blob/89fa370bd1bb346be643c9ab2a0d79fab857e4d7/packages/%40react-spectrum/toast/package.json#L57

mofojed:
Should update the peer dependencies for React across all of our packages.

Issue #1493, opened 2023-09-07
Update to React v18


mofojed:
https://react.dev/blog/2022/03/08/react-18-upgrade-guide

Some improvements we can expect:
- Better typings in generally, specifically from `useCallback`
  - Used to be [typed as any in v17](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v17/index.d.ts#L1105), it's [more strict in v18](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1126).

mattrunyon:
More useful info including a codemod to help w/ types that may break w/ the upgrade https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

mattrunyon:
`react-beautiful-dnd` may or may not work w/ React 18. They added it to the peer deps at least, but I vaguely recall some other issues w/ react 18 that people were having. If it doesn't work, we would either need to use a forked version (since it is no longer updated) or switch it out. We use `@dnd-kit/core` for the column grouping sidebar

mofojed:
Moved to DH-19098

PR #1864, opened 2024-03-10
docs: Update the @deephaven/components docs

mofojed:
- Update the docs to actually have a working example

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -49,8 +49,8 @@
     "shortid": "^2.2.16"
   },
   "peerDependencies": {
-    "react": "^17.x",
-    "react-dom": "^17.x"
+    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0",
+    "react-dom": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0"
   },
   "devDependencies": {
     "@deephaven/mocks": "file:../mocks"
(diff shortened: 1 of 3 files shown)


packages/components/package.json
```diff
@@ -49,8 +49,8 @@
     "shortid": "^2.2.16"
   },
   "peerDependencies": {
-    "react": "^17.x",
-    "react-dom": "^17.x"
+    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0",
+    "react-dom": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0"
```
mattrunyon, line 53: Just use `>= 16.8.0`? No need for the 17 release candidates IMO

I use https://semver.npmjs.com/ to check what the actual published versions are that will match my version range

mattrunyon (changes requested):
Should we update all our peer dependencies on react and react-dom?

packages/components/package.json
mofojed: Seems to be what other libs do: https://github.com/atlassian/react-beautiful-dnd/blob/013bfceac04ff48548c33cdc468dd2927446fc1b/package.json#L148
Though I think you're right, doesn't really seem necessary at this point...

mofojed: I'm going to leave this in another change with #1493 and #1863 , I'll just update the docs with this PR.

mattrunyon (approved):

PR #1873, opened 2024-03-15
refactor: Change embed-grid and embed-chart to redirects

mattrunyon:
Publishing an alpha with the `embed-refactor` label so we can test with a local DHC build

This changes embed-grid and embed-chart to just be simple redirects to `/iframe/widget`. This should reduce our footprint in the install size as this is about half of the size we contribute currently.

BREAKING CHANGE: `@deephaven/embed-grid` does not handle messages to the iframe for filtering or sorting the grid any more


packages/embed-chart/.gitignore
```diff
@@ -0,0 +1 @@
+!build
```
mofojed, line 1: I'd add a comment here (or somewhere?) about why we're ignoring this folder and/or why we're putting `index.html` in the `build` folder instead of just at the root.

mofojed (changes requested):
Minor correction with `document.baseURI` instead of `window.location`.
I see you published the alpha, you tried it out as well and it worked yeah?

packages/embed-chart/.gitignore
mattrunyon: Ya I can put it in the gitignore. It's so the build script in core doesn't need to change at all since eventually we'll want to just remove this and change the build script then. Since there's no build any more, just commit the file directly as it will be published

packages/embed-chart/build/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <body>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
+
+    <script>
+      // Needs to be in JS so we can keep the search params
+      const url = new URL(window.location);
+      url.pathname = '/iframe/widget/';
```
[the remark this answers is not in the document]
mattrunyon: Ya I think I need to adjust it to a relative path and add a base tag then. In the existing version, we have `base=./` so `baseURI` is actually `/iframe/chart` or `/iframe/table`

What this doesn't allow (and neither does our existing base) is specifying a base that isn't `./` or rewriting it with a web server. If our base is ever rewritten, the jsapi and iframes will try to load from paths that likely don't exist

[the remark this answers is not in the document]
mattrunyon: Also, if I combine instead of just changing `url.pathname` after, then the URL params get dropped

mattrunyon:
Changed this to just use vite to build embed-grid and embed-chart still. All it does is copy the index into the build folder. This way there's no issues of people accidentally committing stale build artifacts since the folder was being un-ignored with the first change.

Will publish an alpha and test everything is working properly

packages/embed-grid/index.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <base href="./" />
```
mofojed, line 4: This should still be `BASE_URL`.

packages/embed-grid/index.html
```diff
@@ -0,0 +1,16 @@
...
+  </head>
+  <body>
+    <noscript>You need to enable JavaScript to run this app.</noscript>
+
+    <script>
+      // Can't be an HTML meta redirect because we need to keep the search params
+      const url = new URL('../../iframe/widget/', document.baseURI);
```
mofojed, line 11: May as well just do `../widget` ...
```suggestion
      const url = new URL('../widget/', document.baseURI);
```

We could have env vars that you set when building to reference where the new iframe path should be, but this is fine... if you change the path, you'll need to rewrite the path in this index.html file.

packages/embed-chart/build/index.html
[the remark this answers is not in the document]
mofojed: Yea true it assumes we're at a path like `iframe/table` already:
```
BASE_URL=./
# We assume embed-grid is served at nested path, e.g. '/iframe/grid'
VITE_CORE_API_URL=../../jsapi
VITE_MODULE_PLUGINS_URL=../../js-plugins
```

Devin has rewriting the base before to have two Deephaven instances running on the same host; but that's beside the point.

Also good point about dropping params.

packages/embed-grid/index.html
mattrunyon: I was just trying to make this simple (using Vite default config). I don't see any reason `BASE_URL` needs to be easily changeable at build time as we don't support anything other than `./`

packages/embed-grid/index.html
mattrunyon: Ya I figured these were simple enough redirects that adding an env file just adds an unnecessary layer of pass through. Not any gain IMO when we're replacing a single use value with a single use variable instead and we're ultimately trying to remove these separate embed packages

mattrunyon:
Just tested again w/ the latest commit alpha and `./gradlew prepareCompose` and the redirect works as expected

packages/embed-grid/index.html
mofojed: Yea fair

mofojed (approved):

PR #1876, opened 2024-03-18
feat: Picker - Table support for key + label columns

### LINKED ISSUE #1858: Picker - table support
In support of deephaven/deephaven-plugins/issues/293, Picker components in DHC web need to support tables +  viewporting.

bmingles:
Initial table support for Picker
- key and value column support
- Scroll to selected item when popover opens
- Viewport data now supports re-use of existing item objects via `reuseItemsOnTableResize` flag

### Scroll to item fix for non-table data (this should work without any changes to plugins repo)

```python
import deephaven.ui as ui
from deephaven.ui import use_state

items = list("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")

@ui.component
def picker():
    value, set_value = use_state("M")

    # Picker for selecting values
    pick = ui.picker(
        label="Text",
        on_selection_change=set_value,
        selected_key=value,
        children=items
    )

    # Show current selection in a ui.text component
    text = ui.text("Selection: " + value)

    # Display picker and output in a flex column
    return ui.flex(
        pick,
        text,
        direction="column",
        margin=10,
        gap=10,
    )


p = picker()
```

### Table support
There is a draft PR in the plugins repo that enables the table support on the plugins side of things (note that it currently will have type errors but should run)
https://github.com/deephaven/deephaven-plugins/pull/382

Here's an example that demonstrates a large number of items that also tick

```python
import deephaven.ui as ui
from deephaven.ui import use_state
import datetime

mylist = ["mars", "pluto", "saturn"]

simple_ticking = time_table("PT2S", start_time=datetime.datetime.now() - datetime.timedelta(seconds=2000)).update([
    'Id=(new Integer(i * 100))',
    "Planet=mylist[ (int) (3 * Math.random()) % 3 ]",
])


@ui.component
def picker():
    value, set_value = use_state(13800)

    print("Test", value)

    # Picker for selecting values
    pick = ui.picker(
        simple_ticking,
        key_column="Id",
        label_column="Planet",
        label="Text",
        on_selection_change=set_value,
        selected_key=value,
    )

    # Show current selection in a ui.text component
    text = ui.text("Selection: " + str(value))

    # Display picker and output in a flex column
    return ui.flex(
        pick,
        text,
        direction="column",
        margin=10,
        gap=10,
    )


picker_table = picker()
```

resolves #1858


packages/jsapi-components/src/spectrum/Picker/Picker.tsx
```diff
@@ -0,0 +1,71 @@
...
+    // the real `key` will be available to show the selection in the UI.
+    function setViewportFromSelectedKey() {
+      if (selectedKey == null) {
+        return;
+      }
+
+      const keyColumn = getPickerKeyColumn(table, keyColumnName);
+
+      table
+        .seekRow(0, keyColumn, keyColumn.type, selectedKey)
```
bmingles, line 55: Looks like passing `keyColumn.type` directly here won't always work. I'm getting a `java.lang.UnsupportedOperationException: Invalid value type for seekRow: java.lang.String` error in certain scenarios. Will have to see about mapping this to expected type.

packages/jsapi-components/src/useGetItemIndexByValue.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (table == null || value == null) {
+      return null;
+    }
+
+    const column = table.findColumn(columnName);
+    const columnValueType = tableUtils.getValueType(column.type);
+
+    const index = await table.seekRow(0, column, columnValueType, value);
+
+    return index === -1 ? null : index;
```
bmingles, line 35: If I'm understanding correctly, the `-1` should be returned for non-matches once deephaven/deephaven-core/pull/5238 is merged. For now, the non-match scenario will return an actual row index. In cases where a table ticks causing a previous selection to drop from the table, this will cause scrolling to a seemingly arbitrary position, but this should be enough of an edge case to not be too worried about it since it is temporary

packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -1,36 +1,53 @@
...
   /**
    * Handler that is called when the selection change.
    * Note that under the hood, this is just an alias for Spectrum's
    * `onSelectionChange`. We are renaming for better consistency with other
    * components.
    */
   onChange?: (key: PickerItemKey) => void;
+
+  /** Handler that is called when the popover is scrolled. */
+  onScroll?: (event: Event) => void;
```
mofojed, line 49:
```suggestion
  /** Handler that is called when the picker is scrolled. */
  onScroll?: (event: Event) => void;
```

packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -82,7 +103,10 @@ export function Picker({
   tooltip = true,
   defaultSelectedKey,
   selectedKey,
+  getInitialScrollPosition,
   onChange,
+  onOpenChange,
+  onScroll = noOp,
```
mofojed, line 109:
```suggestion
  onScroll = EMPTY_FUNCTION,
```

packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -70,6 +87,10 @@ function createTooltipContent(content: ReactNode) {
   return content;
 }
 
+function noOp(): void {
+  // No-op
+}
+
```
mofojed, line 93: Use `EMPTY_FUNCTION` from utils instead of redefining.

packages/jsapi-components/src/spectrum/Picker/Picker.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { getPickerKeyColumn } from './PickerUtils';
+import { usePickerItemRowDeserializer } from './usePickerItemRowDeserializer';
+
+export interface PickerProps extends Omit<PickerPropsBase, 'children'> {
+  table: DhType.Table;
+  /* The column of values to use as item keys. Defaults to the first column. */
+  keyColumn?: string;
+  /* The column of values to display as primary text. Defaults to the `keyColumn` value. */
+  labelColumn?: string;
+}
```
mofojed, line 22: // TODO #1890 : descriptionColumn, iconColumn

packages/jsapi-components/src/spectrum/Picker/Picker.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(
+    // Set viewport to include the selected item so that its data will load and
+    // the real `key` will be available to show the selection in the UI.
+    function setViewportFromSelectedKey() {
+      getItemIndexByValue().then(index => {
+        if (index != null) {
+          setViewport(index);
+        }
+      });
+    },
```
mofojed, line 85: We should be cancelling this effect in a cleanup and checking that `isCancelled` before calling `setViewport` again. Otherwise could have a race (in theory).

mofojed (changes requested):
Great stuff! Some little cleanup.
As you mentioned in the ticket, it does not handle `description_column` or `icon_column` - I feel like those should be pretty easy to wire up, just need to include it in normalized items. I added a comment on #1890 to include `icon_column` as well. Using a snippet like:
```python
import deephaven.ui as ui
from deephaven.ui import use_state
import datetime

mylist = ["mars", "pluto", "saturn"]
colors = ["salmon", "lemonchiffon", "purple", "green", "red"]
icons = ["vsEdit", "vsTrash"]

simple_ticking = time_table("PT2S", start_time=datetime.datetime.now() - datetime.timedelta(seconds=2000)).update([
    'Id=(new Integer(i * 100))',
    "Planet=(String)mylist[ (int) (3 * Math.random()) % 3 ]",
    "Color=(String)colors[ (int) (5 * Math.random()) % 5 ]",
    "Icon=(String)icons[ (int) (2 * Math.random()) % 2 ]"
])


@ui.component
def picker():
    value, set_value = use_state(13800)

    print("Test", value)

    # Picker for selecting values
    pick = ui.picker(
        simple_ticking,
        key_column="Id",
        label_column="Planet",
        description_column="fdsafdsa",
        icon_column="Icon",
        label="Text",
        on_selection_change=set_value,
        selected_key=value,
    )

    # Show current selection in a ui.text component
    text = ui.text("Selection: " + str(value))

    # Display picker and output in a flex column
    return ui.flex(
        pick,
        text,
        direction="column",
        margin=10,
        gap=10,
    )


picker_table = picker()
```

That can be after this merges though.

packages/jsapi-components/src/spectrum/Picker/Picker.tsx
```diff
@@ -77,11 +79,19 @@ export function Picker({
...
+
       getItemIndexByValue().then(index => {
-        if (index != null) {
-          setViewport(index);
+        if (index == null || isCanceled) {
+          return;
         }
+
+        setViewport(index);
       });
```
mofojed, line 90: Should we also be catching here? Otherwise we could have an uncaught promise. Probably just want to log the error anyway I guess.

mofojed (approved):

packages/jsapi-components/src/spectrum/Picker/Picker.tsx
bmingles: I went ahead and added an explicit catch / log

packages/jsapi-components/src/spectrum/Picker/Picker.tsx
```diff
@@ -6,13 +6,16 @@ import {
...
 import { Formatter } from '@deephaven/jsapi-utils';
+import Log from '@deephaven/log';
 import { PICKER_ITEM_HEIGHT, PICKER_TOP_OFFSET } from '@deephaven/utils';
 import { useCallback, useEffect, useMemo } from 'react';
 import useGetItemIndexByValue from '../../useGetItemIndexByValue';
 import { useViewportData } from '../../useViewportData';
 import { getPickerKeyColumn } from './PickerUtils';
 import { usePickerItemRowDeserializer } from './usePickerItemRowDeserializer';
 
+const log = Log.module('Picker');
```
mofojed, line 17: One problem with log API is it won't disambiguate between any other `'Picker'` components. Could ensure it's "unique" by prefixing it with `jsapi-components`... doesn't matter much to me.

mofojed (approved):

mofojed (approved):

Issue #1245, opened 2023-04-24
Stop exporting client from support logs


mattrunyon:
We currently export the entire redux store in support logs. This includes the client which can become quite large in enterprise support logs. I saw a 200MB redux store today and it was 99% a massive client object. Something in the client (iterator of some sort or linked list) tracks the prev item and ends up with massively nested objects when trying to stringify

Something in client.configs has a deeply nested structure when stringifying

The client object probably isn't particularly useful in the support logs, so stop exporting it and only export the parts of the redux store which might be useful

`LogExport.ts` should probably have a blacklist of redux keys for the `getReduxDataString` function

mattrunyon:
We only set the client in redux in enterprise apparently, but we store connection and sessionWrapper in a few places in DHC. We should check if those are needed or can balloon in size and not export them if that's the case

I see `dashboardData.defaultLayout.connection` and `layoutStorage.storageService.connection` at least. Not sure if those are useful or should be exported

PR #1881, opened 2024-03-19
feat: log export blacklist

wusteven815:
- Adds #1245 
  - Add a replacer function to all `JSON.stringify` to filter out blacklisted path
  - Add the following paths to blacklist: `api`, `client`, `dashboardData.defaultLayout.connection`, `layoutStorage`, `storage`

### DIFF
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -10,8 +10,10 @@ const FILENAME_DATE_FORMAT = 'yyyy-MM-dd-HHmmss';
 const KEY_BLACKLIST: string[] = ['client'];
 // Blacklist specific paths (e.g. blacklist foo will only blacklist foo but NOT a.foo or b.foo)
 const PATH_BLACKLIST: string[] = [
+  'api',
   'dashboardData.defaultLayout.connection',
-  'layoutStorage.storageService.connection',
+  'layoutStorage',
+  'storage',
 ];
 
 /**
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -31,6 +31,8 @@ function replacerWithPath(
   r: (key: string, value: unknown, path: string) => unknown
 ) {
   const m = new Map();
+  // code source and explanation;
+  // https://stackoverflow.com/questions/61681176/json-stringify-replacer-how-to-get-full-path
   return function replacer(this: unknown, field: string, value: unknown) {
     const path =
       m.get(this) + (Array.isArray(this) ? `[${field}]` : `.${field}`);
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -36,13 +36,11 @@ function stringifyReplacer(blacklist: string[][]) {
         currPath.every((v, index) => v === blacklist[i][index])
       ) {
         // blacklist match
-        console.log('Blacklist match', currPath);
         return undefined;
       }
     }
 
     // not in blacklist, return value
-    console.log('No blacklist match', currPath);
     return value;
   };
 }
--- a/packages/code-studio/src/log/LogExport.ts
+++ b/packages/code-studio/src/log/LogExport.ts
@@ -9,12 +9,12 @@ const FILENAME_DATE_FORMAT = 'yyyy-MM-dd-HHmmss';
 // List of objects to blacklist
 // '' represents the root object
 export const PATH_BLACKLIST: string[][] = [
-  ['', 'api'],
-  ['', 'client'],
-  ['', 'dashboardData', 'defaultLayout', 'connection'],
-  ['', 'layoutStorage'],
-  ['', 'storage'],
-];
+  ['api'],
+  ['client'],
+  ['dashboardData', 'defaultLayout', 'connection'],
+  ['layoutStorage'],
+  ['storage'],
+].map(path => ['', ...path]);
 
 function stringifyReplacer(blacklist: string[][]) {
   // modified from:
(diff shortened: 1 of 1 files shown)


mofojed:
@wusteven815 yes we should blacklist connection as well.

mattrunyon:
We should blacklist these top level keys as well as they're JS API items and don't seem to add anything useful from what I can tell

- api
- storage
- layoutStorage

I think client is stored in redux in enterprise only as I don't see it in the exported logs on latest

packages/code-studio/src/log/LogExport.ts
```diff
@@ -6,6 +6,48 @@ import { logHistory } from './LogInit';
 
 const FILENAME_DATE_FORMAT = 'yyyy-MM-dd-HHmmss';
 
+// Blacklist applies to all keys (e.g. blacklist foo will blacklist foo, a.foo, and b.foo)
+const KEY_BLACKLIST: string[] = ['client'];
```
mofojed, line 10: I don't think we want a `KEY_BLACKLIST` at all, certainly not with something as general as `client`. If it was something like `__keyValue` or a key that is clearly intended to be private, perhaps, but this blacklisting is a bit broad I think. Was there a reason to do a KEY blacklist instead of just by path?

mofojed (changes requested):
We should be using JSON pointers. Also allow passing in a blacklist for Enterprise to use.

Here's an example exported from Enterprise for your testing benefit:
[redux.json](https://github.com/deephaven/web-client-ui/files/14774937/redux.json)

`client` is at the root, so no need to blacklist everything.

packages/code-studio/src/log/LogExport.ts
```diff
@@ -89,13 +130,14 @@ export async function exportLogs(
     FILENAME_DATE_FORMAT,
     new Date()
   )}_support_logs`,
-  metadata?: Record<string, unknown>
+  metadata?: Record<string, unknown>,
+  blacklist: string[][] = PATH_BLACKLIST
```
mofojed, line 134: We should add JSDocs for these parameters.

packages/code-studio/src/log/LogExport.ts
```diff
@@ -6,6 +6,42 @@ import { logHistory } from './LogInit';
...
 const FILENAME_DATE_FORMAT = 'yyyy-MM-dd-HHmmss';
 
+// List of objects to blacklist
+// '' represents the root object
+export const PATH_BLACKLIST: string[][] = [
+  ['', 'api'],
+  ['', 'client'],
+  ['', 'dashboardData', 'defaultLayout', 'connection'],
+  ['', 'layoutStorage'],
+  ['', 'storage'],
```
mofojed, line 16: Any way we can define it so we don't have the extra `''` in front? Annoying to have to define that for every item, and it doesn't help for any valid case (blacklisting the root wouldn't make sense, and even if it did it could be `[]` or `['']`).

mofojed (changes requested):

packages/code-studio/src/log/LogExport.ts
```diff
@@ -6,6 +6,42 @@ import { logHistory } from './LogInit';
...
 
+// List of objects to blacklist
+// '' represents the root object
+export const PATH_BLACKLIST: string[][] = [
+  ['api'],
+  ['client'],
+  ['dashboardData', 'defaultLayout', 'connection'],
+  ['layoutStorage'],
+  ['storage'],
+].map(path => ['', ...path]);
```
mofojed, line 17: This isn't good either - you'd need to do the same thing when passing in your own blacklist. If I'm using the `exportLogs` function below, calling `exportLogs('myFilePrefix', {}, ['client'])`, I'd expect the path `client` to not appear in the exported logs. Requiring the consumer of the API to pass in `['', 'client']` to achieve that is a [footgun](https://en.wiktionary.org/wiki/footgun).

I'd remove the `''` string from the current path when you compare it to the blacklist. Why is there an empty string at the front anyway?

packages/code-studio/src/log/LogExport.ts
```diff
@@ -61,41 +100,46 @@ function makeSafeToStringify(
...
   };
 
-  return JSON.stringify(metadata, null, 2);
+  return JSON.stringify(metadata, stringifyReplacer(blacklist), 2);
 }
 
 /**
  * Export support logs with the given name.
  * @param fileNamePrefix The zip file name without the .zip extension. Ex: test will be saved as test.zip
+ * @param blacklist List of JSON paths to blacklist. A JSON path is a list representing the path to that value (e.g. client.data would be `['client', 'data']`)
```
mofojed, line 128: Add the JSDoc for `metadata` param as well (it was already missing).
It's just any extra metadata you want to export along with the archive, will be saved as a `metadata.json` file in the archive.

Should also add a `@returns` here - returns a promise that resolves successfully if the log archive is created and downloaded successfully, rejected if there's an error.

mofojed (changes requested):

packages/code-studio/src/log/LogExport.ts
wusteven815: The empty string represents the key of the initial object

mofojed:
@wusteven815 yes it is `dashboardData.default.connection`. Also `storage.fileStorage.storageService.connection`, `dashboardData.default.sessionWrapper.dh` should be blacklisted as well.

mofojed (dismissed):
Looks good. Should blacklist `dashboardData.default.connection` and `dashboardData.default.sessionWrapper.dh` as well.

mattrunyon (commented):
We should move this into one of the util packages so DHE can use it with its own blacklist, but that can be a separate PR.

PR #1883, opened 2024-03-20
fix: Add isInvalid prop to Select component

### LINKED ISSUE #1882: isInvalid prop is missing in the Select component
Enterprise uses `isInvalid` prop on input components to set an invalid state. Most inputs have this prop, except the `Select` component.

Invalid selects display a warning in the browser console:
```
Warning: React does not recognize the `isInvalid` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isinvalid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
at select
```

vbabich:
FIxes #1882


dsmmcken (changes requested):
Please add an example in the styleguide after `<Select>` showing a `<Select isInvalid>`

mofojed (commented):
Looks good, just add to styleguide

mofojed (approved):
Annoying we had to update all those snapshots...

Issue #1865, opened 2024-03-11
Improve loading state while seeking table


mofojed:
When scrolling with a large table from a slow backend (or has slow to compute rows), if the data takes a long time to come in, the grid just appears blank and there is no loading indicator. We should have some sort of subtle loading indicator to show there is still a request in progress, "Waiting for viewport...".

Suggested fix:
Display a status message (the `loadingtext` at bottom of table) with a delay of ~0.5s before showing. We don’t want the loader to show in most scenarios, because it’s normally snappy. Show the text “Waiting for viewport…” with the back and forth loader.

![image](https://github.com/deephaven/web-client-ui/assets/1576283/968bfc65-89de-4705-82f6-b8864bfc3c49)

Example query (slow when scrolling because of sleep):
```py
from deephaven import empty_table
import time

def f(a):
    time.sleep(0.1)
    return a

t = empty_table(1000).update("X = i").update_view("Y= f(X)")
```

PR #1898, opened 2024-03-25
feat: improve table loading

wusteven815:
- Adds #1865
  - Add a check for if there is still data being loaded in the viewport
  - Add a new loading message if the above is true for >500ms
  - Add state to determine whether `startLoading` will block the grid or show the cancel button

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -610,7 +610,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       this.handleGotoValueSelectedFilterChanged.bind(this);
     this.handleGotoValueChanged = this.handleGotoValueChanged.bind(this);
     this.handleGotoValueSubmitted = this.handleGotoValueSubmitted.bind(this);
-    this.onViewportUpdate = this.onViewportUpdate.bind(this);
+    this.handleViewportUpdate = this.handleViewportUpdate.bind(this);
     this.makeQuickFilter = this.makeQuickFilter.bind(this);
 
     this.grid = null;
@@ -2513,7 +2513,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     onError(error);
   }
 
-  onViewportUpdate(): void {
+  handleViewportUpdate(): void {
     const { model } = this.props;
     // pending and no timer already exists
     if (model.isViewportPending && this.viewportLoadingTimeout === null) {
@@ -4754,7 +4754,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
                 frozenColumns={frozenColumns}
                 columnHeaderGroups={columnHeaderGroups}
                 partitionConfig={partitionConfig}
-                onViewportUpdate={this.onViewportUpdate}
+                onViewportUpdate={this.handleViewportUpdate}
               />
             )}
             {!isMenuShown && (
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -382,6 +382,7 @@ export interface IrisGridState {
   loadingScrimProgress: number | null;
   loadingSpinnerShown: boolean;
   loadingCancelShown: boolean;
+  loadingBlocksGrid: boolean;
 
   movedColumns: readonly MoveOperation[];
   movedRows: readonly MoveOperation[];
@@ -582,6 +583,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.handleSelectDistinctChanged =
       this.handleSelectDistinctChanged.bind(this);
     this.handlePendingDataUpdated = this.handlePendingDataUpdated.bind(this);
+    this.handleViewportUpdated = this.handleViewportUpdated.bind(this);
     this.handlePendingCommitClicked =
       this.handlePendingCommitClicked.bind(this);
     this.handlePendingDiscardClicked =
@@ -611,7 +613,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       this.handleGotoValueSelectedFilterChanged.bind(this);
     this.handleGotoValueChanged = this.handleGotoValueChanged.bind(this);
     this.handleGotoValueSubmitted = this.handleGotoValueSubmitted.bind(this);
-    this.handleViewportUpdate = this.handleViewportUpdate.bind(this);
+    this.handleViewportUpdated = this.handleViewportUpdated.bind(this);
     this.makeQuickFilter = this.makeQuickFilter.bind(this);
 
     this.grid = null;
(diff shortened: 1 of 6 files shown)


wusteven815:
Didn't use a new event, wired it up to `IrisGridModelUpdater`'s `updateViewport` instead

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -608,6 +608,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       this.handleGotoValueSelectedFilterChanged.bind(this);
     this.handleGotoValueChanged = this.handleGotoValueChanged.bind(this);
     this.handleGotoValueSubmitted = this.handleGotoValueSubmitted.bind(this);
+    this.onViewportUpdate = this.onViewportUpdate.bind(this);
```
mofojed, line 611: Name this `handleViewportUpdate` instead. That's the naming convention that React tends to use in their docs: https://react.dev/learn/responding-to-events
`on*` for the prop name, and then `handle*` for the handler's name.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -484,6 +484,13 @@ abstract class IrisGridModel<
    */
   abstract commitPending(): Promise<void>;
 
+  /**
+   * Check if viewport is still loading data
+   */
+  get hasPendingOperations(): boolean {
+    return false;
+  }
+
```
mofojed, line 493: I know I said to name it `hasPendingOperations`, but this is really viewport specific - I'd rename it `isViewportPending` or `isViewportLoading` instead.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2504,6 +2508,20 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    const { model } = this.props;
+    // pending and no timer already exists
+    if (model.hasPendingOperations && this.pendingOperationsTimeout === null) {
+      this.pendingOperationsTimeout = setTimeout(() => {
+        // still pending after timeout
+        if (model.hasPendingOperations) {
+          this.startLoading('Waiting for viewport...');
+        }
+        this.pendingOperationsTimeout = null;
+      }, 500);
```
mofojed, line 2521: Set this value as a `const` at the top with the other constants.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2504,6 +2508,20 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     onError(error);
   }
 
+  onViewportUpdate(): void {
+    const { model } = this.props;
+    // pending and no timer already exists
+    if (model.hasPendingOperations && this.pendingOperationsTimeout === null) {
+      this.pendingOperationsTimeout = setTimeout(() => {
```
mofojed, line 2515: You'll need to clear this timeout on unmount as well. Similar to my other suggestion, I'd suggest a name `viewportLoadingTimeout` instead.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -455,6 +455,21 @@ class IrisGridTableModelTemplate<
...
+
+    // offset + row.length is last row of loaded data
+    // offset is first row of loaded data
+    // last row of loaded data < last row of viewport
+    // first row of viewport < first row of loaded data
+    return (
+      this.viewportData.offset + this.viewportData.rows.length <
+        this.viewport.bottom || this.viewport.top < this.viewportData.offset
+    );
+  }
```
mofojed, line 471: We should also check that the columns requested match the columns in the viewport for horizontal scrolling with a large table. 

A table with 100 columns: 
```
import time

def f(a):
    time.sleep(0.1)
    return a

t = empty_table(100).update([f"X{c} = i" for c in range(100)]).update_view("Y = f(i)")
```

mofojed (changes requested):
`iris-grid-loading` element that gets loaded blocks all user interaction while scrolling... also, the `Cancel` button doesn't make sense in this context. It will try and rollback the state of filters etc. which doesn't make sense, and you can't really "cancel" a viewport operation.

Instead, I don't think we should show the cancel button, and we shouldn't stop the user from interacting/scrolling the table either. We just want the loading status in the bottom. So we may need to add another loading state value or something.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -75,6 +75,7 @@ abstract class IrisGridModel<
     RECONNECT: 'RECONNECT',
     TOTALS_UPDATED: 'TOTALS_UPDATED',
     PENDING_DATA_UPDATED: 'PENDING_DATA_UPDATED',
+    VIEWPORT_UPDATED: 'VIEWPORT_UPDATED',
```
mofojed, line 78: Let's add some JSDocs for this event. In retrospect, we should have them for all of the events, but we can at least be proactive and add some here to make it clear what the event is for. 

When writing good docs for a function, you want to address possible questions a future reader for have; think about what kind of questions a reader of this code (or user of the event) may have. 

For example, is this fired when the set viewport in the model is changed, when the viewport is applied to the table, or when the viewport data has come back? Is there any data included with the event? Another common question is "Why?", as in "Why do we need this event?"
```suggestion

    /** Fired when the viewport is applied to the table and we're waiting for a response. */
    VIEWPORT_UPDATED: 'VIEWPORT_UPDATED',
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2504,6 +2534,29 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+            loadingCancelShown: false,
+            loadingBlocksGrid: false,
+          });
+        }
+        this.viewportLoadingTimeout = null;
+      }, VIEWPORT_LOADING_DELAY);
+    } else if (loadingText === loadingMessage && !model.isViewportPending) {
+      // extra conditions because timeout might get cleared by update
+      this.stopLoading();
+    }
```
mofojed, line 2557: What you're doing here is throttling the function. This is a pretty common pattern. You can see `handleDownloadProgressUpdate` in this file (or `IrisGridTableModelTemplate.applyViewport`) for an example of a throttled function. But you could rewrite it like this:

```
componentWillUnmount() {
  ...
  // You at least need to stop the timer, you're not doing that currently.
  this.showViewportLoading.cancel();
  ...
}

showViewportLoading = throttle(
  (): void => {
    const { model } = this.props;
    const { loadingSpinnerShown } = this.state;
    if (model.isViewportPending && !loadingSpinnerShown) {
      // We only want to show the viewport loading if the viewport is still loading and we're not already showing a loader for something else
      this.startLoading("Waiting for viewport...", {
        loadingCancelShown: false,
        loadingBlocksGrid: false,
      });
    }
  },
  VIEWPORT_LOADING_DELAY
);


handleViewportUpdated(): void {
  this.showViewportLoading();
}

handleUpdate() {
  ...
  this.showViewportLoading.cancel();
  ...
}
```

mofojed (changes requested):
Just some cleanup/suggestions.

If we wanted to get _really_ fancy, we'd keep a list of the pending operations and their type, then decide what loading text/spinner to display at render time based on all the current pending operations. In theory, this could also help us with our rollback before (right now we just rollback to the last loaded state). I think that's all out of scope for right now though, this is good.

mofojed (approved):
Great work!

PR #1903, opened 2024-03-25
feat: wrap spectrum View, Text and Heading to accept custom colors

dsmmcken:
- Exposes our extra semantic colors to these components
- Created colorUtils in theme, containing theme variables
- Move spectrumUtils to spectrum/utils (was originally going to put it here)
- Added eslint config to allow using UNSAFE_styles/UNSAFE_classname

Tested with:

```html
<Text color="red">css red</Text>
<Text color="red-1000">theme red-1000</Text>
<Text color="accent">accent</Text>
<Text color="positive">positive</Text>
<Text color="negative">negative</Text>
<Text color="info">info</Text>
<Text color="notice">notice</Text>
<Text color="bg">bg</Text>
<Text color="color-mix(in srgb, red, blue)">purple</Text>
```


packages/components/src/spectrum/Heading.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * @returns The Heading component
+ *
+ */
+
+export function Heading(props: HeadingProps): JSX.Element {
+  const { color, UNSAFE_style, ...rest } = props;
+  if (color != null) {
+    return (
+      <SpectrumHeading
+        {...(rest as SpectrumHeadingProps)}
```
bmingles, line 27: No need for the type assertion

packages/components/src/spectrum/Heading.tsx
```diff
@@ -0,0 +1,39 @@
...
+        UNSAFE_style={{
+          ...UNSAFE_style,
+          color: colorValueStyle(color),
+        }}
+      />
+    );
+  }
+
+  return <SpectrumHeading {...(props as SpectrumHeadingProps)} />;
```
bmingles, line 36: No need for the type assertion

packages/components/src/theme/colorUtils.ts
```diff
@@ -0,0 +1,250 @@
...
+  'accent-1300',
+  'accent-1400',
+  'bg',
+  'content-bg',
+  'subdued-content-bg',
+  'surface-bg',
+  'fg',
+] as const;
+
+export type ColorValue =
```
bmingles, line 227: Suggested change:

```typescript
type DHColorValue = (typeof ColorValues)[number];

export type ColorValue =
  | DHColorValue
  // This allows autocomplete to work properly and not collapse the above options into just `string`.
  // See https://github.com/microsoft/TypeScript/issues/29729.
  | CSSProperties['color'];

/** Typeguard to encapsulate the type assertion */
export function isDHColorValue(value: unknown): value is DHColorValue {
  return (
    typeof value === 'string' && ColorValues.includes(value as DHColorValue)
  );
}

export function colorValueStyle(value: string): string {
  if (isDHColorValue(value)) {
    return `var(--dh-color-${value})`;
  }

  return value;
}
```

packages/eslint-config/index.js
```diff
@@ -54,6 +54,13 @@ module.exports = {
     'react/require-default-props': 'off',
     'react/jsx-no-bind': 'off',
     'react-refresh/only-export-components': 'warn',
+    camelcase: [
+      'error',
+      {
+        allow: ['^UNSAFE_'], // allow UNSAFE_styles UNSAFE_className
```
bmingles, line 60: nice. Glad you thought of this.

packages/components/src/theme/colorUtils.test.ts
```diff
@@ -0,0 +1,14 @@
+import { colorValueStyle } from './colorUtils';
+
+describe('ColorValues', () => {
+  test('should return the correct color style', () => {
+    // dh-color variables
+    expect(colorValueStyle('blue-1000')).toBe('var(--dh-color-blue-1000)');
```
bmingles, line 6: This is a good case for `it.each` since you have an array of all possible values, easy to incorporate

packages/components/src/spectrum/Text.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * @returns The Text component
+ *
+ */
+
+export function Text(props: TextProps): JSX.Element {
+  const { color, UNSAFE_style, ...rest } = props;
+  if (color != null) {
+    return (
+      <SpectrumText
+        {...(rest as SpectrumTextProps)}
```
bmingles, line 27: No need for the type assertion

packages/components/src/spectrum/Text.tsx
```diff
@@ -0,0 +1,39 @@
...
+        UNSAFE_style={{
+          ...UNSAFE_style,
+          color: colorValueStyle(color),
+        }}
+      />
+    );
+  }
+
+  return <SpectrumText {...(props as SpectrumTextProps)} />;
```
bmingles, line 36: No need for the type assertion

packages/components/src/spectrum/View.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * @returns The View component
+ *
+ */
+
+export function View(props: ViewProps): JSX.Element {
+  const { backgroundColor, UNSAFE_style, ...rest } = props;
+  if (backgroundColor != null) {
+    return (
+      <SpectrumView
+        {...(rest as SpectrumViewProps<6>)}
```
bmingles, line 27: No need for the type assertion

bmingles:
General idea looks good. I think you may need to memoize the converted props since the `UNSAFE_style` prop gets re-created on every render. Possible Spectrum handles this gracefully, but IMO it's good to not assume that will be the case.

I think you could incorporate a general util here. Something like:

```typescript
export function wrapColorProps<
  TProps extends Record<string, unknown> & {
    UNSAFE_style?: React.CSSProperties;
  },
>(props: TProps): Omit<TProps, 'backgroundColor' | 'color'> {
  const backgroundColor = props.backgroundColor as string | undefined;
  const color = props.color as string | undefined;

  if (backgroundColor == null && color == null) {
    return props;
  }

  return {
    ...props,
    UNSAFE_style: {
      ...props.UNSAFE_style,
      backgroundColor:
        backgroundColor == null ? undefined : colorValueStyle(backgroundColor),
      color: color == null ? undefined : colorValueStyle(color),
    },
  };
}
```

And then consuming it could be like:

```jsx
const colorProps = useMemo(() => wrapColorProps(props), [props]);
return <SpectrumView {...colorProps} />;
```

Alternatively, we could make a custom hook like `useWrappedColorProps` to include the memoization.

bmingles (changes requested):
Left some comments. The suggestion for `wrapColorProps` function may impact the other suggestions, so I would look at that comment first to see what you think.

packages/components/src/theme/colorUtils.test.ts
dsmmcken: Meh, this is less code. If I was doing more in each check or setup sure.

dsmmcken:
Kept it simple. Just did a useMemo for style, and allowed colorValueStyle to return undefined. Didn't seem worth a separate abstraction.

bmingles (dismissed):
LGTM

dsmmcken:
I didn't do an exhaustive search, but replaced some usages. We should add a ` no-restricted-modules` eslint rule later to not allow importing from @adobe/spectrum from outside of the components/spectrum and theme/ folders.

bmingles (approved):
LGTM

PR #1906, opened 2024-03-27
feat: Picker - Get formatter settings from context

### LINKED ISSUE #1889: Picker - handle JS API column types in display labels
#1858 is implementing Table support in the Picker. Handling of data types is pretty naive at the moment and will need to be able to handle additional column types (DateWrapper, Long, etc.)

Current implementation looks like this:
```typescript
function formatValue(value: unknown): string | number | boolean {
  if (
    typeof value === 'string' ||
    typeof value === 'number' ||
    typeof value === 'boolean'
  ) {
    return value;
  }

  // TODO: Add better support for different data types #1889
  return String(value);
}
```

bmingles:
Likely going to replace this PR with simpler version https://github.com/deephaven/web-client-ui/pull/1907
resolves #1889


bmingles:
Replaced by deephaven/web-client-ui/pull/1907

PR #1907, opened 2024-03-28
feat: Picker - formatter settings

### LINKED ISSUE #1889: Picker - handle JS API column types in display labels

bmingles:
- useFormatter hook + optional settings prop in jsapi Picker
- Fixed a bug with `bindAllMethods` function. It now excludes getters

resolves #1889


packages/jsapi-components/src/useGetItemIndexByValue.ts
```diff
@@ -30,9 +33,13 @@ export function useGetItemIndexByValue<TValue>({
...
 
-    const index = await table.seekRow(0, column, columnValueType, value);
-
-    return index === -1 ? null : index;
+    try {
+      const index = await table.seekRow(0, column, columnValueType, value);
+      return index === -1 ? null : index;
+    } catch (err) {
+      log.debug('Error seeking row', { column, value, columnValueType });
+      return null;
```
mofojed, line 41: Any reason not to still throw? Consumer should be handling promise rejection anyway, no?

bmingles: fixed

packages/jsapi-components/src/useGetItemIndexByValue.ts
```diff
@@ -38,7 +38,7 @@ export function useGetItemIndexByValue<TValue>({
       return index === -1 ? null : index;
     } catch (err) {
       log.debug('Error seeking row', { column, value, columnValueType });
-      return null;
+      throw err;
```
mofojed, line 41: I wouldn't bother logging it either - consumer either catches it and logs it or does something with it in which case the log would be extra spam. If it's uncaught would be logged as an uncaught promise anyway.

mofojed (approved):

mofojed (approved):

Issue #1785, opened 2024-02-08
Allow re-opening a closed tab panel


dsmmcken:
As a user, I want to be able to re-open a closed panel so that I can restore with it's state if I accidently close a tab.

Suggested changes:
- Remember the last 15 (arbitrary number?) closed panels PER dashboard, their location and their the state when closed. This can be ephemeral between sessions, doesn't need to be persisted.
- Add a configurable shortcut (win: alt+shift+t / mac: opt+shift+t) to re-open the closed tab
- Add a context menu action to the panel tabs, with a right click menu option for "Re-open closed panel". It should appear in it's own group just above the close group, and display the appropriate shortcut.

Re-opening a closed tab should restore state (ie. filters, sorts, linker etc should be resorted) and pop it from the list of closed tabs. We may not be able to re-open in the same place if the stack the tab was part of is removed, so it should instead re-launch in a new position if required (but ideally opens in the same location).

mofojed:
@dsmmcken Context menu on the panel tab? Shouldn't text be "Re-open closed panel" instead? What about if you remove the last panel from the dashboard, what do you right click on then?

dsmmcken:
I suppose we could conditionally show a button on the "Dashboard is Empty" screen (as button centered below that text) if there's panels available to be re-opened. Does that screen exist now in DHC?

PR #1912, opened 2024-04-02
feat: reopen closed tabs

wusteven815:
- Add #1785 
  - Add keyboard shortcut (Alt+Shift+A) to reopen last closed tab

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -236,6 +236,13 @@ export class AppMainContainer extends Component<
         },
         {
           action: () => {
+            this.sendReopenLast();
+          },
+          shortcut: GLOBAL_SHORTCUTS.REOPEN_CLOSED_PANEL,
+          isGlobal: true,
+        },
+        {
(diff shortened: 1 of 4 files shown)


packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
```diff
@@ -34,6 +41,11 @@ class PanelContextMenu extends PureComponent<
     return [...(glContainer?.tab?.header?.tabs ?? [])];
   }
 
+  handleReopenLast(): void {
```
mofojed, line 44: Should actually be re-opening the last panel in this _stack_, which means we also need to store which stack the panel was closed. I think pull that functionality into a separate PR, leave this PR just for the global re-open last panel action (which will show up as a keyboard shortcut under Settings). 

We also will want a context menu on dashboards, but dashboards in community do not have context menu items (yet), could be another ticket as well.

This was after our discussion with @dsmmcken yesterday.

mofojed (changes requested):

mofojed (approved):
Very nice.

PR #1916, opened 2024-04-03
feat: re-export spectrum useStyleProp util

dsmmcken:
`useStyleProp` takes common spectrum style props and resolves them to values. In my case I need the resolved version of a dimensionValue ("size-100") as var(--XXX), and things like paddingY resolved as paddingTop, paddingBottom for use externally in dh.ui.


bmingles (approved):
LGTM

PR #1917, opened 2024-04-03
fix: adjust alignment of search input next/previous buttons

dsmmcken:
Fixes alignment of next/previous buttons in `SearchInput`. Looks like they became mis-aligned during spectrum switchover.

Before:
image

After:
image


bmingles (approved):
LGTM

PR #1919, opened 2024-04-04
feat: ListView components

bmingles:


mofojed (changes requested):

packages/code-studio/src/styleguide/ListViews.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { useCallback, useState } from 'react';
+import { Grid, Item, ListView, ItemKey, Text } from '@deephaven/components';
+import { vsAccount, vsPerson } from '@deephaven/icons';
+import { Icon } from '@adobe/react-spectrum';
```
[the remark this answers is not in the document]
bmingles: I just picked one of the 2 supported. Is `illustration` preferable over `icon` ?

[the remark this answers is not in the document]
bmingles: Not 100% sure, but I think illustration must be added on the icons coming from `@spectrum-icons/illustrations/xxxxx`
https://react-spectrum.adobe.com/react-spectrum/ListView.html#complex-items

I think you are right that the default slot for `Icon` is `'icon'`. Unfortunately the ListView has to have `image` or `illustration` vs Picker which supports `icon`. Maybe a Spectrum bug but not sure.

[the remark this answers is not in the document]
dsmmcken: I think the expectation is you wouldn't put decorative icons in that slot, as it would be confusing with icons in the action slot (which I think in this example overflow because there is more than one)

image

Where as the primary slot is for images/illustrations, not icons.

packages/jsapi-components/package.json
```diff
@@ -22,6 +22,7 @@
     "build:sass": "sass --embed-sources --load-path=../../node_modules ./src:./dist"
   },
   "dependencies": {
+    "@adobe/react-spectrum": "^3.34.1",
```
mofojed, line 25: Should import everything from `@deephaven/components`, even in `jsapi-components`. Seems we're importing `useProvider` in `ListView.tsx`.

packages/code-studio/src/styleguide/ListViews.tsx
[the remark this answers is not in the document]
bmingles: I guess the lack of exporting Illustration might be an oversight:
https://github.com/adobe/react-spectrum/blob/main/packages/%40adobe/react-spectrum/src/index.ts#L32

Maybe should include `export {Icon} from '@react-spectrum/icon';`

packages/code-studio/src/styleguide/ListViews.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { generateNormalizedItems, sampleSectionIdAndClasses } from './utils';
+
+// Generate enough items to require scrolling
+const itemsSimple = [...generateNormalizedItems(52)];
+
+function AccountIllustration(): JSX.Element {
+  return (
+    // Images in ListView items require a slot of 'image' or 'illustration' to
+    // be set in order to be positioned correctly:
+    // https://github.com/adobe/react-spectrum/blob/784737effd44b9d5e2b1316e690da44555eafd7e/packages/%40react-spectrum/list/src/ListViewItem.tsx#L266-L267
```
mofojed, line 21: Any reason they don't have a slot provider for `icon`? We could even add a PR for it... seems odd they support `image` and `illustration` but not `icon`.

mofojed (changes requested):
One package cleanup note, everything looking good otherwise.

packages/jsapi-components/package.json
bmingles: @mofojed  hmm, we aren't re-exporting currently. Any opinions on how we want to do this? `useProvider` seems too generic, but `useSpectrumProvider` seems a bit odd since we are trying to abstract away Spectrum.

packages/code-studio/src/styleguide/ListViews.tsx
bmingles: I'm guessing @dsmmcken suggestion may be accurate here. Maybe they are trying to not overload the fact that icons can be included in actions.

dsmmcken: Yeah, that's a key difference between listview and listbox, https://react-spectrum.adobe.com/react-spectrum/ListBox.html#complex-items

packages/code-studio/src/styleguide/ListViews.tsx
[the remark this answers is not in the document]
mofojed: Perhaps, they also don't export `UIIcon`. Unsure if intentional or not. Could open a PR and see how they respond.

packages/jsapi-components/package.json
mofojed: Well, you're exporting `SpectrumThemeProvider`, `useSpectrumThemeProvider` to mirror that would be fine with me.
Unsure if we want to get more abstract than that, we already have `ThemeProvider`.

mofojed (approved):

PR #1920, opened 2024-04-04
fix: package-lock missing pinned dependency

dsmmcken:
`@react-spectrum/utils` some how ended up with a * for dependency in the last push. Fixing. Unsure how that happened.


bmingles (approved):
LGTM

PR #1925, opened 2024-04-05
feat: Update panels from field change events

### LINKED ISSUE #1924: Opened panels should match the state of panels in the panels menu
Panels in the panels menu currently get updated whenever field update events are fired on the connection (`connection.subscribeToFieldUpdates` in `AppMainContainer`). These respond to any changes on the server not just the current session. In contrast, panels that are opened only update when commands are issued to the current console session. They do not update if a command is issued from a different session.

It would be preferred if open panels always reflected the same state as the panels in the panels menu.
1. This avoids a mismatch such that clicking on a panel in the menu results in a different state than a currently opened panel
2. This opens up some interesting options for other sessions to update panels such as IDE extensions

bmingles:
resolves #1924

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -35,6 +35,7 @@ import {
   DehydratedDashboardPanelProps,
   getAllDashboardsData,
   getDashboardData,
+  LayoutUtils,
   listenForCreateDashboard,
   PanelEvent,
   setDashboardData as setDashboardDataAction,
@@ -317,6 +318,10 @@ export class AppMainContainer extends Component<
 
     this.widgetListenerRemover = connection.subscribeToFieldUpdates(updates => {
       log.debug('Got updates', updates);
+
+      const { activeTabKey } = this.state;
+      const layout = this.dashboardLayouts.get(activeTabKey);
+
       this.setState(({ widgets }) => {
         const { updated, created, removed } = updates;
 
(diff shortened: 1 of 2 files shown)


packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -335,6 +340,35 @@ export class AppMainContainer extends Component<
...
+        updated.forEach(toUpdate => {
+          let panelId: string | undefined = undefined;
+
+          if (layout) {
+            const widgetContentItem = LayoutUtils.getContentItemInRoot(
+              layout.root,
+              {
+                title: toUpdate.title,
+              }
+            );
```
mofojed, line 352: We shouldn't be using the title here. The title changes when we [clone panels](https://github.com/deephaven/web-client-ui/blob/309ff795dcb367fc9b7b6f2abb4a07b10cf8ab55/packages/dashboard/src/layout/LayoutUtils.ts#L675) and when we [rename](https://github.com/deephaven/web-client-ui/blob/309ff795dcb367fc9b7b6f2abb4a07b10cf8ab55/packages/dashboard/src/layout/LayoutUtils.ts#L647) [panels](https://github.com/deephaven/web-client-ui/blob/309ff795dcb367fc9b7b6f2abb4a07b10cf8ab55/packages/golden-layout/src/items/AbstractContentItem.ts#L345).

We should instead be matching by `panelId` (which we should already know if the panel was opened by `AppMainContainer`) or by the `metadata`. Problem is as you've identified below, `ConsolePanel` is doing a similar thing for panels that it has opened, so you can get conflicts.

Perhaps the panel/component itself should be listening for field updates, and update the widget if the underlying field changes (based somewhat on Raffi's comment that he would always want/expect the widget to reflect the currently named widget, indicating our behaviour when opening a panel and it not updating is incorrect).

bmingles: I like the idea of the panel listening for changes and updating itself. I'll look into that.

AkshatJawne (commented):
@bmingles Are we still actively making code changes here? if not, we could probably close this ticket since it has been open for a couple of months now as a draft?

bmingles:
Closing as we'll probably need a different approach and this has been de-prioritized since vscode extension now uses embed widget.

PR #1927, opened 2024-04-10
fix: Fix issues when auto-size columns/rows is false, and when row headers are not 0

mofojed:
- Shouldn't throw in those cases, we should handle when calculatedSizes is empty
- Found when using out [deephaven-doom](https://github.com/mofojed/deephaven-doom), couldn't resize/move columns like expected
- Also tested in the styleguide - dragging columns when there were row headers visible did not work correctly

### DIFF
--- a/packages/grid/src/GridRenderer.ts
+++ b/packages/grid/src/GridRenderer.ts
@@ -2354,12 +2354,14 @@ export class GridRenderer {
       }
     );
 
-    // Now move to the edge of the "grid" in the so we can draw the contents of the dragging column in the clipped region
+    // Now move to the edge of the "grid" (top-left of top-left cell). We then draw the
+    // grid background, but only the clipped region will be drawn where the dragging column is.
     context.translate(gridX, gridY);
     context.font = font;
 
     this.drawGridBackground(context, state);
 
+    // Then draw the contents of the column that is being dragged
     for (let i = startIndex; i <= endIndex; i += 1) {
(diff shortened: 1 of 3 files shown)


packages/grid/src/GridRenderer.ts
```diff
@@ -2335,6 +2335,8 @@ export class GridRenderer {
       visibleColumns.push(i);
     }
 
+    // The headers expect the context to be at the edge...
+    context.translate(-gridX, 0);
```
mattrunyon, line 2339: The translate before this is `draggingLeft - originalLeft`. Is this the same as `-gridX` in all cases? Also clarify what "the edge" is. Edge of the canvas? Edge of some bounding box?

mattrunyon (changes requested):

packages/grid/src/GridRenderer.ts
```diff
@@ -2352,7 +2354,8 @@ export class GridRenderer {
       }
     );
 
-    context.translate(0, gridY);
+    // Now move to the edge of the "grid" in the so we can draw the contents of the dragging column in the clipped region
```
mattrunyon, line 2357: Not sure if this is what you meant, but looks like something is missing in the comment

```suggestion
    // Now move to the edge of the "grid" in the clipped region so we can draw the contents of the dragging column in the clipped region
```

mattrunyon (changes requested):

mattrunyon (approved):

Issue #1931, opened 2024-04-12
Panel context menu should have option to reopen from its stack


wusteven815:
Continuation of https://github.com/deephaven/web-client-ui/pull/1912#discussion_r1550385541. The context menu of panels should have an option to reopen the last closed panel from that stack only.

Issue #1922, opened 2024-04-05
Replace deprecated shortid package with nanoid


mofojed:
[`shortid`](https://www.npmjs.com/package/shortid) is deprecated. Should replace with [`nanoid`](https://github.com/ai/nanoid/)

dsmmcken:
what about `useId`? https://react.dev/reference/react/useId

PR #1932, opened 2024-04-12
feat: context menu reopen for stack only

wusteven815:
- Adds #1931 
  - Add new context menu option for panels
  - Give IDs to stacks and helper function to get stacks by ID
  - Add ability to open component to specific stack
  - Fix unit tests related to IDs and layouts, added E2E tests

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
+++ b/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
@@ -1,9 +1,5 @@
 import React, { PureComponent, ReactElement } from 'react';
-import {
-  ContextAction,
-  ContextActions,
-  GLOBAL_SHORTCUTS,
-} from '@deephaven/components';
+import { ContextAction, ContextActions } from '@deephaven/components';
 import type {
   Container,
   EventEmitter,
--- a/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
+++ b/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
@@ -142,7 +142,6 @@ class PanelContextMenu extends PureComponent<
 
     contextActions.push(() => ({
       title: 'Re-open closed panel',
-      shortcut: GLOBAL_SHORTCUTS.REOPEN_CLOSED_PANEL,
       order: 10,
       group: ContextActions.groups.reopen,
       action: this.handleReopenLast,
--- a/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
+++ b/packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
@@ -1,11 +1,6 @@
 import React, { PureComponent, ReactElement } from 'react';
 import { ContextAction, ContextActions } from '@deephaven/components';
-import type {
-  Container,
-  EventEmitter,
-  ReactComponentConfig,
-  Tab,
-} from '@deephaven/golden-layout';
+import type { Container, EventEmitter, Tab } from '@deephaven/golden-layout';
 import {
   CustomizableWorkspace,
   RootState,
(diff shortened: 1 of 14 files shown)


packages/golden-layout/src/config/ItemConfig.ts
```diff
@@ -86,6 +86,11 @@ export interface ReactComponentConfig extends ItemConfig {
    * Properties that will be passed to the component and accessible using this.props.
    */
   props?: any;
+
+  /**
+   * The stack the component is in.
+   */
+  parentStackId?: string | string[];
```
mofojed, line 93: I don't think we want this to be part of the `ItemConfig` - this is something that changes as the component is moved around, and should be derived from it's position in the layout, not specified on the config itself.
Similar reasoning for why `ItemConfig` doesn't have a link to it's parent.

packages/golden-layout/src/LayoutManager.ts
```diff
@@ -513,6 +513,7 @@ export class LayoutManager extends EventEmitter {
       };
     }
 
+    config.id = config.id ?? getUniqueId();
```
mofojed, line 516: Update `getUniqueId` function to use `shortid` to generate the ID (or `nanoid` even, as we're replacing `shortid` soon: https://github.com/deephaven/web-client-ui/issues/1922)

mofojed (dismissed):

packages/dashboard-core-plugins/src/panels/PanelContextMenu.tsx
```diff
@@ -102,11 +123,26 @@ class PanelContextMenu extends PureComponent<
...
+      panel => (panel as ReactComponentConfig).parentStackId === stackId
+    );
+
+    contextActions.push(() => ({
+      title: 'Re-open closed panel',
+      order: 10,
+      group: ContextActions.groups.reopen,
+      action: this.handleReopenLast,
+      disabled: !hasPanelToReopen,
+    }));
```
mofojed, line 145: The check for `hasPanelToReopen` is fairly expensive - you should plumb this up like the disabled check below for `this.canCloseTabsRight()`, for example.

Right now you're running the check (which does `LayoutUtils.getStackForConfig`, `workspace.data.closed?.some(...)`, both expensive operations depending on how many components in the layout/workspace data) on every `render`. Try setting a breakpoint in the panel and see how often it re-renders (moving panels, updating table filters, maybe some other things?).

You only need to check when the context menu appears, which you can do by just passing in a function action and checking in there.

packages/dashboard/src/PanelManager.ts
```diff
@@ -291,12 +296,40 @@ class PanelManager {
...
+      config,
+      replaceConfig,
+      stack: stack ?? undefined,
+    });
   }
 
-  handleReopenLast(): void {
+  /**
+   *
+   * @param glContainer Only reopen panels that were closed from the stack of this container, is defined
```
mofojed, line 313:
```suggestion
   * @param glContainer Only reopen panels that were closed from the stack of this container, if defined
```

packages/dashboard/src/PanelManager.ts
```diff
@@ -322,12 +355,15 @@ class PanelManager {
...
     if (config && isReactComponentConfig(config)) {
       const dehydratedConfig = this.dehydrateComponent(
         config.component,
         config
       );
+      (dehydratedConfig as ClosedPanel).parentStackId =
+        LayoutUtils.getStackForConfig(root, config)?.config.id;
       if (dehydratedConfig != null) {
         this.closed.push(dehydratedConfig);
       }
```
mofojed, line 369: We shouldn't need the `null` check here, the `dehydratedConfig` cannot be `null`.

mofojed (dismissed):
Some small cleanup, looks great!

mofojed (approved):

Issue #1902, opened 2024-03-25
Fira fonts require GitHub to install


mofojed:
All of our dependencies are stored in the npmjs registry, except for fira, which references the [GitHub URL](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#github-urls) directly: https://github.com/deephaven/web-client-ui/blob/b49b5069d637ac136578ce839d9fc0416f468adf/packages/app-utils/package.json#L51

That could be a problem when installing the package, as a) They may have access to pulling from GitHub blocked/locked down, and/or b) That repository is not guaranteed to always exist; the packages published on npmjs should always be there.

dsmmcken:
I forget the reason we did this, but there was one.

dsmmcken:
In enterprise we use to embed them in the repo directly. I think we switched to a package to get license provenance.

(non-accessible commit switching it in core repo before it became deephaven-core, while it was still being split)
https://github.com/deephaven/core/pull/344/files

At the time there was no versions of it on npm. I looked at @fontsource/fira-sans and that would probably be fine. I can confirm it is correctly publishing the font with feature settings for tabular numbers preserved.

PR #1944, opened 2024-04-17
fix: change fira source

wusteven815:
- Adds #1902
- Use `@fontsource` instead of GitHub as source
- Lock to Fira Sans to 5.0.20 and Fira Mono to 5.0.13 to prevent kerning problems


mofojed:
@dsmmcken would these differences be expected/okay? Fonts appear okay to me, tabular numbering etc.

**Actual:**
typography-actual

**Expected:**
typography-expected

**Diff:**
typography-diff


From in the app on a table with numbers:
image

dsmmcken (dismissed):
Missing `Fira Mono`, which is being substituted. The previous package included both Fira Sans and Mono, you'll have to add those in as well.

That's frustrating that the kerning tables must be different. Still looks acceptable, and has correct font-features-settings. 🤷

dsmmcken:
null

We are probably also missing an italic version in that weight.

packages/app-utils/src/components/FontBootstrap.tsx
```diff
@@ -1,5 +1,11 @@
 import React, { createContext, useEffect, useState } from 'react';
-import 'fira';
+import '@fontsource/fira-mono/400.css';
+import '@fontsource/fira-sans/300.css';
+import '@fontsource/fira-sans/400.css';
+import '@fontsource/fira-sans/400-italic.css';
+import '@fontsource/fira-sans/500.css';
+import '@fontsource/fira-sans/600.css';
+import '@fontsource/fira-sans/700.css';
```
dsmmcken, line 8:
```suggestion
import '@fontsource/fira-sans/300.css';
import '@fontsource/fira-sans/300-italic.css';
import '@fontsource/fira-sans/400.css';
import '@fontsource/fira-sans/400-italic.css';
import '@fontsource/fira-sans/500.css';
import '@fontsource/fira-sans/600.css';
import '@fontsource/fira-sans/700.css';
import '@fontsource/fira-sans/700-italic.css';
```

There's a couple places with 300 italic and 700 italic

dsmmcken (approved):

mofojed (approved):

mofojed (changes requested):
Need to re-run `npm install` to update the `package-lock.json` now.

dsmmcken (approved):

PR #1953, opened 2024-04-22
fix: DH-16865: Fix spinner on empty partitions list

### LINKED ISSUE #1904: DH-16865: Partitioned table with no data displays a spinner indefinitely
Steps to reprodude:

Run this snippet to create a partitioned table with no data:
```
t = empty_table(0).update("X=ii").partition_by("X")
```
Expected:
- Table loads with no data rows

Actual:
- Table shows a spinner, doesn't load

vbabich:
Closes #1904


vbabich:
This PR should fix the issue with partitioned tables with no partitions in DHE. But I think I found an edge case with widgets of type `VariableType.PARTITIONEDTABLE` without partitions in DHC, where a `null` returned from `this.partitionedTable.getTable(partitions)` sends `IrisGrid` into an infinite `initState` loop.

mofojed (changes requested):
Not sure why my comments didn't post before

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2043,39 +2043,23 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
-            keyTable.close();
-            resolve(newPartition);
-          } catch (e) {
-            keyTable.close();
-            reject(e);
-          }
-        }
-      );
-    });
+    const { rows } = await keyTable.getViewportData();
```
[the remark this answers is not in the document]
vbabich: Yep, I only tested with DHE partitioned tables, didn't realize `PARTITIONEDTABLE ` in DHC triggers a different code path. Will fix and make sure to test both DHE and DHC.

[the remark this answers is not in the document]
vbabich: > We shouldn't be returning the initial partition until it's available.
What should we return in that case? Ok to show the Spinner on DHC partitioned tables? In Enterprise we want to show an empty table when there's no available partitions.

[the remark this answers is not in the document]
mofojed: We should be showing the empty table in the DHC case as well if we already know there's no partitions.

packages/iris-grid/src/EmptyIrisGridModel.ts
```diff
@@ -1,42 +1,21 @@
...
-    super(dh);
-
-    this.modelFormatter = formatter;
-  }
-
-  modelFormatter: Formatter;
-
-  get rowCount(): number {
-    return 0;
+    super(dh, []);
```
mofojed, line 12: Need to pass the `formatter` through.
```suggestion
    super(dh, [], formatter);
```

That being said, it's not like this really _needs_ a formatter since it's an empty grid, but :shrug:

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -872,6 +875,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     const { model } = this.props;
     this.initState();
     this.startListening(model);
+    this.startListeningForPartitionChanges(model);
```
mofojed, line 878: I'd probably break down the current `startListening`/`stopListening` methods, e.g.

```
startListening(model: IrisGridModel): void {
  this.startListeningForEvents(model);
  this.startListeningForPartitionChanges(model);
}
```
As you're calling both startListening methods everywhere, may as well combine it into just one top level one.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2304,6 +2276,69 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    keyTable.applySort(sorts);
+    // Select the last partition by default
+    keyTable.setViewport(0, 0);
+  }
+
+  stopListeningForPartitionChanges(
+    model: IrisGridModel,
+    resetState = false
+  ): void {
+    const { keyTable } = this.state;
```
mofojed, line 2327: There is a race condition here - in `startListeningForPartitionChanges`, we fetch the keyTable asynchronously, so in theory it's possible that `stopListening` is called before `startListening` completes/can set the key table.

May need to keep that promise around explicitly such that it can be cancelled/unset if stopListening is called.

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -68,6 +68,8 @@ class IrisGridPartitionSelector extends Component<
         t => t.close()
       );
 
+      // TODO: find out how partitionTables can be null after this
+
```
mofojed, line 72: Can they be? Remove TODO

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -232,7 +234,13 @@ class IrisGridPartitionSelector extends Component<
    */
   updatePartitionFilters(): void {
     const { partitionTables } = this.state;
-    assertNotNull(partitionTables);
+    // TODO: cleanup, partitionTables can be null now
+    // assertNotNull(partitionTables);
+
```
mofojed, line 239: TODO cleanup

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -253,8 +261,11 @@ class IrisGridPartitionSelector extends Component<
     log.debug('getPartitionFilters', partitionConfig);
 
     if (partitions.length !== partitionTables.length) {
-      throw new Error(
-        `Invalid partition config set. Expected ${partitionTables.length} partitions, but got ${partitions.length}`
+      // throw new Error(
+      //   `Invalid partition config set. Expected ${partitionTables.length} partitions, but got ${partitions.length}`
+      // );
```
mofojed, line 266: Why not throw here?

mofojed (changes requested):

AkshatJawne:
Closing PR since this bug has been resolved in https://github.com/deephaven/web-client-ui/pull/2110

Issue #1933, opened 2024-04-12
Playwright CI tests can be sharded to run faster


wusteven815:
See https://playwright.dev/docs/test-sharding

mofojed:
Timebox to a day for research, then re-evaluate.

mattrunyon:
Could try doing a build, uploading the build as an artifact (+ node_modules maybe), then using a matrix task based on browser. Each of those tasks just downloads the build artifact to start so we don't duplicate the build

PR #1955, opened 2024-04-22
ci: shard e2e tests

wusteven815:
- Adds #1933 
- Shards Playwright E2E tests by browser
  - Webkit into 2 shards because it runs slower

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -13,22 +13,20 @@ on:
       - 'release/**'
 env:
   DOCKER_TAG: edge
+
 jobs:
+
   build:
     runs-on: ubuntu-22.04
     concurrency:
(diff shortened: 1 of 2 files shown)


mofojed (dismissed):
I'd like the StyleGuide changes separated into their own PR, and sharding in it's own PR. Any reason not to?
I'm curious how much the savings are the StyleGuide changes vs. the sharding.
Overall savings look very impressive - 8 minutes instead of 32 minutes, you're cutting the run time to 25%! Sign me up!

wusteven815:
> Can you open a PR that based off of this PR and introduce something that will fail the tests? Want to see what the report looks like in failure when sharding.

https://github.com/deephaven/web-client-ui/actions/runs/8819362643?pr=1961

mofojed (approved):
Excellent work.

PR #1960, opened 2024-04-24
test: update styleguide to isolate by state

wusteven815:
- Refactor styleguide page to filter components by if it's rendered, instead of by CSS variables
- Refactor styleguide test to have sections be soft steps instead of its own step
  - Refactor buttons section to be a function component
- Refactor styleguide test to run in parallel instead of serial


packages/code-studio/src/styleguide/utils.ts
```diff
@@ -5,6 +5,10 @@ import { NormalizedItem } from '@deephaven/components';
 export const HIDE_FROM_E2E_TESTS_CLASS = 'hide-from-e2e-tests';
 export const SAMPLE_SECTION_CLASS = 'sample-section';
 
+export interface IsHashProp {
```
bmingles, line 8: Should be `isHashProp`

packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -76,63 +79,91 @@ function StyleGuide(): React.ReactElement {
...
 
-        <Typograpy />
+        {isHash('typography') && <Typograpy />}
 
-        <SampleMenuCategory data-menu-category="Colors" />
-        <Colors />
-        <ThemeColors />
+        {isHash('') && <SampleMenuCategory data-menu-category="Colors" />}
+        {isHash('colors') && <Colors />}
+        {(isHash('theme-color-palette') ||
```
bmingles, line 111: I would consider allowing multiple values passed to `isHash` instead of || chaining

tests/styleguide.spec.ts
```diff
@@ -1,159 +1,158 @@
...
 
-buttonSectionIds.forEach((id, i) => {
-  test(`Buttons regression test - ${id}`, async ({ page }) => {
-    // Isolate the section
-    await page.goto(`/ide/styleguide?isolateSection=true#${id}`);
+test('Buttons regression test', async ({ page }) => {
+  await goToStyleguide(page);
 
-    const sampleSection = page.locator(`#${id}`);
+  for (let i = 0; i < buttonSectionIds.length; i += 1) {
```
bmingles, line 97: Same comment here. Running multiple tests in same test block may cause issues.

packages/code-studio/src/styleguide/StyleGuide.tsx
```diff
@@ -48,25 +48,28 @@ const stickyProps = {
 } as const;
 
 function StyleGuide(): React.ReactElement {
-  const isolateSection = window.location.search.includes('isolateSection=true');
   const { themes } = useTheme();
+  const [hash, setHash] = useState(window.location.hash);
   const hasMultipleThemes = themes.length > 1;
 
+  function isHash(label: string, processLabel = false): boolean {
```
bmingles, line 55: Instead of passing this around via props, I would consider encapsulating the hash matching functionality in a custom hook. Something like:

```typescript
function useIsHash() {
  const [hash, setHash] = useState(window.location.hash);

  useEffect(() => {
    const hashChangeHandler = () => setHash(window.location.hash);
    window.addEventListener('hashchange', hashChangeHandler);
    return () => window.removeEventListener('hashchange', hashChangeHandler);
  }, []);

  const isHash = useCallback(
    (label: string, processLabel = false): boolean => {
      const newLabel = processLabel
        ? label.toLocaleLowerCase().replaceAll(' ', '-')
        : label;

      return hash === `#${newLabel}` || hash === '';
    },
    [hash]
  );

  return isHash;
}
```

bmingles: Then you can just use it in any component like:
```
const isHash = useIsHash();
```

tests/styleguide.spec.ts
```diff
@@ -1,159 +1,158 @@
...
-// Iterate over all sample sections and take a screenshot of each one.
-sampleSectionIds.forEach(id => {
-  test(`UI regression test - Styleguide section - ${id}`, async ({ page }) => {
-    // Isolate the section
-    await page.goto(`/ide/styleguide?isolateSection=true#${id}`);
+test('UI regression - Styleguide sections', async ({ page }) => {
+  await goToStyleguide(page);
 
-    const sampleSection = page.locator(`#${id}`);
+  for (let i = 0; i < sampleSectionIds.length; i += 1) {
```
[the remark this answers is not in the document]
wusteven815: Using soft assertions this time so any failures will continue test execution.

packages/code-studio/src/styleguide/StyleGuide.tsx
dsmmcken: is there something that can be done so that it doesn't have to be littered all over the place?

tests/styleguide.spec.ts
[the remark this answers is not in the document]
bmingles: I'm not familiar with soft assertions. Curious how those are used?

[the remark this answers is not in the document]
wusteven815: All the expects within the loop are soft so even if one fails, we continue onto the next one as if it didn't fail. However, it's still reported at the end. So this prevents the case where 1 test failing early means the remaining are missed.

PR #1965, opened 2024-04-25
feat: Create an ErrorView that can be used to display errors

mofojed:
- Added to the StyleGuide as well
- Displays errors in a textarea, with a copy button and an expand button


packages/components/src/ErrorView.scss
```diff
@@ -0,0 +1,90 @@
...
+    width: 100%;
+
+    .error-view-header-text {
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      padding-left: $spacer;
+      padding-right: $spacer;
+      font-weight: bold;
+      text-overflow: ellipsis;
```
dsmmcken, line 33: not needed

packages/components/src/ErrorView.scss
```diff
@@ -0,0 +1,90 @@
...
+
+      .error-view-copy-button {
+        min-width: 3rem;
+      }
+
+      .error-view-expand-button {
+        svg {
+          margin-right: $spacer-1;
+        }
+      }
```
dsmmcken, line 73: remove, button has defautl margin on svg now

packages/components/src/ErrorView.tsx
```diff
@@ -0,0 +1,92 @@
...
+ */
+function ErrorView({
+  message,
+  isExpanded: isExpandedProp = false,
+  type = 'Error',
+}: ErrorViewerProps): JSX.Element {
+  const [isExpandable, setIsExpandable] = useState(false);
+  const [isExpanded, setIsExpanded] = useState(false);
+  const viewRef = useRef<HTMLDivElement>(null);
+  const textareaRef = useRef<HTMLPreElement>(null);
```
dsmmcken, line 35: rename, not a textarea

packages/components/src/ErrorView.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const handleResize = useCallback(() => {
+    if (isExpanded || textareaRef.current == null) {
```
dsmmcken, line 38: should also check `isExpandedProp` ?

dsmmcken (changes requested):
webkit styleguide snapshot looks wrong

packages/components/src/ErrorView.tsx
mofojed: Doesn't need to, `isExpandedProp` being true will never show the expand more/less button. Don't need to change this logic.

packages/components/src/ErrorView.scss
```diff
@@ -0,0 +1,83 @@
...
+      border-radius: 0 $border-radius 0 0;
+      flex-shrink: 0;
+
+      .btn-danger {
+        border-radius: 0;
+        color: $black;
+        opacity: 0.8;
+        padding: $spacer-1;
+        &:active {
+          color: $black;
```
dsmmcken, line 60:
```suggestion
        color: var(--dh-color-contrast-dark);
        opacity: 0.8;
        padding: $spacer-1;
        &:active {
          color: var(--dh-color-contrast-dark);
```

Got to watch those color variables for light theme.

dsmmcken (changes requested):

packages/components/src/ErrorView.tsx
dsmmcken: Yeah but aren't you doing an unnecessary work still? Checking height can trigger a forced reflow.

packages/components/src/ErrorView.scss
mattrunyon: We could try adding a stylelint rule to ban sass variables basically. I think we could use [this rule](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/dollar-variable-pattern/README.md) and set the regex to something like `/iWillNotUseSassVariables/` which would effectively ban sass variables I think

mofojed: Would need rules to allow for $spacer as well.

dsmmcken (approved):

PR #1981, opened 2024-05-02
feat: XComponent framework, allow WidgetPanelTooltip to be overridden

mofojed:
I must have screwed up when creating feature/core-plugins-refactor, because this code isn't in there. Previously approved with https://github.com/deephaven/web-client-ui/pull/1926

- Add XComponent (eXtendable Component) framework to allow components to be wrapped/replaced
  - Similar to Swizzling in Docusaurus
  - Useful for components that we need to replace at the Enterprise level (e.g. `WidgetPanelTooltip` needs to display the query name in Enterprise)
- Pass a `VariableDescriptor` to `WidgetPanel` and `WidgetPanelTooltip`
  - Allows for more information to be included, e.g. a `QueryVariableDescriptor` which extends `VariableDescriptor`
- Tested with branch `bender_DH-16263-core-plugins` on Enterprise


mattrunyon (approved):
Looks like the target branch wasn't updated in the original PR

mofojed:
@mattrunyon yea, but then I created `feature/core-plugins-refactor` after and deleted the `1936-core-plugins-refactor` branch. I must have accidentally created `feature/core-plugins-refactor` based off main instead of that 1936 branch.

PR #1982, opened 2024-05-02
perf: remove workspace dependancy from iris-grid-panel and memoize settings redux selector

dsmmcken:
Discovered as part of #1977 investigation. Typing in a notebook triggers a workspace update, which was triggering all tables to re-render. Removed deprecated dependeny of workspace from iris-grid.

Further settings is derived from workspace and returns a new settings object. Iris-grid depends on settings, but the object had no changes. Memoize the selector so it always returns the same object if settings haven't changed.

Removes ~100ms of lag that was occuring 400ms after the debounce of notebook key press events.

BREAKING CHANGE: getPluginContent deprecatedProps have been removed from iris-grid


mofojed (approved):
Tested with Enterprise, improves things bigly. Thanks!

mattrunyon:
Was this tested with enterprise table plugins? I think it would break any customers using table plugins

PR #1986, opened 2024-05-03
fix: Fixed ListView props

bmingles:
Wired up ListView props I missed on refactor.

### DIFF
--- a/packages/components/src/spectrum/listView/ListView.tsx
+++ b/packages/components/src/spectrum/listView/ListView.tsx
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 import { SpectrumListViewProps } from '@adobe/react-spectrum';
+import cl from 'classnames';
 import { EMPTY_FUNCTION } from '@deephaven/utils';
 import {
   ItemKey,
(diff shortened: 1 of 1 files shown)


packages/components/src/spectrum/listView/ListView.tsx
```diff
@@ -71,8 +72,18 @@ export function ListView({
     <ListViewWrapper
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...spectrumListViewProps}
-      UNSAFE_className="dh-list-view"
+      UNSAFE_className={cl('dh-list-view', UNSAFE_className)}
+      selectedKeys={
+        selectedKeys as ListViewWrapperProps<unknown>['selectedKeys']
+      }
```
mofojed, line 78: Do we have a ticket or PR with Spectrum to include boolean in their keys? Kind of annoying we need to do this casting.

mofojed (approved):

packages/components/src/spectrum/listView/ListView.tsx
bmingles: We have a related one https://github.com/adobe/react-spectrum/issues/6103 but not one for the types.

PR #1987, opened 2024-05-03
perf: Improve performance of lots of grids in a dashboard

mofojed:
- Each GridRenderer had an independent cache for checking various colour operations, instead use a shared cache
  - Grids are likely to share the same colours because they have similar themes anyway
- Improves performance when opening dashboards with lots of grids


dsmmcken:
Another idea I've had is making a shared cache for every measure text call. The cache would need to be keyed on both the current context.font + string being measured, but tables frequently repeat values internally and across tables, and measureText is rather expensive.

mattrunyon (changes requested):

mattrunyon (approved):

PR #1992, opened 2024-05-07
feat: add middle click dashboard tab deletion

### LINKED ISSUE #1990: Allow middle clicking a dashboard top tab to close the dashboard
We allow middle clicking to close a tab within a dashboard, we should also allow middle clicking the dashboard's tab itself to close the whole dashboard.

Example for creating a dashboard:
```python
from deephaven import ui, empty_table

my_dash = ui.dashboard([
  empty_table(10).update("x = i"),
  empty_table(10).update("x = i"),
])
```

AkshatJawne:
Resolves #1990
- Add functionality to close dashboard tab with middle mouse click by specifying onAuxClick event
- Create NavBar component test cases that test changes

### DIFF
--- a/packages/components/src/navigation/NavTab.tsx
+++ b/packages/components/src/navigation/NavTab.tsx
@@ -67,6 +67,12 @@ const NavTab = memo(
               data-testid={`btn-nav-tab-${title}`}
               role="tab"
               tabIndex={0}
+              onAuxClick={event => {
+                // e.button equaling 1 corresponds to middle mouse button being clicked
+                if (isClosable && event.button === 1) {
+                  onClose?.(key);
+                }
+              }}
               onClick={e => {
                 (e.target as HTMLDivElement).focus();
                 // focus is normally set on mousedown, but dnd calls preventDefault for drag purposes
--- a/packages/components/src/navigation/NavTab.tsx
+++ b/packages/components/src/navigation/NavTab.tsx
@@ -67,17 +67,21 @@ const NavTab = memo(
               data-testid={`btn-nav-tab-${title}`}
               role="tab"
               tabIndex={0}
-              onAuxClick={event => {
-                // e.button equaling 1 corresponds to middle mouse button being clicked
-                if (isClosable && event.button === 1) {
+              onAuxClick={e => {
+                // button equaling 1 corresponds to middle mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
+                if (isClosable && e.button === 1 && e.buttons === 0) {
                   onClose?.(key);
                 }
               }}
               onClick={e => {
+                // have to have seperate check onClick for Safari not supporting AuxClick
+                if (isClosable && e.button === 1 && e.buttons === 0) {
+                  onClose?.(key);
+                  return;
+                }
                 (e.target as HTMLDivElement).focus();
                 // focus is normally set on mousedown, but dnd calls preventDefault for drag purposes
                 // so we can call focus on the firing of the actual click event manually
-
                 onSelect(key);
               }}
               onKeyPress={event => {
--- a/packages/components/src/navigation/NavTab.tsx
+++ b/packages/components/src/navigation/NavTab.tsx
@@ -79,10 +79,14 @@ const NavTab = memo(
                   onClose?.(key);
                   return;
                 }
-                (e.target as HTMLDivElement).focus();
-                // focus is normally set on mousedown, but dnd calls preventDefault for drag purposes
-                // so we can call focus on the firing of the actual click event manually
-                onSelect(key);
+                // button equaling 0 corresponds to left mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
+                if (e.button === 0 && e.buttons === 0) {
+                  // focus is normally set on mousedown, but dnd calls preventDefault for drag purposes
+                  // so we can call focus on the firing of the actual click event manually
+                  (e.target as HTMLDivElement).focus();
+
+                  onSelect(key);
+                }
               }}
               onKeyPress={event => {
                 if (event.key === 'Enter') onSelect(key);
--- a/packages/components/src/navigation/NavTab.tsx
+++ b/packages/components/src/navigation/NavTab.tsx
@@ -68,7 +68,7 @@ const NavTab = memo(
               role="tab"
               tabIndex={0}
               onAuxClick={e => {
-                // button equaling 1 corresponds to middle mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
+                // Middle mouse button was clicked, and no buttons remain pressed
                 if (isClosable && e.button === 1 && e.buttons === 0) {
                   onClose?.(key);
                 }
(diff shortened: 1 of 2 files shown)


packages/components/src/navigation/NavTab.tsx
```diff
@@ -67,6 +67,12 @@ const NavTab = memo(
               data-testid={`btn-nav-tab-${title}`}
               role="tab"
               tabIndex={0}
+              onAuxClick={event => {
```
dsmmcken, line 70: I don't think onAuxClick works in safari.

AkshatJawne:
recheck

packages/components/src/navigation/NavTab.tsx
AkshatJawne: Will take a look into this

mattrunyon: In that case you probably will need `onMouseUp` or `onMouseDown` (check behavior of the golden-layout tabs) and check `event.buttons===4` I think

dsmmcken: It's 1 not 4, and you should be able to get it from just the click event I would think.

bmingles: `Tab` component seems to use:

```
this.element.on('click', this._onTabClick);
this.element.on('auxclick', this._onTabClick);
```

@dsmmcken any idea if jQuery is making `auxclick` work in Safari, or is there something else happening here?

mattrunyon: `button === 1` only checks the last pressed button. `buttons` (plural) the value is 4. The plural is all buttons currently being pressed. So you can verify it's only a middle click instead of multiple buttons being held.

The plural uses a bitmask approach so every combination of buttons has a unique number. Left mouse = 1, right mouse = 2, middle = 4. So then left + right = 3, left + middle = 5

mattrunyon: Actually if `mouseUp` then `button === 1 && buttons === 0` would indicate the middle mouse button being released without any other buttons currently held. `buttons === 4` would be used on mouse down

bmingles: @mattrunyon  This seems overly complicated. What differentiates this from the `Tab` component implementation?

mattrunyon: Can you check middle click to close actually works on Safari with golden layout? If so, then jQuery is polyfilling the functionality. If not, then we just never tested it works in Safari

You would want to check both button and buttons on mouseUp because you could start dragging, then middle click on accident and possibly get into a very weird state. You just need to check 2 values, so I'm not sure how it's that complicated

AkshatJawne: Looking into it currently, will update

bmingles: Middle click does work with Safari golden layout.

bmingles: I'm following now. Seems like `mouseUp` event with a `button === 1 && buttons === 0` check is the simplest approximation. Just means that if you click down elsewhere and drag onto the tab and then release it would close the tab, but that doesn't really seem like a concern.

AkshatJawne:

packages/components/src/navigation/NavTab.tsx
AkshatJawne: Spoke to Matt, pushing changes shortly.

mattrunyon:
recheck

packages/components/src/navigation/NavTab.tsx
```diff
@@ -67,17 +67,21 @@ const NavTab = memo(
               data-testid={`btn-nav-tab-${title}`}
               role="tab"
               tabIndex={0}
-              onAuxClick={event => {
-                // e.button equaling 1 corresponds to middle mouse button being clicked
-                if (isClosable && event.button === 1) {
+              onAuxClick={e => {
```
bmingles, line 70: I think this needs to go in the `onMouseUp` prop to work in Safari right?

packages/components/src/navigation/NavTab.tsx
```diff
@@ -67,17 +67,21 @@ const NavTab = memo(
...
+                // button equaling 1 corresponds to middle mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
+                if (isClosable && e.button === 1 && e.buttons === 0) {
                   onClose?.(key);
                 }
               }}
               onClick={e => {
+                // have to have seperate check onClick for Safari not supporting AuxClick
+                if (isClosable && e.button === 1 && e.buttons === 0) {
```
bmingles, line 78: I don't think `onClick` this will fire in Safari for middle mouse button right?

AkshatJawne: Seems to be working

packages/components/src/navigation/NavTab.tsx
AkshatJawne: Spoke to Matt, and the issue with the onMouseUp is that there is no track of if you started the click somewhere else, tried this on Safari and seems to be working.

AkshatJawne:
recheck

packages/components/src/navigation/NavTab.tsx
bmingles: I guess `onClick` behaves differently than the jQuery `click` where I tested in `Tab`

AkshatJawne: Yeah, this behaviour is definitely weird, but there was something about Safari will call regular click for a middle click, so I'm pretty sure that's why this works.

bmingles: Yeah, 'click' fires in Safari for me now. I thought it was not when I tested earlier. Looks good.

packages/components/src/navigation/NavTab.tsx
```diff
@@ -67,12 +67,26 @@ const NavTab = memo(
               data-testid={`btn-nav-tab-${title}`}
               role="tab"
               tabIndex={0}
+              onAuxClick={e => {
+                // button equaling 1 corresponds to middle mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
```
bmingles, line 71:
```suggestion
                // Middle mouse button was clicked, and no buttons remain pressed
                button equaling 1 corresponds to middle mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
```

packages/components/src/navigation/NavTab.tsx
```diff
@@ -67,12 +67,26 @@ const NavTab = memo(
...
               onClick={e => {
-                (e.target as HTMLDivElement).focus();
-                // focus is normally set on mousedown, but dnd calls preventDefault for drag purposes
-                // so we can call focus on the firing of the actual click event manually
+                // have to have seperate check onClick for Safari not supporting AuxClick
+                if (isClosable && e.button === 1 && e.buttons === 0) {
+                  onClose?.(key);
+                  return;
+                }
+                // button equaling 0 corresponds to left mouse button being clicked, and buttons equaling 0 corresponds to it being the only button pressed
```
bmingles, line 82:
```suggestion
                // Left mouse button was clicked, and no buttons remain pressed
```

bmingles (changes requested):
Changes look good. I verified everything works as expected on Chrome + Safari.

Left minor suggestion for comment tweaks

bmingles (approved):
Looks good

Issue #1964, opened 2024-04-25
Embed widget should load workspace settings


bmingles:
Some deephaven ui components need formatter settings from the Redux store (e.g. [Picker](https://github.com/deephaven/deephaven-plugins/blob/main/plugins/ui/src/js/src/elements/Picker.tsx#L26)).

As-is, the following query crashes in embed widget:
```python
# import deephaven.ui as ui
# from deephaven.ui import use_state
from deephaven import ui
from deephaven import time_table
from deephaven.plot.figure import Figure
import datetime

default_value=5
initial_row_count=20

column_types = time_table("PT2S", start_time=datetime.datetime.now() - datetime.timedelta(seconds=initial_row_count)).update([
    "Int=new Integer(i)",
    "Double=new Double(i+i/10)",
    "Sin=Math.sin(i)",
    "Cos=Math.cos(i)",
])

@ui.component
def picker():
    value, set_value = ui.use_state(default_value)

    print("Test", value)

    text = ui.text(value, key="text1")

    # Picker for selecting values
    pick = ui.picker(
        column_types,
        key_column="Int",
        label_column="Int",
        label="Picker",
        selection_mode="multiple",
        on_change=set_value,
        selected_key=value,
    )

    # Display picker and output in a flex column
    return ui.flex(
        pick,
        text,
        direction="column",
        margin=10,
        min_height=0, # necessary to avoid overflowing container height
        gap=10,
    )

picker_column_types = picker()

figure = Figure().plot_xy(series_name="Demo", t=column_types, x="Sin", y="Cos").show()
```

<img width="712" alt="image" src="https://github.com/deephaven/web-client-ui/assets/1900643/02902bb1-8498-425f-8a8a-f8d8b1c01a79">

PR #1995, opened 2024-05-07
feat: Make grid widget respect global formatter settings

mattrunyon:
Fixes https://github.com/deephaven/deephaven-plugins/issues/435

Does not fix the issue in embed-widget. I think that should be a separate ticket. I'll keep https://github.com/deephaven/web-client-ui/issues/1964 open for tracking that if that seems fine

Tested by creating a dh.ui component that included a table and then changed the formatter settings. Tried global column formats too (the formats for specifically named columns)

Looks like ui.table already loads these settings, so this will fix for normal tables.


packages/dashboard-core-plugins/src/GridWidgetPlugin.tsx
```diff
@@ -7,11 +7,17 @@ import {
...
   type IrisGridModel,
 } from '@deephaven/iris-grid';
+import { useSelector } from 'react-redux';
+import { getSettings, RootState } from '@deephaven/redux';
 
 export function GridWidgetPlugin(
   props: WidgetComponentProps<dh.Table>
 ): JSX.Element | null {
   const dh = useApi();
+  const settings = useSelector<
```
bmingles, line 17: Is this because default workspace settings is a partial in plugins context?

bmingles (approved):
LGTM

Issue #1868, opened 2024-03-12
Do not show "Delete Selected Rows" for input tables without columns


mofojed:
Spun off from 

**Steps to Reproduce:**
1. Run snippet:
```python
from deephaven import empty_table, time_table, input_table, merge
from deephaven import dtypes as dht

historical = empty_table(10).update(["Timestamp = atMidnight(now(), 'ET') + ii*SECOND", "X = ii%2==0 ? `A` : `B`", "Y = ii"])
live = time_table("PT1s").update(["X = ii%2==0 ? `A` : `B`", "Y = ii"])

it = input_table({"X":dht.string})

t = merge([historical.where_in(it, "X"), live.where_in(it, "X")])
```
2. Right click on a row in the table and select "Delete Selected Rows"

**Expected Results:**
2. "Delete Selected Rows" menu item should not be available, since the table does not have a key column

**Actual Results:**
2. "Delete Selected Rows" menu item appears, but selecting it just errors

mofojed:
This would be a good one for Akshat to start with.

PR #1996, opened 2024-05-08
fix: "Delete Selected Rows" bug for tables with no key columns

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1868 

Changes:
- Added DeletableGridModel interface and corresponding isDeletableRange and isDeletableRanges methods as well as isDeletableGridModel and assertIsDeletableGridModel
- Logic is shifted such that IrisGridProxyModel implements DeletableGridModel, and IrisGridTableModelTemplate implements the DeletableGridModel

### DIFF
--- a/packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
+++ b/packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
@@ -335,6 +335,14 @@ class MockIrisGridTreeModel
     throw new Error('Not defined in mock');
   }
 
+  isDeletableRange(range: GridRange): boolean {
+    return false;
+  }
+
+  isDeletableRanges(ranges: readonly GridRange[]): boolean {
+    return false;
+  }
+
   valuesTable(columns: DhType.Column | DhType.Column[]): Promise<never> {
     throw new Error('Not defined in mock');
   }
--- a/packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
+++ b/packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
@@ -335,14 +335,6 @@ class MockIrisGridTreeModel
     throw new Error('Not defined in mock');
   }
 
-  isDeletableRange(range: GridRange): boolean {
-    return false;
-  }
-
-  isDeletableRanges(ranges: readonly GridRange[]): boolean {
-    return false;
-  }
-
   valuesTable(columns: DhType.Column | DhType.Column[]): Promise<never> {
     throw new Error('Not defined in mock');
   }
--- a/packages/grid/src/DeletableGridModel.ts
+++ b/packages/grid/src/DeletableGridModel.ts
@@ -1,0 +1,42 @@
+import GridRange from './GridRange';
+import GridModel from './GridModel';
+import { AssertionError } from './errors';
+
+export function isDeletableGridModel(
+  model: GridModel
+): model is DeletableGridModel {
+  return (model as DeletableGridModel)?.isDeletable !== undefined;
+}
+
+export function assertIsDeletableGridModel(
+  model: GridModel
+): asserts model is DeletableGridModel {
+  if (!isDeletableGridModel(model)) {
+    throw new AssertionError(
+      `Expected 'model' to be deletable, but received ${model}`
+    );
+  }
+}
+
+/**
+ * Model for an deletable grid
+ */
+export interface DeletableGridModel extends GridModel {
+  isDeletable: boolean;
+
+  /**
+   * Check if a given range is deletable
+   * @param range The range to check if it is deletable
+   * @returns True if the range is deletable
+   */
+  isDeletableRange: (range: GridRange) => boolean;
+
+  /**
+   * Check if an array of ranges is deletable
+   * @param ranges The ranges to check if they are deletable
+   * @returns True if the ranges is deletable
+   */
+  isDeletableRanges: (ranges: readonly GridRange[]) => boolean;
+}
+
+export default DeletableGridModel;
--- a/packages/grid/src/DeletableGridModel.ts
+++ b/packages/grid/src/DeletableGridModel.ts
@@ -19,9 +19,10 @@ export function assertIsDeletableGridModel(
 }
 
 /**
- * Model for an deletable grid
+ * Model for a deletable grid
  */
 export interface DeletableGridModel extends GridModel {
+  /** Whether this model is deletable or not */
   isDeletable: boolean;
 
   /**
@@ -32,11 +33,18 @@ export interface DeletableGridModel extends GridModel {
   isDeletableRange: (range: GridRange) => boolean;
 
   /**
-   * Check if an array of ranges is deletable
+   * Check if an array of ranges are deletable
    * @param ranges The ranges to check if they are deletable
-   * @returns True if the ranges is deletable
+   * @returns True if the ranges are deletable
    */
   isDeletableRanges: (ranges: readonly GridRange[]) => boolean;
+
+  /**
+   * Delete ranges from an input grid.
+   * @param ranges The ranges to delete
+   * @returns A promise that resolves successfully when the operation is complete or rejects if there's an error
+   */
+  delete: (ranges: readonly GridRange[]) => Promise<void>;
 }
 
 export default DeletableGridModel;
--- a/packages/iris-grid/src/EmptyIrisGridModel.ts
+++ b/packages/iris-grid/src/EmptyIrisGridModel.ts
@@ -152,6 +152,14 @@ class EmptyIrisGridModel extends IrisGridModel {
     // No-op
   }
 
+  isDeletableRange(range: GridRange): boolean {
+    return false;
+  }
+
+  isDeletableRanges(ranges: readonly GridRange[]): boolean {
+    return false;
+  }
+
   export(): Promise<DhType.Table> {
     throw new Error('Method not implemented.');
   }
--- a/packages/iris-grid/src/EmptyIrisGridModel.ts
+++ b/packages/iris-grid/src/EmptyIrisGridModel.ts
@@ -152,14 +152,6 @@ class EmptyIrisGridModel extends IrisGridModel {
     // No-op
   }
 
-  isDeletableRange(range: GridRange): boolean {
-    return false;
-  }
-
-  isDeletableRanges(ranges: readonly GridRange[]): boolean {
-    return false;
-  }
-
   export(): Promise<DhType.Table> {
     throw new Error('Method not implemented.');
   }
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -45,6 +45,7 @@ import {
   isEditableGridModel,
   BoundedAxisRange,
   isExpandableGridModel,
+  isDeletableGridModel,
 } from '@deephaven/grid';
 import {
   dhEye,
@@ -3722,11 +3723,13 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
    */
   deleteRanges(ranges: readonly GridRange[]): void {
     const { model } = this.props;
-    this.pending.add(model.delete(ranges)).catch(e => {
-      if (!PromiseUtils.isCanceled(e)) {
-        log.error('Unable to delete ranges', ranges, e);
-      }
-    });
+    if (isDeletableGridModel(model) && model.isDeletable) {
+      this.pending.add(model.delete(ranges)).catch(e => {
+        if (!PromiseUtils.isCanceled(e)) {
+          log.error('Unable to delete ranges', ranges, e);
+        }
+      });
+    }
   }
 
   resetColumnSelection(): void {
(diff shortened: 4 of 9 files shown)


packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -925,6 +926,10 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
           title: 'Delete Selected Rows',
           group: IrisGridContextMenuHandler.GROUP_EDIT,
           order: 50,
+          disabled:
+            !isIrisGridTableModelTemplate(model) ||
+            model.inputTable == null ||
+            model.inputTable.keyColumns.length === 0,
```
mofojed, line 932: Should just be able to check `selectedRanges.every(range => model.isEditableRange(range)` here. You can't delete rows that are not editable.

mofojed (changes requested):

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
AkshatJawne: Just tried doing that, it fixes disables the option for the code mentioned in the ticket (where there is no key columns). However, it also disables the option for the snippet below, when there are key columns defined: 

`from deephaven import empty_table, input_table
source = empty_table(10).update(
    [
        "Sym = (i % 2 == 0) ? `A` : `B`",
        "Marker = (i % 3 == 2) ? `J` : `K`",
        "X = i",
        "Y = sin(0.1 * X)",
    ]
)
input_source = input_table(init_table=source, key_cols=["X", "Y"])`

dsmmcken:
> Need to clarify whether the desired behaviour is for the the tab to be disabled as shown below, or should it not be present at all

If we know it can never be shown for a given table it should be hidden rather than disabled. Only disable it if there's a condition in which it could be come enabled.

AkshatJawne:
> > Need to clarify whether the desired behaviour is for the the tab to be disabled as shown below, or should it not be present at all
> 
> If we know it can never be shown for a given table it should be hidden rather than disabled. Only disable it if there's a condition in which it could be come enabled.

Will make this change right now

AkshatJawne:
Will look into e2e tests failing

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -919,7 +919,8 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
       if (
         isEditableGridModel(model) &&
         model.isEditable &&
-        selectedRanges.length > 0
+        selectedRanges.length > 0 &&
+        model.isDeletableRanges(selectedRanges)
```
mofojed, line 923: From [Don's comment](https://github.com/deephaven/web-client-ui/pull/1996#issuecomment-2102734094):
> If we know it can never be shown for a given table it should be hidden rather than disabled. Only disable it if there's a condition in which it could be come enabled.

In this case you're just hiding it if the selected range isn't deletable, which if you have a totals row visible isn't quite correct - in that case, we should be showing the menu item, just disabled (since you _can_ delete some rows, just not the totals rows). Other tables where you can't delete anything is where it shouldn't be shown at all.

So there's a few different types of tables we should be testing here ([input_table docs here](https://deephaven.io/core/docs/how-to-guides/input-tables/)):
```python
from deephaven import agg, empty_table, input_table

# Immutable table - should not show Delete Selected Rows, should not be editable
immutable_table = empty_table(10).update(["X = i", "Y=Math.sin(i)"])

# Append only table - should be able to add new rows, Delete Selected Rows should not be visible
append_only_table = input_table(init_table=source)

# Keyed row table - should be able to add new rows, Delete Selected Rows should be visible
keyed_table = input_table(init_table=source, key_cols=["X"])
```

Then, after making the `keyed_table`, you should be able to go to the table menu, hit Aggregate Columns, and add an aggregation (such as Sum). (BUG NOTE: Looks like you'll need to position it at the "Top" or you can't see the totals - should log that as a separate issue). Right clicking a keyed row should show "Delete Selected Rows" and enabled, but right-clicking the totals row should just show "Delete Selected Rows" and have it disabled.

I think to achieve all that we'll change up the design a little bit... it will be a bit more complicated, but it will be a cleaner design.
- Add a `DeletableGridModel.ts` next to `EditableGridModel.ts`, and define an interface for a grid model that is deletable (e.g. an `isDeletable: boolean;` property, and `isDeletableRange` and `isDeletableRanges` methods). Also create the type predicates (`isDeletableGridModel` and `assertIsDeletableGridModel`. See `isEditableGridModel` and `assertIsEditableGridModel` in `EditableGridModel` for examples).
- Have `IrisGridProxyModel` implement `DeletableGridModel`, and delegate to it's underlying model if the underlying model is `DeletableGridModel` (much like the `EditableGridModel` methods work in `IrisGridProxyModel`)
- Have `IrisGridTableModelTemplate` implement the `DeletableGridModel`.
  - No need to change the other models, since they won't implement `DeletableGridModel`.
- Check `isDeleteable` for adding the menu, then disable with `!isDeletableRange`

Let me know if you have questions. Making this a little more complicated, but now it will be more robust.

mofojed (changes requested):

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
mofojed: Also update your PR description after updating the content.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -300,6 +300,10 @@ abstract class IrisGridModel<
 
   abstract set totalsConfig(totalsConfig: UITotalsTableConfig | null);
 
+  abstract isDeletableRange(range: GridRange): boolean;
+
+  abstract isDeletableRanges(ranges: readonly GridRange[]): boolean;
+
```
mofojed, line 306: Shouldn't need this anymore. `IrisGridModel` does not implement `DeletableGridModel` (or `EditableGridModel`).

packages/code-studio/src/styleguide/MockIrisGridTreeModel.ts
```diff
@@ -335,6 +335,14 @@ class MockIrisGridTreeModel
...
   }
 
+  isDeletableRange(range: GridRange): boolean {
+    return false;
+  }
+
+  isDeletableRanges(ranges: readonly GridRange[]): boolean {
+    return false;
+  }
+
```
mofojed, line 345: Don't need this, it doesn't implement the `DeletableGridModel`.
```suggestion
```

packages/grid/src/DeletableGridModel.ts
```diff
@@ -0,0 +1,42 @@
...
+      `Expected 'model' to be deletable, but received ${model}`
+    );
+  }
+}
+
+/**
+ * Model for an deletable grid
+ */
+export interface DeletableGridModel extends GridModel {
+  isDeletable: boolean;
```
mofojed, line 25: We should still add a JS doc for this. It's possible to implement the method, but for that particular instance itself to not be deletable (e.g. `IrisGridTableModelTemplate` with an input table with no key columns).
```suggestion
  /** Whether this model is deletable or not */
  isDeletable: boolean;
```

packages/grid/src/DeletableGridModel.ts
```diff
@@ -0,0 +1,42 @@
...
+): asserts model is DeletableGridModel {
+  if (!isDeletableGridModel(model)) {
+    throw new AssertionError(
+      `Expected 'model' to be deletable, but received ${model}`
+    );
+  }
+}
+
+/**
+ * Model for an deletable grid
```
mofojed, line 22:
```suggestion
 * Model for a deletable grid
```

packages/grid/src/DeletableGridModel.ts
```diff
@@ -0,0 +1,42 @@
...
+   * Check if a given range is deletable
+   * @param range The range to check if it is deletable
+   * @returns True if the range is deletable
+   */
+  isDeletableRange: (range: GridRange) => boolean;
+
+  /**
+   * Check if an array of ranges is deletable
+   * @param ranges The ranges to check if they are deletable
+   * @returns True if the ranges is deletable
```
mofojed, line 37:
```suggestion
   * Check if an array of ranges are deletable
   * @param ranges The ranges to check if they are deletable
   * @returns True if the ranges are deletable
```

packages/grid/src/DeletableGridModel.ts
```diff
@@ -0,0 +1,42 @@
...
+   */
+  isDeletableRanges: (ranges: readonly GridRange[]) => boolean;
+}
```
mofojed, line 40: We should also add the `delete` method here, and remove it from `IrisGridModel`.
```suggestion
  /**
   * Delete ranges from a grid.
   * @param ranges The ranges to delete
   * @returns A promise that resolves successfully when the operation is complete or rejects if there's an error
   */
  abstract delete(ranges: readonly GridRange[]): Promise<void>;
}
```

One thing we're missing in general is the delete mode (e.g. shift cells up, shift cells left, or leave in place), but we'll ignore that for simplicity for now.

packages/iris-grid/src/EmptyIrisGridModel.ts
```diff
@@ -152,6 +152,14 @@ class EmptyIrisGridModel extends IrisGridModel {
...
   }
 
+  isDeletableRange(range: GridRange): boolean {
+    return false;
+  }
+
+  isDeletableRanges(ranges: readonly GridRange[]): boolean {
+    return false;
+  }
+
```
mofojed, line 162: Not a DeletableGridModel, so remove.
```suggestion
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -455,6 +455,10 @@ class IrisGridTableModelTemplate<
     return !this.isSaveInProgress && this.inputTable != null;
   }
 
+  get isDeletable(): boolean {
+    return this.isDeletable;
+  }
+
```
mofojed, line 461: This will cause an infinite loop, since `this.isDeletable` will call the `get isDeletable()` method. Shouldn't need this method anyway since this doesn't implement `DeletableGridModel`.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3722,11 +3723,13 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
-        log.error('Unable to delete ranges', ranges, e);
-      }
-    });
+    if (isDeletableGridModel(model) && model.isDeletable) {
+      this.pending.add(model.delete(ranges)).catch(e => {
+        if (!PromiseUtils.isCanceled(e)) {
+          log.error('Unable to delete ranges', ranges, e);
+        }
+      });
+    }
```
mofojed, line 3732: This is a case where the `deleteRanges` method should not be called if the model is not deletable. Right now this will just silently fail; instead we want to fail hard (throw an exception) so the caller knows to fix it.
I'd probably write it something like:
```suggestion
    if (!isDeletableGridModel(model) || !model.isDeletable) {
      throw new Error('Model does not support deleting ranges');
    }
    this.pending.add(model.delete(ranges)).catch(e => {
      if (!PromiseUtils.isCanceled(e)) {
        log.error('Unable to delete ranges', ranges, e);
      }
    });
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -455,6 +455,13 @@ class IrisGridTableModelTemplate<
     return !this.isSaveInProgress && this.inputTable != null;
   }
 
+  get isDeletable(): boolean {
+    if (this.inputTable !== null) {
+      return this.inputTable?.keyColumns.length > 0;
+    }
+    return false;
+  }
+
```
mofojed, line 464: Ah I see what's missing. At the top, we have `IrisGridTableModelTemplate` which `extends IrisGridModel`, but we should also specify that it `implements DeletableGridModel, EditableGridModel`. That way we can be sure the types of each method match correctly, and we're not missing any methods.
Add that in, you may need to adjust some of the types (mostly just adding a `readonly` in front of arrays).

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -2054,11 +2062,11 @@ class IrisGridTableModelTemplate<
     }
   }
 
-  editValueForCell(x: ModelIndex, y: ModelIndex): string | null | undefined {
-    return this.textValueForCell(x, y);
+  editValueForCell(column: ModelIndex, row: ModelIndex): string {
+    return this.textValueForCell(column, row) as string;
```
mofojed, line 2066: We should avoid casting whenever possible. In this case, it would be bad to return possibly `null` or `undefined` 

```suggestion
    return this.textValueForCell(column, row) ?? '';
```

mofojed (changes requested):

mofojed (approved):

AkshatJawne:
Merging now!

PR #1998, opened 2024-05-09
feat: e2e combined improvements

dsmmcken:
Contains several changes that all touch many snapshots. Combined to reduce snapshot churn.

The primary goal was to address webkit kerning issues caused by: https://github.com/microsoft/playwright/issues/20203

Recommend to review using `.` vscode mode and initially collapsing all snapshot folders.

1. Moves e2e test to execute inside docker container so enviroment is consitent between ci and local
   - changes e2e.yml to build a docker container, then uses that for each matrix test
   - dockerfile remove deja vu stuff, environment is now the same and grid.tsx
   - dockercompose adds service that runs the image from CI with matrix config via env variables
   - Fixes dockerfile and compose to *actually use the node version installed by nvm* (existing code didn't actually use the version installed by nvm, only the commands inside the -i interactive session would)
       - Get rid of interactive mode and make source work
       - Sym link nvm node binaries to clobber the version installed by playwright jammy container
   - Churns snapshots 

  
2. Removes deprecated ButtonOld
   - removed button old component
   - changes to style guide to no longer display that component
   - removed usage in chart filter overlay
   - renames snapshots to by id instead of index for buttons
   - Churns snapshots 
   
3.  Disable kerning and smoothing in webkit playwright tests:
    - theory is webkit kerning may be non-deterministic, disabling entirely may avoid those diffs
    - adds Playwright.css dynamically imported based of a env VITE_PLAYWRIGHT variable to AppRoot       
    - initially tried using built in https://playwright.dev/docs/test-snapshots#stylepath however, results were inconsistent as canvas grids may or may not re-render after playwright appended the styles
    - Updated playwright version to 1.44
    - Churns snapshots 
   
4.  Optimize ThemeColors section in style guide
    - Makes styleguide section load faster 
    - Remove all the ContrastColor() which have poor performance and were being done of every swatch. Instead text next to swatch
    - Move slow NormailzeColors() call in tooltip into props so it doesn't run unless tooltip is shown
    - Churns snapshots 
    
 5. Run golden-layout tests in parallel
     - I don't see a reason why these need to be serial, so might be slightly faster in parallel
  

BREAKING CHANGE: Removed ButtonOld component, use Button instead.


mofojed (changes requested):
Just marking as needing change so that this doesn't merge until we cut a release/branch for V+.

packages/code-studio/src/AppRoot.tsx
```diff
@@ -6,6 +6,11 @@ import { DownloadServiceWorkerUtils } from '@deephaven/iris-grid';
 import MonacoWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
 import AppRouter from './main/AppRouter';
 
+// load addional css for playwright docker tests
+if (import.meta.env.VITE_PLAYWRIGHT === '1') {
+  await import('./Playwright.css');
+}
```
mattrunyon, line 12: Doesn't matter since the file is so small, but this always gets included in the prod bundle. Just not loaded without this flag at buildtime. It should also get split into its own chunk that is never loaded.

tests/docker-scripts/Dockerfile
```diff
@@ -50,4 +42,4 @@ RUN SKIP_POSTINSTALL=1 npm ci
 COPY . .
 
 # Now build the app
-RUN npm run build
\ No newline at end of file
+RUN VITE_PLAYWRIGHT=1 npm run build
```
mattrunyon, line 45: Could change this env variable to something more descriptive like `DISABLE_FONT_FEATURES` or something. I don't think the `VITE_` is necessary since we load all env variables in our `vite.config.ts`, but doesn't hurt to have it.

mattrunyon (changes requested):

mattrunyon (commented):
I only reviewed the CI and docker files

tests/context-menu.spec.ts-snapshots/advanced-filters-6-firefox-linux.png
bmingles, line 1: Any idea why this screenshot is only a small square compared to the other browser versions?

packages/code-studio/src/AppRoot.tsx
dsmmcken: Yeah, I didn't know how do exclude from from a prod build, it doesn't get loaded but would still exist on the server. But it's also tiny, so 🤷

tests/docker-scripts/Dockerfile
dsmmcken: > To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code. e.g. for the following env variables

> https://vitejs.dev/guide/env-and-mode.html#env-files

I assumed it was needed from this. I could name it `PLAYWRIGHT_CSS` if you think that's better? I was leaving it slightly more generic incase we need to slip in any other css to mask anything else in playwright tests ever.

mattrunyon: It normally is (and their recommendation makes sense), but we change that in the vite config and everything is available. We only use it for `npm_package_version` though I think.

`VITE_PLAYWRIGHT_CSS` works

tests/context-menu.spec.ts-snapshots/advanced-filters-6-firefox-linux.png
dsmmcken: ? Guessing you are reviewing in github and have an error loading it. The default viewer is garbage for big reviews. Try hitting period to load the vs code reviewer.

bmingles: ah yeah, shows in vs code reviewer.

bmingles (dismissed):
LGTM

packages/code-studio/src/AppRoot.tsx
```diff
@@ -6,6 +6,11 @@ import { DownloadServiceWorkerUtils } from '@deephaven/iris-grid';
 import MonacoWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
 import AppRouter from './main/AppRouter';
 
+// load addional css for playwright docker tests
+if (import.meta.env.VITE_PLAYWRIGHT_CSS === '1') {
+  await import('./Playwright.css');
+}
+
```
mofojed, line 13: We should follow up to see if we can fix this in Playwright with some sort of global hook to inject CSS to a page as it loads
Maybe we can use https://playwright.dev/docs/api/class-page#page-add-style-tag

mofojed (approved):
Good to merge, added a comment to remove the VITE_PLAYWRIGHT_CSS junk

packages/code-studio/src/AppRoot.tsx
bmingles: I believe this gets statically replaced and then stripped out at build time if the condition resolves to `false`, so may not be an issue to leave this here.

bmingles:
@dsmmcken tests pass locally for me now

tests/docker-scripts/docker-compose.yml
```diff
@@ -24,15 +41,15 @@ services:
...
+    entrypoint: '/bin/bash --login -i -c "npm run e2e -- --config=playwright-ci.config.ts"'
     depends_on:
       dhc-server:
         condition: service_healthy
 
   web-ui-update-snapshots:
     extends:
       service: web-ui-tests
-    entrypoint: 'npm run e2e:update-snapshots -- --config=playwright-ci.config.ts'
+    entrypoint: '/bin/bash --login -i -c "npm run e2e:update-snapshots -- --config=playwright-ci.config.ts"'
```
bmingles, line 52: It seems we can no longer pass in additional args to constrain this command e.g. before I could

```sh
npm run e2e:update-ci-snapshots styleguide
```
or 

```sh
npm run e2e:update-ci-snapshots -- -g "some test name"
```

Not sure if this is the culprit, but it doesn't seem to be working anymore

mattrunyon: This is the culprit. The params just get passed to the entrypoint, so now it's passing those params at the end of the bash command instead of in the npm command.

Might be fixed by adding $@ to the end of the npm command in the quotes

tests/docker-scripts/Dockerfile
```diff
@@ -19,28 +26,20 @@ RUN fc-list : family;
 COPY .nvmrc .
 
 # Set the default shell so the correct node/npm is used after install
-# https://stackoverflow.com/a/60137919
+# Based on https://stackoverflow.com/a/60137919
```
bmingles, line 29: I think I found away to avoid the interactive shell here:

```dockerfile
SHELL ["/bin/bash", "--login", "-c"]
# Nvm appends its config to `.bashrc` if it exists, but the default `.bashrc` 
# bails if not in an interactive shell, which causes `nvm` to never actually get
# configured. To avoid having to use `-i` interactive shell, we overwrite the
# `.bashrc` file with a version that doesn't bail.
# https://github.com/nvm-sh/nvm/issues/1985#issuecomment-813189002
RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash
RUN echo 'export NVM_DIR="$HOME/.nvm"' > /root/.bashrc
RUN echo 'source "$NVM_DIR/nvm.sh" --install' >> /root/.bashrc
RUN source /root/.bashrc
RUN nvm install
```

tests/docker-scripts/docker-compose.yml
bmingles: I've tried this a few different ways but can't get the $@ args to get included

tests/docker-scripts/Dockerfile
```diff
@@ -1,44 +1,55 @@
...
-COPY package.json package-lock.json skip.js ./
-# cd first so the cp doesn't include /tmp/web-client-ui in the paths
-RUN cd /tmp/web-client-ui && cp --parents ./packages/*/package.json /work/
 
-FROM playwright AS build
-WORKDIR /work/
+# Now clobber the default node installed by the playwright
+# image so that the commands in docker compose entrypoint will 
+# also use the correct version of node without running in bash
+RUN ln -s $(which node) /usr/local/bin/nodejs
```
bmingles, line 42: Kudos. This simplifies things.

bmingles (approved):
Tests are passing in CI and locally except for a timeout when I ran everything. Re-running the test in isolation passed, so I think we are good to go.

packages/code-studio/src/AppRoot.tsx
bmingles: I'm pretty sure this actually gets removed from the prod bundle. The env check should get statically replaced in the build, and tree shaking should remove it. Based on some testing, I think it still exists in the source map but gets removed from the .js code.

mattrunyon (approved):

mofojed (approved):

PR #2010, opened 2024-05-13
feat: update @vscode/codicons to 0.36

dsmmcken:
Exposes several new icons. Added ability to exclude icons from vscode. 

Exclude `repo-fetch` due to it not being convertible to a single shape path with svgo.

Rounded rects don't convert with https://github.com/svg/svgo/blob/main/plugins/convertShapeToPath.js


packages/icons/scripts/build.js
```diff
@@ -47,7 +51,12 @@ async function getFiles(src) {
   const files = await getFilesInFolder(src.dir);
   const contents = await Promise.all(
     files
-      .filter(file => file.slice(file.length - 1) !== '.svg')
+      .filter(
+        file =>
+          // file is svg and not in excluded list
+          file.slice(file.length - 1) !== '.svg' &&
```
mofojed, line 57: Uhhhh this expression `file.slice(file.length - 1) !== '.svg'` will _always_ be true. Should be able to just do `.endsWith('.svg')` instead, or `file.slice(-4)`.

mofojed (changes requested):

packages/icons/scripts/build.js
dsmmcken: shows that line is entirely unnecessary, will fix though instead of deleting incase that ever changes.

Issue #1505, opened 2023-09-12
PouchStorageTable using incorrect $ne operator


bmingles:
**Description**

The operator used in `packages/pouch-storage/src/PouchStorageTable.ts` `makePouchFilter` for `FilterType.notEq` is incorrect. 

We may also want to update the return type to use `PouchDB.Find.ConditionOperators`, but there is a mismatch in the `$regex` type as well that will need adjusting. It's possible that the RegExp values are stringified internally. We could possibly do this explicitly to satisfy the types.

**Expected results**

`return { $ne: value };` should be specified

**Actual results**

`return { $neq: value };` is specified

PR #2011, opened 2024-05-14
fix: PouchStorageTable using incorrect $ne operator

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1505

Changes:
- Modified `PouchFilter` type to use $ne instead of $neq
- Modified `$regex` prop to be able to accept `RegExp` and `string`, rather than just string by default

Steps Moving Forward:
- Code change is implemented, having difficulty with imports when trying to run Unit Tests (did not commit the file yet)

Update: Apparently, there are some issues with PouchDB and jest tests, since it has not been updated to an ES6 module. (https://stackoverflow.com/questions/64905287/import-error-when-running-jest-tests-that-use-the-pouchdb-asyncstorage-adapter). Could create a `babel.config.js`, but not sure if we want to do that just for a function that is never used beyond an internal scope. Thus, going to **omit** unit tests for now.

### DIFF
--- a/packages/pouch-storage/src/PouchStorageTable.test.ts
+++ b/packages/pouch-storage/src/PouchStorageTable.test.ts
@@ -1,0 +1,26 @@
+import { makePouchFilter } from './PouchStorageTable';
+
+describe('PouchStorageTable - Filter Functions', () => {
+  describe('makePouchFilter', () => {
+    it.each([
+      ['eq', 'value', { $eq: 'value' }],
+      ['notEq', 'value', { $ne: 'value' }],
+      ['greaterThan', 'value', { $gt: 'value' }],
+      ['greaterThanOrEqualTo', 'value', { $gte: 'value' }],
+      ['lessThan', 'value', { $lt: 'value' }],
+      ['lessThanOrEqualTo', 'value', { $lte: 'value' }],
+    ])(
+      'should create a PouchFilter for %s operator',
+      (type, value, expected) => {
+        const filter = makePouchFilter(type, value);
+        expect(filter).toEqual(expected);
+      }
+    );
+
+    it('should throw an error for unsupported filter types', () => {
+      expect(() => makePouchFilter('unsupported', 'value')).toThrow(
+        'Unsupported type: unsupported'
+      );
+    });
+  });
+});
--- a/packages/pouch-storage/src/PouchStorageTable.test.ts
+++ b/packages/pouch-storage/src/PouchStorageTable.test.ts
@@ -9,6 +9,9 @@ describe('PouchStorageTable - Filter Functions', () => {
       ['greaterThanOrEqualTo', 'value', { $gte: 'value' }],
       ['lessThan', 'value', { $lt: 'value' }],
       ['lessThanOrEqualTo', 'value', { $lte: 'value' }],
+      ['startsWith', 'val', { $regex: `/^val.*/` }],
+      ['contains', 'value', { $regex: '/value/' }],
+      ['inIgnoreCase', ['value1', 'value2'], { $regex: '/value1,value2/i' }],
     ])(
       'should create a PouchFilter for %s operator',
       (type, value, expected) => {
--- a/packages/pouch-storage/src/PouchStorageTable.ts
+++ b/packages/pouch-storage/src/PouchStorageTable.ts
@@ -43,8 +43,7 @@ type PouchFilter = OnlyOneProp<{
   $lt: FilterValue | FilterValue[];
   $lte: FilterValue | FilterValue[];
   $regex: RegExp | string;
-}> &
-  Omit<PouchDB.Find.ConditionOperators, '$regex'>;
+}>;
 
 export function makePouchFilter(
   type: string,
--- a/packages/pouch-storage/src/PouchStorageTable.ts
+++ b/packages/pouch-storage/src/PouchStorageTable.ts
@@ -35,15 +35,12 @@ export type PouchDBSort = Array<
   string | { [propName: string]: 'asc' | 'desc' }
 >;
 
-type PouchFilter = OnlyOneProp<{
-  $eq: FilterValue | FilterValue[];
-  $ne: FilterValue | FilterValue[];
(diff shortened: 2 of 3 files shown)


AkshatJawne:
Geting Playwright failures here, not sure if this is related to the issues that were being previously discussed last week?

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -35,20 +35,18 @@ export type PouchDBSort = Array<
...
   $lt: FilterValue | FilterValue[];
   $lte: FilterValue | FilterValue[];
-  $regex: RegExp;
-}>;
+  $regex: RegExp | string;
+}> &
+  Omit<PouchDB.Find.ConditionOperators, '$regex'>;
 
-function makePouchFilter(
+export function makePouchFilter(
```
bmingles, line 49: @AkshatJawne are there unit tests for this?

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -35,20 +35,18 @@ export type PouchDBSort = Array<
...
+  $ne: FilterValue | FilterValue[];
   $gt: FilterValue | FilterValue[];
   $gte: FilterValue | FilterValue[];
   $lt: FilterValue | FilterValue[];
   $lte: FilterValue | FilterValue[];
-  $regex: RegExp;
-}>;
+  $regex: RegExp | string;
+}> &
+  Omit<PouchDB.Find.ConditionOperators, '$regex'>;
```
bmingles, line 47: I was thinking about this after I suggested it, and I don't think the intersection type is actually providing anything useful. Namely, if you change `$ne` back to `$neq`, the compiler doesn't complain, so it's not really protecting anything. I would just remove the `& Omit<PouchDB.Find.ConditionOperators...` piece.

packages/pouch-storage/src/PouchStorageTable.ts
AkshatJawne: Created unit tests, but as I mentioned above, there are some issues with PouchDB I believe where Jest cannot understand the import, can speak to you about this offline. 

Essentially, from what I can find, the viable solution is creating a `babel.config.js `and doing the steps outlined here https://stackoverflow.com/questions/54461150/jest-not-parsing-es6-syntaxerror-unexpected-token-import . 

I found this from the comment thread of the original issue outlining difficulties using a different type of PouchDB storage https://stackoverflow.com/questions/64905287/import-error-when-running-jest-tests-that-use-the-pouchdb-asyncstorage-adapter , but again, not sure if that is what we want to do here given scope of code.

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -35,20 +35,17 @@ export type PouchDBSort = Array<
...
   $eq: FilterValue | FilterValue[];
-  $neq: FilterValue | FilterValue[];
+  $ne: FilterValue | FilterValue[];
   $gt: FilterValue | FilterValue[];
   $gte: FilterValue | FilterValue[];
   $lt: FilterValue | FilterValue[];
   $lte: FilterValue | FilterValue[];
-  $regex: RegExp;
+  $regex: RegExp | string;
 }>;
```
mofojed, line 46: Here we should use the `ConditionOperators` types defined in `PouchDB` themselves instead of redefining them:

```
type PouchFilter = OnlyOneProp<
  Pick<PouchDB.Find.ConditionOperators, '$eq' | '$ne' | ...etc... >
>;

Note that the `RegExp` type is defined as a `string`, so we'll need to modify the switch-case below to `.toString()` the `RegExp`s (I think this is implicitly being done somewhere right now, we should be explicit with it though since this type doesn't match currently).

On a side note, the types for some of the operators like `$gt` are just `any`. Ideally those operators would have a more strict type, but we'll leave that up to the pouchdb-find package to enforce.
```

mofojed (changes requested):

packages/pouch-storage/src/PouchStorageTable.ts
bmingles: @mofojed I was the one who encouraged this. The `ConditionOperators` type uses a number of `any` types such that we lose type safety.

packages/pouch-storage/src/PouchStorageTable.test.ts
```diff
@@ -0,0 +1,26 @@
+import { makePouchFilter } from './PouchStorageTable';
+
+describe('PouchStorageTable - Filter Functions', () => {
+  describe('makePouchFilter', () => {
+    it.each([
```
bmingles, line 5: This doesn't seem to include all of the operators?

AkshatJawne:
@bmingles Have added the test cases for the regex types, however,a fter adding them I kept getting this error: 
image

Figured that there was an issue with our old regex (but may have never been picked up given the any type), so I modified it. Let me know what you think about this

packages/pouch-storage/src/PouchStorageTable.ts
```diff
@@ -72,7 +65,7 @@ function makePouchFilter(
     case FilterType.lessThanOrEqualTo:
       return { $lte: value };
     case FilterType.startsWith:
-      return { $regex: new RegExp(`^(?${value}).*`) };
+      return { $regex: new RegExp(`^${value}.*`).toString() };
```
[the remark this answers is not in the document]
AkshatJawne: @bmingles Did testing, appears to be working fine. Tried the "starts with" as an isolated filter, and then also in conjunction with other filters, and appears to be working normally. Attached screenshots of some of my tests:

Screenshot 2024-05-15 at 8 51 57 AM
Screenshot 2024-05-15 at 8 52 40 AM
Screenshot 2024-05-15 at 8 53 50 AM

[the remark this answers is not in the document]
AkshatJawne: Oh okay, yeah that makes sense, sorry about that.

bmingles (approved):
LGTM

mofojed (approved):

Issue #1970, opened 2024-04-29
iris grid partition selector can only display 1000 partitions


dsmmcken:
```
from deephaven import empty_table
t = empty_table(15000).update(["x = i", "y = i"]).partition_by(["x"])
```

Viewing multiple years of a daily partitioned table can easily exceed the 1000 item limit. The partition select is limited by us to 1000 items for performance. (And is capped by the browser at 10000 in chrome).

We should switch the selector to use the spectrum select which is virtualized and can handle a much higher number of rows (hundreds of thousands).

mofojed:
Should use the `Picker` component from `@deephaven/jsapi-components`.

PR #2012, opened 2024-05-14
fix: Use picker for iris grid partition selector 

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1970

**Changes Implemented:**
- Replace TableDropdown with Picker for the Partition Selector for Iris Grid
- Change corresponding test, since role is no longer "combobox"

**Testing Information:**

Code used:
```
t = empty_table(15000).update(["x = i", "y = i"]).partition_by(["x"]) 
```

Sample visual of new Picker:
image

### DIFF
--- a/packages/iris-grid/src/IrisGridPartitionSelector.tsx
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.tsx
@@ -306,13 +306,13 @@ class IrisGridPartitionSelector extends Component<
         {partitionTables?.[index] && (
           <Picker
             label={column.name}
+            labelPosition="side"
             table={partitionTables[index]}
             direction="bottom"
             shouldFlip={false}
             keyColumn={partitionTables[index].columns[index].name}
             placeholder={'Loading...' as string}
             labelColumn={partitionTables[index].columns[index].name}
-            labelPosition="side"
             onChange={this.getCachedChangeCallback(index)}
             defaultSelectedKey={partitions[index] as ItemKey}
             isDisabled={
--- a/packages/iris-grid/src/IrisGridPartitionSelector.tsx
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.tsx
@@ -1,10 +1,10 @@
 import React, { Component } from 'react';
 import memoizee from 'memoizee';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Button } from '@deephaven/components';
+import { Button, ItemKey } from '@deephaven/components';
 import { vsChevronRight, vsMerge, vsKey } from '@deephaven/icons';
 import Log from '@deephaven/log';
-import { TableDropdown } from '@deephaven/jsapi-components';
+import { Picker } from '@deephaven/jsapi-components';
 import type { dh } from '@deephaven/jsapi-types';
 import { TableUtils } from '@deephaven/jsapi-utils';
 import { assertNotNull, Pending, PromiseUtils } from '@deephaven/utils';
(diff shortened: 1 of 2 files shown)


dsmmcken (changes requested):
Spectrum picker doesn't have a small size, so we should remove the `btn-sm` classes from the merge and keys buttons

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -294,23 +294,31 @@ class IrisGridPartitionSelector extends Component<
...
+            keyColumn={partitionTables[index].columns[index].name}
+            placeholder={'Loading...' as string}
+            labelColumn={partitionTables[index].columns[index].name}
+            onChange={this.getCachedChangeCallback(index)}
+            defaultSelectedKey={partitions[index] as ItemKey}
+            isDisabled={
+              (index > 0 && partitionConfig.mode !== 'partition') || isLoading
+            }
+          />
+        )}
```
mofojed, line 321: As this is used currently, there's a warning logged in console:
```
If you do not provide a visible label, you must specify an aria-label or aria-labelledby attribute for accessibility 
    at $51664a9007a3c6f9$var$Picker (http://localhost:4000/ide/node_modules/.vite/deps/@adobe_react-spectrum.js?v=8e48b768:36010:57)
    at PickerNormalized (http://localhost:4000/ide/@fs/home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/components/src/spectrum/picker/PickerNormalized.tsx:23:40)
    at Picker (http://localhost:4000/ide/@fs/home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/jsapi-components/src/spectrum/Picker.tsx:21:26)
    at div
    at div
    at IrisGridPartitionSelector (http://localhost:4000/ide/@fs/home/bender/dev/deephaven/iris-oss/web-client-ui/main/packages/iris-grid/src/IrisGridPartitionSelector.tsx?t=1715716828868:310:9)
```
It's telling us that we're not labelling this component correctly. Instead, the label should be set on the Picker itself, instead of displayed separately in the `column-name` div we have outside of it. Add `label={column.name}` and `labelPosition="side"` to add the label to the `Picker` directly.

With that there will be a slight difference in styles:
image
image

Just want to make sure that extra width is okay, @dsmmcken may have a comment about that. 
Test with multiple partitions as well just to make sure it's still groovy:

```
t = empty_table(15000).update(["x = i", "y = i", "z=i"])
pt = t.partition_by(["x"])
pt2 = t.partition_by(["x", "y"])
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridPartitionSelector.tsx
dsmmcken: label difference is fine with me

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -303,15 +303,16 @@ class IrisGridPartitionSelector extends Component<
...
         {partitionTables?.[index] && (
           <Picker
+            label={column.name}
             table={partitionTables[index]}
             direction="bottom"
             shouldFlip={false}
             keyColumn={partitionTables[index].columns[index].name}
             placeholder={'Loading...' as string}
             labelColumn={partitionTables[index].columns[index].name}
+            labelPosition="side"
```
mofojed, line 315: I would've probably grouped the `label` and `labelPosition` props together (since they're both labelling the component), but :shrug:

mofojed (dismissed):

packages/iris-grid/src/IrisGridPartitionSelector.tsx
AkshatJawne: My bad, I thought I should have `labelColumn` and `labelPosition` together, but what you have mentioned makes sense, since the `labelColumn` is referring to something else. Should be good now

mofojed (approved):

AkshatJawne:
@dsmmcken Could you take another look at this whenever you have the time, so that I can merge?

dsmmcken (approved):

AkshatJawne:
Thanks Don, merging now!

PR #2013, opened 2024-05-14
feat: Refactor console objects menu

### LINKED ISSUE #1884: deephaven.ui widgets do not appear in Console dropdowns
**Description**

When you create deephaven.ui components, they do not appear under the Console table or widget dropdowns. They're only under the Panels menu.

**Steps to reproduce**

1. Create a deephaven.ui widget in the console:
```
from deephaven import ui

foo = ui.text("Bar")
```
2. Check the table/figure dropdowns in the Console panel

**Expected results**
2. `foo` should appear in a dropdown and be openable

**Actual results**
2. `foo` does not appear in either dropdown. It only appears under "Panels"

We should consider having one dropdown for all the objects, with a filter in that dropdown for type? The dropdown is redundant of the Panels menu, get rid of it entirely (for Core)?

**Versions**
Engine Version: 0.34.0-SNAPSHOT
Web UI Version: 0.69.1
Java Version: 11.0.22
Barrage Version: 0.6.0
Browser Name: Chrome 122
OS Name: Linux

mofojed:
- Split ConsoleMenu up so objects is it's own component ConsoleObjectsMenu
- Convert to functional component
- Display all widgets in the menu
- Add a flag to hide the menu, which we set for DHC, but will not be set in DHE so current behaviour is not broken there
- Fixes #1884


packages/console/src/ConsoleObjectsMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+            setFilterText(e.target.value);
+          }}
+          className="console-menu"
+          ref={searchRef}
+        />
+      ),
+    };
+    const filteredObjects = filterText
+      ? objects.filter(
+          ({ title }: { title?: string }) =>
```
mattrunyon, line 43: Don't think you need to specify the type here. The JSAPI types say it's always defined and the source looks to confirm that

packages/console/src/ConsoleObjectsMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+            title != null &&
+            title.toLowerCase().indexOf(filterText.toLowerCase()) >= 0
```
mattrunyon, line 45:
```suggestion
            title.toLowerCase().includes(filterText.toLowerCase())
```

packages/console/src/ConsoleObjectsMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+            transform="right-5 down-5"
+          />
+          <FontAwesomeIcon icon={vsTriangleDown} transform="right-8 down-6" />
+        </div>
+      }
+    >
+      <DropdownMenu
+        actions={actions}
+        onMenuOpened={() => searchRef.current?.focus()}
+        onMenuClosed={() => setFilterText('')}
```
mattrunyon, line 81: I wonder if we should leave the filter and just have the search select the search text when it's reopened. I guess if this is the behavior it has always been and nobody has complained then it's inconsequential. I was just thinking if I'm searching maybe I wanted to open several matching the same search or something

packages/console/src/ConsoleStatusBar.tsx
```diff
@@ -1,136 +1,110 @@
...
-  handleCommandCompleted(error: unknown): void {
-    // Don't update state if the promise was canceled
-    if (!(error instanceof CanceledPromiseError)) {
-      this.setState({
-        isCommandRunning: this.pending.pending.length > 0,
-      });
+    try {
+      await result;
+    } catch (error) {
+      // No-op, fall through
```
mattrunyon, line 44: Why the change to the logic here? Previously looks like we did something w/ the error

packages/console/src/ConsoleStatusBar.tsx
```diff
@@ -1,136 +1,110 @@
...
+  let tooltipText = null;
+  if (isCommandRunning) {
+    // Connected, Pending
+    statusIconClass = 'console-status-icon-pending';
+    tooltipText = 'Worker is busy';
+  } else {
+    // Connected, Idle
+    statusIconClass = 'console-status-icon-idle';
+    tooltipText = 'Worker is idle';
+  }
```
mattrunyon, line 77: No disconnected state after the refactor? This component is consumed by DHE, but I'm not sure how to test/get the disconnected state there.

packages/console/src/ConsoleStatusBar.tsx
```diff
@@ -1,136 +1,110 @@
...
+      {children}
+      {hasActions && (
+        <Button
+          kind="ghost"
+          icon={vsKebabVertical}
+          tooltip="More Actions..."
+          aria-label="More Actions..."
+          onClick={() => {
+            // no-op: click is handled in `DropdownMenu`
+          }}
```
mattrunyon, line 98: EMPTY_FUNCTION

mattrunyon (changes requested):

packages/console/src/ConsoleObjectsMenu.tsx
mofojed: Yeah, could have it on a timer to reset it to blank after a few seconds. I could have sworn we had that logic on the query widget list on Enterprise, but I can't find it.

I'll just have it select text on open, that's how the QueryWidgetList behaves on Enterprise now.

packages/console/src/ConsoleStatusBar.tsx
mofojed: Actually nevermind, `isDisconnected` isn't used at all in this component. It's always `false`, and it never gets set anywhere. That's why I removed it, it was never used.

packages/console/src/ConsoleStatusBar.tsx
mofojed: Before we just used the error to check the pending count. This just cleans up the logic.
I did change it to use a count instead of a flag in case of multiple commands pending.

packages/console/src/ConsoleStatusBar.tsx
```diff
@@ -1,136 +1,109 @@
...
+        dh.IdeSession.EVENT_COMMANDSTARTED,
+        handleCommandStarted
+      );
+
+      return function stopListening() {
+        session.removeEventListener(
+          dh.IdeSession.EVENT_COMMANDSTARTED,
+          handleCommandStarted
+        );
+      };
```
mattrunyon, line 63: Looks like `session.addEventListener` returns a cleanup function we could use instead

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -440,6 +440,7 @@ export class ConsolePanel extends PureComponent<
                 />
               </>
             }
+            // showObjectsMenu={false}
```
mattrunyon, line 443: Guessing this is unintentional

mattrunyon (changes requested):

mattrunyon (approved):

Issue #1777, opened 2024-02-06
Markdown widgets do not persist refresh if they were not edited


mattrunyon:
**Description**

Markdown widgets only seem to get saved in the layout if they are edited. If they are not, their space is replaced w/ blank space in the layout on page reload

**Steps to reproduce**

1. Create a markdown widget from the controls panel.
2. Refresh the page

**Expected results**

3. Markdown panel reloads where it was

**Actual results**

3. Markdown panel space is blank and no panel is loaded (bottom right quadrant in this screenshot had a markdown panel before reload)
![image](https://github.com/deephaven/web-client-ui/assets/8619789/f37d10f5-cc78-4733-899e-82d14e81059b)

**Versions**

Engine Version: 0.33.0-SNAPSHOT
Web UI Version: 0.62.0
Java Version: 11.0.21
Barrage Version: 0.6.0
Browser Name: Chrome 121
OS Name: Linux

PR #2019, opened 2024-05-15
fix: Unedited markdown widgets not persisting

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1777

Changes Implemented:
- Removed check to return null if the content is the default content (meaning it is unedited), which was resulting in the window disappearing on refresh


packages/dashboard-core-plugins/src/MarkdownPlugin.tsx
```diff
@@ -31,8 +31,7 @@ export function MarkdownPlugin(props: MarkdownPluginProps): JSX.Element | null {
       title == null ||
       panelState == null ||
       panelState.content == null ||
-      panelState.content.length === 0 ||
-      panelState.content === MarkdownUtils.DEFAULT_CONTENT
+      panelState.content.length === 0
```
mofojed, line 34: This is okayyyyy, but _ideally_ we don't want to save any data still, but we want the markdown widget when re-hydrated to display the default text ("This note can be edited..."). So when rehydrating, it should recognize there is no data, and display the default text again. That way, if we wanted to display something different for that default text (such as adding a blurb about double-clicking to edit, or change the styling or something), the component would display that newly defined default if it still hadn't been edited instead of the old default.

This is a fairly extreme edge case with marginal utility though, but still a good exercise for working with dehydration/hydration.

packages/dashboard-core-plugins/src/MarkdownPlugin.tsx
```diff
@@ -27,16 +27,27 @@ export function MarkdownPlugin(props: MarkdownPluginProps): JSX.Element | null {
...
+          metaData: {},
+          panelState: {
+            content: MarkdownUtils.DEFAULT_CONTENT,
+          },
+        },
+        localDashboardId: configProps.id,
+        title: MarkdownUtils.DEFAULT_TITLE,
+      };
+      return hydrate(configNew);
+    }
```
mofojed, line 50: We should never be calling `hydrate` from within a dehydration function.

Dehydration is taking a panel and outputting a serializable object such that we can serialize the layout and save it; hydration is the opposite, where we take a serialized layout, and hydrate it into objects (such as a grid model).

When we're dehydrating the layout, we check if the returned dehydration is `null`, and if it is, we omit the component from the layout completely: https://github.com/deephaven/web-client-ui/blob/6cf124012a19b122b3a96f620886b970c12b6d29/packages/dashboard/src/layout/LayoutUtils.ts#L340

In this case though, we _do_ want to save the panel to the layout, we just don't want to save the actual content with it. I don't think there's a case where we would want to drop it completely from the layout (which returning `null` does).

Digging into it a bit ... I think you should actually remove this custom `dehydrateMarkdown` function completely, and just use the default hydrate/dehydrate functions (omitting the last two parameters from the call to `registerComponent` below). Then need to change `MarkdownPanel.getClosedMarkdowns` to filter out any panel that doesn't have a markdown set.

Actually found another bug too - if we return `null` from the dehydration function (as we're doing above), PanelManager throws an error:
```
Error while emitting event: TypeError: Cannot set properties of null (setting 'parentStackId')
    at PanelManager.addClosedPanel (PanelManager.ts:365:54)
    at PanelManager.handleClosed (PanelManager.ts:347:10)
    at EventHub.emit (EventEmitter.ts:69:22)
```
You can see this by just opening and closing a markdown panel before your changes. We should have a check there for if the dehydrated component is `null`, and if it is, skip pushing it onto the `closed` array.

mofojed (changes requested):

packages/dashboard-core-plugins/src/MarkdownPlugin.tsx
AkshatJawne: Thanks for the information on hydration was very helpful! Will make the changes and will take a look at the other bug you mentioned

packages/dashboard-core-plugins/src/MarkdownPlugin.tsx
```diff
@@ -52,16 +32,12 @@ export function MarkdownPlugin(props: MarkdownPluginProps): JSX.Element | null {
...
-        MarkdownPanel.COMPONENT
-      );
       const usedTitles = openedMarkdowns.map(markdown => markdown.title ?? '');
       const panelTitle =
         title != null && title !== ''
           ? title
           : MarkdownUtils.getNewMarkdownTitle(usedTitles);
-      const content =
-        closedMarkdowns.length > 0 ? null : MarkdownUtils.DEFAULT_CONTENT;
+      const content = MarkdownUtils.DEFAULT_CONTENT;
```
mofojed, line 40: We don't want to do this - right now this change is causing a regression, which is re-opening closed markdown widgets.

If you open a markdown panel, edit it to add some text, then close it, and then open a new markdown panel, you get a starting page prompting you to re-open one of the previously closed markdown panels or to just create a new one:
image

I think what you _should_ do is just set `null` for content from here, then in MarkdownPanel is where it will actually decide whether to show the starting page or the default content (when it calls `getClosedMarkdowns`).

mofojed (changes requested):

packages/dashboard-core-plugins/src/MarkdownPlugin.tsx
AkshatJawne: Just pushed changes, if I understood correctly when you said 

> then in MarkdownPanel is where it will actually decide whether to show the starting page or the default content (when it calls getClosedMarkdowns).

It now works such that on initial opening of a Markdown, you get this panel, given that when `getClosedMarkdowns` is called in MarkdownPanel, the condition for the starting page appearing is whether the state isStatePageShown is set to true:
image


However, after refreshing after editing, we see the default content: 

image

AkshatJawne: *** Something seems to be breaking, need to inspect, will update ***

AkshatJawne:
Looking into e2e failures

AkshatJawne:
Opted to not create utility for filtering given that it was only used in one spot, and I would just be creating that function and calling it inside the memoization in the `MarkdownPanel`

AkshatJawne:
Still failing on e2e with Mathjax, investigating why

packages/dashboard-core-plugins/src/panels/MarkdownPanel.tsx
```diff
@@ -123,10 +128,6 @@ export class MarkdownPanel extends Component<
     }
   }
 
-  getClosedMarkdowns = memoize((closedPanels: ClosedPanels) =>
-    closedPanels.filter(panel => panel.component === 'MarkdownPanel').reverse()
-  );
-
```
mofojed, line 129: We still want to memoize this, but it can call into the `MarkdownUtils.getClosedMarkdowns` method

packages/dashboard-core-plugins/src/controls/markdown/MarkdownUtils.ts
```diff
@@ -19,6 +22,16 @@ class MarkdownUtils {
 
     return title;
   }
+
+  static getClosedMarkdowns = memoize((closedPanels: ClosedPanels) =>
+    closedPanels
+      .filter(
+        panel =>
+          panel.component === 'MarkdownPanel' &&
+          panel.props.panelState.content !== ''
```
mofojed, line 31: Also need to filter out `props.panelState.content !== null`.

Right now if I open a Markdown, don't edit it, close it, then open another Markdown, it's showing the start page when it shouldn't.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/MarkdownPanel.tsx
```diff
@@ -128,6 +130,10 @@ export class MarkdownPanel extends Component<
     }
   }
 
+  getClosedMarkdowns = memoize((closedPanels: ClosedPanels) =>
+    MarkdownUtils.getClosedMarkdowns(closedPanels)
+  );
```
mofojed, line 135: I think you could in theory just shorten this since it's a static function and you're just passing the args through, but no big deal:
```suggestion
  getClosedMarkdowns = memoize(MarkdownUtils.getClosedMarkdowns);
```

mofojed (approved):

packages/dashboard-core-plugins/src/panels/MarkdownPanel.tsx
AkshatJawne: Just changed it, I agree, looks a lot cleaner.

AkshatJawne:
@mofojed changed the getClosedMarkdowns, but the approving review has gone stale, so if you could approve it again after reviewing, that would be great!

mofojed (dismissed):

mofojed (approved):

Issue #2020, opened 2024-05-15
Replace DH `RadioGroup` with Spectrum version


bmingles:
We want to replace our `@deephaven/components` `RadioGroup` component wit Spectrum's

- Delete `RadioGroup` and `RadioItem` components
- Re-export Spectrum `RadioGroup` and `Radio` compnents
- Update all references to DH `RadioGroup`

PR #2021, opened 2024-05-15
feat: Replaced `RadioGroup` with Spectrum's (#2020)

bmingles:
- Removed DH `RadioGroup` and `RadioItem` components
- Re-exported Spectrum `RadioGroup` and `Radio` components
- Updated references to use new version

Here's diffs of what the components look like after migrating to Spectrum RadioGroup:

**FilterSetManager**
image

**ChartBuilder**
image

**TableCsvExporter**
image

**Aggregations**
image

BREAKING CHANGE: `RadioGroup` has been replaced by Spectrum `RadioGroup`. `RadioItem` has been replaced by Spectrum `Radio`


packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -186,18 +183,23 @@ function Inputs(): React.ReactElement {
...
+              value={radioValue}
+            >
+              <Radio value="1">Toggle this custom radio</Radio>
+              <Radio value="2">Or toggle this other custom radio</Radio>
+              <Radio value="3" isDisabled>
                 Disabled radio
-              </RadioItem>
-              <RadioItem value="4" isInvalid>
+              </Radio>
+              <Radio value="4" UNSAFE_className="is-invalid">
```
mofojed, line 197: I don't think we should do this in the styleguide - it doesn't seem like a real use case, and we aren't doing it anywhere.
In the Spectrum component, you can set `isInvalid` on the `RadioGroup` itself, not an individual option, which makes more sense IMO - if a single option is invalid, then it should just be disabled.
Also we can show off the error message functionality as well, and description:
```

            <RadioGroup
              aria-labelledby="inputs-radios-heading"
              onChange={handleRadioChange}
              value={radioValue}
              isInvalid={radioValue === '4'}
              description="Select a radio item"
              errorMessage={
                radioValue === '4' ? 'Invalid radio selected' : undefined
              }
            >
              <Radio value="1">Toggle this custom radio</Radio>
              <Radio value="2">Or toggle this other custom radio</Radio>
              <Radio value="3" isDisabled>
                Disabled radio
              </Radio>
              <Radio value="4">Invalid radio</Radio>
              <Radio isHidden={!check4} value="5">
                Extra radio item
              </Radio>
            </RadioGroup>
```

mofojed (changes requested):
Just want to clean up the example in the styleguide

packages/code-studio/src/styleguide/Inputs.tsx
bmingles: Fixed

packages/code-studio/src/styleguide/Inputs.tsx
```diff
@@ -188,15 +188,18 @@ function Inputs(): React.ReactElement {
               aria-labelledby="inputs-radios-heading"
               onChange={handleRadioChange}
               value={radioValue}
+              isInvalid={radioValue === '4'}
+              description="Select a radio item"
+              errorMessage={
+                radioValue === '4' ? 'Invalid radio selected' : undefined
+              }
```
mofojed, line 195: Actually I don't think you necessarily need a ternary here, as the error message is only shown when it's invalid anyway. But no matter.
```suggestion
              errorMessage="Invalid radio selected"
```

mofojed (approved):

Issue #1256, opened 2023-05-01
Handle reconnect auth fail in the embed-widget app


mofojed:
**Description**

Right now if the JS API throws a reconnect auth fail error, the embed-widget (embed-grid/embed-chart deprecated) packages do not handle it. The change #1149 to wire up the reconnect auth fail event only added it to `code-studio` package, so the `embed-widget` package would not react the same.

Possibly add it to `AppBootstrap` or some other common component between the three packages.

Follow these steps to reproduce an auth failure:
1. Start up server with deephaven-core
2. Use ngrok to start a tunnel to that port, e.g.: ngrok http 10000
3. Start up Web UI connecting to that tunnel, e.g.: VITE_CORE_API_URL=http://acfc-23-233-0-34.ngrok.io/jsapi npm start
4. Open the Web UI in Firefox, run some commands to make sure initial connection is fine.
5. Press Alt and from the File menu, select "Work Offline"
6. See it transition to a disconnected state. Disconnected message should appear and should not be able to enter new commands
7. Reconnect by deselecting "Work Offline" option from Step 5
8. See it transition to connected state. Should be able to run commands and have the results appear.
9. Kill the server (Ctrl+C). See it transition to a Shutdown state, app unloaded.

mofojed:
Follow these steps to reproduce an auth failure:
1. Start up server with deephaven-core
2. Use ngrok to start a tunnel to that port, e.g.: ngrok http 10000
3. Start up Web UI connecting to that tunnel, e.g.: VITE_CORE_API_URL=http://acfc-23-233-0-34.ngrok.io/jsapi npm start
4. Open the Web UI in Firefox, run some commands to make sure initial connection is fine.
5. Press Alt and from the File menu, select "Work Offline"
6. See it transition to a disconnected state. Disconnected message should appear and should not be able to enter new commands
7. Reconnect by deselecting "Work Offline" option from Step 5
8. See it transition to connected state. Should be able to run commands and have the results appear.
9. Kill the server (Ctrl+C). See it transition to a Shutdown state, app unloaded.

AkshatJawne:
**_Update:_** 

After trying and testing different approaches, feel that the approach where I add an event listener and corresponding functions to handle the Auth Failure event in `ConnectionBootstrap.tsx` is a viable approach. 

[...]

PR #2023, opened 2024-05-16
fix: Reconnect Auth Fail Fix - embed-widget

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1256

**Changes Implemented:** 
- Refactored disconnection, shutdown, and auth fail handling logic to now be `ConnectionBootstrap` in order to properly handle auth fail event in the `embed-widget` 

**Visual of modal that appears on auth fail:** 
image


AkshatJawne:
@dsmmcken Failing webkit test here, should I try and rerun it like you did for the other ticket?

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -1,12 +1,13 @@
...
+import { BasicModal } from '@deephaven/components';
 import {
   ObjectFetcherContext,
   sanitizeVariableDescriptor,
   useApi,
   useClient,
 } from '@deephaven/jsapi-bootstrap';
-import type { dh } from '@deephaven/jsapi-types';
+import dh from '@deephaven/jsapi-shim';
+import type { dh as DhType } from '@deephaven/jsapi-types';
```
mattrunyon, line 10: `@deephaven/jsapi-shim` shouldn't be used here. It's an old glue package that we should basically never be using (definitely not adding uses of) in DHC.

Previously `dh` was a globally provided module when you connected to the server. We used the shim which is just `export dh = window.dh`. This is still the case, but we have moved to `dh` being a module from the server that doesn't have to be global.

Instead here, use `api` from the `useApi` hook. That is the JS API and the same as `dh`.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -29,8 +30,10 @@ export function ConnectionBootstrap({
 }: ConnectionBootstrapProps): JSX.Element {
   const api = useApi();
   const client = useClient();
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [error, setError] = useState<unknown>();
```
mattrunyon, line 34: If it's unused then remove it. But looks like this should be used still to display when the server shuts down

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -56,47 +59,63 @@ export function ConnectionBootstrap({
     [api, client]
   );
 
-  useEffect(
-    function listenForShutdown() {
-      if (connection == null) return;
+  useEffect(() => {
+    if (connection == null) return;
```
mattrunyon, line 63: We don't have a linter rule for this, but on longer effects we prefer to provide named functions. Gives a little more insight into what the effect lifecycle is handling without needing to read the implementation. This was previously named `listenForShutdown`.

Change back to a named function or update the name if there's a better fitting name

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -56,47 +59,63 @@ export function ConnectionBootstrap({
...
-    [api, connection]
-  );
+      setIsAuthFailedState(true);
+    }
+    const authFailed = connection.addEventListener(
+      dh.CoreClient.EVENT_RECONNECT_AUTH_FAILED,
+      handleAuthFailed
+    );
+
+    return authFailedState ? authFailed : removerFn;
```
mattrunyon, line 90: These should be separate effects. The return here is not correct.

The return function of an effect is run when the effect is changed before it runs the next version of the effect. So in this case if the connection changes we are removing the old listener by returning `removerFn` and then attaching a new listener. This implementation would only remove 1 of the event listeners and leave the other attached.

Also, effects should generally only cover 1 thing/lifecycle so they are more clear.

packages/embed-widget/src/index.scss
```diff
@@ -1,14 +1 @@
 @import '@deephaven/components/scss/custom.scss';
-
-body {
```
mattrunyon, line 3: You can just remove this file. This doesn't cause any style changes between the app and embed-widget?

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -56,47 +59,63 @@ export function ConnectionBootstrap({
...
+    log.info('Refreshing application');
+    window.location.reload();
   }
 
   return (
-    <ConnectionContext.Provider value={connection}>
+    <ConnectionContext.Provider value={connection ?? null}>
       <ObjectFetcherContext.Provider value={objectFetcher}>
-        {children}
+        <>
```
mattrunyon, line 109: This previously returned a `LoadingOverlay` if the connection was null so that we didn't try to load the next things which depend on `connection`. Will need to check this doesn't effect loading or remove the spinner too soon.

That's also how we showed the server shutdown state which is not being used right now. Should add that back.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -56,47 +59,63 @@ export function ConnectionBootstrap({
...
+          {children}
+          <BasicModal
+            confirmButtonText="Refresh"
+            onConfirm={handleRefresh}
+            isOpen={authFailedState}
+            headerText="Authentication failed"
+            bodyText="Credentials are invalid. Please refresh your browser to try and reconnect."
+          />
+        </>
```
mattrunyon, line 118: The logic handling all of this should be removed from `AppMainContainer` now so it's not duplicated.

mattrunyon (changes requested):

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -32,7 +31,7 @@ export function ConnectionBootstrap({
   const client = useClient();
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [error, setError] = useState<unknown>();
-  const [connection, setConnection] = useState<DhType.IdeConnection>();
+  const [connection, setConnection] = useState<dh.IdeConnection>();
   const [authFailedState, setIsAuthFailedState] = useState<boolean>(false);
```
mattrunyon, line 35: The state and setter should have consistent names. Also you don't need the type annotation as TS can infer it's a boolean.

```suggestion
  const [isAuthFailed, setIsAuthFailed] = useState(false);
```

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -59,39 +58,53 @@ export function ConnectionBootstrap({
...
+        log.warn(
+          'Reconnect authentication failed',
+          `${JSON.stringify(detail)}`
+        );
+        setError(
+          `Reconnect authentication failed: ${detail ?? 'Unknown reason'}`
+        );
+        setIsAuthFailedState(true);
+      }
+      const authFailed = connection.addEventListener(
```
mattrunyon, line 96: Change to a more descriptive name. `addEventListener` in our API returns a function to remove the event listener

```suggestion
      const removerFn = connection.addEventListener(
```

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -966,11 +950,7 @@ export class AppMainContainer extends Component<
                   )}
                 </span>
               }
-              tooltip={
-                isDisconnected && !isAuthFailed
-                  ? 'Server disconnected'
-                  : 'User Settings'
-              }
+              tooltip="Server disconnected"
```
mattrunyon, line 953: This should still be `User Settings` if we're not disconnected.

That said, I'm not even sure it's possible to see/use this button in the disconnected state any more since we're showing a full screen loader

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -83,20 +111,24 @@ export function ConnectionBootstrap({
...
   );
 
-  if (connection == null || error != null) {
-    return (
-      <LoadingOverlay
-        data-testid="connection-bootstrap-loading"
-        isLoading={connection == null}
-        errorMessage={error != null ? `${error}` : undefined}
-      />
-    );
```
mattrunyon, line 93: This shouldn't be removed

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -1015,10 +995,7 @@ export class AppMainContainer extends Component<
           onChange={this.handleImportLayoutFiles}
           data-testid="input-import-layout"
         />
-        <DebouncedModal
-          isOpen={isDisconnected && !isAuthFailed}
-          debounceMs={1000}
-        >
+        <DebouncedModal isOpen={isDisconnected} debounceMs={1000}>
```
mattrunyon, line 998: I'm curious if this should be pulled out as well. I'd think embed-widget doesn't handle disconnect/reconnect (or doesn't display it at least) since it doesn't handle these events

mattrunyon (changes requested):

packages/app-utils/src/components/ConnectionBootstrap.tsx
AkshatJawne: If I don't remove this, then we cannot see the Authentication failed modal, since the overlay goes over it.

packages/code-studio/src/main/AppMainContainer.tsx
AkshatJawne: Not sure how this will be shown, right now it does not show up since we have the overlay and the auth failed modal on top of it, so I can't see it regardless.

packages/code-studio/src/main/AppMainContainer.tsx
AkshatJawne: Yeah, can't see this button in the disconnected state, so removing.

packages/embed-widget/src/index.tsx
```diff
@@ -5,7 +5,6 @@ import { store } from '@deephaven/redux';
 import '@deephaven/components/scss/BaseStyleSheet.scss';
 import { LoadingOverlay, preloadTheme } from '@deephaven/components';
 import { ApiBootstrap } from '@deephaven/jsapi-bootstrap';
-import './index.scss';
```
mattrunyon, line 8: I can't remember why this was removed to begin with. I would just add it back as I'm not sure if it does anything, so not sure it can safely be removed

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -83,7 +160,34 @@ export function ConnectionBootstrap({
...
+    return (
+      <BasicModal
+        confirmButtonText="Refresh"
+        onConfirm={handleRefresh}
+        isOpen={isAuthFailed}
+        headerText="Authentication failed"
+        bodyText="Credentials are invalid. Please refresh your browser to try and reconnect."
+      />
+    );
+  }
```
mattrunyon, line 189: I believe both of these cases should actually render the children still.

@mofojed we don't want the children completely gone when trying to reconnect, right?

This PR
image

Prior to this PR
image

mattrunyon (changes requested):
I also noticed this doesn't stop listening when the server is shutdown which is different behavior than what we currently have. I tested this by just stopping/starting the core server. The app showed the "auth failed" modal in that case. Previously this would just stay on the "Server shutdown" component.

packages/embed-widget/src/index.tsx
AkshatJawne: It was because it was not providing styling: https://github.com/deephaven/web-client-ui/pull/2023#discussion_r1605666964

AkshatJawne:
> I also noticed this doesn't stop listening when the server is shutdown which is different behavior than what we currently have. I tested this by just stopping/starting the core server. The app showed the "auth failed" modal in that case. Previously this would just stay on the "Server shutdown" component.

@mattrunyon Have been testing on my local branch, not able to reproduce, when I shut down the core server and start it again, I do not see the "auth failed" modal, it just stays on the "Server shutdown".

packages/app-utils/src/components/ConnectionBootstrap.tsx
mofojed: Yes children should still render

packages/app-utils/package.json
```diff
@@ -40,6 +40,7 @@
     "@deephaven/iris-grid": "file:../iris-grid",
     "@deephaven/jsapi-bootstrap": "file:../jsapi-bootstrap",
     "@deephaven/jsapi-components": "file:../jsapi-components",
+    "@deephaven/jsapi-shim": "file:../jsapi-shim",
```
mofojed, line 43:
```suggestion
```

mofojed (changes requested):

packages/embed-widget/src/index.tsx
mattrunyon: Right but that was after removing the style in the index file in this PR. I wasn't sure why the style was removed.

The only thing I can think of would be this is needed for some reason in jupyter, but I'm not sure.

AkshatJawne: Okay yeah I can add it back, doesn't do anything for my logic so don't really mind either way.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -205,7 +174,50 @@ export function ConnectionBootstrap({
   return (
     <ConnectionContext.Provider value={connection ?? null}>
       <ObjectFetcherContext.Provider value={objectFetcher}>
-        {children}
+        {(() => {
+          if (isShutdown) {
+            return (
```
mattrunyon: You don't need an IIFE (immediately invoked function expresssion) here. Instead you can just do this. The modals don't show unless `isOpen` is true, so it's fine to mount them here.

```ts
if (isShutdown) {
  return <LoadingOverlay />
}

return (
  <>
    {children}
    <DebouncedModal isOpen={isReconnecting} />
    <BasicModal isOpen={isAuthFailed />
  </>
)
```

If you want conditional rendering inside of JSX, you can do this (just for future reference)

```ts
return (
  {children}
  {isReconnecting && <DebouncedModal isOpen={isReconnecting} />
)
```

But that's not needed here since the modal won't show if `isOpen` is false.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -115,6 +118,8 @@ export function ConnectionBootstrap({
         log.info('Shutdown', `${JSON.stringify(detail)}`);
         setError(`Server shutdown: ${detail ?? 'Unknown reason'}`);
         setIsShutdown(true);
+        setIsReconnecting(false);
+        setIsAuthFailed(false);
```
mattrunyon: You could have a string represent the connection state instead. Since it should only be 1 of the states at a time. This would reduce the needed to set multiple states any time the connection state changes.

```ts
const [connectionState, setConnectionState] = useState<'connected' | 'connecting' | 'reconnecting' | 'failed' | 'shutdown'>('connecting');

const isAuthFailed = connectionState === 'failed';
const isShutdown = connectionState === 'shutdown';
```

The `isAuthFailed` is cheap to calculate here, so deriving the state is a good way to keep the change of state simpler here.

mattrunyon (changes requested):

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -104,21 +186,43 @@ export function ConnectionBootstrap({
...
+  function handleRefresh(): void {
+    log.info('Refreshing application');
+    window.location.reload();
+  }
+
+  if (isShutdown) {
     return (
       <LoadingOverlay
-        data-testid="connection-bootstrap-loading"
-        isLoading={connection == null}
```
mattrunyon, line 111: This `testid` should not be removed. It makes the couple assertions for tests unable to find this.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -104,21 +186,43 @@ export function ConnectionBootstrap({
     [objectFetcher]
   );
 
-  if (connection == null || error != null) {
+  function handleRefresh(): void {
+    log.info('Refreshing application');
+    window.location.reload();
+  }
+
+  if (isShutdown) {
```
mattrunyon, line 194: This should also show if the status is `connecting` (i.e. the initial connection). Previously that's what was shown here

Also, there is nothing that sets the status to `connected` on the first connection. Add that where `setConnection` is called so you can then do `isShutdown || connectionStatus === 'connecting'`. Previously this was `if connection == null || error != null` which was initial connection or shutdown. The reconnection and auth failed were handled in the app.

packages/app-utils/src/components/AppBootstrap.test.tsx
```diff
@@ -173,8 +173,8 @@ it('should log in automatically when the anonymous handler is supported', async
   });
 
   expect(screen.queryByTestId('auth-base-loading')).toBeNull();
-  expect(screen.queryByTestId('connection-bootstrap-loading')).not.toBeNull();
-  expect(screen.queryByText(mockChildText)).toBeNull();
+  expect(screen.queryByTestId('connection-bootstrap-loading')).toBeNull();
+  expect(screen.queryByText(mockChildText)).not.toBeNull();
```
mattrunyon, line 177: The test shouldn't need any changes. The test was failing because of a change in behavior. With the other comments about loading being shown when connecting, the test passes as it was originally written

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -31,8 +38,15 @@ export function ConnectionBootstrap({
 }: ConnectionBootstrapProps): JSX.Element {
   const api = useApi();
   const client = useClient();
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mattrunyon, line 41: Remove this comment

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -950,7 +885,7 @@ export class AppMainContainer extends Component<
               icon={
                 <span className="fa-layers">
                   <FontAwesomeIcon icon={vsGear} transform="grow-3" />
-                  {isDisconnected && !isAuthFailed && (
+                  {connection == null && (
                     <>
                       <FontAwesomeIcon
                         icon={dhSquareFilled}
```
mattrunyon, line 891: This icon is actually never used anymore. When the connection is null, the loader is shown. This was previously shown when the client was attempting to reconnect.

@dsmmcken Is it fine to not show this little disconnect tag on the settings menu when disconnected? We are showing the "Attempting to reconnect" modal when disconnected, so I think it's pretty clear what the state is

The logic for reconnect is now handled outside of this component, so we would need to create a context or pass down the prop just to show this little badge on the settings icon when there's a giant modal that tells you it's reconnecting.

mattrunyon (changes requested):

AkshatJawne:
Changed test case based on new logic, until the connection is resolved, it will show the modal, and hence for the first assertion of the connection bootstrap, should not be null.

mattrunyon (changes requested):

mattrunyon (approved):
Will still need @mofojed to review again. I want to make sure he's ok w/ the state names as well.

Changes for the states could be `not_connecting` -> `errored`. And `failed` -> `authFailed` instead. I'm indifferent though

mofojed (approved):

PR #2025, opened 2024-05-17
fix: Replace shortid package with nanoid package

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1922

Changes Implemented:
- Replaced shortid with nanoid, given that shortid is deprecated (Reran all tests to ensure that functionality upholds)

Info on Nanoid: https://github.com/ai/nanoid/blob/main/README.md


AkshatJawne:
Looking into errors here

mattrunyon:
Remove `esmify` from golden-layout dev deps and add `babelify` as a dev dependency.

In `karma.conf.cjs` remove the `plugins` line and add this. For some reason `esmify` isn't transforming node_modules even though it says it does (it might not understand the package.json of nanoid).

```js
transform: [
  ['babelify', { presets: ['@babel/preset-env'], global: true }],
],
```

This seems to fix the issue by using babel to transpile the ESM before the bundling for the test.

bmingles:
LGTM. Will wait for @mattrunyon to approve since he suggested the babelify change, and I'm less familiar here

AkshatJawne:
Sounds good, thanks @bmingles

packages/golden-layout/karma.conf.cjs
```diff
@@ -21,10 +21,13 @@ module.exports = function (config) {
     preprocessors: {
       // '../src/**': 'coverage'
       './test/helper.js': ['browserify'],
+      nanoid: ['browserify'],
```
mattrunyon, line 24: This shouldn't be needed

packages/golden-layout/package.json
```diff
@@ -50,8 +50,8 @@
     "access": "public"
   },
   "devDependencies": {
+    "babelify": "^10.0.0",
```
mattrunyon, line 53: golden-layout should have `nanoid` in its normal dependencies. Looks like `shortid` wasn't added there initially which was a mistake

packages/golden-layout/package.json
```diff
@@ -53,6 +53,7 @@
     "babelify": "^10.0.0",
     "browserify": "^17.0.0",
     "karma-browserify": "^8.1.0",
+    "nanoid": "^5.0.7",
```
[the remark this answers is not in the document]
AkshatJawne: Ahh makes sense, sorry about that. Pushing changes now

mattrunyon (commented):
There's a `@types/shortid` in the root `package.json` still. Remove it and run `npm install` again so it's removed from package-lock too.

mattrunyon (approved):

PR #2028, opened 2024-05-23
feat: re-export Spectrum ButtonGroup

### LINKED ISSUE #2016: Re-export Spectrum `ButtonGroup` and make dh `Button` compatible
We want to re-export Spectrum's `ButtonGroup` from @deephaven/components, however, We are not planning to re-export Spectrum's `Button`, so we need to make it compatible with Spectrum's ButtonGroup

- Apply Spectrum styling to DH `Button` via `useSlotProps` hook
- Re-export Spectrum `ButtonGroup`

e.g. We should be able to add something like this to our `Button` component.
```typescript
const { UNSAFE_className } = useSlotProps<{ UNSAFE_className?: string }>(
  {},
  'button'
);

...
<button
  ...
  className={classNames(
     ...
     className,
     UNSAFE_className
  )}
```

bmingles:
- Added Spectrum ButtonGroup support to DH Button
- Re-exported Spectrum ButtonGroup

resolves #2016


packages/code-studio/src/styleguide/Buttons.tsx
mofojed, line 1: Nice

tests/styleguide.spec.ts-snapshots/buttons-hover-section-sample-section-buttons-socketed-3-firefox-linux.png
mofojed, line 1: Just noticed we have the same snapshots taken a few times for the socketed buttons section - is that necessary? Seems like we should only be taking one snapshot of the section to verify everything?

In any case, not related to this change necessarily. Can be a follow up if needed to clean up.

mofojed (approved):

PR #2030, opened 2024-05-24
feat: DH-16737 Add ObjectManager, `useWidget` hook

mofojed:
- Hook for loading a widget that makes it easy to use
- `ObjectManager` context allows for implementation to handle loading the object
  - On Core side, we have a very simple `ObjectManager`, as we just have one connection
  - On Enterprise side, the `ObjectManager` provided to the context manager will need to handle fetching from queries, and will be able to handle more scenarios (such as when a query is restarting)
- Use with https://github.com/deephaven/deephaven-plugins/pull/502

### DIFF
--- a/packages/app-utils/src/components/ConnectionBootstrap.tsx
+++ b/packages/app-utils/src/components/ConnectionBootstrap.tsx
@@ -93,7 +93,7 @@ export function ConnectionBootstrap({
         // We send an update with the fetch right away
         onUpdate({
           fetch: () => objectFetcher(descriptor),
-          error: null,
+          status: 'ready',
         });
         return () => {
           // no-op
--- a/packages/app-utils/src/components/ConnectionBootstrap.tsx
+++ b/packages/app-utils/src/components/ConnectionBootstrap.tsx
@@ -1,7 +1,9 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { LoadingOverlay } from '@deephaven/components';
 import {
   ObjectFetcherContext,
+  ObjectFetchManager,
+  ObjectFetchManagerContext,
   sanitizeVariableDescriptor,
   useApi,
   useClient,
@@ -31,6 +33,7 @@ export function ConnectionBootstrap({
   const client = useClient();
   const [error, setError] = useState<unknown>();
   const [connection, setConnection] = useState<dh.IdeConnection>();
+
   useEffect(
     function initConnection() {
       let isCanceled = false;
@@ -83,6 +86,24 @@ export function ConnectionBootstrap({
     [connection]
   );
 
+  /** We don't really need to do anything fancy in Core to manage an object, just fetch it  */
+  const objectManager: ObjectFetchManager = useMemo(
+    () => ({
+      subscribe: (descriptor, onUpdate) => {
+        // We send an update with the fetch right away
+        onUpdate({
+          fetch: () => objectFetcher(descriptor),
+          error: null,
+        });
+        return () => {
+          // no-op
+          // For Core, if the server dies then we can't reconnect anyway, so no need to bother listening for subscription or cleaning up
+        };
+      },
+    }),
+    [objectFetcher]
+  );
+
   if (connection == null || error != null) {
     return (
       <LoadingOverlay
@@ -96,7 +117,9 @@ export function ConnectionBootstrap({
   return (
     <ConnectionContext.Provider value={connection}>
       <ObjectFetcherContext.Provider value={objectFetcher}>
-        {children}
+        <ObjectFetchManagerContext.Provider value={objectManager}>
+          {children}
+        </ObjectFetchManagerContext.Provider>
       </ObjectFetcherContext.Provider>
     </ConnectionContext.Provider>
   );
--- a/packages/jsapi-bootstrap/src/useObjectFetch.ts
+++ b/packages/jsapi-bootstrap/src/useObjectFetch.ts
@@ -1,5 +1,6 @@
 import { createContext, useContext, useEffect, useState } from 'react';
 import type { dh } from '@deephaven/jsapi-types';
+import { ObjectFetcherContext } from './useObjectFetcher';
 
 /** Function for unsubscribing from a given subscription */
 export type UnsubscribeFunction = () => void;
@@ -23,6 +24,10 @@ export type ObjectFetchUpdate<T = unknown> = {
   error: unknown | null;
 };
 
+export type ObjectFetchUpdateCallback<T = unknown> = (
+  update: ObjectFetchUpdate<T>
+) => void;
+
 /** ObjectFetchManager for managing a subscription to an object using a VariableDescriptor */
 export type ObjectFetchManager = {
   /**
@@ -33,7 +38,7 @@ export type ObjectFetchManager = {
    */
   subscribe: <T = unknown>(
     descriptor: dh.ide.VariableDescriptor,
-    onUpdate: (update: ObjectFetchUpdate<T>) => void
+    onUpdate: ObjectFetchUpdateCallback<T>
   ) => UnsubscribeFunction;
 };
 
@@ -58,14 +63,22 @@ export function useObjectFetch<T = unknown>(
     })
   );
 
+  const objectFetcher = useContext(ObjectFetcherContext);
   const objectFetchManager = useContext(ObjectFetchManagerContext);
 
   useEffect(() => {
     if (objectFetchManager == null) {
-      setCurrentUpdate({
-        fetch: null,
-        error: new Error('No ObjectFetchManager available in context'),
-      });
+      if (objectFetcher == null) {
+        setCurrentUpdate({
+          fetch: null,
+          error: new Error('No ObjectFetchManager available in context'),
+        });
+      } else {
+        setCurrentUpdate({
+          fetch: () => objectFetcher(descriptor),
+          error: null,
+        });
+      }
       return;
     }
     // Signal that we're still loading
@@ -74,7 +87,7 @@ export function useObjectFetch<T = unknown>(
       error: null,
     });
     return objectFetchManager.subscribe(descriptor, setCurrentUpdate);
-  }, [descriptor, objectFetchManager]);
+  }, [descriptor, objectFetcher, objectFetchManager]);
 
   return currentUpdate;
 }
--- a/packages/jsapi-bootstrap/src/useObjectFetch.ts
+++ b/packages/jsapi-bootstrap/src/useObjectFetch.ts
@@ -1,0 +1,80 @@
+import { createContext, useContext, useEffect, useState } from 'react';
+import type { dh } from '@deephaven/jsapi-types';
+
+/** Function for unsubscribing from a given subscription */
+export type UnsubscribeFunction = () => void;
+
+/**
+ * Update with the current `fetch` fuonction and status of the object.
+ * - If both `fetch` and `error` are `null`, it is still loading the fetcher
+ * - If `fetch` is not `null`, the object is ready to be fetched
+ * - If `error` is not `null`, there was an error loading the object
+ */
+export type ObjectFetchUpdate<T = unknown> = {
+  /**
+   * Function to fetch the object. If `null`, the object is still loading or there was an error.
+   */
+  fetch: (() => Promise<T>) | null;
+
+  /**
+   * Error that occurred while fetching the object. If `null`, there was no error.
+   * Will automatically retry when possible while the subscribed.
+   */
+  error: unknown | null;
+};
+
+/** ObjectFetchManager for managing a subscription to an object using a VariableDescriptor */
+export type ObjectFetchManager = {
+  /**
+   *
+   * @param descriptor Descriptor object to fetch the object from. Can be extended by a specific implementation to include more details necessary for the ObjectManager.
+   * @param onUpdate Callback function to be called when the object is updated.
+   * @returns An unsubscribe function to stop listening for updates and clean up the object.
+   */
+  subscribe: <T = unknown>(
+    descriptor: dh.ide.VariableDescriptor,
+    onUpdate: (update: ObjectFetchUpdate<T>) => void
+  ) => UnsubscribeFunction;
+};
+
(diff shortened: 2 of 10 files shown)


packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -0,0 +1,93 @@
...
+  /** Widget object to retrieve */
+  widget: T | null;
+
+  /** Error status if there was an issue fetching the widget */
+  error: unknown | null;
+};
+
+/**
+ * Retrieve a widget for the given variable descriptor. Note that if the widget is successfully fetched, ownership of the widget is passed to the consumer and will need to close the object as well.
+ * @param descriptor Descriptor to get the widget for
```
mattrunyon, line 22: I would add here or in the description that the descriptor must be stable to avoid infinite re-fetching

mattrunyon (changes requested):
Should we deprecate/remove `useObjectFetcher` with this? We still use it in at least 1 place internally and I'm not sure if that should be replaced with this

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -0,0 +1,93 @@
...
+      // If not though, the consumer of the widget is expected to take ownership and close the widget appropriately.
+      let isCancelled = false;
+      async function loadWidgetInternal() {
+        try {
+          assertNotNull(fetch);
+          const newWidget = await fetch();
+          if (isCancelled) {
+            log.debug2('loadWidgetInternal cancelled', descriptor, newWidget);
+            newWidget.close();
+            newWidget.exportedObjects.forEach(
```
mattrunyon, line 64: I this this needs to be `newWidget.exportedObjects?.forEach`

`Widget` doesn't encompass the 1st party widgets like `Table`, `Figure`, etc. Just the plugin registered widgets

Or we need to check the descriptor type and throw/error if we try to use this with `Table`, `Figure`, etc.

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Function to fetch the object. If `null`, the object is still loading or there was an error.
+   */
+  fetch: (() => Promise<T>) | null;
+
+  /**
+   * Error that occurred while fetching the object. If `null`, there was no error.
+   * Will automatically retry when possible while the subscribed.
+   */
+  error: unknown | null;
```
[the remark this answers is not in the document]
mofojed: Good catch

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -0,0 +1,93 @@
...
+/** Function for unsubscribing from a given subscription */
+export type UnsubscribeFunction = () => void;
+
+/**
+ * Update with the current `fetch` fuonction and status of the object.
+ * - If both `fetch` and `error` are `null`, it is still loading the fetcher
+ * - If `fetch` is not `null`, the object is ready to be fetched
+ * - If `error` is not `null`, there was an error loading the object
+ */
+export type ObjectFetchUpdate<T = unknown> = {
```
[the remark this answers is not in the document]
mofojed: Yea I thought about that, but decided against it at this time. Didn't _need_ it, and wasn't sure what statuses I wanted (`running` vs `disconnected`? `ok` vs `error`? `success`/`failure`?)

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export function useObjectFetch<T = unknown>(
+  descriptor: dh.ide.VariableDescriptor
+): ObjectFetchUpdate<T> {
+  const [currentUpdate, setCurrentUpdate] = useState<ObjectFetchUpdate<T>>(
+    () => ({
+      fetch: null,
+      error: null,
+    })
+  );
```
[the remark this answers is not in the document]
mofojed: Why not?

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -0,0 +1,93 @@
...
+        setCurrentUpdate({
+          fetch: null,
+          error: new Error('No ObjectFetchManager available in context'),
+        });
+      } else {
+        setCurrentUpdate({
+          fetch: () => objectFetcher(descriptor),
+          error: null,
+        });
+      }
```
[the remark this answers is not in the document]
mofojed: Hmm yea they could be combined... ObjectFetcher would be if you just want to fetch an object once, not subscribe to it and listen for updates... though not sure when you actually would want to do that.

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -0,0 +1,93 @@
...
+      return;
+    }
+    // Signal that we're still loading
+    setCurrentUpdate({
+      fetch: null,
+      error: null,
+    });
```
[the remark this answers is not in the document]
mofojed: Yea, I guess technically I could check if the old update is still `fetch: null; error: null`... it would technically matter if you are just looking at the object instead of destructuring it right away, as it's a new object...
I guess there's no real reason to fallback to the objectFetcher either. I should just take that out.

packages/jsapi-bootstrap/src/useObjectFetch.ts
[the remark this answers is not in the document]
mofojed: `loading`/`loaded`/`error`... I'll do that.

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -0,0 +1,93 @@
...
+type WidgetWrapper<T extends dh.Widget = dh.Widget> = {
+  /** Widget object to retrieve */
+  widget: T | null;
+
+  /** Error status if there was an issue fetching the widget */
+  error: unknown | null;
+};
+
+/**
+ * Retrieve a widget for the given variable descriptor. Note that if the widget is successfully fetched, ownership of the widget is passed to the consumer and will need to close the object as well.
```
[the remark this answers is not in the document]
mofojed: Yea, with a widget you can get updates that include new exported objects (like we do with deephaven.ui), those need to be cleaned up as well; or you may want to re-export exported objects rather than fetch them, I don't think we want to automatically close things here as it's not clear if the objects were actually fetched by the consumer or not.

packages/jsapi-bootstrap/src/useWidget.ts
mofojed: Let's throw an error if the descriptor is not for a widget.

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -93,7 +93,7 @@ export function ConnectionBootstrap({
         // We send an update with the fetch right away
         onUpdate({
           fetch: () => objectFetcher(descriptor),
-          error: null,
+          status: 'ready',
```
mattrunyon, line 96: The statuses should probably be consts (or an enum, but just a JS object "enum")

packages/jsapi-bootstrap/src/useObjectFetch.ts
```diff
@@ -31,7 +39,7 @@ export type ObjectFetchUpdateCallback<T = unknown> = (
 /** ObjectFetchManager for managing a subscription to an object using a VariableDescriptor */
 export type ObjectFetchManager = {
   /**
-   *
+   * Subscribe to an object using a variable descriptor.
```
mattrunyon, line 42: Just to be clear, this doesn't actually subscribe to the object, but subscribes to the function to fetch the object? It's still up to the consumer to actually fetch the object if they were using this hook.

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -29,27 +30,50 @@ export function useWidget<T extends dh.Widget = dh.Widget>(
...
+    () => [
+      api.VariableType.TABLE,
+      api.VariableType.TREETABLE,
+      api.VariableType.HIERARCHICALTABLE,
+      api.VariableType.TABLEMAP,
+      api.VariableType.PARTITIONEDTABLE,
+      api.VariableType.FIGURE,
+      api.VariableType.PANDAS,
+      api.VariableType.TREEMAP,
+    ],
```
mattrunyon, line 44: Is there any reason we couldn't/shouldn't support these? The `fetch` works the same, regardless of descriptor. It's just the returned type doesn't extend `JsWidget` so we would only close `exportedObjects` if that field exists.

I don't think we would call this hook with these types since they are all handled by other plugins right now, but I could see an argument for replacing the subscription code for `IrisGridPanel` and `ChartPanel` with this centralized listener code

mattrunyon (changes requested):

mattrunyon:
I think the status messages are fine (as long as they're constants). I've used tanstack query some which has each state as a boolean value in the returned object. When I was using it, it seemed to be a fine pattern as well. In the case of tanstack query it's ajax/browser fetch requests with optional refetch intervals/events.

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

packages/jsapi-bootstrap/src/useWidget.ts
mattrunyon: I think the main thing that would prevent us from using this hook in those panels is at least 1 requires a fetch function and not the widget I think (I can't remember which). IIRC it was something to do with reloading or resubscribing to the widget properly and the way the models interacted w/ the panel/component

packages/jsapi-bootstrap/src/useObjectFetch.ts
mofojed: You might be thinking of charts built using `ChartBuilder`, I think that currently needs the fetch. Pretty sure I addressed that with my core plugins/xcomponent change if I recall correctly. 
My intent for this was useWidget is just for widgets. Could have another hook for `useTable`, `useFigure`, etc that is just built on top of `useObjectFetch`.
In any case, I wasn't going to replace the existing panels until Grizzly. This PR is intended for V+.

packages/jsapi-bootstrap/src/useWidget.ts
mofojed: You might be thinking of charts built using `ChartBuilder`, I think that currently needs the fetch. Pretty sure I addressed that with my core plugins/xcomponent change if I recall correctly. 
My intent for this was useWidget is just for widgets. Could have another hook for `useTable`, `useFigure`, etc that is just built on top of `useObjectFetch`.
In any case, I wasn't going to replace the existing panels until Grizzly. This PR is intended for V+.

packages/app-utils/src/components/ConnectionBootstrap.tsx
mofojed: Any reason to use constants for that instead of just defined in the type in TS?

mattrunyon: I guess just the convention we've normally used, but a TS string literal union works too.

packages/jsapi-bootstrap/src/useWidget.ts
mattrunyon: I think you're right, it was chart builder.

And that's fine we can think about it at a later point if we add those hooks. One reason against it would be you can't call hooks conditionally (although you can with `use` in react 19 I think, but not sure on the specifics). So if you had a descriptor that could be any of the types, we wouldn't be able to fetch it with the hook. Or we would have to call multiple hooks and disregard the results of all but 1.

packages/app-utils/src/components/ConnectionBootstrap.tsx
mofojed: We use string literal unions in lots of places, like ButtonTypes, not sure we wouldn't use them here?
We use constants when it's defined from the JS API, but this status is purely client side.

packages/jsapi-bootstrap/src/useWidget.ts
mofojed: Yeaaaa you know what, I think you're right - I'll just have it support all those widget types. Let the hook handle the messiness a little bit.

packages/jsapi-bootstrap/src/useWidget.ts
mofojed: I instead opted to check if the `exportedObjects` is on the object and close if it is, so we handle all types now.

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -85,11 +71,13 @@ export function useWidget<T extends dh.Widget = dh.Widget>(
...
             log.debug2('loadWidgetInternal cancelled', descriptor, newWidget);
             newWidget.close();
-            newWidget.exportedObjects.forEach(
-              (exportedObject: dh.WidgetExportedObject) => {
-                exportedObject.close();
-              }
-            );
+            if ('exportedObjects' in newWidget) {
+              newWidget.exportedObjects.forEach(
+                (exportedObject: dh.WidgetExportedObject) => {
```
mattrunyon, line 76: TS can infer this param type, so no need to be explicit about it here

packages/jsapi-bootstrap/src/useWidget.test.tsx
```diff
@@ -78,30 +66,119 @@ describe('useWidget', () => {
...
+
+    expect(result.current).toEqual({ widget: null, error: null });
+    await act(TestUtils.flushPromises);
+    unmount();
+    expect(widget.close).not.toHaveBeenCalled();
+    expect(widget.exportedObjects[0].close).not.toHaveBeenCalled();
+    expect(widget.exportedObjects[1].close).not.toHaveBeenCalled();
+  });
+
+  it('should handle a Table being fetched', async () => {
```
mattrunyon, line 157: Do we want a test for closing a table too? Idk if it's necessary since it would just be skipping the `exportedObjects` close block if it were a table

mattrunyon (changes requested):

packages/jsapi-bootstrap/src/useWidget.test.tsx
mofojed: May as well

mattrunyon (approved):

PR #2034, opened 2024-05-24
chore: Upgraded to node lts 20.13.1

bmingles:
Upgraded to node lts 20.13.1


mofojed (approved):
Verified a few things:
- `npm install` produced no changes
- `npm start` works correctly
- `npm test` runs all tests in watch more, all tests pass

Be sure to notify folks in the team channel - depending on how they have nvm set up they may need to change directories first before they switch to the new node/npm (though it should still work correctly without switching).

PR #2041, opened 2024-05-29
fix: DH-17076 LayoutHints on TreeTables were not being applied

### LINKED ISSUE #2035: Column grouping layout hints are ignored on rollup tables
A customer noticed that column groupings from layout hints are respected on regular tables, but ignored on rollup tables. (The columns can be grouped manually afterward and it works; the issue is just the layout hints being ignored.)

DH-17076

Only reproducible in Groovy:

```java
import io.deephaven.engine.util.LayoutHintBuilder

source = emptyTable(100).update(
    "K1 = ii",
    "K2 = K1 % 10",
    "Y = randomDouble(0, 10)",
    "Z = randomDouble(0, 10)",
)

result_norollup = source.setLayoutHints(LayoutHintBuilder.get().columnGroup("MyCols", ["Y", "Z"]).build())

import io.deephaven.api.agg.spec.AggSpec
import io.deephaven.api.agg.Aggregation


result_rollup = source
    .rollup([Aggregation.of(AggSpec.avg(), "Y", "Z")], false, "K2")
    .setLayoutHints(LayoutHintBuilder.get().columnGroup("MyCols", ["Y", "Z"]).build())
```

For some reason, `layout_hints` is not an available method on `RollupTable` in Python: https://github.com/deephaven/deephaven-core/blob/dbf2b2db31ac3e982fc6c432b66d3cb688fda431/py/server/deephaven/table.py#L169

mofojed:
- Related to DH-17076
- Needed a Core patch: https://github.com/deephaven/deephaven-core/pull/5555
- Tested using both the Groovy snippet from the ticket
- Fixes #2035

### DIFF
--- a/packages/iris-grid/src/IrisGridTreeTableModel.test.ts
+++ b/packages/iris-grid/src/IrisGridTreeTableModel.test.ts
@@ -4,7 +4,7 @@ import IrisGridTreeTableModel from './IrisGridTreeTableModel';
 
 const irisGridTestUtils = new IrisGridTestUtils(dh);
 
-describe('IrisGridTreeTableModel', () => {
+describe('IrisGridTreeTableModel virtual columns', () => {
   const expectedVirtualColumn = expect.objectContaining({
     name: '__DH_UI_GROUP__',
     displayName: 'Group',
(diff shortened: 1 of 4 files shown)


packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -230,6 +230,12 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
     return [this.virtualColumns.length, this.groupedColumns.length];
   }
 
+  get layoutHints(): DhType.LayoutHints | null | undefined {
+    // TODO: Update this when JS API types are updated with this.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return (this.table as any).layoutHints;
```
mattrunyon, line 236: Should we feature detect here? If this had to be added to the DHC JSAPI then it's probably not part of the DHE JSAPI

Although since it's just 1 level deep it shouldn't throw and should just return undefined like it would have before

mattrunyon (changes requested):
Did you want to merge the DHC change before merging this? Or order shouldn't matter?

Also, did you want to put the Jira ticket in the title so we know when it goes into DHE?

packages/iris-grid/src/IrisGridTreeTableModel.ts
mofojed: Yea you're right, I should be checking here.

mofojed:
@mattrunyon order shouldn't matter here, I added feature detection as well.

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -231,9 +240,10 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
   }
 
   get layoutHints(): DhType.LayoutHints | null | undefined {
-    // TODO: Update this when JS API types are updated with this.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return (this.table as any).layoutHints;
+    if (isLayoutTreeTable(this.table)) {
+      return this.table.layoutHints;
+    }
+    return undefined;
```
mattrunyon, line 246: I realized it shouldn't really matter since DHE would just return undefined for `this.table.layoutHints`. But still not bad to feature detect it.

mattrunyon (changes requested):
Just 2 small things

packages/iris-grid/src/IrisGridTreeTableModel.test.ts
```diff
@@ -27,3 +27,48 @@ describe('IrisGridTreeTableModel', () => {
...
+
+  test('layoutHints property does not exist should not crash', () => {
+    const columns = irisGridTestUtils.makeColumns();
+    const table = irisGridTestUtils.makeTreeTable(columns, columns, 100, []);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    delete (table as any).layoutHints;
+    const model = new IrisGridTreeTableModel(dh, table);
+
+    expect(model.layoutHints).toEqual(undefined);
+  });
```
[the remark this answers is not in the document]
mofojed: Not quite. `undefined` vs. property not existing, it is technically different. e.g.

```javascript
var o = { foo: 'bar', fiz: undefined }
var o2 = { ...o }
delete o2.fiz

console.log(o.hasOwnProperty('fiz')) // true
console.log(o2.hasOwnProperty('fiz')) //false
```

packages/iris-grid/src/IrisGridTreeTableModel.ts
mofojed: Yea my thoughts as well

mattrunyon (approved):

PR #2043, opened 2024-05-29
feat: Add ruff linter and formatter for Python

mattrunyon:
This adds the linter/formatter, but does not add user configuration yet. So I created a feature branch for it.

This is lazy loaded and only loaded if there is a Python monaco instance. You can verify by looking for wasm files in the network tab.

One thing to note is since embed-widget has its own copy of everything, I believe there would be 2 copies of the wasm in DHC which is not great.


mofojed (approved):

mofojed (changes requested):
Boom

packages/console/src/monaco/ruff/ruff_wasm.js
mofojed, line 1: Where'd this file come from? I think a README in this `ruff` folder would be beneficial as to where this code came from/how to generate it.

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -959,6 +975,27 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
...
+  async handleFormat(): Promise<void> {
+    if (this.canFormat()) {
+      await this.notebook?.editor
+        ?.getAction('editor.action.formatDocument')
+        ?.run();
+    }
+  }
+
+  async handleSaveFromShortcut(): Promise<void> {
+    await this.handleFormat();
```
mofojed, line 995: I don't think we should auto format on save by default unless the user has opted in, configured the formatting settings, or has otherwise been notified this will be occurring. Imposing our preferred formatting settings on others without their approval is likely to annoy people.
I know this is going into a feature branch, but I think important to call that out.

mofojed (changes requested):

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -14,13 +16,119 @@ interface MonacoProviderProps {
...
+  static lintPython(model: monaco.editor.ITextModel): void {
+    if (!MonacoProviders.workspace) {
+      return;
+    }
+
+    monaco.editor.setModelMarkers(
+      model,
+      'ruff',
+      MonacoProviders.workspace.check(model.getValue()).map((d: Diagnostic) => {
+        const isUnnecessary = d.code === 'F401' || d.code === 'F841';
```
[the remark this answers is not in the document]
mattrunyon: This is unused variable and unused import. It is mainly to add the unnecessary tag in monaco which makes the text appear grayed-out. I took this from ruff's playground code

Ruff doesn't distinguish between error and warning diagnostics (there is pycodestyle error and warning, but that's just the rule names). So it would be up to us to determine if we wanted configuration for more as warnings instead of errors

packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
mattrunyon: I commented this out for now and added a TODO comment to check if the user enabled format on save.

You can still format from the overflow menu on this branch

packages/console/src/monaco/MonacoProviders.tsx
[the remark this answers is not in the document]
mofojed: Okay, was just expecting implicitly concatenated strings to be a warning not an error.

mofojed (approved):

Issue #2029, opened 2024-05-23
Opening context menu on tree tables throws an exception in the browser console


vbabich:
**Description**

Opening context menu on tree tables throws an exception in the browser console

**Steps to reproduce**

Run this snippet to create a tree table:
```
from deephaven.constants import NULL_INT
from deephaven import empty_table

source = empty_table(100).update_view(
    ["ID = i + 2", "Parent = i == 0 ? NULL_INT : i % 9"]
)

result_w_orphans = source.tree(id_col="ID", parent_col="Parent", promote_orphans=True)
```

Right-click on a table cell.

Expected:
- No exceptions in the console

Actual:
```
LogProxy.js:87 [ContextMenu] Error: Value is null or undefined
    at I7 (Asserts.js:33:28)
    at IrisGridTreeTableModel.js:133:13
    at Generator.next (<anonymous>)
    at nr (IrisGridTreeTableModel.js:1:102)
    at l (IrisGridTreeTableModel.js:2:194)
    at IrisGridTreeTableModel.js:2:364
    at new Promise (<anonymous>)
    at IrisGridTreeTableModel.js:2:97
    at Ul.snapshot (IrisGridTreeTableModel.js:140:6)
    at Kl.snapshot (IrisGridProxyModel.js:182:26)
    at IrisGridContextMenuHandler.js:462:34
    at Generator.next (<anonymous>)
    at Or (IrisGridContextMenuHandler.js:3:102)
    at o (IrisGridContextMenuHandler.js:4:194)
    at IrisGridContextMenuHandler.js:4:364
    at new Promise (<anonymous>)
    at IrisGridContextMenuHandler.js:4:97
    at b.getCellFilterActions (IrisGridContextMenuHandler.js:566:6)
    at b.onContextMenu (IrisGridContextMenuHandler.js:680:27)
    at F.handleContextMenu (Grid.js:1439:24)
    at Object.Qqe (react-dom.production.min.js:52:317)
    at Xqe (react-dom.production.min.js:52:471)
    at Jqe (react-dom.production.min.js:53:35)
    at pfe (react-dom.production.min.js:100:68)
    at Qbe (react-dom.production.min.js:101:380)
    at react-dom.production.min.js:113:65
    at Sbe (react-dom.production.min.js:292:189)
    at Yqe (react-dom.production.min.js:50:57)
    at Jbe (react-dom.production.min.js:105:469)
    at Yae (react-dom.production.min.js:75:265)
[...]

PR #2047, opened 2024-05-30
fix: Console error when opening context menu on tree table

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2029

**Changes Implemented:** 
- Added optional chaining when call formatValue function, to ensure that it is only called when it is defined

I had also considered the approach below, which achieved the **same result** of removing the exception in the console. Ultimately opted to not use this logic, under the assumption that we are okay with the `resultRow` being `undefined`.

However, if we want `resultRow` to be defined, this approach ensures that it is, via a default function that performs no formatting.

```
// Approach #2: Default function that performs no formatting if formatValue is undefined
const formatVal =
formatValue ?? ((value: unknown, column: DhType.Column) => value);
resultRow.push(
formatVal(viewportRow.data.get(c)?.value, this.columns[c])
);

```


packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -181,9 +181,8 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
           c <= intersection.endColumn;
           c += 1
         ) {
-          assertNotNull(formatValue);
           resultRow.push(
-            formatValue(viewportRow.data.get(c)?.value, this.columns[c])
+            formatValue?.(viewportRow.data.get(c)?.value, this.columns[c])
```
bmingles, line 185: Seems if `formatValue` is not defined, there is no point in entering the for loop at all. I would check before the loop.

AkshatJawne: Good catch, just tested and it seems like this also works, but definitely more efficient (since we are not going into the loop if its not defined anymore). Pushed change

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -176,15 +176,16 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
...
-        for (
-          let c = intersection.startColumn;
-          c <= intersection.endColumn;
-          c += 1
-        ) {
-          assertNotNull(formatValue);
-          resultRow.push(
-            formatValue(viewportRow.data.get(c)?.value, this.columns[c])
-          );
+        if (formatValue !== undefined) {
```
bmingles, line 179: Usually preferred to use loose equality `!= null` here since it will catch `null` and `undefined` scenarios.

bmingles: Unless of course you have a reason for explicitly catching `undefined` and ignoring null.

AkshatJawne: Don't have a specific reason for explicitly catching undefined and ignoring null, actually in my first commit I had it set to null, so can definitely change it back.

AkshatJawne: I had only changed it to undefined to be more in alignment with your comment, but yeah I think it is a better choice to use null here

bmingles (approved):
Code changes look good. I confirmed the error no longer shows up when opening the context menu on tree table.

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -176,15 +176,16 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
...
-            formatValue(viewportRow.data.get(c)?.value, this.columns[c])
-          );
+        if (formatValue != null) {
+          for (
+            let c = intersection.startColumn;
+            c <= intersection.endColumn;
+            c += 1
+          ) {
+            resultRow.push(
+              formatValue(viewportRow.data.get(c)?.value, this.columns[c])
```
vbabich, line 186: We still want to populate the `resultRow` array with unformatted values if `formatValue` is not defined.
With this change, the `resultRow` ends up empty.

PR #2048, opened 2024-05-30
object manager vplus

mofojed:
Same as https://github.com/deephaven/web-client-ui/pull/2030 but just on 0.78 instead


mofojed:
https://github.com/deephaven/web-client-ui/pull/2030 was merged to v0.78 already

Issue #2037, opened 2024-05-28
CSV Drag and Drop Upload Hits Console Error Log


nbauernfeind:
To reproduce, drag and drop a CSV and upload.

What happens:
1. The csv uploads successfully and shows a blue-pill button that I can click on to look at the table.
2. The browser's log and the console log show this error:
```
LogProxy.js:87 [CsvInputBar] Error: Value is null or undefined
    at I7 (Asserts.js:33:28)
    at qr.openWidget (ConsolePanel.js:265:5)
    at qr.handleOpenObject (ConsolePanel.js:221:12)
    at zt.handleOpenCsvTable (Console.js:616:5)
    at nl.openTable (CsvInputBar.js:258:5)
    at CsvInputBar.js:167:75
```
3. The table does not open automatically; I must click the button.

What I expected to happen:
The uploaded csv opens in a new tab and there are no errors in the log.

PR #2052, opened 2024-06-03
fix: CSV Drag and Drop Console Error

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2037

**Changes Implemented:** 
- Changed the destructuring in the `ConsolePanel.tsx` `openWidget` method to set `title` to be  the `name` if `title` is undefined

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
@@ -298,9 +298,9 @@ export class ConsolePanel extends PureComponent<
     assertNotNull(sessionWrapper);
 
     const { config, session } = sessionWrapper;
-    const { title } = widget;
-    assertNotNull(title);
-    const panelId = this.getItemId(title);
+    const { name } = widget;
+    assertNotNull(name);
+    const panelId = this.getItemId(name);
     const openOptions = {
       fetch: () => session.getObject(widget),
       panelId,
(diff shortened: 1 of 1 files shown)


packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -298,9 +298,9 @@ export class ConsolePanel extends PureComponent<
     assertNotNull(sessionWrapper);
 
     const { config, session } = sessionWrapper;
-    const { title } = widget;
-    assertNotNull(title);
-    const panelId = this.getItemId(title);
+    const { name } = widget;
+    assertNotNull(name);
+    const panelId = this.getItemId(name);
```
mofojed, line 303: We want to use `title` if it's available, then fallback to `name` if `title` is not available.

Other objects we open are actually a `VariableDefinition` type, and the `title` is the friendly name for the object and what we want to display to the user.

mofojed (changes requested):

packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
AkshatJawne: Right, but then would it not make sense to pass in` title: title` instead of `name: title `? Given the declaration below, I was under the impression that the intention **was to pass in the `name`**? Since, even if I add a fallback to `name` if `title` is not available, it would not matter since `title` will never be available, given that the `title` is assigned to `name` here

https://github.com/deephaven/web-client-ui/blob/139388d77f94f561938bb4ce61da4780387ef4a9/packages/console/src/Console.tsx#L787

mofojed: You'd get a type error then, because a `VariableDescriptor` is required in `openWidget`.
The `title` vs. `name` thing is kind of annoying and due to the types provided from the JS API which we're not totally abstracted from here. We should try and use `title` if available, otherwise fallback to `name`.

mofojed (approved):

Issue #2036, opened 2024-05-28
Can't edit the correct cells if key columns are not the first columns


mofojed:
**Steps to Reproduce:**

1. Run the following snippet:

```python
from deephaven import empty_table, input_table

_source1 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])

result1 = input_table(init_table=_source1, key_cols=["A", "B", "E", "F"])
```
2. In table result1, attempt to edit something in editable column C
3. In table result1, attempt to edit something in key column E

**Expected Results:**

2. Should be able to edit the value
3. Should not be able to edit the value

**Actual Results:**

2. Cannot edit the value
3. Can edit the value, then it saves as a new row

DH-17095

AkshatJawne:
Created PR with proposed information, here were the cases I tested with: 

```
# Base Testing
from deephaven import empty_table, input_table

_source1 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])

result1 = input_table(init_table=_source1, key_cols=["A", "B", "E", "F"])

# No Key Columns

from deephaven import empty_table, input_table

_source2 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])
result2 = input_table(init_table=_source2, key_cols=[])

# All Key Columns 

from deephaven import empty_table, input_table

_source3 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])
result3 = input_table(init_table=_source3, key_cols=["A", "B", "C", "D", "E", "F"])

# Key Columns at the start

from deephaven import empty_table, input_table

_source4 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])
result4 = input_table(init_table=_source4, key_cols=["A", "B"])

# Key Columns at the end

from deephaven import empty_table, input_table

_source5 = empty_table(5).update(["A=`` + i", "B=`` + i * 2", "C=i", "D=i * 2", "E=`` + i", "F=`` + i * 2",])
result5 = input_table(init_table=_source5, key_cols=["E", "F"])

# Singular key column in the middle

[...]

PR #2053, opened 2024-06-03
fix: Editing issues when key columns are not first columns

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2036

**Changes Implemented:**
- Changed check for key columns within range **to no longer check** whether the _start of the range is > # of key columns_ and _end of range is > # of key columns_ , **but rather**, iterate through and check _if any of the columns are key columns_

- The check priorly was making it so that the key columns are always the first columns, which was causing issues with editing columns if they were not

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModelTemplate.ts
+++ b/packages/iris-grid/src/IrisGridTableModelTemplate.ts
@@ -1622,10 +1622,7 @@ class IrisGridTableModelTemplate<
   }
 
   isKeyColumn(x: ModelIndex): boolean {
-    if (this.columns[x] != null) {
-      return this.keyColumnSet.has(this.columns[x].name);
-    }
-    return false;
+    return this.keyColumnSet.has(this.columns[x].name);
   }
 
   isRowMovable(): boolean {
--- a/packages/grid/src/GridMetricCalculator.ts
+++ b/packages/grid/src/GridMetricCalculator.ts
@@ -800,6 +800,10 @@ export class GridMetricCalculator {
     const { model } = state;
     const { columnCount, floatingRightColumnCount } = model;
 
+    if (columnCount === 0) {
+      return 0;
+    }
+
     let lastLeft = Math.max(0, columnCount - floatingRightColumnCount - 1);
     if (right != null) {
       lastLeft = right;


mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1644,12 +1644,24 @@ class IrisGridTableModelTemplate<
...
       this.isPendingRow(range.startRow) && this.isPendingRow(range.endRow);
+
+    let isKeyColumnInRange = false;
+    // Check if any of the columns in grid range are key columns
+    for (
+      let column = range.startColumn;
+      column <= range.endColumn;
+      column += 1
+    ) {
+      if (this.inputTable.keyColumns.includes(this.columns[column])) {
```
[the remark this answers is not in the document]
AkshatJawne: Sounds good, will start work on implementation. In terms of working with Java code, I would love to do so, so would be very grateful for the opportunity.

AkshatJawne:
Investigating unit test failures

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -321,6 +321,13 @@ abstract class IrisGridModel<
     return EMPTY_ARRAY;
   }
 
+  /**
+   * @returns Names of key columns
+   */
+  get keyColumnSet(): Set<ColumnName> {
+    return new Set();
```
mofojed, line 328: We don't want to create a new set each time this property is accessed. Since it's just an empty set, you can declare an empty set as a constant at the top (similar to `EMPTY_ARRAY` which we use in other locations; would be `const EMPTY_SET: Set<never> = new Set();`

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,6 +78,24 @@ class IrisGridTableModel
...
+            inputTableKeys.includes(column.name)
+          )
+          .map((column: DhType.Column) => column.name)
+      )
+  );
+
+  get keyColumnSet(): Set<ColumnName> {
+    return this.getMemoizedKeyColumnSet(
+      this.table.columns,
+      this.inputTable?.keys ?? []
```
mofojed, line 95: We should use `EMPTY_ARRAY` here instead of `[]`. Looks like we're making that same mistake in a couple other methods below (`frontColumns`).
The problem with falling back to `[]` is that if `inputTable` is undefined, then we're creating a new array each time, which means the memoization is _always_ going to fail, since two different array instances do not equal each other (e.g. `[] == []` is always `false`).

Looks like `backColumns`/`getMemoizedBackColumns` has it (mostly) correct below - pass in `undefined` if it's not defined, and then in the memoized function return an array (though the `getMemoizedBackColumns` method should fallback to returning `EMPTY_ARRAY`, not `[]`).

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,6 +78,24 @@ class IrisGridTableModel
...
+  getMemoizedKeyColumnSet = memoize(
+    (tableColumns: DhType.Column[], inputTableKeys: string[]) =>
+      new Set(
+        tableColumns
+          .filter((column: DhType.Column) =>
+            inputTableKeys.includes(column.name)
+          )
+          .map((column: DhType.Column) => column.name)
+      )
+  );
```
mofojed, line 90: I don't know why I suggested memoizing on `table.columns` and keys... we really only need the names of the columns here. Should just memoize on `inputTableKeys`. Also should take the array input as a `readonly` array to be safe. In tandem with my other column below, I think it should just be:

```typescript
getMemoizedKeyColumnSet = memoize((inputTableKeys?: readonly string[]) => new Set(inputTableKeys ?? EMPTY_ARRAY))
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1617,15 +1617,18 @@ class IrisGridTableModelTemplate<
...
     return this.columns[modelIndex].isSortable ?? true;
   }
 
-  isKeyColumn(x: ModelIndex): boolean {
-    return x < (this.inputTable?.keyColumns.length ?? 0);
+  isKeyColumn(column: DhType.Column): boolean {
+    if (column == null) {
+      return false;
+    }
+    return this.keyColumnSet.has(column.name);
```
mofojed, line 1631: We shouldn't need to pass in the column here, just pass in the `ModelIndex` as we were before.
```typescript
isKeyColumn(x: ModelIndex): boolean {
  return this.keyColumnSet.has(this.columns[modelIndex].name);
}
```
Then you shouldn't need to change the calls to `isKeyColumn` elsewhere.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -459,7 +459,7 @@ class IrisGridTableModelTemplate<
 
   get isDeletable(): boolean {
     if (this.inputTable !== null) {
-      return this.inputTable?.keyColumns.length > 0;
+      return this.keyColumnSet.size > 0;
     }
     return false;
```
mofojed, line 464: Don't need to check if `this.inputTable != null`, we can just do `return this.keyColumnSet.size > 0`

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -16,6 +16,8 @@ import { PartitionedGridModelProvider } from './PartitionedGridModel';
 
 const log = Log.module('IrisGridTableModel');
 
+const EMPTY_ARRAY: readonly string[] = [];
```
mofojed, line 19: You can import `EMPTY_ARRAY` from `@deephaven/utils` and use that, we don't need to redefine it.

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,21 +80,32 @@ class IrisGridTableModel
     return this.table.applyCustomColumns != null;
   }
 
+  getMemoizedKeyColumnSet = memoize(
+    (inputTableKeys?: readonly string[]) =>
+      new Set(inputTableKeys ?? EMPTY_ARRAY)
+  );
+
+  get keyColumnSet(): Set<ColumnName> {
+    return this.getMemoizedKeyColumnSet(this.inputTable?.keys ?? EMPTY_ARRAY);
```
mofojed, line 89: Don't need to fallback here, since the `getMemoizedKeyColumnSet` can handle `undefined` being passed in
```suggestion
    return this.getMemoizedKeyColumnSet(this.inputTable?.keys);
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,21 +80,32 @@ class IrisGridTableModel
...
   }
 
   getMemoizedBackColumns = memoize(
     (layoutHintsBackColumns: ColumnName[] | undefined) =>
-      layoutHintsBackColumns ?? []
+      layoutHintsBackColumns ?? EMPTY_ARRAY
   );
 
-  get backColumns(): ColumnName[] {
+  get backColumns(): ColumnName[] | readonly ColumnName[] {
```
mofojed, line 108:
```suggestion
  get backColumns(): readonly ColumnName[] {
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,21 +80,32 @@ class IrisGridTableModel
...
+
   getMemoizedFrontColumns = memoize(
     (layoutHintsFrontColumns: ColumnName[] | undefined) =>
-      layoutHintsFrontColumns ?? []
+      layoutHintsFrontColumns ?? EMPTY_ARRAY
   );
 
-  get frontColumns(): ColumnName[] {
-    return this.getMemoizedFrontColumns(this.layoutHints?.frontColumns ?? []);
+  get frontColumns(): ColumnName[] | readonly ColumnName[] {
```
mofojed, line 97:
```suggestion
  get frontColumns(): readonly ColumnName[] {
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,21 +80,32 @@ class IrisGridTableModel
...
+  getMemoizedKeyColumnSet = memoize(
+    (inputTableKeys?: readonly string[]) =>
+      new Set(inputTableKeys ?? EMPTY_ARRAY)
+  );
+
+  get keyColumnSet(): Set<ColumnName> {
+    return this.getMemoizedKeyColumnSet(this.inputTable?.keys ?? EMPTY_ARRAY);
+  }
+
   getMemoizedFrontColumns = memoize(
```
mofojed, line 92: Input should be `readonly ColumnName[]`

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -78,21 +80,32 @@ class IrisGridTableModel
...
 
-  get frontColumns(): ColumnName[] {
-    return this.getMemoizedFrontColumns(this.layoutHints?.frontColumns ?? []);
+  get frontColumns(): ColumnName[] | readonly ColumnName[] {
+    return this.getMemoizedFrontColumns(
+      this.layoutHints?.frontColumns ?? undefined
+    );
   }
 
   getMemoizedBackColumns = memoize(
```
mofojed, line 103: Input should be `readonly ColumnName[]`

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -102,10 +115,11 @@ class IrisGridTableModel
     (
       layoutHintsFrozenColumns?: ColumnName[],
       userFrozenColumns?: ColumnName[]
-    ): ColumnName[] => userFrozenColumns ?? layoutHintsFrozenColumns ?? []
+    ): ColumnName[] | readonly ColumnName[] =>
+      userFrozenColumns ?? layoutHintsFrozenColumns ?? EMPTY_ARRAY
   );
 
-  get frozenColumns(): ColumnName[] {
+  get frozenColumns(): ColumnName[] | readonly ColumnName[] {
```
mofojed, line 122:
```suggestion
  get frozenColumns(): readonly ColumnName[] {
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -102,10 +115,11 @@ class IrisGridTableModel
     (
       layoutHintsFrozenColumns?: ColumnName[],
       userFrozenColumns?: ColumnName[]
-    ): ColumnName[] => userFrozenColumns ?? layoutHintsFrozenColumns ?? []
+    ): ColumnName[] | readonly ColumnName[] =>
```
mofojed, line 118: Input parameters should also be `readonly ColumnName[]`
```suggestion
    ): readonly ColumnName[] =>
```

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1625,7 +1622,11 @@ class IrisGridTableModelTemplate<
   }
 
   isKeyColumn(x: ModelIndex): boolean {
-    return x < (this.inputTable?.keyColumns.length ?? 0);
+    if (x < 0 || x >= this.columnCount) {
+      return false;
+    }
+
```
mofojed, line 1628: I don't think this check is necessary - we should never be passed an `x` that is outside of the column count range.

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModel.ts
AkshatJawne: Shouldn't this be a set though, because the getMemoizedKeyColumnSet returns a Set,  not an array?

AkshatJawne:
Have to add check in `isKeyColumn`, or else, it tries to access the `name` property of the column at index `x`, and then that raises an error when the `columns` array is empty.

mofojed (approved):

Issue #1930, opened 2024-04-11
Remove TreeTableViewportUpdater and TableViewportUpdater


mofojed:
It's not used anywhere, and the table on it is typed incorrectly.

PR #2057, opened 2024-06-04
fix: Remove TreeTableViewportUpdater, TableViewportUpdater, and StorageTableViewportUpdater

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1930

BREAKING CHANGE: Removed `TreeTableViewportUpdater`,`TableViewportUpdater`, and `StorageTableViewportUpdater`. If wanting to continue using them, copy the deleted files from this PR.


mattrunyon:
These are actually consumed in DHE, so this should be marked as breaking. There should be info on this in the `CONTRIBUTING.md` file in the root. Basically the PR body will need `BREAKING CHANGE: Explanation` at the end

Also, `StorageTableViewportUpdater` is used in DHE as well.

mofojed (changes requested):
May as well get rid of `StorageTableViewportUpdater` as well in this PR, add it to the BREAKING CHANGE list

mofojed (approved):

PR #2061, opened 2024-06-06
feat: Allow custom renderer to be passed into IrisGrid

mofojed:
- For adding your own grid/cell renderer


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1430,6 +1430,15 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     ): readonly GridMouseHandler[] => [...mouseHandlers, ...this.mouseHandlers]
   );
 
+  getCachedRenderer = memoize(
+    (rendererProp?: IrisGridRenderer) => rendererProp ?? new IrisGridRenderer()
+  );
```
mattrunyon, line 1435: Should this have `{ max: 1 }`? No need to keep other renderers around if it changed (even though it should be uncommon)

mattrunyon (changes requested):

mattrunyon (approved):

PR #2062, opened 2024-06-06
fix: A few small cleanups for DateTimeInput

mofojed:
- Don't allow an undefined to be returned
- Export more types

### DIFF
--- a/packages/components/src/DateTimeInput.tsx
+++ b/packages/components/src/DateTimeInput.tsx
@@ -46,7 +46,7 @@ const EXAMPLES = [addSeparators(DEFAULT_VALUE_STRING)];
 export const DateTimeInput = React.forwardRef<
   HTMLInputElement,
   DateTimeInputProps
->((props: DateTimeInputProps, ref) => {
+>((props, ref) => {
   const {
     className = '',
     onChange = () => undefined,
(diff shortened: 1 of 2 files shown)


packages/components/src/DateTimeInput.tsx
```diff
@@ -43,64 +43,65 @@ function removeSeparators(value: string): string {
...
-      'data-testid': dataTestId,
-    } = props;
-    const [value, setValue] = useState(
-      defaultValue.length > 0 ? addSeparators(defaultValue) : ''
-    );
-    const [selection, setSelection] = useState<SelectionSegment>();
+export const DateTimeInput = React.forwardRef<
+  HTMLInputElement,
+  DateTimeInputProps
+>((props: DateTimeInputProps, ref) => {
```
mattrunyon, line 49: You shouldn't need to define the props type if using the `forwardRef` generic for it already. It looks like the generic automatically types it as `PropsWithChildren<DateTimeInputProps>`

```suggestion
export const DateTimeInput = React.forwardRef<
  HTMLInputElement,
  DateTimeInputProps
>((props, ref) => {
```

mattrunyon (changes requested):

mattrunyon (approved):

PR #2063, opened 2024-06-07
fix: view border styling

ethanalvizo:
Exposed border color style props for view component.

Had to define border color styling using shorthand properties instead to avoid re-rendering bugs. Left comment in code explaining why this approach was taken

image

### DIFF
--- a/packages/components/src/spectrum/View.tsx
+++ b/packages/components/src/spectrum/View.tsx
@@ -43,6 +43,8 @@ export const View = forwardRef<DOMRefValue<HTMLElement>, ViewProps>(
       ...rest
     } = props;
 
+    // Using shorthand to define border styling. Removing a style property during rerender can lead to styling bugs
+    // ex. changing from borderXColor = blue to borderColor = blue will cause the left and right borders to disappear
     const defaultBorderColor = colorValueStyle(borderColor) ?? 'transparent';
     const defaultBorderXColor =
       colorValueStyle(borderXColor) ?? defaultBorderColor;
@@ -71,7 +73,7 @@ export const View = forwardRef<DOMRefValue<HTMLElement>, ViewProps>(
       borderXColor,
       borderYColor,
     ]);
-    
+
     return <SpectrumView {...rest} ref={forwardedRef} UNSAFE_style={style} />;
   }
 );
--- a/packages/components/src/spectrum/View.tsx
+++ b/packages/components/src/spectrum/View.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import { forwardRef, useMemo } from 'react';
+import { forwardRef, useMemo, useEffect, CSSProperties } from 'react';
 import {
   View as SpectrumView,
   type ViewProps as SpectrumViewProps,
@@ -9,6 +9,13 @@ import { type ColorValue, colorValueStyle } from '../theme/colorUtils';
 
 export type ViewProps = Omit<SpectrumViewProps<6>, 'backgroundColor'> & {
   backgroundColor?: ColorValue;
+  borderColor?: ColorValue;
+  borderStartColor?: ColorValue;
+  borderEndColor?: ColorValue;
+  borderTopColor?: ColorValue;
+  borderBottomColor?: ColorValue;
+  borderXColor?: ColorValue;
+  borderYColor?: ColorValue;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


packages/components/src/spectrum/View.tsx
```diff
@@ -9,6 +9,13 @@ import { type ColorValue, colorValueStyle } from '../theme/colorUtils';
 
 export type ViewProps = Omit<SpectrumViewProps<6>, 'backgroundColor'> & {
```
mofojed, line 10: Should add the `border` colour props that you're redefining to the `Omit` here (like `backgroundColor` is), e.g.
```
export type ViewProps = Omit<SpectrumViewProps<6>, 'backgroundColor' | 'borderColor' | ...>
```

packages/components/src/spectrum/View.tsx
```diff
@@ -23,13 +30,46 @@ export type ViewProps = Omit<SpectrumViewProps<6>, 'backgroundColor'> & {
```
mofojed, line 30: Can you delete this blank line so the comment is actually with the function? Also the last blank line of the comment... and check that comment for typos ("paletee" should be "palette"

packages/components/src/spectrum/View.tsx
```diff
@@ -23,13 +30,46 @@ export type ViewProps = Omit<SpectrumViewProps<6>, 'backgroundColor'> & {
...
+      borderTopColor,
+      borderBottomColor,
+      borderXColor,
+      borderYColor,
+      UNSAFE_style,
+      ...rest
+    } = props;
+
+    // Using shorthand to define border styling. Removing a style property during rerender can lead to styling bugs
+    // ex. changing from borderXColor = blue to borderColor = blue will cause the left and right borders to disappear
```
mofojed, line 47: Weird, I can't find the error you mentioned in the [React spectrum source code](https://github.com/adobe/react-spectrum/blob/main/packages/@react-spectrum/utils/src/styleProps.ts) at all. But the error also says "replace the shorthand with separate values", which is the opposite of what you're doing here (you're converting everything to [shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties)).
Also, we shouldn't be setting these properties at all if they're not set; leave them `undefined`, not `'transparent'`.

mofojed (changes requested):

packages/components/src/spectrum/View.tsx
ethanalvizo: My bad on this one I think I was misusing the useMemo hook earlier. The new way I define borderStyle defines everything into separate values and unset properties are no longer set to transparent

ethanalvizo:
Connected with https://github.com/deephaven/deephaven-plugins/pull/506 so that the border color styling props are applied

packages/components/src/spectrum/View.tsx
```diff
@@ -43,34 +52,48 @@ export const View = forwardRef<DOMRefValue<HTMLElement>, ViewProps>(
...
-      colorValueStyle(borderYColor) ?? defaultBorderColor;
-    const topColor = colorValueStyle(borderTopColor) ?? defaultBorderYColor;
-    const bottomColor =
-      colorValueStyle(borderBottomColor) ?? defaultBorderYColor;
-    const leftColor = colorValueStyle(borderStartColor) ?? defaultBorderXColor;
-    const rightColor = colorValueStyle(borderEndColor) ?? defaultBorderXColor;
+    const style = useMemo(() => {
+      const borderStyle: CSSProperties = {};
+      if (borderColor !== undefined) {
+        borderStyle.borderColor = colorValueStyle(borderColor);
```
mofojed, line 58: It would be nice if we could just pass our own conversion function into `View`... Right now `View` is using `useStyleProps` with the `viewStyleProps` handlers to convert functions: https://github.com/adobe/react-spectrum/blob/bd458c1ed166a5ff1fd93f0c1a397e1540b3d880/packages/%40react-spectrum/view/src/View.tsx#L26C43-L26C57
If we could just pass our own color conversion handler in there somehow, then we wouldn't have to rewrite these props for every color: https://github.com/adobe/react-spectrum/blob/bd458c1ed166a5ff1fd93f0c1a397e1540b3d880/packages/%40react-spectrum/utils/src/styleProps.ts#L263
I'm tempted to file a ticket/PR in the Adobe GitHub repo to add this functionality, since we're ending up having to duplicate a bunch of things here just to handle the borders... but this should be fine for now.

packages/components/src/spectrum/View.tsx
```diff
@@ -43,34 +52,48 @@ export const View = forwardRef<DOMRefValue<HTMLElement>, ViewProps>(
...
+      if (borderXColor !== undefined) {
+        borderStyle.borderLeftColor = colorValueStyle(borderXColor);
+        borderStyle.borderRightColor = colorValueStyle(borderXColor);
+      }
+      if (borderYColor !== undefined) {
+        borderStyle.borderTopColor = colorValueStyle(borderYColor);
+        borderStyle.borderBottomColor = colorValueStyle(borderYColor);
+      }
+      if (borderStartColor !== undefined) {
+        borderStyle.borderLeftColor = colorValueStyle(borderStartColor);
```
mofojed, line 69: Technically, this would be incorrect for RTL languages. We should be using `useLocale()` to get the direction of text, and setting the appropriate side depending on if it's RTL or LTR: https://github.com/adobe/react-spectrum/blob/bd458c1ed166a5ff1fd93f0c1a397e1540b3d880/packages/%40react-spectrum/utils/src/styleProps.ts#L259
So something like:
```
const { direction } = useLocale();

...
if (borderStartColor !== undefined) {
  borderStyle[direction === 'rtl' ? 'borderRightColor' : 'borderLeftColor'] = convertValueStyle(borderStartColor);
}
```

Alternatively you could find a way to use `useStyleProps` or `convertStyleProps` directly, and not have to add all this stuff... though that would mean passing in our own [viewStyleProps](https://github.com/adobe/react-spectrum/blob/bd458c1ed166a5ff1fd93f0c1a397e1540b3d880/packages/%40react-spectrum/utils/src/styleProps.ts#L66) with the `convertValueStyle` as the handler... probably simpler to just check rtl and do it here.

mofojed (changes requested):
It's fine as is, but would be incorrect for RTL languages. We may as well add that support now.

packages/components/src/spectrum/View.tsx
dsmmcken: I was able to use useStyleProps/convertStyleProps previously when I was doing background, so that is possible but it ultimately ended up simpler to not. 

As far as I am aware, there's no other components where this applies, so I would probably just do what you are suggesting instead.

mofojed (approved):

PR #2064, opened 2024-06-07
fix: Embed-widget with multiple panels not showing panel headers

mattrunyon:
This should fix embed-widget not showing the panel headers when there are multiple panels in a widget, but it's not a dashboard.

The issue is embed-widget is loading the layout from `dashboardData` and the `DEFAULT_DASHBOARD_ID` stores its data in `workspaceData` while all other IDs use `dashboardData`.

Tested with widget with just 1 panel, widget with multiple panels, and widget that is a dashboard.


bmingles (approved):
LGTM. Verified all tabs show now in http://localhost:4010/?name=sti for the ui table example

PR #2067, opened 2024-06-10
feat: ComboBox - @deephaven/components

### LINKED ISSUE #2065: ComboBox - @deephaven/components
@deephaven/components ComboBox component to support Deephaven UI

bmingles:
* New `ComboBox` component in @deephaven/components
* Split out some shared logic from `Picker` since `ComboBox` is basically a subclass
* Deleted old `ComboBox` component and replaced usage (condition column + row formatting. Also tested to make sure it still works as before)
* Updated Styleguide

resolves #2065

BREAKING CHANGE: ComboBox component has been replaced. 
To migrate to new version:
- Passing children is used instead of `options` prop to define dropdown items. For cases where option value and display are the same, passing an array of values as `children` will work. For cases where value and display differ, `Item` elements must be passed as children. e.g. `<Item key={value}>{display}</Item>`
e.g.
```typescript
// values will be used for display + value
const items = useMemo(
  () => ['Aaa', 'Bbb', 'Ccc'], 
  []
)
<ComboBox>{items}</ComboBox>
```
```typescript
<ComboBox>
  <Item key="aaa">Aaa</Item>
  <Item key="bbb">Bbb</Item>
  <Item key="ccc">Ccc</Item>
</ComboBox>
```
- The `spellcheck=false` prop is no longer supported or needed
- `searchPlaceholder` and `inputPlaceholder` props are no longer supported and should be omitted. There is an optional `description` prop for cases where a descriptive label is desired. There is also a `label` prop for the primary component label.


packages/components/src/ComboBox.tsx
mofojed, line 1: This change should be marked BREAKING, with migration instructions to transition to the new `ComboBox`.
Pretty much just mapping the `options` to `children` and wrapping them in `Item`? Would be good if you had an example.
I notice we don't have a `spellCheck` prop anymore (seems to be `false` by default, which is probably what we want anyway), and no more `placeholder` props (Spectrum doesn't seem to like them) - what should we do when migrating those props?

packages/components/src/spectrum/comboBox/ComboBox.tsx
```diff
@@ -0,0 +1,44 @@
+import {
+  ComboBox as SpectrumComboBox,
+  SpectrumComboBoxProps,
+} from '@adobe/react-spectrum';
+import type { FocusableRef } from '@react-types/shared';
+import cl from 'classnames';
+import type { NormalizedItem } from '../utils';
+import { PickerPropsT, usePickerProps } from '../picker';
+
+export type ComboBoxProps = PickerPropsT<SpectrumComboBoxProps<NormalizedItem>>;
```
mofojed, line 10: I think we still want to template the props here so items get passed through correctly no?
Something like:
```suggestion
export type ComboBoxProps<T extends NormalizedItem = NormalizedItem> = PickerPropsT<SpectrumComboBoxProps<T>>;
```

tests/styleguide.spec.ts-snapshots/inputs-chromium-linux.png
mofojed, line 1: See in these demos we have no `placeholder` anymore. We should have instructions how to deal with that (default to the desired value instead? Display a validation error if blank with an appropriate warning?) and show those cases in the styleguide.

mofojed (changes requested):
Looks good, just needs to be labelled a breaking change and then migration support should be a little more detailed (particularly around `placeholder`)

packages/components/src/spectrum/comboBox/ComboBox.tsx
bmingles: What's the use case? `SpectrumComboBoxProps<NormalizedItem>` should already pass things through. We aren't extending `NormalizedItem` so I think it should be ok?

mofojed: I guess you're right - I was thinking if you pass in items that extended items and wanted the original item back, but that's not how this works anyway.

mofojed (approved):

PR #2068, opened 2024-06-11
feat: Embed widget loading workspace settings

### LINKED ISSUE #1964: Embed widget should load workspace settings

mattrunyon:
Fixes #1964. Also removed a bunch of unnecessary code in `AppInit` where we were using the redux `connect` API with a functional component instead of just using the redux hooks

Note that even though this moves a bunch of files to `app-utils`, I don't consider it breaking since they are being moved out of `code-studio`. They wouldn't be consumed by anything else since `code-studio` is published as a bundle.

Due to local storage requiring a port match, this can only be tested w/ the local server in 2 steps (this should be fine in prod since the app and embed-widget run on the same port there).

1. Start the dev server and go to the app
2. Run some code to create a table.
3. Change user formatting settings (e.g. show timezones and have a timestamp column)
4. Stop the dev server.
5. Run `PORT=4000 npm run start:embed-widget`. This will let us access the saved workspace from the app
6. Go to `localhost:4000/?name=<your-variable-name>`
7. The formatting settings should be applied


packages/app-utils/src/storage/grpc/GrpcFileStorageTable.test.ts
```diff
@@ -1,20 +1,22 @@
...
-function makeStorageService(): StorageService {
+function makeStorageService(): dh.storage.StorageService {
   return {
     listItems: jest.fn(async () => []),
     loadFile: jest.fn(async () => {
       throw new Error('No file loaded');
     }),
     deleteItem: jest.fn(async () => undefined),
-    saveFile: jest.fn(async () => undefined),
+    saveFile: jest.fn(
```
bmingles, line 17: This might be a good spot for 

```typescript
TestUtils.createMockProxy<dh.storage.StorageService>({
  loadFile: jest.fn(async () => {
      throw new Error('No file loaded');
  }),
})
```

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -233,51 +164,4 @@ function AppInit(props: AppInitProps): JSX.Element {
   );
 }
 
-AppInit.propTypes = {
```
bmingles, line 236: yay for removing `propTypes`

packages/embed-widget/src/App.tsx
```diff
@@ -60,6 +88,43 @@ function App(): JSX.Element {
             throw new Error('Missing URL parameter "name"');
           }
 
+          const sessionDetails = await getSessionDetails();
```
bmingles, line 91: There seems to be a lot of duplicate code here as `AppInit` `initApp` effect. Any chance we could split out the common code into a `useInitApp` hook or something?

mattrunyon: Probably. I thought about it and initially had most of the new code in a separate effect, but it actually caused a race condition with loading the table. I think I'd name it `useInitRedux` or something like that.

I did also try moving some of the redux setters to their respective `*Bootstrap` components. But would need to hoist the redux store up since right now it's a child of all the bootstraps. Shouldn't have an impact though if that's what we went with

bmingles:
Tested and confirmed this works as described.

bmingles (approved):
Left 2 suggestions. Nothing big enough to block the PR

packages/code-studio/src/main/AppInit.tsx
```diff
@@ -1,124 +1,69 @@
...
+import { setLayoutStorage } from '../redux/actions';
 import App from './App';
-import LocalWorkspaceStorage from '../storage/LocalWorkspaceStorage';
-import LayoutStorage from '../storage/LayoutStorage';
-import GrpcLayoutStorage from '../storage/grpc/GrpcLayoutStorage';
-import GrpcFileStorage from '../storage/grpc/GrpcFileStorage';
 
 const log = Log.module('AppInit');
 
-interface AppInitProps {
```
mofojed, line 59: Woo that's some nice cleanup

packages/embed-widget/src/App.tsx
mofojed: Instead of hook could just be an async helper function that gets called with all these params passed in, then `await initializeApp(...)` here; or a Bootstrap component as you suggested. Either way this is fine for now.

mofojed (approved):

packages/embed-widget/src/App.tsx
mattrunyon: True. Either way pulling it out then requires moving some redux things out of code-studio and I think this is fine as is for now

PR #2070, opened 2024-06-12
test: Added jest.clearAllMocks and expect.hasAssertions to global test setup

### LINKED ISSUE #2069: Add `jest.clearAllMocks` and `expect.hasAssertions()` to global test setup
I find I put the same boilerplate setup / teardown in every unit test I write:

```typescript
beforeEach(() => {
  // Good practice to avoid tests accidentally affecting other tests
  jest.clearAllMocks();
});

afterEach(() => {
  // Good practice to avoid forgetting to actually test something
  expect.hasAssertions();
});
```

We should add these to the `jest.setup.ts` to globally configure for all tests and cut down on boilerplate.

bmingles:
Added jest.clearAllMocks and expect.hasAssertions to global test setup

resolves #2069

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -88,3 +88,11 @@ Object.defineProperty(window.CSS, 'supports', {
   value: () => true,
   writable: true,
 });
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+afterEach(() => {
+  expect.hasAssertions();
+});
--- a/packages/code-studio/src/main/AppMainContainer.test.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.test.tsx
@@ -167,7 +167,7 @@ afterEach(() => {
 });
 
 it('mounts and unmounts AppMainContainer without crashing', () => {
-  renderAppMainContainer();
+  expect(renderAppMainContainer()).toBeDefined();
 });
 
 it('listens for widgets properly', async () => {
--- a/packages/components/src/BasicModal.test.tsx
+++ b/packages/components/src/BasicModal.test.tsx
@@ -23,7 +23,7 @@ function makeWrapper(partialProps: Partial<BasicModalProps> | undefined = {}) {
 }
 
 it('mounts and unmounts without failing', () => {
-  makeWrapper();
+  expect(makeWrapper()).toBeDefined();
 });
 
 it('focuses default button on first render after opening', () => {
--- a/packages/components/src/ComboBox.test.tsx
+++ b/packages/components/src/ComboBox.test.tsx
@@ -7,5 +7,5 @@ function makeWrapper(options = []) {
 }
 
 it('mounts and unmounts without failing', () => {
-  makeWrapper();
+  expect(makeWrapper()).toBeDefined();
 });
--- a/packages/components/src/DateTimeInput.test.tsx
+++ b/packages/components/src/DateTimeInput.test.tsx
@@ -27,6 +27,7 @@ function makeDateTimeInput({
 it('mounts and unmounts properly', () => {
   const { unmount } = makeDateTimeInput();
   unmount();
+  expect('ok').toBe('ok');
 });
 
 it('trims trailing mask and spaces in the input', async () => {
--- a/packages/components/src/EditableItemList.test.tsx
+++ b/packages/components/src/EditableItemList.test.tsx
@@ -30,6 +30,7 @@ function makeWrapper({
 it('mounts and unmounts without failing', () => {
   const { unmount } = makeWrapper();
   unmount();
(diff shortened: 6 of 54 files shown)


bmingles:
@mofojed I've updated the ticket description on #2069

jest.setup.ts
```diff
@@ -88,3 +88,11 @@ Object.defineProperty(window.CSS, 'supports', {
...
   writable: true,
 });
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+afterEach(() => {
+  expect.hasAssertions();
+});
```
mofojed, line 98: I think these should both be in the `beforeEach` call. In the examples they show it being called before testing code: https://jestjs.io/docs/expect#expecthasassertions
```suggestion
beforeEach(() => {
  jest.clearAllMocks();
  expect.hasAssertions();
});
```
Also, I'm not 100% sure I like doing the `expect.hasAssertions()`... for example, if the code throws before it gets to your assertion, the test fails, but then afterwards it prints out "Expected at least one assertion to be called but received none", but the test has already failed and it's kind of extra noise. For example, I changed `UISwitch.test.tsx` to have a failing test:
```
it('mounts and unmounts properly', () => {
  const { unmount } = makeUISwitch();
  // There's nothing with role `'menu'` here, so it throws before it gets that, and before our expect runs
  // (Can change to `'button'` to get it to pass)
  expect(screen.getByRole('menu')).toBeInTheDocument();
  unmount();
});
```

Then in the error output, I get two blocked printed out, with the second being slightly more visible but less useful:
```
Summary of all failing tests
 FAIL  packages/components/src/UISwitch.test.tsx
  ● mounts and unmounts properly

    TestingLibraryElementError: Unable to find an accessible element with the role "menu"

    Here are the accessible roles:

      button:

      Name "":
      <button
        class="btn btn-switch active"
        data-testid="TestUISwitch"
        type="button"
      />

      --------------------------------------------------

    Ignored nodes: comments, <script />, <style />
    <body>
      <div>
        <button
          class="btn btn-switch active"
          data-testid="TestUISwitch"
          type="button"
        >
          <div
            class="handle"
          />
        </button>
      </div>
    </body>

      16 | it('mounts and unmounts properly', () => {
      17 |   const { unmount } = makeUISwitch();
    > 18 |   expect(screen.getByRole('menu')).toBeInTheDocument();
         |                 ^
      19 |   unmount();
      20 | });
      21 |

      at Object.getElementError (../../node_modules/@testing-library/dom/dist/config.js:38:19)
      at ../../node_modules/@testing-library/dom/dist/query-helpers.js:90:38
      at ../../node_modules/@testing-library/dom/dist/query-helpers.js:62:17
      at ../../node_modules/@testing-library/dom/dist/query-helpers.js:111:19
      at Object.getByRole (src/UISwitch.test.tsx:18:17)

  ● mounts and unmounts properly

    expect.hasAssertions()

    Expected at least one assertion to be called but received none.

      95 |
      96 | afterEach(() => {
    > 97 |   expect.hasAssertions();
         |          ^
      98 | });
      99 |

      at Object.hasAssertions (../../jest.setup.ts:97:10)
```

mofojed (changes requested):
I don't like the addition of `expect.hasAssertions()` to the global setup.
- It's main use is for asynchronous testing/assertions, as shown in [the documentation](https://jestjs.io/docs/expect#expecthasassertions). Using it for synchronous code doesn't make a lot of sense.
- Many of these tests are simply to check that the code path doesn't throw, which now we need to add a useless assertion around, which just results in more boilerplate/confusing test output
  - We could update those tests to actually check that something useful is rendered to the screen, and then check unmount, but we're still not doing that even with the addition of the `expect.hasAssertions` here, I don't think adding it is helpful and just adds more noise/confusion.

bmingles:
> I don't like the addition of `expect.hasAssertions()` to the global setup.
> 
> * It's main use is for asynchronous testing/assertions, as shown in [the documentation](https://jestjs.io/docs/expect#expecthasassertions). Using it for synchronous code doesn't make a lot of sense.
> * Many of these tests are simply to check that the code path doesn't throw, which now we need to add a useless assertion around, which just results in more boilerplate/confusing test output
>   
>   * We could update those tests to actually check that something useful is rendered to the screen, and then check unmount, but we're still not doing that even with the addition of the `expect.hasAssertions` here, I don't think adding it is helpful and just adds more noise/confusion.

Yeah, that's fair. I didn't realize that it also errored when there were other failures which is annoying. I'm fine to reject this PR if it's not adding any real value. At least things are explicit that way.

bmingles:
Closing as this causes some undesirable side effects that outweigh the benefits

Issue #2060, opened 2024-06-06
Console scroll bar won't follow dynamic output


jjbrosnan:
**Description**

It's pretty common that, when running a query on dynamic data, the scroll bar will not follow the dynamic output of said query in the console window. If I scroll up to see a previous message, and then scroll back down all the way to the bottom, I'd like the scroll bar to stay at the bottom to follow output. However, most of the time, it just stays at the current spot, not following output at all. The actual scroll bar on the RHS of the console slowly moves up, since it's no longer at the bottom of all of the output text.

To repro, you should be able to just create a real-time query that logs some kind of output with ticking data (table listener that prints on update, for example), scroll up in the console window, and scroll back to the bottom. For me, it rarely stays at the bottom and follows output in lock-step.

**Expected results**

The scroll bar to stay at the bottom, which makes the console window follow text output as it's generated in real-time.

**Actual results**

The scroll bar slowly moves up, while the console window stays fixed on the latest line of output that was there when the scroll bar was previously moved to the bottom.

In this case, the only way to keep the console up with new text is to either constantly use the scroll wheel on the mouse, or click and hold the scroll bar at the bottom. The latter also requires moving the mouse constantly while held clicked.

**Additional details and attachments**

None.

**Versions**

 - Engine Version: 0.34.4 (latest)
 - Web UI Version: Latest as of writing this ticket. I have seen older versions exhibit this behavior as well.
 - Java Version: Unknown
 - Barrage Version: Unknown
 - OS: OS X
 - Browser: Chrome
 - Docker: 20.10.13

PR #2076, opened 2024-06-13
fix: Console scroll bar following dynamic output

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2060

**Changes Implemented:**
- Added `isStuckToBottom` state that is updated on each scroll (in `handleScrollPaneScroll`), then in `componentDidUpdate`, call the `scrollConsoleHistoryToBottom` based on `isStuckToBottom`

### DIFF
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -262,7 +262,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
         this.consolePane.current.scrollHeight -
           this.consolePane.current.clientHeight -
           this.consolePane.current.scrollTop
-      ) < 1
+      ) < 0.5
     ) {
       this.scrollConsoleHistoryToBottom();
     }
@@ -669,7 +669,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
           scrollPane.scrollHeight -
             scrollPane.clientHeight -
             scrollPane.scrollTop
-        ) <= 1,
+        ) <= 0.5,
     });
   }
 
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -662,16 +662,16 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
   handleScrollPaneScroll(): void {
     const scrollPane = this.consoleHistoryScrollPane.current;
     assertNotNull(scrollPane);
-    const isAtBottom =
-      Math.abs(
-        scrollPane.scrollHeight - scrollPane.clientHeight - scrollPane.scrollTop
-      ) >= 1;
-
     this.setState({
       isScrollDecorationShown:
         scrollPane.scrollTop > 0 &&
         scrollPane.scrollHeight > scrollPane.clientHeight,
-      isStuckToBottom: isAtBottom,
+      isStuckToBottom:
+        Math.abs(
+          scrollPane.scrollHeight -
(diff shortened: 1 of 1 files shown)


packages/console/src/Console.tsx
```diff
@@ -251,6 +253,10 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
     if (props.objectMap !== prevProps.objectMap) {
       this.updateObjectMap();
     }
+
+    if (state.isStuckToBottom) {
+      this.scrollConsoleHistoryToBottom();
+    }
```
mattrunyon, line 259: You should be able to remove the calls to `scrollConsoleHistoryToBottom` in the 2 spots it's called w/o force and then actually drop the `force` param I think since everywhere would be calling it w/ force.

The function as written is really `queueScrollConsoleHistoryToBottomIfCurrentlyAtBottomOrForced`

So instead the function could just be something like this

```ts
scrollConsoleHistoryToBottom(): void {
  const pane = this.consoleHistoryScrollPane.current;
  if (pane == null) {
    return;
  }
  pane.scrollTo({ top: pane.scrollHeight });
```

packages/console/src/Console.tsx
```diff
@@ -656,14 +662,17 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
...
-      scrollPane.scrollHeight > scrollPane.clientHeight
-    ) {
-      this.setState({ isScrollDecorationShown: true });
-    } else {
-      this.setState({ isScrollDecorationShown: false });
-    }
+    this.setState({
+      isScrollDecorationShown:
+        scrollPane.scrollTop > 0 &&
+        scrollPane.scrollHeight > scrollPane.clientHeight,
```
mattrunyon, line 668: Since I was looking at this trying to figure out what it actually did, can you add a comment to the interface `ConsoleState` for `isScrollDecorationShown` that says `Scroll decoration is a drop shadow across the top border of the console`

packages/console/src/Console.tsx
```diff
@@ -251,6 +253,10 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
     if (props.objectMap !== prevProps.objectMap) {
       this.updateObjectMap();
     }
+
+    if (state.isStuckToBottom) {
+      this.scrollConsoleHistoryToBottom();
```
mattrunyon, line 258: I don't know if it's an optimization that matters, but we could also check that it's not already scrolled to the bottom before scrolling here.

Not sure if telling an element to scroll to the position it's already at is expensive in any way in the DOM.

mattrunyon (changes requested):

packages/console/src/Console.tsx
AkshatJawne: I tried looking into this, don't feel like this adds much value, given that the ways I was trying seemed quite inefficient. Unless this is really needed, I think the way we have it currently is simpler to understand + does not require unnecessary DOM manipulations / operations.

mattrunyon: If it does pose a performance issue we notice in the future it should be easy enough to fix.

You would just take the `scrollHeight - clientHeight >= 1` or w/e the check is and put it here as well so it's `isStuckToBottom && notAtBottom`

AkshatJawne: Fair enough, will retest with change and repush

mattrunyon (approved):

PR #2077, opened 2024-06-13
feat: ComboBox - @deephaven/jsapi-components

### LINKED ISSUE #2074: ComboBox - @deephaven/jsapi-components
We need a ComboBox component in the @deephaven/jsapi-components package to provide support for table data sources. This depends on the ComboBoxNormalized component in https://github.com/deephaven/web-client-ui/issues/2071

bmingles:
Jsapi support for ComboBox. Includes some splitting out of existing Picker logic to make code re-usable. 

Should be testable with plugins PR https://github.com/deephaven/deephaven-plugins/pull/588
I have also deployed an alpha [0.83.1-alpha-combobox.8](https://www.npmjs.com/package/@deephaven/components/v/0.83.1-alpha-combobox.8) if you need it, although should only matter for types I think.

```python
from deephaven import empty_table, ui, time_table
import datetime


# Change this to test different data types
key_column="Timestamp"


initial_row_count=5 * 8760 # 5 years in hours

# Tick every 6 hours (makes it easier to test Timestamp filters for a whole day like `2024-01-02`)
_items = time_table("PT6H", start_time=datetime.datetime.now() - datetime.timedelta(hours=initial_row_count)).update([
    # Timestamp column also implicitly included in `time_table`
    "Int=new Integer(i)",
    "Long=new Long(i)",
    "BigInt=new java.math.BigInteger(``+i)",
    "String=new String(`a`+i * 1000)",
])


@ui.component
def ui_combo_box(items):
    value, set_value = ui.use_state("")

    combo = ui.combo_box(
        ui.item_table_source(
            items,
            key_column=key_column,
            label_column=key_column,
        ),
        label=key_column,
        on_selection_change=set_value,
        menu_trigger="focus",
        selected_key=value,
    )

    # Show current selection in a ui.text component
    text = ui.text("Selection: " + str(value))

    return combo, text


my_combo_box = ui_combo_box(_items)
```

There is a known issue with inconsistent open as you type for table data sources.
https://github.com/deephaven/web-client-ui/issues/2115

resolves #2074


packages/jsapi-components/src/spectrum/utils/usePickerProps.ts
```diff
@@ -0,0 +1,187 @@
...
+  // for explicit `undefined` to identify uncontrolled mode.
+  const isUncontrolled = props.selectedKey === undefined;
+  const [uncontrolledSelectedKey, setUncontrolledSelectedKey] = useState<
+    ItemKey | null | undefined
+  >(props.defaultSelectedKey);
+
+  // Copy table so we can apply filters without affecting the original table.
+  // (Note that this call is not actually applying any filters. Filter will be
+  // applied in `useSearchableViewportData`.)
+  const { data: tableCopy } = usePromiseFactory(
```
bmingles, line 75: @mofojed for reference, I think this is where we could apply addition filters as part of @AkshatJawne partition table work

bmingles: I guess it might make sense to just add it to this PR, but I'm fine either way

mofojed (approved):

mofojed (changes requested):
Only thing that needs updating is the PR description
Show passing in a `ui.item_table_data_source` instead of passing in props for `key_column` directly on `combo_box`.

mofojed (approved):

Issue #1771, opened 2024-02-05
Dashboard plugin that throws error crashes UI


mofojed:
**Description**

If you run a dashboard plugin that throws in the render cycle, it will crash DH. Instead, it should just log an error and not crash the application.

**Steps to reproduce**
1. Modify a DashboardPlugin to throw in render
2. Start up the UI

**Expected results**
2. UI should start, with an error logged about the dashboard plugin that threw

**Actual results**
2. Deephaven UI crashes.

PR #2080, opened 2024-06-13
fix: Dashboard plugin crashing UI on throw

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/1771

**Changes Implemented:** 
- Wrap `ErrorBoundary` around cloning of Dashboard Plugin in DashboardLayout

### DIFF
--- a/packages/dashboard/src/DashboardLayout.tsx
+++ b/packages/dashboard/src/DashboardLayout.tsx
@@ -15,6 +15,7 @@ import type {
 } from '@deephaven/golden-layout';
 import Log from '@deephaven/log';
 import { usePrevious, useThrottledCallback } from '@deephaven/react-hooks';
+import { ErrorBoundary } from '@deephaven/components';
 import { RootState } from '@deephaven/redux';
 import { useDispatch, useSelector } from 'react-redux';
 import PanelManager, { ClosedPanels } from './PanelManager';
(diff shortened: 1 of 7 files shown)


AkshatJawne:
Will look into failures here

AkshatJawne:
Will remove commented out code if approach works

packages/dashboard/src/DashboardLayout.tsx
```diff
@@ -343,14 +344,16 @@ export function DashboardLayout({
       {isDashboardEmpty && emptyDashboard}
       {layoutChildren}
       {React.Children.map(children, child =>
-        child != null
-          ? React.cloneElement(child as ReactElement, {
+        child != null ? (
+          <ErrorBoundary fallback={[]}>
```
mattrunyon, line 348: Add a comment that the fallback is so we do not show the error message over the entire app and we should do something like a toast message w/ the error in the future

mattrunyon (changes requested):
Update the PR body as well. Looks like what you listed was the original fix attempt

mattrunyon (approved):
Looks good. Also at least prevents the crash in https://github.com/deephaven/deephaven-plugins/issues/510 but that ticket still documents bad behavior in Deephaven UI, so it should stay open

PR #2085, opened 2024-06-18
refactor: Make IrisGridProxyModel use an actual Proxy

mattrunyon:
This makes `IrisGridProxyModel` use an actual Proxy instead of requiring us pass through most of the calls to the underlying model.

The proxy will check if the `IrisGridProxyModel` class implements a method first. If it does, then it calls the implemented method. If it does not, it delegates the call to the current `model`

Many of the methods which had a check for a specific model type first can also be handled by the proxy. A caller is must use that check first anyway or else TS will complain. So the check was really being done twice before

Didn't look into it too hard, but I disabled the TS check for `extends IrisGridModel`. There are many abstract methods in `IrisGridModel` which are implemented by the proxy handling them, but TS doesn't know that.

### DIFF
--- a/packages/iris-grid/src/GotoRow.tsx
+++ b/packages/iris-grid/src/GotoRow.tsx
@@ -22,15 +22,9 @@ import classNames from 'classnames';
 import './GotoRow.scss';
 import { nanoid } from 'nanoid';
 import IrisGridModel from './IrisGridModel';
-import IrisGridProxyModel from './IrisGridProxyModel';
 import IrisGridBottomBar from './IrisGridBottomBar';
 import { ColumnName } from './CommonTypes';
-
-function isIrisGridProxyModel(
-  model: IrisGridModel
-): model is IrisGridProxyModel {
-  return (model as IrisGridProxyModel).model !== undefined;
-}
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/GotoRow.tsx
```diff
@@ -90,7 +84,7 @@ const GotoRow = forwardRef<GotoRowElement, GotoRowProps>(
     const [isGotoRowActive, setIsGotoRowActive] = useState(false);
     let columns: dh.Column[] = [];
 
-    if (isIrisGridProxyModel(model) && model.table !== undefined) {
+    if (isIrisGridTableModelTemplate(model)) {
```
mofojed, line 87: Side note, `IrisGridModel` should just be an interface, and then the models could be classes or just objects with the correct properties or whatever.
Not for this ticket though, that'd be a bigger overhaul (as `GridModel` would also need to be an interface), and we'd need to do something about needing the model to be an EventListener (could just another type like `type UpdatingGridModel = { onUpdate: () => void }`

mofojed (approved):
Nice.

PR #2086, opened 2024-06-20
fix: DH-17292 Handle disconnect from GridWidgetPlugin

mofojed:
- When the model is disconnected, we should just display an error. There's no option to reconnect, as the widget schema could change entirely
- By unloading the IrisGrid component, it's no longer throwing an error by trying to access table methods after a table.close()
- Fixes DH-17292 from Enterprise
  - Tested using the steps in the description


packages/dashboard-core-plugins/src/GridWidgetPlugin.tsx
```diff
@@ -36,6 +40,36 @@ export function GridWidgetPlugin(
...
+  useEffect(
+    function startListeningModel() {
+      if (!model) {
+        return;
+      }
+
+      // If the table inside a widget is disconnected, then don't bother trying to listen to reconnect, just close it and show a message
+      // Widget closes the table already when it is disconnected, so no need to close it again
+      function handleDisconnect() {
+        setError(new Error('Model disconnected'));
```
mattrunyon, line 52: Just curious if we maybe want this to say `Table disconnected` which might be slightly more user friendly?

mattrunyon (commented):
Does this need to be fixed with `ChartWidgetPlugin` too? I'd guess yes since it has basically the same setup as `GridWidgetPlugin`

mattrunyon:
And possibly `PandasWidgetPlugin`, but not sure if that model emits disconnects

mofojed:
Having a hard time debugging the Pandas thing because of a Pandas Enterprise issue: https://deephaven.atlassian.net/issues/DH-17081
I'll submit a fix for that first...

mofojed:
Now that we no longer show the widget after a disconnect/error (https://github.com/deephaven/deephaven-plugins/pull/585) this PR is no longer relevant and will be closed.

mofojed:
Actually I realized there's some refactoring here (`useIrisGridModel`) which cleans some things up and could still be merged, so keeping it open.

packages/dashboard-core-plugins/src/useIrisGridModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  );
+  await waitForNextUpdate();
+  const fetchResult = result.current;
+  expect(fetchResult.status).toBe('success');
+  fetch.mockClear();
+  await act(async () => {
+    fetchResult.reload();
+  });
+  expect(fetch).toHaveBeenCalledTimes(1);
+});
```
mattrunyon, line 76: Should also add a few tests for reload lifecycle. I.e. reload goes to loading state, then success or reload goes to loading then failure

mattrunyon (changes requested):

mattrunyon (approved):

PR #2092, opened 2024-06-20
useSearchableViewportData - support non-text filters

### LINKED ISSUE #2091: useSearchableViewportData - Support searching for non-text column types
`useSearchableViewportData` can be used for components that provide search text in order to filter a windowed table viewport. It currently only supports string column types but will need to support boolean, numeric, and date column types in support of UI ComboBox.

bmingles:
resolves #2091

BREAKING CHANGE: The contract of useSearchableViewportData has changed and will need to be updated by consumers

### DIFF
--- a/packages/jsapi-utils/src/TableUtils.ts
+++ b/packages/jsapi-utils/src/TableUtils.ts
@@ -7,6 +7,7 @@ import {
 import Log from '@deephaven/log';
 import type { dh as DhType } from '@deephaven/jsapi-types';
 import {
+  assertNotNull,
   bindAllMethods,
   CancelablePromise,
   PromiseUtils,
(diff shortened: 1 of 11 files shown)


packages/jsapi-utils/src/TableUtils.ts
```diff
@@ -1825,6 +1887,11 @@ export class TableUtils {
         dh.LongWrapper.ofString(TableUtils.removeCommas(value))
       );
     }
+    if (TableUtils.isDateType(type)) {
+      return dh.FilterValue.ofNumber(
+        DateUtils.trimDateTimeStringTimeZone(value) as unknown as number
```
mofojed, line 1892: Does this actually work? I wouldn't expect a date `String` to work being passed into `ofNumber` (just looking at the code).

bmingles: It does work, but yeah that does look strange. I had based it on the call to `makeQuickDateFilterWithOperation` inside of `makeQuickDateFilter` which passes a string.

bmingles: Oh actually, I think I had actually based this on the fact that this if check wasn't there before and the fallthrough case is just passing to `dh.FilterValue.ofNumber`. The code I added was simply to trim the date string and no longer strip commas

bmingles: hmm, Looks like maybe this function doesn't actually get called on Date columns which would explain why it works

bmingles: I removed that block of code since it's never actually hit

mofojed (approved):

Issue #2066, opened 2024-06-10
DH-17242: PartitionedTable display widget does not render properly when there are non-key, non-constituent columns


rcaudy:
**Description**

When a `PartitionedTable` is created that has extra columns (not just key columns and the constituent columns), the display widget fails to render the partition filter widget or any table data. A box at the bottom of the component informs the user that it is "Reloading partition".

**Steps to reproduce**

You can reproduce with this script in a Groovy console:
```
tt = timeTable("PT00:00:01") // Make a TimeTable
it = tt.update("A=ii%2==0", "B=k%10") // Add some columns
pt = it.partitionBy("A", "B") // Partition by those columns
ptt = pt.table() // Grab the underlying table
pttu = ptt.update("C=A ? B : -B") // Derive a new column
pt2 = io.deephaven.engine.table.PartitionedTableFactory.of(pttu, List.of("A","B"), true, "__CONSTITUENT__", pt.constituentDefinition(), false) // Make a new PartitionedTable with the same key columns
```

Another way to reproduce is with a `partitionedAggBy`:
```
pt = it.partitionedAggBy([AggLast("Last=Timestamp")], false, null, "A", "B")
```

In both cases an extra column in the underlying table seems to be confusing the UI. This is an important, permitted use case for `PartitionedTables`.

**Expected results**

I expect a partition filtering widget to show as normal. Ideally, though, we'd have some better way to display this, since showing the extra columns of the partitioned table would be valuable. I'm not certain that `PartitionedTable` and `PartitionAwareSourceTable` should be displayed in the same way.

**Additional details and attachments**

DH-17242

**Versions**

 - Engine Version: latest
 - Web UI Version: ...
 - Java Version: ...
 - Barrage Version: ...
 - OS: ...
 - Browser: ...
 - Docker: ...

dsmmcken:
> Ideally, though, we'd have some better way to display this, since showing the extra columns of the partitioned table would be valuable. I'm not certain that PartitionedTable and PartitionAwareSourceTable should be displayed in the same way.

Can you elaborate on what you are thinking?

vbabich:
There's a ticket for differentiating the UI for PartitionedTable and PartitionAwareSourceTable - https://github.com/deephaven/web-client-ui/issues/2049

rcaudy:
[...]

Issue #2079, opened 2024-06-13
EPIC: Partitioned Table UI improvements


dsmmcken:
Summarizing changes to be made:

- [x] Fix table dying under the agg condition (#2066)

New changes For Partitioned Tables:
- [ ] #2103
- [ ] #2104
- [ ] #2105
- [ ] #2106
- [ ] #2107

New changes for Partitioned Aware Source Table
- [ ] #2108
- [ ] Re-label merge as "Coalesce"  (#2049)
- [ ] #2109

AkshatJawne:
Close to completing all enhancements outlined in epic, currently working through adding double click functionality through a mouse handler.

AkshatJawne:
Someone with permissions, need to reopen ticket, only have merged in Java side changes, UI changes still in review / adding functionality for empty table

AkshatJawne:
Only have merged in Java side changes, UI changes still in review / adding functionality for empty table bug

PR #2096, opened 2024-06-21
fix: DH-17242: Partition Table Pickers not partitioning table properly

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2066

This PR is solely to solve the bugs of the partition table; first that the widget was not appearing, and once appearing, the partition selectors not working properly. The `PartitionedTable` design changes and additional functionality specified in https://github.com/deephaven/web-client-ui/issues/2079 to be brought up here https://github.com/deephaven/web-client-ui/pull/2110

### DIFF
--- a/packages/iris-grid/src/IrisGridPartitionSelector.scss
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.scss
@@ -19,7 +19,7 @@
   .partition-button-group {
     display: flex;
     border-right: 1px solid var(--dh-color-hr);
-    padding: 0 $spacer-2;
+    padding-right: $spacer-2;
     gap: $spacer-2;
   }
 }
--- a/packages/iris-grid/src/IrisGridPartitionSelector.scss
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.scss
@@ -18,8 +18,7 @@
   }
   .partition-button-group {
     display: flex;
-    border: 1px var(--dh-color-hr);
-    border-style: none solid;
+    border-right: 1px solid var(--dh-color-hr);
     padding: 0 $spacer-2;
     gap: $spacer-2;
   }
--- a/packages/iris-grid/src/IrisGridPartitionSelector.tsx
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.tsx
@@ -145,18 +145,12 @@ class IrisGridPartitionSelector extends Component<
     index: number,
     selectedValue: unknown
   ): Promise<void> {
-    const { model, partitionConfig: prevConfig } = this.props;
+    const { partitionConfig: prevConfig } = this.props;
 
     log.debug('handlePartitionSelect', index, selectedValue, prevConfig);
 
     const newPartitions = [...prevConfig.partitions];
     newPartitions[index] = selectedValue;
-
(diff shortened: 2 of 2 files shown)


packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -327,7 +321,9 @@ class IrisGridPartitionSelector extends Component<
     ));
     return (
       <div className="iris-grid-partition-selector">
-        <div className="table-name">Partitioned Table</div>
+        <div className="table-name">
+          {model.isPartitionRequired ? 'Partitioned Table' : ''}
+        </div>
```
mattrunyon, line 326: This leaves a weird divider on the left side when it's not a partitioned table. Why the empty string?

It looks like the epic ticket says to just remove this label

mattrunyon (changes requested):

mattrunyon:
I think your e2e tests might be failing due to a core change. Looks like it's failing to apply date filters

AkshatJawne:
Hmm interesting, I guess I will try and pull the newest version of main?

AkshatJawne:
I do see that is due to a screenshot, but I thought that it may be due to the change in the Partition Table label that I made here

mattrunyon (changes requested):
Ok so looks like `IrisGridPartitionSelector.scss` needs some modification. There's still the divider line to the left of the buttons on the left side of this image
image

Looks like you need to change `.partition-button-group` for `border-style` to `border-right-style: solid` . And then `padding` to `padding-right: $spacer-2`

mattrunyon:
Also, if this is the PR that fixes the linked ticket (there are 2 and I think this is the real one as the other PR just includes these commits right now) then change the title to `fix: DH-17242: Partition...`. That way when we update the web package versions in enterprise, we see the Jira ticket in the changelog and know to update its status

packages/iris-grid/src/IrisGridPartitionSelector.scss
```diff
@@ -18,8 +18,7 @@
   }
   .partition-button-group {
     display: flex;
-    border: 1px var(--dh-color-hr);
-    border-style: none solid;
+    border-right: 1px solid var(--dh-color-hr);
     padding: 0 $spacer-2;
```
mattrunyon, line 22: Need to change the `padding` too. There's too much gap on the left with this padding. Should just be `padding-right: $spacer-2`

mattrunyon (changes requested):

mattrunyon (approved):

AkshatJawne:
Will hold off on merging until we can add filter prop to the Picker component, to avoid having to set the viewport on each render to 0,50 -- @bmingles mentioned that will have to wait for #2077 to merge before making these changes.

mattrunyon (changes requested):
Switching to request changes so it doesn't get accidentally merged

AkshatJawne:
Closing this PR given that we are no longer opting to manually set the viewport to 0,50 -- and we ultimately did not add a filter prop to Picker, but rather changed how the options are populated in the pickers.

Refer to https://github.com/deephaven/web-client-ui/pull/2110 for all details and updated changes

PR #2098, opened 2024-06-21
fix: Fix missing scrim background on LoadingOverlay

vbabich:
Background class wasn't applied when LoadingOverlay has error or is loading

### DIFF
--- a/packages/components/src/LoadingOverlay.tsx
+++ b/packages/components/src/LoadingOverlay.tsx
@@ -41,7 +41,7 @@ function LoadingOverlay({
           className={classNames(
             'iris-panel-message-overlay',
             'fill-parent-absolute',
-            { 'iris-panel-scrim-background': isLoaded }
+            { 'iris-panel-scrim-background': !isLoaded }
           )}
         >
           <div className="message-content">
--- a/packages/components/src/LoadingOverlay.test.tsx
+++ b/packages/components/src/LoadingOverlay.test.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import LoadingOverlay from './LoadingOverlay';
+
+describe('LoadingOverlay', () => {
+  it('renders loading spinner and scrim when isLoading is true', () => {
+    const { container } = render(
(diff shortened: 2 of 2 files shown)


mofojed (commented):
I believe not having an overlay in this case was intentional - the scrim changes the colour unnecessarily of the panel if it hasn't loaded a widget yet.
What ticket are we fixing here? This is for the error covering a dashboard?

vbabich:
> I believe not having an overlay in this case was intentional - the scrim changes the colour unnecessarily of the panel if it hasn't loaded a widget yet. What ticket are we fixing here? This is for the error covering a dashboard?

This is for DH-16180 where I need a spinner with an overlay while the panel content is updating. 
I see that we do something like this in a couple of places:
```
<LoadingOverlay isLoading={showSpinner} isLoaded={!showSpinner} ... />
```
This displays the spinner without the overlay, then on the fade-out the black overlay is shown for a split second, so there's a bit of a flicker.

I guess the right usage should be just `<LoadingOverlay isLoading={showSpinner} />` to show the spinner without the overlay, and `<LoadingOverlay isLoaded isLoading={showSpinner} />` to show the spinner with the overlay. In that case, my change isn't needed.

For my use case, Don suggested adding a background-colored overlay immediately with no fade-in, and do a quick fade-out when loaded. I'm going to update the PR to parameterize background color, fade-in, and fade-out duration.

packages/components/src/LoadingOverlay.tsx
```diff
@@ -41,7 +47,11 @@ function LoadingOverlay({
           className={classNames(
             'iris-panel-message-overlay',
             'fill-parent-absolute',
-            { 'iris-panel-scrim-background': isLoaded }
+<<<<<<< Updated upstream
+            { 'iris-panel-scrim-background': !isLoaded }
+=======
+            { [scrimClassName]: isLoaded }
+>>>>>>> Stashed changes
```
mofojed, line 54: Conflict

mofojed (changes requested):

mofojed (approved):

PR #2099, opened 2024-06-21
fix: `isElementOfType` Improved type inference

### LINKED ISSUE #2094: isElementOfType - Not passing derived `T` type to `ReactElement`
The `isElementOfType` util is not currently providing the 2nd generic arg to `ReactElement`. This results in certain cases failing to narrow types which is the whole point of the type guard.

```typescript
export function isElementOfType<T>(
  node: ReactNode,
  type: T
): node is ReactElement<InferComponentProps<T>> {
  return isValidElement(node) && node.type === type;
}
```

We should be able to fix this by constraining T a bit more and then passing it as the 2nd arg.

```typescript
export function isElementOfType<
  T extends string | JSXElementConstructor<any> =
    | string
    | JSXElementConstructor<any>,
>(node: ReactNode, type: T): node is ReactElement<InferComponentProps<T>, T> {
  return isValidElement(node) && node.type === type;
}
```

bmingles:
I missed the 2nd generic argument when I first wrote this util which made the `type` property resolve to `string | React.JSXElementConstructor<any>` in certain cases instead of the actual type. This should fix that.

resolves #2094

### DIFF
--- a/packages/react-hooks/src/ElementUtils.ts
+++ b/packages/react-hooks/src/ElementUtils.ts
@@ -1,4 +1,9 @@
-import { isValidElement, ReactElement, ReactNode } from 'react';
+import {
+  isValidElement,
+  JSXElementConstructor,
+  ReactElement,
+  ReactNode,
+} from 'react';
 import { InferComponentProps } from '@deephaven/utils';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/react-hooks/src/ElementUtils.ts
```diff
@@ -7,10 +12,13 @@ import { InferComponentProps } from '@deephaven/utils';
  * @param type The type to check against
  * @returns True if the node is a React element of the specified type
  */
-export function isElementOfType<T>(
-  node: ReactNode,
-  type: T
-): node is ReactElement<InferComponentProps<T>> {
+export function isElementOfType<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T extends string | JSXElementConstructor<any> =
```
AkshatJawne, line 17: Just curious for learning purposes, what is the reasoning for having to separately extend string here, not any other types?

bmingles: Element types are either `string` or a factory function. e.g. `'div' | 'span' | SomeComponent`. This also matches the signature of the `ReactElement` 2nd generic arg

mofojed, line 17: Could have `string | JSXElementConstructor<any>` defined as a type above to make this a little nicer.

mofojed (approved):

PR #2110, opened 2024-06-26
feat: Partitioned Table UI Enhancements

### LINKED ISSUE #1904: DH-16865: Partitioned table with no data displays a spinner indefinitely

### LINKED ISSUE #2049: UI should differentiate between Partitioned Tables and Partition-Aware Source Tables
If I have a PartitionAwareSourceTable, I get the partitioned table viewer in the UI, which is very helpful, but a bit confusing because it tells me I have a "Partitioned Table":

image

It _is_ a partitioned table in the sense that "it is a table, and it has partitions", but it's not an actual `io.deephaven.engine.table.PartitionedTable`.

It would be helpful if there were some way to see in the UI what the actual table type is (like some small text under "where it says "Partitioned Table" with the class name...or even a tooltip when you hover over "Partitioned Table", or the table's tab).

### LINKED ISSUE #2066: DH-17242: PartitionedTable display widget does not render properly when there are non-key, non-constituent columns

### LINKED ISSUE #2079: EPIC: Partitioned Table UI improvements

### LINKED ISSUE #2103: Replace the keys table, with a view of the underlying "table" (which includes additional columns)

### LINKED ISSUE #2104: Re-name the "keys" button as "Partitions", and remove the "Partitioned Table" label.

### LINKED ISSUE #2105: Make the "Partitions" table shown by default

### LINKED ISSUE #2106: Allow double clicking on said table to populate the selected keys drop downs

### LINKED ISSUE #2107: Add context menu to rows in said table "View constituent table" as last item in list performing same action

### LINKED ISSUE #2108: Default view should remain as current (selecting the last key for each drop down)

### LINKED ISSUE #2109: Inherit the above changes to "keys" table ("Partitions")

### LINKED ISSUE #2120: Modify Picker logic responsible for updating options from table
In order to remove manual setting of viewport after applying filter proposed here: https://github.com/deephaven/web-client-ui/pull/2096

AkshatJawne:
Resolves #2079  ( Resolves #2066 , Resolves #2103 , Resolves #2104 , Resolves #2105 , Resolves #2106 , Resolves #2107 , Resolves #2108 , Resolves #2109 , Resolves #2049 ), Resolves #2120, Resolves #1904

**Changes Implemented:**
- Replaced `keysTable` with `baseTable` 
- Made small UI changes to differentiate between Partition Table and Partition-aware source table
- Allow double clicking to set partition
- Add context menu action to set partition

**Sample Visuals:**
_Default Partition Table View_
Screenshot 2024-06-26 at 2 57 09 PM
_Default Partition-aware source table view:_
Screenshot 2024-06-26 at 2 56 59 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -51,5 +51,8 @@
         "description": "Test pattern or filename"
       }
     ]
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -51,8 +51,5 @@
         "description": "Test pattern or filename"
       }
     ]
-  },
-  "[typescriptreact]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
 }
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -899,8 +899,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     const changedInputFilters =
       inputFilters !== prevProps.inputFilters
         ? inputFilters.filter(
-          inputFilter => !prevProps.inputFilters.includes(inputFilter)
-        )
+            inputFilter => !prevProps.inputFilters.includes(inputFilter)
+          )
         : [];
     if (changedInputFilters.length > 0) {
       const { advancedSettings } = this.props;
@@ -1384,11 +1384,11 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       advancedFilters: ReadonlyAdvancedFilterMap,
       searchFilter: DhType.FilterCondition | undefined
     ) => [
-        ...(customFilters ?? []),
-        ...IrisGridUtils.getFiltersFromFilterMap(quickFilters),
-        ...IrisGridUtils.getFiltersFromFilterMap(advancedFilters),
-        ...(searchFilter !== undefined ? [searchFilter] : []),
-      ],
+      ...(customFilters ?? []),
+      ...IrisGridUtils.getFiltersFromFilterMap(quickFilters),
+      ...IrisGridUtils.getFiltersFromFilterMap(advancedFilters),
+      ...(searchFilter !== undefined ? [searchFilter] : []),
+    ],
     { max: 1 }
   );
 
@@ -2108,12 +2108,13 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     assertNotNull(rowIndex);
     keyTable.setViewport(rowIndex, rowIndex);
     try {
-      const data = await this.pending.add(keyTable.getViewportData());
-      // Core JSAPI returns undefined for null table values, IrisGridPartitionSelector expects null
-      // https://github.com/deephaven/deephaven-core/issues/5400
-      const values = keyTable.columns.map(
-        column => data.rows[0].get(column) ?? null
-      );
+      const data = this.getRowDataMap(rowIndex);
+      const values = Object.entries(data)
+        .filter(([key, value]) =>
+          model.partitionColumns.map(column => column.name).includes(key)
+        )
+        .map(column => column[1].value);
+
       const newPartition: PartitionConfig = {
         partitions: values,
         mode: 'partition',
@@ -3160,10 +3161,10 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         pendingRowCount = Math.max(
           0,
           bottomViewport -
-          (model.rowCount - model.pendingRowCount) -
-          model.floatingTopRowCount -
-          model.floatingBottomRowCount -
-          1
+            (model.rowCount - model.pendingRowCount) -
+            model.floatingTopRowCount -
+            model.floatingBottomRowCount -
+            1
         );
       }
     }
@@ -3494,7 +3495,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.clearAllFilters();
 
     this.startLoading(
-      `Selecting distinct values in ${columnNames.length > 0 ? columnNames.join(', ') : ''
+      `Selecting distinct values in ${
+        columnNames.length > 0 ? columnNames.join(', ') : ''
       }...`
     );
 
@@ -4302,9 +4304,9 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
 
     const debounceMs = metrics
       ? Math.min(
-        Math.max(IrisGrid.minDebounce, Math.round(metrics.rowCount / 200)),
-        IrisGrid.maxDebounce
-      )
+          Math.max(IrisGrid.minDebounce, Math.round(metrics.rowCount / 200)),
+          IrisGrid.maxDebounce
+        )
       : IrisGrid.maxDebounce;
 
     if (isFilterBarShown && focusedFilterBarColumn != null && metrics != null) {
@@ -4500,19 +4502,19 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
           const xFilterBar = gridX + columnX + columnWidth - 20;
           const style: CSSProperties = isFilterBarShown
             ? {
-              position: 'absolute',
-              top: columnHeaderHeight,
-              left: xFilterBar,
-              width: 20,
-              height: theme.filterBarHeight,
-            }
+                position: 'absolute',
+                top: columnHeaderHeight,
+                left: xFilterBar,
+                width: 20,
+                height: theme.filterBarHeight,
+              }
             : {
-              position: 'absolute',
-              top: 0,
-              left: xColumnHeader,
-              width: columnWidth,
-              height: columnHeaderHeight,
-            };
+                position: 'absolute',
+                top: 0,
+                left: xColumnHeader,
+                width: columnWidth,
+                height: columnHeaderHeight,
+              };
           const modelColumn = this.getModelColumn(columnIndex);
           if (modelColumn != null) {
             const column = model.columns[modelColumn];
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1783,15 +1783,16 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
    * Rebuilds all the current filters. Necessary if something like the time zone has changed.
    */
   rebuildFilters(): void {
-    log.debug('Rebuilding filters');
-
     const { model } = this.props;
-    const { advancedFilters, quickFilters, partitionConfig } = this.state;
+    const { advancedFilters, quickFilters } = this.state;
     const { columns, formatter } = model;
 
     if (advancedFilters.size === 0 && quickFilters.size === 0) {
+      // No need to rebuild filters if no filters are set
       return;
     }
+
+    log.debug('Rebuilding filters');
 
     const newAdvancedFilters = new Map();
     const newQuickFilters = new Map();
@@ -1818,10 +1819,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         filter: this.makeQuickFilter(column, text, formatter.timeZone),
       });
     });
-    // If partitionConfig is empty, no need to rebuild filters
-    if (partitionConfig?.mode !== 'empty' && newQuickFilters.size > 0) {
-      this.startLoading('Rebuilding filters...', { resetRanges: true });
-    }
+    this.startLoading('Rebuilding filters...', { resetRanges: true });
+
     this.setState({
       quickFilters: newQuickFilters,
       advancedFilters: newAdvancedFilters,
@@ -2032,11 +2031,15 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
 
   async loadPartitionsTable(model: PartitionedGridModel): Promise<void> {
     try {
-      const partitionConfig = await this.getInitialPartitionConfig(model);
-      this.setState(
-        { isSelectingPartition: true, partitionConfig },
-        this.loadTableState
-      );
+      const { partitionConfig } = this.state;
+      if (!partitionConfig) {
+        this.startLoading('Loading partitions...', {
+          loadingCancelShown: false,
+          resetRanges: true,
+        });
+        this.initializePartitionConfig(model);
+      }
+      this.setState({ isSelectingPartition: true }, this.loadTableState);
     } catch (error) {
       if (!PromiseUtils.isCanceled(error)) {
         this.handleTableLoadError(error);
@@ -2045,19 +2048,9 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   }
 
   /**
-   * Gets the initial partition config for the currently set model.
-   * Sorts the key table and gets the first key.
-   * If the table is ticking, it will wait for the first tick.
+   * Initialize the partition config to the default partition.
    */
-  async getInitialPartitionConfig(
-    model: PartitionedGridModel
-  ): Promise<PartitionConfig> {
-    const { partitionConfig } = this.state;
-    if (partitionConfig !== undefined) {
-      // User already has a partition selected, just use that
-      return partitionConfig;
-    }
-
+  async initializePartitionConfig(model: PartitionedGridModel): Promise<void> {
     const keyTable = await this.pending.add(
       model.partitionKeysTable(),
       resolved => resolved.close()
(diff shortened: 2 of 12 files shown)


AkshatJawne:
Looking into unit test failures

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2093,6 +2097,38 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     });
   }
 
+  async setPartitionConfig(
```
mofojed, line 2100: Name this `selectPartitionKeyFromTable`, and add JSDocs as well.
`set` implies you're passing in the config to set, whereas `select` is more accurate in what we're doing here (selecting the row).
Also don't need to pass the `model` in... `IrisGrid` already has the model.

packages/iris-grid/src/mousehandlers/IrisGridPartitionedTableMouseHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+  onDoubleClick(gridPoint: GridPoint, grid: Grid): EventHandlerResult {
+    const { column, row } = gridPoint;
+    if (
+      row == null ||
+      column == null ||
+      !isPartitionedGridModel(this.irisGrid.props.model)
+    ) {
+      return false;
+    }
+
```
mofojed, line 32: We also only want this to trigger when the `partitionConfig` is in `keys` mode; we don't want double-clicking on another table to trigger this behaviour. Get the `partitionConfig` from `irisGrid.state` and check it first, if it's not in `keys` mode then just `return false`.

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -296,6 +300,10 @@ class IrisGridTableModel
     );
   }
 
+  get isPartitionAwareSourceTable(): boolean {
+    return true;
```
mofojed, line 304: To be safe, should check `isPartitionRequired` as well:
```suggestion
    return this.isPartitionRequired && true;
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2093,6 +2097,38 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    model: PartitionedGridModel,
+    rowIndex: GridRangeIndex
+  ): Promise<void> {
+    const keyTable = await this.pending.add(
+      model.partitionKeysTable(),
+      resolved => resolved.close()
+    );
+    assertNotNull(rowIndex);
+    keyTable.setViewport(rowIndex, rowIndex);
```
mofojed, line 2109: You don't need to use the `keyTable` at all in this method, take it out. Just get the data from `this.getRowDataMap` like you're doing below, filter the values, and then set `partitionConfig`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2093,6 +2097,38 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    try {
+      const data = this.getRowDataMap(rowIndex);
+      const values = Object.entries(data)
+        .filter(([key, value]) =>
+          model.partitionColumns.map(column => column.name).includes(key)
+        )
+        .map(column => column[1].value);
```
mofojed, line 2116: We can build a set here at the start so we don't need to remap every column table for every key here. Right now this is `O(k*(c^2))` (iterates through each key, then iterates through each column remapping it, then iterates through each column again to check if it matches the key). That's way more complex than required; we don't need to be super optimized as this is only done on a double click, but we can definitely do better than that.
Just create a set of the column names before you do this, e.g.
```suggestion
      const partitionColumnSet = new Set(model.partitionColumns.map(column => column.name));
      const values = Object.entries(data)
        .filter(([key, _]) => partitionColumnSet.has(key))
        .map(([key, columnData]) => columnData.value);
```

Side note - we _could_ have the model also add a `partitionColumnSet` method, but it's not really necessary since we're not iterating over it a lot... but if the time comes then we can add it.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
AkshatJawne: Yes this does make sense, definitely a more time-efficient approach. Thanks for the suggestion.

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -212,6 +212,10 @@ class IrisGridTableModel
     return t;
   }
 
+  async partitionBaseTable(): Promise<DhType.Table> {
+    return this.valuesTable(this.columns);
```
mofojed, line 216:
```suggestion
    return this.valuesTable(this.partitionColumns);
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridPartitionSelector.tsx
```diff
@@ -322,7 +322,7 @@ class IrisGridPartitionSelector extends Component<
                 ? null
                 : (partitionConfig.partitions[index] as ItemKey)
             }
-            placeholder={'Select an option' as string}
+            placeholder={'Select a key' as string}
```
mofojed, line 325: Well this is odd, I don't know why it was like this (casting a string to a `string`). The casting is unnecessary.
```suggestion
            placeholder="Select a key"
```

packages/iris-grid/src/PartitionedGridModel.ts
```diff
@@ -43,9 +43,14 @@ export interface PartitionedGridModelProvider extends IrisGridModel {
   /** Whether a partition is required to be set on this model */
   get isPartitionRequired(): boolean;
 
+  /** Whether the table is a partition-aware source table or a partition table */
+  get isPartitionAwareSourceTable(): boolean;
+
   /** Get a keys table for the partitions */
   partitionKeysTable: () => Promise<dh.Table>;
 
+  partitionBaseTable: () => Promise<dh.Table>;
```
mofojed, line 52: Need a JS doc here. Something like:
```suggestion
  /** Get the base table for the partitions. Includes key columns and possibly other columns, such as constituent columns */
  partitionBaseTable: () => Promise<dh.Table>;
```

packages/iris-grid/src/PartitionedGridModel.ts
```diff
@@ -43,9 +43,14 @@ export interface PartitionedGridModelProvider extends IrisGridModel {
   /** Whether a partition is required to be set on this model */
   get isPartitionRequired(): boolean;
 
+  /** Whether the table is a partition-aware source table or a partition table */
+  get isPartitionAwareSourceTable(): boolean;
+
   /** Get a keys table for the partitions */
```
mofojed, line 49:
```suggestion
  /** Get a keys table for the partitions. Only includes the key columns. */
```

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -296,6 +300,10 @@ class IrisGridTableModel
     );
   }
 
+  get isPartitionAwareSourceTable(): boolean {
+    return this.isPartitionRequired && true;
```
mofojed, line 304: I'm dumb, the `&& true` is unnecessary. Just need to check `isPartitionRequired`.
```suggestion
    return this.isPartitionRequired;
```

packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -317,7 +317,7 @@ class IrisGridProxyModel extends IrisGridModel implements PartitionedGridModel {
     ) {
       if (partitionConfig.mode === 'keys') {
         modelPromise = this.originalModel
-          .partitionBaseTable()
+          .partitionKeysTable()
```
mofojed, line 320: This should still be returning the `partitionBaseTable()`. It is kind of confusing that the mode is `'keys'` and we're getting the base table here, but we want to show the user the base table when they're in `keys` mode; we use the `keysTable` for getting the options for the partition selector itself.

```suggestion
          .partitionBaseTable()
```

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2070,16 +2070,11 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
           try {
             const { detail: data } = event;
-            if (data.rows.length === 0) {
-              // Table is empty, wait for the next updated event
-              return;
+            let values: unknown[] = [];
+            if (data.rows.length > 0) {
+              const row = data.rows[0];
+              values = keyTable.columns.map(column => row.get(column));
             }
```
mofojed, line 2077: What are we changing this section for? It purposely returned when `data.rows.length === 0` before so it didn't do an update before a row had ticked in.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2060,37 +2062,62 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+            return partitionConfig;
           }
+          const row = data.rows[0];
+          const values = keyTable.columns.map(column => row.get(column));
+          const newPartition: PartitionConfig = {
+            partitions: values,
+            mode: model.isPartitionAwareSourceTable ? 'partition' : 'keys',
+          };
+          keyTable.close();
+          return newPartition;
```
mofojed, line 2085: Should be calling `setState` here now. Returning the `newPartition` from here isn't doing anything.

mofojed (changes requested):

AkshatJawne:
Looking into failures here

AkshatJawne:
Weird that the failures are only occurring on chromium, but I believe the reason of the failure is that we expect to have the pickers accessible in **1.5s**, but now, with the new logic, the pickers are disabled while the `partitionConfig.mode === 'empty'`, which in some cases (like the vid below), is not the case.

The reason we do this is so that when we have an empty partition table, we wanted to have the pickers disabled. But even for a table with rows, the table will have 0 rows of length (and thus, will be considered empty) momentarily.

We could either change the benchmark to be **20000ms** instead of **15000ms**, or we could look at adding a specific `"temporarily empty"` config mode, which would somehow differ then the `"empty"` mode (which would now act as more of a **forever empty** mode).

The latter will be difficult to implement though, since the event listener listens for when the table is updated and then changes the mode accordingly, so I am not sure how before the event, it will differentiate between a table that will be **forever empty**, and one that is **just empty until the data loads in**.

@mofojed @dsmmcken any thoughts?

[screen-capture (4).webm](https://github.com/deephaven/web-client-ui/assets/69530774/625bbfda-f617-4341-ab6a-18227a86d967)

mofojed:
We've already got the `loading` state for when it's still loading. The `empty` state is for when the table is empty, there's no need for a `forever empty` state; an empty table could eventually tick in data (it could be in one second, one minute, one hour, we don't really know).
That being said - the failures in the e2e tests are not with the partitioned tables? Where are you seeing that the disabled pickers is affecting the e2e tests? Look at the report attached: https://github.com/deephaven/web-client-ui/actions/runs/9845185336/artifacts/1682155129
Looking at the report it's looking for a context menu that does not appear for some reason.

.vscode/settings.json
```diff
@@ -51,5 +51,8 @@
         "description": "Test pattern or filename"
       }
     ]
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
mofojed, line 56: Shouldn't be a part of this change, please remove.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1812,8 +1814,10 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         filter: this.makeQuickFilter(column, text, formatter.timeZone),
       });
     });
-
-    this.startLoading('Rebuilding filters...', { resetRanges: true });
+    // If partitionConfig is empty, no need to rebuild filters
+    if (partitionConfig?.mode !== 'empty' && newQuickFilters.size > 0) {
```
mofojed, line 1818: We can likely just return early from this function if there are no filters, then nothing to rebuild. At the beginning of the function, can just check:
```
if (advancedFilters.size() === 0 && quickFilters.size() === 0) {
  return;
}
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4466,7 +4498,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
           const xFilterBar = gridX + columnX + columnWidth - 20;
           const style: CSSProperties = isFilterBarShown
             ? {
-                position: 'absolute',
+                position: 'ßabsolute',
```
mofojed, line 4501:
```suggestion
                position: 'absolute',
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4226,7 +4258,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       gotoValueSelectedFilter,
       partitionConfig,
     } = this.state;
-    if (!isReady) {
+    if (!isReady || partitionConfig?.mode === 'loading') {
```
mofojed, line 4261: I'm going to need to think about this a bit more - we should be showing the loading spinner while it's still loading.

mofojed (changes requested):
I'm going to need a bit more time to think about this one.

packages/iris-grid/src/IrisGrid.tsx
AkshatJawne: Yeah saw this, have already made change in local branch, not sure how it got there, I do not even know how to get that letter from my keyboard

.vscode/settings.json
AkshatJawne: Weird, I thought that the settings.json would be on the Git Ignore list, will remove rn.

AkshatJawne: Another thing that is non-intuitive with this that I just realized, is that when you modify the `settings.json`, it does not appear in the Source Control Tab of VSCode, which is interesting. Weird that they have an inconsistency between files changed there and in the actual PR.

AkshatJawne (commented):
@mofojed, what if we did something like this? 

From line 4365-5370 of `IrisGrid.tsx`, where we set the `show` of the loading text based on `loadingSpinnerShown or the `configMode === "loading"`

Screenshot 2024-07-12 at 8 00 09 AM

AkshatJawne:
No actually, that does not fix the issue we are facing with the spinner, ignore.

AkshatJawne:
Been doing some more investigations, the firefox and cjromium failures here are **not consistent**. On one run they fail, but then if I re-run those jobs, they pass. The inconsistent behaviour makes me think there may be a race condition or some weird code circumstance in the code responsible for rendering the context menu.

The only change I made to the Context Menu was adding the _View Constituent Table_ option, but I am unsure why that would result in the whole context menu not loading, need to dig more.

AkshatJawne:
Messed up by accidentally merging main, working on reverting changes

mofojed (changes requested):
This is looking good, but the e2e tests are failing: https://github.com/deephaven/web-client-ui/actions/runs/9977913349/job/27578476354?pr=2110
I _think_ it's because of our early return in `rebuildFilters()` if the filters aren't set, causing it to right-click in the test before the table has appeared.
We can debug in our 1:1, seems kind of tricky

AkshatJawne:
Yeah I have been trying to debug this, but it is slightly tricky, will continue to debug and then we can debug together as well in our 1:1, will make the requested change with -1 rn.

mofojed (approved):

Issue #2112, opened 2024-06-27
Make text_value prop default to key for Item component


AkshatJawne:
As per discussion and review on https://github.com/deephaven/deephaven-plugins/pull/489

PR #2113, opened 2024-06-27
fix: make textValue default to key for Normalized Item

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2112

### DIFF
--- a/packages/components/src/spectrum/utils/itemUtils.ts
+++ b/packages/components/src/spectrum/utils/itemUtils.ts
@@ -137,7 +137,10 @@ export function getItemKey<
  */
 export function getItemTextValue<T>(item: ItemElement<T>): string | undefined {
   if (item.props.textValue == null) {
-    return item.key == null ? ITEM_EMPTY_STRING_TEXT_VALUE : String(item.key);
+    const itemKey = item.key !== undefined ? String(item.key) : undefined;
+    return ['string', 'boolean', 'number'].includes(typeof item.props.children)
+      ? String(item.props.children)
+      : itemKey;
   }
 
   return item.props.textValue === ''
--- a/packages/components/src/spectrum/utils/itemUtils.ts
+++ b/packages/components/src/spectrum/utils/itemUtils.ts
@@ -137,10 +137,10 @@ export function getItemKey<
  */
 export function getItemTextValue<T>(item: ItemElement<T>): string | undefined {
   if (item.props.textValue == null) {
-    const itemKey = item.key !== undefined ? String(item.key) : undefined;
+    const itemKeyStr = item.key == null ? undefined : String(item.key);
     return ['string', 'boolean', 'number'].includes(typeof item.props.children)
       ? String(item.props.children)
-      : itemKey;
+      : itemKeyStr;
   }
 
   return item.props.textValue === ''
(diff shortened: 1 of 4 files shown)


packages/components/src/spectrum/utils/itemUtils.ts
```diff
@@ -137,9 +137,10 @@ export function getItemKey<
  */
 export function getItemTextValue<T>(item: ItemElement<T>): string | undefined {
   if (item.props.textValue == null) {
+    const itemKey = item.key !== undefined ? String(item.key) : undefined;
```
bmingles, line 140: I would use the looser == null check and check for equality instead of inequality. As a pattern, negation checks are slightly harder to read since they require an extra boolean flip in your brain

```typescript
const itemKeyStr = item.key == null ? undefined : String(item.key)`
```

packages/components/src/spectrum/utils/useRenderNormalizedItem.tsx
```diff
@@ -50,7 +50,8 @@ export function useRenderNormalizedItem({
     (normalizedItem: NormalizedItem) => {
       const itemKey = getItemKey(normalizedItem);
       const content = wrapPrimitiveWithText(normalizedItem.item?.content);
-      const textValue = normalizedItem.item?.textValue ?? '';
+      const textValue =
```
bmingles, line 53: This should also be the same check as the first one. Something like:
```typescript
const textValue =
  normalizedItem.item?.textValue ??
  (itemKey == null ? undefined : String(itemKey));
```

bmingles: Namely the `as string` may not actually be true since it could be a `boolean` or a `number`. You need the `String` to ensure it is actually a string

packages/components/src/spectrum/utils/itemUtils.ts
bmingles: Regarding your Slack question about `ITEM_EMPTY_STRING_TEXT_VALUE`, you could incorporate that in `getItemTextValue` so that it also gets applied to the case where it falls back to key, if key were `''`. 

I would leave the `useRenderNormalizedItem` as-is in regards to `ITEM_EMPTY_STRING_TEXT_VALUE`. It's slightly different in that you are dealing with a `NormalizedItem` instead of `ItemElement`, so I think you'd have to do more refactoring that probably isn't necessary

bmingles (changes requested):
Left comments

packages/components/src/spectrum/utils/useRenderNormalizedItem.test.tsx
```diff
@@ -105,7 +105,7 @@ describe.each([
     it.each([
       [
         { key: 'mock.key', textValue: undefined },
-        'Empty',
+        'mock.key',
```
bmingles, line 108: Minor test coverage regression:

Before:
image

Afer (lines 55,104 missing coverage):
image

AkshatJawne:
Ready for re-review

packages/components/src/spectrum/utils/useRenderNormalizedItem.test.tsx
bmingles: Do we have a test case for `{ key: '', textValue: undefined }` resulting in `Empty` ?

AkshatJawne: I have one for key being undefined and textvalue being undefined resulting in Empty, but not for key as an empty string, can add that as well if you would like: currently coverage is 100% though.

bmingles (approved):
Changes look good. Verify test coverage is still 100%

PR #2118, opened 2024-06-28
feat: Read file separator from server config

### LINKED ISSUE #2046: Treat files as opaque strings to support different file systems
Related to https://github.com/deephaven/deephaven-core/issues/5068 and https://github.com/deephaven/deephaven-core/pull/5532

Currently Web UI assumes file paths all use a `/` as a delimiter and start with a slash. On native Windows and in some other situations (such as S3 buckets) that may not necessarily be the case. Instead of assuming there is a `/` delimiter, we should be treating the file path as an opaque string.

mofojed:
- Uses the file separator specified in the server config 
- Fixes #2046
- Add context `FileStorageContext` for getting the `FileStorage`

BREAKING CHANGE: Implementations of `FileStorage` must now provide the `readonly separator: string` property


mofojed:
Not currently a priority.

PR #2121, opened 2024-07-03
fix: Allow ComboBox to accept the FocusableRef for ref

mofojed:
- Wasn't allowing a ref to be passed in to ComboBox


bmingles (approved):
Ah yeah. Good catch.

Guessing we probably need this for Picker as well.

PR #2122, opened 2024-07-03
fix: Ref was not being passed through for Picker

mofojed:
- Pass through ref to SpectrumPicker correctly
- Added a `useMultiRef` hook for handling multiple refs being passed in

### DIFF
--- a/packages/components/src/spectrum/picker/Picker.tsx
+++ b/packages/components/src/spectrum/picker/Picker.tsx
@@ -2,7 +2,9 @@ import {
   Picker as SpectrumPicker,
   SpectrumPickerProps,
 } from '@adobe/react-spectrum';
+import type { DOMRef } from '@react-types/shared';
 import cl from 'classnames';
+import React from 'react';
 import type { NormalizedItem } from '../utils';
 import type { PickerProps } from './PickerProps';
 import { usePickerProps } from './usePickerProps';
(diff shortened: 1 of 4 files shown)


packages/components/src/spectrum/picker/Picker.tsx
```diff
@@ -14,17 +16,18 @@ import { usePickerProps } from './usePickerProps';
...
+): JSX.Element {
   const { defaultSelectedKey, disabledKeys, selectedKey, ...pickerProps } =
-    usePickerProps<PickerProps, HTMLDivElement>(props);
+    usePickerProps(props);
 
   return (
     <SpectrumPicker
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...pickerProps}
+      ref={ref}
```
bmingles, line 30: I just realized the `ref` prop returned by `usePickerProps` needs to be incorporated here as it is used to wire up the scroll handling.

bmingles: Possible `useMappedRef` could be incorporated here to map to the forwardRef and the scrollRef

bmingles (changes requested):
Left a comment regarding the scroll ref

packages/components/src/spectrum/picker/Picker.tsx
bmingles: This is going to be a little annoying, but looks like the ref originates in `usePopoverOnScrollRef` and will need to have it's type signature changed from `useRef<T>(null)` to `useRef<T | null>(null);` to make it mutable.

mofojed:
I tested this with DHE, looks to be working correctly.

packages/components/src/spectrum/picker/useMultiRef.ts
```diff
@@ -0,0 +1,22 @@
+import { MutableRefObject, Ref, RefCallback, useCallback } from 'react';
+
+/**
+ * Takes in multiple refs and then returns one ref that can be assigned to the component.
+ * In turn all the refs passed in will be assigned when the ref returned is assigned.
+ * @param refs The refs to assign
+ */
+function useMultiRef<T>(...refs: readonly Ref<T>[]): RefCallback<T> {
```
bmingles, line 8: Very nice.

packages/components/src/spectrum/picker/useMultiRef.test.ts
```diff
@@ -0,0 +1,29 @@
+import { renderHook } from '@testing-library/react-hooks';
+import useMultiRef from './useMultiRef';
+
+describe('useMultiRef', () => {
+  it('should assign the ref to all refs passed in', () => {
```
bmingles, line 5: Should we add a test case for non function refs?

bmingles (dismissed):
Looks good. Left question about additional test case but I wouldn't hold up PR for it.

bmingles (approved):
LGTM

Issue #885, opened 2022-11-15
Option to increase information density


mofojed:
As a power user, I want increase the information density to allow for more information to appear on the screen at once. Things like reducing column/row padding in grids, decreasing font size, decreasing padding in plots, decreasing tab padding/size, etc.

dsmmcken:
We default slightly higher in font-size and in cell y-padding, and marginally higher in x-padding. A higher density option that lowers font-size and padding would increase density, we should also offer the reverse.

![image](https://user-images.githubusercontent.com/1576283/201960015-37e7bac3-be44-4dfc-909b-bb104f3bbc76.png)

dsmmcken:
In terms of just the table:
![image](https://user-images.githubusercontent.com/1576283/201962643-499e7c67-d3f6-4935-9d21-c984c1a3d083.png)

1px taller text, and overall cell height is 3px taller per cell. 

Horizontally it's mixed, some cells are wider by default some are narrower, but padding is still larger. Overall though we come out slightly wider.
![image](https://user-images.githubusercontent.com/1576283/201964060-55edabba-164f-42c8-9bd6-11ed9c59b52c.png)

dsmmcken:
In practice, on a fresh console running a single command, we are showing 6 rows less than swing. We lose a bit of space from browser chrome, but also we are proportionally allocating more space to the console (we have less ui-chrome in the console panel). So we could reduce it's default height and still be comparable. Currently it's 50/50 split, but maybe 40/60 or something would be a better default. Our tabs are slightly taller, and so are out header columns, but I think there's value in that in terms of click target size and would prefer to squeeze elsewhere.

![ui-chrome-comparison](https://user-images.githubusercontent.com/1576283/201971597-bbca0c4d-2233-4e87-bfec-494bc2947d3f.jpg)

rbasralian:
Just adding another example to this:

![image](https://user-images.githubusercontent.com/9516585/232853952-d4b172a6-39ec-47b5-919e-62fc65ae9e1e.png)

![image](https://user-images.githubusercontent.com/9516585/232854011-83fb9964-abcf-4553-a4fb-09965a55ed25.png)

These two screenshots were taken w/ the boundaries between the charts/tables exactly lined up, more or less to the pixel.

Here is the second screenshot again, with font size increased to 12:

![image](https://user-images.githubusercontent.com/9516585/232854513-2e1e94e9-dc3a-4491-bc26-f8b33d80193b.png)


[...]

PR #2123, opened 2024-07-03
feat: IrisGridTheme iconSize

mattrunyon:
This is part of the work needed for #885 that I felt would be better as a separate PR.

Snapshot changes are going to be slightly shifted sort icon and vertical center + slight left shift on tree table arrows.


tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
mofojed, line 1: The reason we do those `0.5` half pixels is so that the lines are crisp, not blurry. In the Deleted snapshot, you can see the vertical line for the expanding tree is crisp, whereas on the Added snapshot it's blurry/anti-aliasing is applied. The icons look okay though, seems to just be the line that is off.

For some details on that: https://stackoverflow.com/questions/9311428/draw-single-pixel-line-in-html5-canvas

packages/grid/src/GridRenderer.ts
```diff
@@ -1085,8 +1085,7 @@ export class GridRenderer {
           columnX +
           i * treeDepthIndent +
           treeDepthIndent * 0.5 +
-          treeHorizontalPadding +
-          0.5;
+          treeHorizontalPadding;
```
mofojed, line 1088: This `0.5` is here intentionally, for crisp lines. When drawing a 1 pixel line, you need to adjust it by half a pixel. https://stackoverflow.com/questions/9311428/draw-single-pixel-line-in-html5-canvas

packages/grid/src/GridRenderer.ts
```diff
@@ -1101,8 +1100,7 @@ export class GridRenderer {
...
```
mofojed, line 1103: Don't remove the 0.5

packages/iris-grid/src/IrisGridCellRendererUtils.ts
```diff
@@ -13,12 +13,15 @@ class IrisGridCellRendererUtils {
...
     const markerIcon = isExpanded
-      ? getIcon('caretDown')
-      : getIcon('caretRight');
-    const iconX = columnX + x1 - 2;
-    const iconY = rowY + y1 + 2.5;
+      ? getIcon('caretDown', iconSize)
+      : getIcon('caretRight', iconSize);
+
+    const iconX = columnX + (x1 + x2) / 2 - iconSize / 2; // Midpoint of the tree box minus half the icon width
+    const iconY = rowY + (y2 - y1 - iconSize) / 2; // y2 - y1 is effecitvely rowHeight
```
mofojed, line 24:
```suggestion
    const iconY = rowY + (y2 - y1 - iconSize) / 2; // y2 - y1 is effectively rowHeight
```

packages/iris-grid/src/IrisGridIcons.ts
```diff
@@ -18,15 +15,15 @@ export type IconName =
   | 'cellOverflow';
 
 const iconMap = new Map<IconName, IconDefinition>([
-  ['sortUp', dhSortUp],
-  ['sortDown', dhSortDown],
+  ['sortUp', vsTriangleUp],
```
mofojed, line 18: Why'd you switch these out? Pretty sure @dsmmcken made the `dhSortUp` and `dhSortDown` icons specifically for the sorting case (and correct sizing).

mofojed (changes requested):

tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
mattrunyon: Right so we might want to do something like floor and then add 0.5 after. If you had an odd size tree indent then the 0.5 would make the final value an int. Or if there were row headers and the `columnX` value was a non-int (right now it's always 0), then you would end up off of a `x.5` value

packages/iris-grid/src/IrisGridIcons.ts
mattrunyon: I talked w/ Don about this. He said those icons were probably just to stay consistent w/ the fontawesome icons.

`dhSortUp` and `dhSortDown` are not centered vertically. They are the top/bottom half of the double sort arrows icon. I was trying to center them based on the icon size, but the actual centering was not something clean like shift by `0.25*iconSize`. It had a +/- 1 needed depending on which arrow and I couldn't figure out why or what would make that consistent regardless of icon size.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1498,7 +1498,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     return rowIndex != null ? modelRows.get(rowIndex) : null;
   }
 
-  getTheme(): Partial<IrisGridThemeType> {
+  getTheme(): IrisGridThemeType {
```
dsmmcken, line 1501: I see the satisfies on our theme, but does this still work if a third-party provided theme is a partial?

Like we want third parties to be able to declare themes that override just one parameter for example.

mattrunyon: Yes. This return type was just incorrect. It returns `getCachedTheme` which spreads the default theme and returns a full theme. The return type needed to be fixed for the cell overflow context menu handler

mofojed (approved):

PR #2125, opened 2024-07-04
WIP debugging issues with filter not being applied

### LINKED ISSUE #2120: Modify Picker logic responsible for updating options from table

AkshatJawne:
Resolves #2120

### DIFF
--- a/packages/iris-grid/src/IrisGridPartitionSelector.tsx
+++ b/packages/iris-grid/src/IrisGridPartitionSelector.tsx
@@ -6,7 +6,7 @@ import { vsChevronRight, vsMerge, vsKey } from '@deephaven/icons';
 import Log from '@deephaven/log';
 import { Picker } from '@deephaven/jsapi-components';
 import type { dh } from '@deephaven/jsapi-types';
-import { TableUtils } from '@deephaven/jsapi-utils';
+import { FilterConditionFactory, TableUtils } from '@deephaven/jsapi-utils';
 import { assertNotNull, Pending, PromiseUtils } from '@deephaven/utils';
 import './IrisGridPartitionSelector.scss';
 import { PartitionConfig, PartitionedGridModel } from './PartitionedGridModel';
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/IrisGridPartitionSelector.tsx
mofojed, line 1: Okay, so to sum up the problem a bit:
- Lets say we have three partition columns `A`,`B`, which results in the following table with four keys, E.g.

| A | B |
|---|---|
| 0 | 0 |
| 0 | 1 |
| 1 | 0 |
| 1 | 2 |

  - If you haven't selected anything yet, the options for `A` should be `0, 1`, and `B` should be `0, 1, 2` (size 0 is duplicated)
  - If you've selected `A=1`, then the options for `B` should just be `0, 2`
- We do a `selectDistinct` with a partition column, so that we only get the unique options for that partition instead of duplicating items (like `0` in `B`)
  - After you do a `.selectDistinct` to get the options for a partition column, we won't be able to filter it after that based on other partition columns, since we've only got that column in the outputted table
- We should call `applyFilter` _before_ we call `selectDistinct`. 

So I think this needs to be re-factored a bit. Right now we're building the `partitionTables` in `componentDidMount` only, but we should also be updating them after `handlePartitionSelect`. Roughly what we should be doing...
- Change `updatePartitionFilters` to `updatePartitionOptions`, and call it from `componentDidMount` instead of initializing `partitionTables` and `partitionFilters` in there (though we could still initialize the `keysTable`, and keep that around)
- In `updatePartitionFilters`, actually build the `partitionTables` by applying a filter to the `keysTable` first, then doing a `.selectDistinct` (I think you can just do `applyFilter` and then call `.selectDistinct` to get a new table)
- Just pass those partitionTables down, no need other filters to be passed down (since it's already passed in).

Let me know if that works.

mofojed (changes requested):

packages/iris-grid/src/IrisGridPartitionSelector.tsx
AkshatJawne: Changes implemented in latest push to https://github.com/deephaven/web-client-ui/pull/2110

AkshatJawne:
Can close this PR, implemented changes in Improvements PR linked in comment, so that could test with all other code changes

PR #2126, opened 2024-07-04
build: Fix the jsapi-types dependency to be more permissive

mofojed:
- If it's locked at a specific version, it becomes a problem when trying to install with other packages that are locked to a specific version (e.g. on Enterprise) and result in multiple `jsapi-types` packages being installed, where the types may not match and you may get errors


bmingles (approved):
Seems like a good change.

PR #2127, opened 2024-07-04
build: Fix the dependency for jsapi-types

mofojed:
- We only need `v0.33.4` of core for v0.78 web packages
- Having the dependency specific can cause issues with multiple versions of the jsapi-types package being installed, causing mismatches


bmingles (approved):
LGTM

Issue #2084, opened 2024-06-18
re-export @react-types/shared types needed by dh ui


bmingles:
As part of removing Spectrum dependencies from the plugins repo, we should re-export any types we need from `@react-types/shared`. 

Types currently used by plugins are:
* Orientation
* PressEvent
* SelectionMode

PR #2132, opened 2024-07-08
fix: re-export remaining types needed by dh ui from @react-types/shared

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2084

**Steps Moving Forward:**
- Once this PR has merged, will go into plugins repo and use the re-exported types, and remove @react-types/shared dependency (in order to further eliminate spectrum dependency)


bmingles (approved):
LGTM

PR #2134, opened 2024-07-08
feat: Add option to disable WebGL rendering

mofojed:
- Advanced Settings section in the Settings menu, with an option to disable WebGL
- Read from the server config for default values
  - web.webgl: default setting for WebGL being enabled or not
  - web.webgl.editable: disallow the user from enabling WebGL
- Contextual Help item to add more info
- Tested by adding the following to the deephaven.prop config:
```
web.webgl=false
web.webgl.editable=false
client.configuration.list=java.version,deephaven.version,barrage.version,http.session.durationMs,file.separator,web.storage.layout.directory,web.storage.notebook.directory,web.webgl,web.webgl.editable
```
Screenshot of when it's enabled/editable:
image

Screenshot of when it's disabled/editable (@dsmmcken I think we may need to change `Switch` styling up here, or the Settings menu background):
image

Screenshot of when it's disabled and not editable on the server:
image

### DIFF
--- a/packages/chart/src/ChartUtils.ts
+++ b/packages/chart/src/ChartUtils.ts
@@ -681,17 +681,22 @@ class ChartUtils {
    * Converts the Iris plot style into a plotly chart type
    * @param plotStyle The plotStyle to use, see dh.plot.SeriesPlotStyle
    * @param isBusinessTime If the plot is using business time for an axis
+   * @param allowWebGL If WebGL is allowedd
    */
   getPlotlyChartType(
     plotStyle: DhType.plot.SeriesPlotStyle,
-    isBusinessTime: boolean
+    isBusinessTime: boolean,
+    allowWebGL: boolean
   ): PlotType | undefined {
     const { dh } = this;
     switch (plotStyle) {
       case dh.plot.SeriesPlotStyle.SCATTER:
       case dh.plot.SeriesPlotStyle.LINE:
-        // scattergl mode is more performant, but doesn't support the rangebreaks we need for businessTime calendars
-        return !isBusinessTime && IS_WEBGL_SUPPORTED ? 'scattergl' : 'scatter';
+        // scattergl mode is more performant (usually), but doesn't support the rangebreaks we need for businessTime calendars
+        // In some cases, WebGL is less performant (like in virtual desktop environments), so we also allow the option of the user explicitly disabling it even if it's supported
+        return !isBusinessTime && IS_WEBGL_SUPPORTED && allowWebGL
+          ? 'scattergl'
+          : 'scatter';
       case dh.plot.SeriesPlotStyle.BAR:
       case dh.plot.SeriesPlotStyle.STACKED_BAR:
         return 'bar';
@@ -871,7 +876,8 @@ class ChartUtils {
     series: DhType.plot.Series,
     axisTypeMap: AxisTypeMap,
     seriesVisibility: boolean | 'legendonly',
-    showLegend: boolean | null = null
+    showLegend: boolean | null = null,
+    allowWebGL = true
   ): Partial<PlotData> {
     const {
       name,
@@ -888,7 +894,7 @@ class ChartUtils {
     const isBusinessTime = sources.some(
       source => source.axis?.businessCalendar
     );
-    const type = this.getChartType(plotStyle, isBusinessTime);
+    const type = this.getChartType(plotStyle, isBusinessTime, allowWebGL);
     const mode = this.getPlotlyChartMode(
       plotStyle,
       isLinesVisible ?? undefined,
@@ -1019,7 +1025,8 @@ class ChartUtils {
 
   getChartType(
     plotStyle: DhType.plot.SeriesPlotStyle,
-    isBusinessTime: boolean
+    isBusinessTime: boolean,
+    allowWebGL: boolean
   ): PlotType | undefined {
     const { dh } = this;
     switch (plotStyle) {
@@ -1028,7 +1035,7 @@ class ChartUtils {
         // plot.ly to calculate the bins and sum values, just convert it to a bar chart
         return 'bar';
       default:
-        return this.getPlotlyChartType(plotStyle, isBusinessTime);
+        return this.getPlotlyChartType(plotStyle, isBusinessTime, allowWebGL);
     }
   }
 
--- a/packages/chart/src/ChartUtils.ts
+++ b/packages/chart/src/ChartUtils.ts
@@ -686,7 +686,7 @@ class ChartUtils {
   getPlotlyChartType(
     plotStyle: DhType.plot.SeriesPlotStyle,
     isBusinessTime: boolean,
-    allowWebGL: boolean
+    allowWebGL = true
   ): PlotType | undefined {
     const { dh } = this;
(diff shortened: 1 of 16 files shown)


dsmmcken:
theme-spectrum-overrides.css
```
span[class*='spectrum-ToggleSwitch-switch'] {
  /* increase contrast of switch in off position */
  background: var(--dh-color-gray-500);
}
```

dsmmcken (changes requested):
theme-spectrum-overrides.css
```css
span[class*='spectrum-ToggleSwitch-switch'] {
  /* increase contrast of switch in off position */
  background: var(--dh-color-gray-500);
}
```

dsmmcken:
Is there a follow on ticket for deephaven-express to check this setting?

mofojed:
@dsmmcken followup ticket: https://github.com/deephaven/deephaven-plugins/issues/612

dsmmcken (approved):

packages/code-studio/src/settings/AdvancedSectionContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {
+  getWebGL,
+  getWebGLEditable,
+  RootState,
+  updateSettings,
+} from '@deephaven/redux';
+
+function AdvancedSectionContent(): JSX.Element {
+  const dispatch = useDispatch();
+  const webgl = useSelector<RootState>(getWebGL) as boolean;
```
bmingles, line 19: Not sure why these don't infer the type, but I think `useSelector<RootState, boolean>` will give us better type safety. Same for next one.

packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
```diff
@@ -111,3 +111,8 @@ label[class*='spectrum-Radio'] {
     --dh-color-negative-key-focus-bg
   );
 }
+
+span[class*='spectrum-ToggleSwitch-switch'] {
+  /* increase contrast of switch in off position */
+  background: var(--dh-color-gray-500);
```
bmingles, line 117: Should we introduce a `--dh-toggle-switch-bg: var(--dh-color-gray-500)` to make this overridable?

packages/redux/src/store.ts
```diff
@@ -57,6 +57,8 @@ export interface WorkspaceSettings {
   defaultNotebookSettings: {
     isMinimapEnabled?: boolean;
   };
+  webgl: boolean;
```
bmingles, line 60: Will these always be defined?

packages/chart/src/ChartUtils.ts
```diff
@@ -681,17 +681,22 @@ class ChartUtils {
    * Converts the Iris plot style into a plotly chart type
    * @param plotStyle The plotStyle to use, see dh.plot.SeriesPlotStyle
    * @param isBusinessTime If the plot is using business time for an axis
+   * @param allowWebGL If WebGL is allowedd
    */
   getPlotlyChartType(
     plotStyle: DhType.plot.SeriesPlotStyle,
-    isBusinessTime: boolean
+    isBusinessTime: boolean,
+    allowWebGL: boolean
```
bmingles, line 689: Should this be mentioned as a BREAKING CHANGE?

packages/chart/src/ChartUtils.ts
```diff
@@ -1019,7 +1025,8 @@ class ChartUtils {
 
   getChartType(
     plotStyle: DhType.plot.SeriesPlotStyle,
-    isBusinessTime: boolean
+    isBusinessTime: boolean,
+    allowWebGL: boolean
```
bmingles, line 1029: Should this be mentioned as a BREAKING CHANGE?

bmingles (changes requested):
Left a couple of suggestions / questions

packages/code-studio/src/settings/AdvancedSectionContent.tsx
bmingles: Actually, looks like this may be a better option ([documented here](https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks)):

```typescript
export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;

...

const webgl = useAppSelector(getWebGL)
```

mofojed: Ohhh nice, much better.

packages/chart/src/ChartUtils.ts
mofojed: I'll just default it to `true` to avoid making it a breaking change. I wanted to make sure I caught all the places where it was called from though.

packages/redux/src/store.ts
mofojed: Yes. See in the `getSettings` selector, we start from all the default settings, and override the ones the user has set. The users settings are a partial: https://github.com/deephaven/web-client-ui/blob/232fb4db121cd507d0303dba6b468ae1edab9135/packages/redux/src/selectors.ts#L71

packages/components/src/theme/theme-dark/theme-dark-semantic.css
```diff
@@ -199,4 +199,7 @@
 
   /* Editing */
   --dh-color-modified: var(--dh-color-yellow-1300);
+
+  /* Components */
+  --dh-toggle-switch-bg: var(--dh-color-gray-500);
```
bmingles, line 204: We actually have `theme-xxx-components.css` for component specific variables. Alternatively, could also just define it as a local var inline
```typescript
span[class*='spectrum-ToggleSwitch-switch'] {
  --dh-toggle-switch-bg: var(--dh-color-gray-500);
  
  /* increase contrast of switch in off position */
  background: var(--dh-toggle-switch-bg);
}
```

bmingles (changes requested):
1 issue with location of the css variable, but otherwise looks good.

packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
dsmmcken: yes, and actually should double check it in light mode, as I didn't.

Should be added to theme-dark-components.css and theme-light-components.css with appropriate values, that use that variable here.

packages/components/src/theme/theme-dark/theme-dark-semantic.css
dsmmcken: yeah should go in the components file, both light and dark. And then check that it looks good in light mode still because I didn't and adjust it if necessary.

bmingles (dismissed):
LGTM

dsmmcken (approved):

packages/components/src/theme/theme-light/theme-light-components.css
```diff
@@ -162,4 +162,7 @@
     var(--dh-color-visual-gray) 5%,
     transparent
   );
+
+  /* Switch */
+  --dh-toggle-switch-bg: var(--dh-color-gray-500);
```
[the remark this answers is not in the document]
dsmmcken: doesn't need to be that dark on light.

bmingles (approved):
LGTM

PR #2138, opened 2024-07-09
refactor: Improve GL item config types

### LINKED ISSUE #2130: Improve types for Golden Layout `createContentItem`
The current types for Golden Layout `LayoutManager.createContentItem` are not aware of additional properties used for RowOrColumn types. This is currently causing ts errors in the plugins repo.

bmingles:
- Broke out explicit `ItemConfig` sub types and converted `ItemConfig` type into a union them all
- Updated `createContentItem` to be more explicit in its mapping of config types to returned content items 
- Updated references to `ItemConfigType` with `ItemConfig`
- Updated `Component`, `RowOrColumn`, and `Stack` classes to allow null `parent`

resolves #2130

### DIFF
--- a/packages/dashboard/src/layout/LayoutUtils.ts
+++ b/packages/dashboard/src/layout/LayoutUtils.ts
@@ -244,7 +244,7 @@ class LayoutUtils {
    */
   static getStackForRoot(
     root: ContentItem,
-    config: Partial<ItemConfig> & { component?: string },
+    config: Partial<ItemConfig>,
     createIfNecessary = true,
     matchComponentType = true,
     allowEmptyStack = true
@@ -253,7 +253,7 @@ class LayoutUtils {
     if (!stack && matchComponentType) {
       stack = this.getStackForConfig(
         root,
-        { component: config.component },
+        { component: (config as { component?: string }).component },
         allowEmptyStack
       );
     }
--- a/packages/dashboard/src/layout/LayoutUtils.ts
+++ b/packages/dashboard/src/layout/LayoutUtils.ts
@@ -105,11 +105,11 @@ class LayoutUtils {
    * @returns The newly created stack.
    */
   static addStack(parent: ContentItem, columnPreferred = true): Stack {
-    const rowOrColConfig: ItemConfig = {
-      type: columnPreferred ? 'column' : 'row',
-    };
-
     if (isRoot(parent)) {
+      const rowOrColConfig: ItemConfig = {
+        type: columnPreferred ? 'column' : 'row',
+      };
+
       if (parent.contentItems == null || parent.contentItems.length === 0) {
         parent.addChild(rowOrColConfig);
       }
--- a/packages/dashboard/src/layout/LayoutUtils.ts
+++ b/packages/dashboard/src/layout/LayoutUtils.ts
@@ -16,7 +16,6 @@ import type {
   Container,
   ContentItem,
   ItemConfig,
-  ItemConfigType,
   ReactComponentConfig,
   Stack,
   Tab,
@@ -38,12 +37,12 @@ export type StackItemConfig = ItemConfig & {
   activeItemIndex?: number;
 };
 
-function isComponentConfig(config: ItemConfigType): config is ComponentConfig {
+function isComponentConfig(config: ItemConfig): config is ComponentConfig {
   return (config as ComponentConfig).componentName !== undefined;
 }
 
 export function isReactComponentConfig(
-  config: ItemConfigType
+  config: ItemConfig
 ): config is ReactComponentConfig {
   const reactConfig = config as ReactComponentConfig;
   // Golden layout sets the type to 'component' and componentName to 'lm-react-component' in `createContentItem`, then changes it back in `toConfig`
@@ -60,13 +59,13 @@ function isHTMLElement(element: Element): element is HTMLElement {
   return (element as HTMLElement).focus !== undefined;
 }
 
-function isStackItemConfig(config: ItemConfigType): config is StackItemConfig {
+function isStackItemConfig(config: ItemConfig): config is StackItemConfig {
   return config.type === 'stack';
 }
 class LayoutUtils {
   static DEFAULT_FOCUS_SELECTOR = 'input, select, textarea, button';
 
-  static activateTab(root: ContentItem, config: Partial<ItemConfigType>): void {
+  static activateTab(root: ContentItem, config: Partial<ItemConfig>): void {
     const stack = LayoutUtils.getStackForRoot(root, config, false);
     if (!stack) {
       log.error('Could not find stack for config', config);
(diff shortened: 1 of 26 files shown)


packages/golden-layout/src/LayoutManager.ts
```diff
@@ -890,6 +931,7 @@ export class LayoutManager extends EventEmitter {
     }
 
     if ($.isPlainObject(contentItemOrConfig) && contentItemOrConfig.type) {
+      assertNotNull(parent);
```
bmingles, line 934: All of the content item constructors called in `createContentItem` require `parent` based on their types, so seemed like a good option to assert this here to allow making the type required in `createContentItem` signature.

bmingles: Well looks like this actually fails e2e tests. Going to put back in draft

packages/golden-layout/src/items/Component.ts
```diff
@@ -66,7 +66,7 @@ export default class Component extends AbstractContentItem {
   }
 
   close() {
-    this.parent.removeChild(this);
+    this.parent?.removeChild(this);
```
bmingles, line 69: Trying to wrap my head around how this didn't cause problems before, but e2e tests would suggest that `Component` item types can be instantiated without a parent.

packages/golden-layout/src/LayoutManager.ts
```diff
@@ -536,8 +566,36 @@ export class LayoutManager extends EventEmitter {
...
+
+    if (config.type === 'column') {
+      if (parent == null) {
+        throw new Error('Columns require a parent');
+      }
+
+      return new RowOrColumn(true, this, config, parent);
+    }
+
+    return new Component(this, config, parent);
```
bmingles, line 598: @mofojed e2e tests fail for `react-component` type if I require `parent` to be defined, so I tweaked the `Component` config type internally. The others (RowOrColumn and Stack) seem to require a parent as far as I can tell. I'm guessing a bit here, so if you have any more concrete insight into which cases would always require a parent vs not, that would be helpful

bmingles: Unit tests don't seem to like requiring parents for rows. It's possible that all of these items should have parent marked as optional and need some runtime refactoring internally, but I'm not sure

bmingles: I ended up updating the content items to allow `null` parent instead since it seems that is what can happen at runtime.

packages/golden-layout/src/items/Stack.ts
```diff
@@ -31,47 +35,33 @@ type ContentAreaDimensions = {
...
+        header?: StackItemHeaderConfig;
       };
     },
-    config: ComponentConfig & {
-      header?: StackHeaderConfig;
-      hasHeaders?: boolean;
-    },
-    parent: RowOrColumn
+    config: StackItemConfig,
+    parent: AbstractContentItem | null
```
bmingles, line 64: I inferred that this needs to be `AbstractContentItem` instead of `RowOrColumn` based on the fact that `Stack` calls `this.layoutManager.createContentItem` passing itself as the parent. We seem to do some checking later to identify if it is a valid row or column item and if not creating one, so seemed ok to broaden the type here.

packages/golden-layout/src/items/Stack.ts
```diff
@@ -276,8 +266,9 @@ export default class Stack extends AbstractContentItem {
     const insertBefore =
       this._dropSegment === 'top' || this._dropSegment === 'left';
     const hasCorrectParent =
-      (isVertical && this.parent.isColumn) ||
-      (isHorizontal && this.parent.isRow);
+      this.parent instanceof RowOrColumn &&
```
bmingles, line 269: I am making a slight assumption adding the `instanceof RowOrColumn` check here, since `AbstractContentItem` has an optional `isColumn` and `isRow` prop. Technically it wouldn't have to be said instance to have passed this check in the past. The other piece is the call to `addChild` on line 200 which does assume the 3rd argument in `addChild` which is only present in `RowOrColumn`

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -106,17 +105,20 @@ class LayoutUtils {
    * @returns The newly created stack.
    */
   static addStack(parent: ContentItem, columnPreferred = true): Stack {
-    const type = columnPreferred ? 'column' : 'row';
+    const rowOrColConfig: ItemConfig = {
+      type: columnPreferred ? 'column' : 'row',
+    };
+
```
mofojed, line 111: Move this to within the `if (isRoot(parent))` block below, which is the only place `rowOrColConfig` is used.

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -248,7 +253,7 @@ class LayoutUtils {
     if (!stack && matchComponentType) {
       stack = this.getStackForConfig(
         root,
-        { component: config.component },
+        { component: (config as { component?: string }).component },
```
mofojed, line 256: This is kind of strange, but seems messier to do it a different way... Could create a type predicate just for this, but seems kind of messy.

packages/golden-layout/src/LayoutManager.ts
```diff
@@ -44,6 +50,14 @@ export type ComponentConstructor<
   new (container: ItemContainer<C>, state: unknown): unknown;
 };
 
+type ItemConfigSupported =
+  | ColumnItemConfig
+  | RowItemConfig
+  | StackItemConfig
+  | ComponentConfig;
+
+const SUPPORTED_ITEM_TYPES = ['column', 'row', 'stack', 'component'] as const;
```
mofojed, line 59: Docstrings for why these are the only supported types? Why isn't `ReactComponentConfig` supported for example? What is this "supported" by? We use `ReactComponentConfig` in a lot of places elsewhere... Maybe a better name would be something like `LayoutItemConfig` ?

packages/golden-layout/src/items/Stack.ts
mofojed: Yea in the examples they do show nested stacks: https://golden-layout.com/examples/#b40c19e79915493e9d3ae0220773358f
So a Stack could be in a Stack. That being said, that example above seems to have bugs as soon as you start moving things around. But I digress, seems fine for these changes here.

packages/golden-layout/src/config/ItemConfig.ts
```diff
@@ -1,27 +1,17 @@
-import type { StackHeaderConfig } from '../items/Stack';
-
-export type ItemConfigType =
-  | ItemConfig
+export type ItemConfig =
```
mofojed, line 1: Add an `export type ItemConfigType = ItemConfig` as well; otherwise this is a breaking change (since the `ItemConfigType` is disappearing).

mofojed (changes requested):

packages/golden-layout/src/LayoutManager.ts
bmingles: @bender I don't actually know the answer as to why we only support these. This was just based on the constructors that were in the previous code. Any ideas?

bmingles: I guess the `ReactComponentConfig` isn't listed due to it being converted out where this is used. I can improve that in the names / docs here to be more explicit, but not sure of any other reasons why other types aren't supported.

packages/dashboard/src/layout/LayoutUtils.ts
bmingles: fixed

packages/dashboard/src/layout/LayoutUtils.ts
bmingles: Yeah, there's a limitation of TypeScript for scenarios where a type is changed as a result of setting something.

packages/golden-layout/src/LayoutManager.ts
bmingles: Updated based on what I know. Not sure why DefaultItemConfig and RootItemConfig aren't included, but should at least clarify how things are used.

packages/golden-layout/src/config/ItemConfig.ts
bmingles: done

mofojed (approved):

PR #2144, opened 2024-07-15
feat: Add functions for PanelEvent.OPEN

mofojed:
- Instead of using PanelEvent.OPEN directly and not getting any type safety, add functions for emitting, listening, and a hook to enforce type safety
- Add xcomponents to the styleguide

### DIFF
--- a/packages/dashboard/src/PanelEvent.ts
+++ b/packages/dashboard/src/PanelEvent.ts
@@ -7,10 +7,13 @@ export type WidgetDescriptor = {
 };
 
 export type PanelOpenEventDetail<T = unknown> = {
-  /** Opening the widget was triggered by dragging from a list. The coordinates are used as the starting location for the drag.  */
+  /**
+   * Opening the widget was triggered by dragging from a list, such as the Panels dropdown.
+   * The coordinates are used as the starting location for the drag, where we will show the panel until the user drops it in the dashboard.
+   */
   dragEvent?: MouseEvent;
 
-  /** ID of this panel to use */
+  /** ID of the panel to re-use. Will replace any existing panel with this ID. Otherwise a new panel is opened with a randomly generated ID. */
   panelId?: string;
 
   /** Descriptor of the widget. */
@@ -75,6 +78,6 @@ export const {
   useListener: usePanelOpenListener,
 } = makeEventFunctions<PanelOpenEventDetail>(PanelEvent.OPEN);
 
-// TODO: Add the rest of the event functions here. Need to create the correct types for all of them.
+// TODO (#2147): Add the rest of the event functions here. Need to create the correct types for all of them.
 
 export default PanelEvent;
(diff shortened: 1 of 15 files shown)


packages/golden-layout/src/utils/EventUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    jest.clearAllMocks();
+  });
+
+  it('listenForEvent', () => {
+    const event = 'test';
+    const handler = jest.fn();
+    const remove = listenForEvent(eventEmitter, event, handler);
+    expect(eventEmitter.on).toHaveBeenCalledWith(event, handler);
+    remove();
+    expect(eventEmitter.off).toHaveBeenCalledWith(event, handler);
```
mattrunyon, line 25: Check off isn't called before the `remove` call just to verify

packages/golden-layout/src/utils/EventUtils.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(eventEmitter.on).toHaveBeenCalledWith(event, handler);
+  });
+
+  it('makeEmitFunction', () => {
+    const event = 'test';
+    const emit = makeEmitFunction(event);
+    const payload = { test: 'test' };
+    emit(eventEmitter, payload);
+    expect(eventEmitter.emit).toHaveBeenCalledWith(event, payload);
+  });
```
mattrunyon, line 42: Add a test for the hook? And that it removes/adds new when the event hub or callback changes

packages/golden-layout/src/utils/index.ts
```diff
@@ -6,3 +6,4 @@ export { default as ReactComponentHandler } from './ReactComponentHandler';
 export * from './ConfigMinifier';
 export { default as BubblingEvent } from './BubblingEvent';
 export { default as EventHub } from './EventHub';
+export * from './EventUtils';
```
mattrunyon, line 9: Do we want to export everything here or just `makeEventFunctions` and the types?

mattrunyon (changes requested):

packages/dashboard/src/PanelEvent.ts
```diff
@@ -7,16 +7,26 @@ export type WidgetDescriptor = {
 };
 
 export type PanelOpenEventDetail<T = unknown> = {
-  dragEvent?: DragEvent;
-  fetch?: () => Promise<T>;
+  /** Opening the widget was triggered by dragging from a list. The coordinates are used as the starting location for the drag.  */
+  dragEvent?: MouseEvent;
```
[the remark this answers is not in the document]
mofojed: When you're dragging from the widget list to add a widget to the dashboard, that's what this drag event is for.

packages/golden-layout/src/utils/index.ts
mofojed: I don't see any reason _not_ to export the other functions.

mattrunyon: My thought was if we expect to use `makeEventFunctions` and not the specific function calls then might as well only export that, but it doesn't hurt anything to export the others

PR #2148, opened 2024-07-16
fix: error when edited cell is out of grid viewport

### LINKED ISSUE #2087: Error when Edited Cell scrolls out of Viewport
**Description**

When an Edited Cell is scrolled out the viewport, the table fails with an error "Unable to get ModelIndex for row N". This was discovered while investigating the Enterprise ticket DH-16515. I have only reproduced this scrolling rows, but I suspect it also happens with columns. Tested with both the Legacy and Core+ engines.

**Steps to reproduce**

1. Create an InputTable.
2. Double click on cell to start editing.
3. While the cell is still in editing mode, scroll it off the screen.

**Expected results**

The table should scroll like normal.

**Actual results**

The displays an error "Unable to get ModelIndex for row N"

image

**Versions**
This was found on Enterprise 1.20231218.146 with Core engine 0.33.4

AkshatJawne:
Resolves #2087 

**Steps Moving Forward:**
- Clarify whether we want to be able to **handle edits that are not in the current viewport** -- with my understanding, this will mean we will have to load in more grid data, which does not seem overly valuable (in terms of the efficiency tradeoff) since we are dealing with an improbable edge case where the user begins to edit in the viewport, continues editing outside of the viewport where the cell originally was, and then wants to commit those changes outside as well

** A better solution for the edge case above IMO  is to **create a new enhancement ticket** where we attempt to have the editing cell be floating/sticky in the grid viewport, similar to how it is done on Google Sheets **

**_Ticket Created:_** https://github.com/deephaven/web-client-ui/issues/2153

Screenshot 2024-07-16 at 10 37 02 AM

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -2167,15 +2167,25 @@ class Grid extends PureComponent<GridProps, GridState> {
           }
         : { opacity: 0 };
 
-    const modelColumn = this.getModelColumn(column);
-    const modelRow = this.getModelRow(row);
+    let modelColumn;
+    let modelRow;
+    try {
+      modelColumn = this.getModelColumn(column);
+      modelRow = this.getModelRow(row);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      return null;
+    }
     const inputStyle: CSSProperties | undefined =
       modelColumn != null && modelRow != null
         ? {
             textAlign: model.textAlignForCell(modelColumn, modelRow),
           }
         : undefined;
-    const isValid = model.isValidForCell(modelColumn, modelRow, value);
+    const isValid =
+      modelColumn != null && modelRow != null
+        ? model.isValidForCell(modelColumn, modelRow, value)
+        : false;
 
     return (
       <div style={wrapperStyle}>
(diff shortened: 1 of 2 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -2167,15 +2167,25 @@ class Grid extends PureComponent<GridProps, GridState> {
...
 
-    const modelColumn = this.getModelColumn(column);
-    const modelRow = this.getModelRow(row);
+    let modelColumn;
+    let modelRow;
+    try {
+      modelColumn = this.getModelColumn(column);
+      modelRow = this.getModelRow(row);
+    } catch (e) {
+      // eslint-disable-next-line no-console
```
mofojed, line 2176:
```suggestion
```

mofojed (changes requested):
Small cleanup. 

Feel free to file an feature request for the floating google docs type behaviour you've described. With the floating cell we'd want to display the cell coordinates next to the floating input, but that could get cluttered with a long column name, and also we don't display the row headers by default. Just to note on the ticket.

image

mofojed (approved):

PR #2151, opened 2024-07-17
feat: Adjustable grid density

### LINKED ISSUE #885: Option to increase information density

mattrunyon:
Fixes #885 

This adds a user setting for grid density in the theme section which applies to all grids without an explicit density prop. A separate dh.ui PR will allow setting the density per table and is not influenced by the global density setting.

A few known issues/things I still need to clean up
- Adjusting quick filter bar icon size and position

Not sure if we should call the option "Choose grid density" or something else ("Choose table density", "Table density")


dsmmcken:
What about "Default table density"?

Implying ui.table() density overrides default density, similar to "Default formatting"

packages/app-utils/src/components/ThemeBootstrap.tsx
```diff
@@ -21,11 +23,17 @@ export function ThemeBootstrap({ children }: ThemeBootstrapProps): JSX.Element {
     [pluginModules]
   );
 
+  const settings = useSelector<RootState, ReturnType<typeof getSettings>>(
```
mofojed, line 26: Use `useAppSelector` instead, it should get typed automatically: https://github.com/deephaven/web-client-ui/blob/58ee88dc92bfe9a283ebc789c93f23639a954ba3/packages/dashboard/src/redux/selectors.ts#L29

packages/redux/src/store.ts
```diff
@@ -59,6 +59,7 @@ export interface WorkspaceSettings {
   };
   webgl: boolean;
   webglEditable: boolean;
+  gridDensity?: 'compact' | 'normal' | 'spacious';
```
mofojed, line 62: This should always be set. Also we may as well match the Spectrum density props, which use `regular` instead of `normal` (though not every component has a `spacious` option).
```suggestion
  gridDensity: 'compact' | 'regular' | 'spacious';
```
Then define what the default value is in `LocalWorkspaceStorage`.
In the `updateWorkspaceData` action, we map the users `CustomizableWorkspaceSettings` on top of the default settings. That way the default values are always set in the `WorkspaceSettings` we get from redux, and there's no ambiguity on what the "default" value is.

Side note, we should probably allow this to be set via a prop as well (for a default value). Can be a separate ticket.

packages/iris-grid/src/IrisGridThemeProvider.tsx
```diff
@@ -14,14 +14,21 @@ export const IrisGridThemeContext =
 
 export interface IrisGridThemeProviderProps {
   children: ReactNode;
+  density?: 'compact' | 'normal' | 'spacious';
```
mofojed, line 17: Add docs/what the default is if unset (assumedly `normal`)

packages/iris-grid/src/IrisGridTheme.ts
```diff
@@ -49,13 +49,60 @@ export type IrisGridThemeType = GridThemeType & {
...
+  minColumnWidth: number;
+  rowHeight: number;
+  font: string;
+  headerFont: string;
+  iconSize: number;
+  columnHeaderHeight: number;
+  filterBarHeight: number;
+}
+
+export const NORMAL_DENSITY_THEME = {
```
mofojed, line 64: We may want to remove these from being defined in the theme created below and just spread `NORMAL_DENSITY_THEME` in there by default. Otherwise we're re-defining these values in two places, if we ever do bother to change one of them might forget to change the other...

That also being said - we should use `IrisGridTheme['header-font']` for `headerFont` here, and all the other things from `IrisGridTheme.module.scss` instead of redefining here.

packages/code-studio/src/settings/ThemeSectionContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ThemePicker,
+  useTheme,
+} from '@deephaven/components';
+import { assertNotNull } from '@deephaven/utils';
+import { useDispatch, useSelector } from 'react-redux';
+import { getSettings, updateSettings, type RootState } from '@deephaven/redux';
+
+export function ThemeSectionContent(): JSX.Element {
+  const theme = useTheme();
+  const settings = useSelector<RootState, ReturnType<typeof getSettings>>(
```
mofojed, line 15: `useAppSelector`

packages/code-studio/src/settings/ThemeSectionContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+    getSettings
+  );
+  const dispatch = useDispatch();
```
mofojed, line 18: `useAppDispatch`

packages/code-studio/src/settings/ThemeSectionContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const density = settings.gridDensity ?? 'normal';
+
+  assertNotNull(theme, 'ThemeContext value is null');
+
+  return (
+    <>
+      <ThemePicker />
+      <Picker
+        label="Choose grid density"
```
mofojed, line 42: Maybe a question for @dsmmcken, but should this density apply to other Spectrum components such as ListView by default as well? https://react-spectrum.adobe.com/react-spectrum/ListView.html#density
Though some items (like `Tabs`) don't have a `spacious` option.

packages/code-studio/src/settings/ThemeSectionContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+        selectedKey={density}
+        onChange={updateDensity}
+      >
+        <Item key="normal">Normal</Item>
```
mofojed, line 46:
```suggestion
        <Item key="normal">Regular</Item>
```

packages/code-studio/src/settings/ThemeSectionContent.tsx
```diff
@@ -0,0 +1,54 @@
...
+  );
+  const dispatch = useDispatch();
+
+  const updateDensity = useCallback(
+    (density: ItemKey | null) => {
+      if (
+        density !== 'normal' &&
+        density !== 'compact' &&
+        density !== 'spacious'
+      ) {
```
mofojed, line 26: If we had these values defined as a constant that we could do a `DENSITY_OPTIONS.includes(density)` instead, that would be ideal. This isn't typed at all right now.
When declaring `gridDensity` type, could do:
```
const DENSITY_OPTIONS = ['compact', 'regular', 'spacious'] as const
type Density = typeof DENSITY_OPTIONS[number]
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1396,17 +1404,31 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+      const { density } = this.props;
+
       return {
-        ...defaultTheme,
+        // Base theme includes global density settings
+        ...baseTheme,
+        // Explicitly set density overrides base theme
+        ...(density === 'normal' ? NORMAL_DENSITY_THEME : {}),
+        ...(density === 'compact' ? COMPACT_DENSITY_THEME : {}),
+        ...(density === 'spacious' ? SPACIOUS_DENSITY_THEME : {}),
```
mofojed, line 1431: Maybe add a function to get the density theme based on density, e.g.

```
getDensityTheme(density) {
  switch (density) {
    case 'compact': return COMPACT_DENSITY_THEME;
    case 'spacious': return SPACIOUS_DENSITY_THEME;
    default: return REGULAR_DENSITY_THEME;
  }
}
```

mofojed (changes requested):

packages/app-utils/src/components/ThemeBootstrap.tsx
mattrunyon: Nice. It was annoying that `useSelector<RootState>(getSettings)` didn't infer the return type

packages/redux/src/store.ts
mattrunyon: Ya I initially made it optional since it wouldn't be set in enterprise right away, but shouldn't matter until we add the settings menu in enterprise

packages/code-studio/src/settings/ThemeSectionContent.tsx
mattrunyon: True. Since it was only 3 values and only in 1 place that I needed to do this (because of Picker saying it can be `string | int`) I was indifferent on doing something like the array. Although I guess I could create the items based on the array too

packages/code-studio/src/settings/ThemeSectionContent.tsx
mattrunyon: One vote for no is users of dh.ui should already be able to specify `density` on those components.

In our UI, we would then need to test anywhere we use list or tabs (or others w/ density) looks ok w/ all density options

packages/code-studio/src/settings/ThemeSectionContent.tsx
mattrunyon: This doesn't seem to work with the thunk action I'm dispatching

mattrunyon: I believe this is because we aren't using `configureStore` with redux and are using the old, deprecated `createStore`

packages/code-studio/src/settings/ThemeSectionContent.tsx
mattrunyon: Actually the types get more annoying than just doing this check IMO. `Array.includes` types based on the array type, so if you use a const string literal array, it only accepts values that are known to be in the array (which is pretty dumb IMO). The reasoning I saw in a closed issue was "you could be checking for a number in an array of strings which is probably wrong".

I'm just going to leave this as is I think

packages/app-utils/src/storage/LocalWorkspaceStorage.ts
```diff
@@ -121,7 +122,9 @@ export class LocalWorkspaceStorage implements WorkspaceStorage {
       ),
     };
 
-    const keys = Object.keys(serverSettings) as Array<keyof typeof settings>;
+    const keys = Object.keys(serverSettings) as Array<
+      keyof typeof serverSettings
```
mofojed, line 126: Nice catch

packages/iris-grid/src/IrisGridTheme.ts
```diff
@@ -61,36 +61,61 @@ export interface IrisGridDensity {
...
+ * Get the density specific theme settings for a density.
+ * @param density Density of the theme to get
+ * @returns Density specific theme settings
+ */
+export function getDensityTheme(
+  density: 'compact' | 'regular' | 'spacious'
+): IrisGridDensity {
+  // This must be read when the function is called and not in the global scope of the module
+  // Otherwise it initializes w/ a bunch of empty values
+  const IrisGridTheme = resolveCssVariablesInRecord(IrisGridThemeRaw);
```
mofojed, line 74: I think we can simplify this. Instead of having `IrisGridDensity` be explicitly defined, it should just be something like:
```
export interface IrisGridDensity = Partial<IrisGridThemeType>;
```
I don't think we need to get fancier than that, but if we do, add in a `Pick`:
```
export interface IrisGridDensity = Partial<Pick<IrisGridThemeType, 'cellHorizontalPadding' | ...>>;
```

Afterward, this `getDensityTheme` can just return a partial for the properties it does override from the normal theme. So for `'regular'`, you'd just return `{}` since you just want the regular theme, `'compact'` you'd return what you are now, and for `'spacious'` you'd just return:
```
{
    cellHorizontalPadding: 7,
    headerHorizontalPadding: 15,
    rowHeight: 28,
}
```

Then this function doesn't need to resolve the CSS variables and the override is simpler.

packages/iris-grid/src/IrisGridTheme.ts
```diff
@@ -183,21 +205,16 @@ export function createDefaultIrisGridTheme(
     gridRowColor: null,
     groupedColumnDividerColor: IrisGridTheme['grouped-column-divider-color'],
     columnHoverBackgroundColor: null,
-    headerHorizontalPadding: 12,
```
mofojed, line 139: Leave these defined in the default theme, this is just the `regular` density.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1422,13 +1422,17 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
 
       const { density } = this.props;
 
+      // Only get theme for explicitly set density
+      let densityTheme = {};
+      if (density != null) {
+        densityTheme = getDensityTheme(density);
```
mofojed, line 1428: If you do as my other suggestion says to just return `{}` for `regular`, then you don't need this `undefined` check and can always just set the density, simplifying this.

mofojed (changes requested):
Lookin good!

packages/code-studio/src/settings/ThemeSectionContent.tsx
mofojed: Yup limit the scope of this ticket, but could be interesting to do later.

mofojed:
We should have a follow up ticket to read the default density from the server.

packages/iris-grid/src/IrisGridTheme.ts
mattrunyon: With the way the global density setting is implemented, I think this wouldn't work for overriding the individual densities.

Right now, the global setting is merged at the `IrisGridThemeContext` to provide the color + density choices as the default for grids. If the densities aren't defined fully, then there's a problem when you change the global setting and also set a density explicitly w/ dh.ui.

For example, set the global density to compact. Then with dh.ui, create a ui.table w/ `density="regular"`. Now we need either a separate context or context value from `IrisGridThemeContext` to apply the global density in `IrisGrid` or we need `IrisGrid` to be able to fetch the actual values for the regular density. But you can't just get the default theme and spread it for default density. Because now you've reverted the color part of the theme set in the context.

So we need to know what to override when `<IrisGrid density="..." />` is used. This seemed like the easiest way. Otherwise `IrisGrid` is dealing with global, app-specific redux state here to fetch the global density instead of just reading its theme context for global theme.

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: Would need to connect  `IrisGrid` to redux to do that or create a new context to provide the default density to `IrisGrid`. Could modify the `IrisGridThemeContext` to contain theme + density, I guess, but this seemed cleaner

packages/iris-grid/src/IrisGridTheme.ts
mofojed: Okay, so sleeping on this the thing that's bugging me is that we're arbitrarily defining the density theme overrides; really you should be able to define all the different densities within the theme itself, and then you pick which densities. Otherwise say you wanted to theme the `regular` theme with a default `cellHorizontalPadding: 15`, but you can't theme the `spacious` density so when the user switches to that, it would actually reduce density (making the "Density" options useless, depending on your originally defined theme). So somehow we should be able to use the theme to define all the different densities, and that can be the theme passed in to the ThemeProvider.
Just taking a look at how Spectrum handles this. In their tabs, they are just setting an extra CSS class if the `compact` density is used: https://github.com/adobe/react-spectrum/blob/9b48ec5d528adeba1a96592945459726def7ba56/packages/%40react-spectrum/tabs/src/Tabs.tsx#L297
Same with ListView: https://github.com/adobe/react-spectrum/blob/9b48ec5d528adeba1a96592945459726def7ba56/packages/%40react-spectrum/list/src/ListView.tsx#L232
Then they define the extra CSS for those classes specifically: https://github.com/adobe/react-spectrum/blob/9b48ec5d528adeba1a96592945459726def7ba56/packages/%40react-spectrum/list/src/styles.css#L548

It would make sense to do something similar... though I don't think a nested class definition in `IrisGridTheme.module.scss` actually generates the correct object when imported, it looks like it gets mangled (I tried adding a `.spacious { }` section within the module.scss and imported):
image

In any case, we can just define it using a `--compact` or `--spacious` suffix as Spectrum does for their class names, and filter out properties based on that suffix for density.
Then the theme defined in the context is defined entirely in SCSS (as it should be) and defines all the different density specs, and then we just pick the appropriate one based on the density provided for a particular grid.
With that in mind, the default density setting should not override the theme itself... really it should be a separate context. However, that's a bit of a pain for `IrisGrid` which is a class component and only has one `contextType`... so then `IrisGridPanel`, `GridWidgetPlugin`, etc. need to read the density from settings and pass in that prop.

mattrunyon: Moved compact and spacious into the `IrisGridTheme` interface. Didn't bother with CSS variables because almost none of the size is currently in CSS variables, so might as well do that all at once if we want to support grid size properties in the main theme

packages/redux/src/store.ts
mofojed: Would be nice to add to both but no urgency for DHC

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1420,15 +1419,17 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         metricCalculator.resetCalculatedRowHeights();
       }
 
-      const { density } = this.props;
+      let densityTheme = {};
+      if (density === 'compact') {
+        densityTheme = baseTheme.density.compact;
+      }
+      if (density === 'spacious') {
```
mofojed, line 1426: Minor, doesn't really matter
```suggestion
      } else if (density === 'spacious') {
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -537,6 +534,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     canDownloadCsv: true,
     frozenColumns: null,
     theme: null,
+    // Do not set a default density prop since we need to know if it overrides the global density setting
```
mofojed, line 537: This doesn't really apply anymore
```suggestion
```

mofojed (approved):

packages/iris-grid/src/IrisGrid.tsx
mofojed: Ah true k that works

mattrunyon:
#2162 and DH-17459 for setting via server and adding the setting to DHE

dsmmcken (approved):

Issue #2089, opened 2024-06-20
Input Tables cannot Paste More than the Number of Visible Rows


dgodinez-dh:
**Description**

Pasting a large number of rows into an InputTable fails with an error "Copy and paste area are not same size."

**Steps to reproduce**

1. Run the following groovy script:
`import io.deephaven.engine.table.impl.util.AppendOnlyArrayBackedInputTable`
`copyHere = emptyTable(1000).update("X = i")`
`pasteHere = AppendOnlyArrayBackedInputTable.make(emptyTable(1).update("X = i"))`

2. From "copyHere" table, select all and copy.
3.  Copy data in the "pasteHere" table.

**Expected results**

Data should paste.

**Actual results**

InputTable displays error: "Copy and paste area are not same size."

**Additional details and attachments**

Problem found in Enterprise ticket DH-11898. Occurs in Grid.tsx pasteValue method.

**Versions**

Enterprise: 1.20231218.359
Core+: 0.33.4

(Note: Found earlier but fix in 1.20231218)

mofojed:
Python equivalent:
```
from deephaven import empty_table, input_table
copy_here = empty_table(1000).update(["X=i"])
paste_here = input_table(init_table=empty_table(1).update(["X = i"]))
```

PR #2152, opened 2024-07-17
fix: Input Tables cannot paste more rows than number of visible rows

AkshatJawne:
Resolves https://github.com/deephaven/web-client-ui/issues/2089

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1497,14 +1497,13 @@ class Grid extends PureComponent<GridProps, GridState> {
       const { columnCount, rowCount } = model;
       let ranges = selectedRanges;
       // If each cell is a single selection, we need to update the selection to map to the newly pasted data
+      // Check for
       if (
         ranges.every(
           range =>
-            (GridRange.cellCount([range]) === 1 &&
-              (range.startColumn ?? 0) + tableWidth <= columnCount &&
-              (range.startRow ?? 0) + tableHeight <= rowCount) ||
-            (GridRange.rowCount([range]) === tableHeight &&
-              GridRange.columnCount([range]) === tableWidth)
+            GridRange.cellCount([range]) === 1 &&
+            (range.startColumn ?? 0) + tableWidth <= columnCount &&
+            (range.startRow ?? 0) + tableHeight <= rowCount
         )
       ) {
         // Remap the selected ranges
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1500,9 +1500,11 @@ class Grid extends PureComponent<GridProps, GridState> {
       if (
         ranges.every(
           range =>
-            GridRange.cellCount([range]) === 1 &&
-            (range.startColumn ?? 0) + tableWidth <= columnCount &&
-            (range.startRow ?? 0) + tableHeight <= rowCount
+            (GridRange.cellCount([range]) === 1 &&
+              (range.startColumn ?? 0) + tableWidth <= columnCount &&
+              (range.startRow ?? 0) + tableHeight <= rowCount) ||
+            (GridRange.rowCount([range]) === tableHeight &&
+              GridRange.columnCount([range]) === tableWidth)
         )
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1661,17 +1663,6 @@ class IrisGridTableModelTemplate<
       return false;
     }
 
-    // Editing the aggregations/totals which are floating above/below is not allowed
```
mofojed, line 1664: As mentioned on our previous call, I think we still need this check - we need to make sure we don't paste and edit the aggregation row. If you have an input table and Add Aggregation, right now you can edit that aggregation row (though it's also not displaying the correct value when placed on the Bottom, which is another bug which should be fixed).

mofojed (changes requested):
A case that's needs the failure case cleaned up: If you copy multiple columns, and then paste in the right-most column you get an error. e.g.
```
from deephaven import empty_table, input_table
copy_here = empty_table(1000).update(["X=i", "Y=-i"])
paste_here = input_table(init_table=empty_table(1).update(["X=0", "Y=0", "Z=0"]))
```
Copy from `copy_here`, then paste in column Z in `paste_here`. The error thrown is an actually a TypeError, when we should be throwing an error saying the Copy/Paste area are not the same size. 

image

packages/grid/src/Grid.tsx
```diff
@@ -1518,16 +1520,6 @@ class Grid extends PureComponent<GridProps, GridState> {
         this.setSelectedRanges(ranges);
       }
 
-      if (
```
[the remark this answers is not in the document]
dsmmcken: I think I agree.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
AkshatJawne: Having this check will automatically fail cases where the pasting size is larger, given that the `endRow` will always be larger than the `topRowCount` + `table size` + `pendingRowCount` -- which is why I had initially opted to remove it, and then after we had spoken, I had tested further, but had not re-added it. With regards, to the aggregation row, is there any other way we can tell if a row is an aggregation row, and then based on that, I can have a check to basically check if the `endRow` is going beyond the aggregation row?

With regards to the separate bug you identified, as we discussed in call, I had created a seperate ticket for that: https://github.com/deephaven/web-client-ui/issues/2156

packages/grid/src/Grid.tsx
```diff
@@ -1500,9 +1500,11 @@ class Grid extends PureComponent<GridProps, GridState> {
...
         ranges.every(
           range =>
-            GridRange.cellCount([range]) === 1 &&
-            (range.startColumn ?? 0) + tableWidth <= columnCount &&
-            (range.startRow ?? 0) + tableHeight <= rowCount
+            (GridRange.cellCount([range]) === 1 &&
+              (range.startColumn ?? 0) + tableWidth <= columnCount &&
+              (range.startRow ?? 0) + tableHeight <= rowCount) ||
+            (GridRange.rowCount([range]) === tableHeight &&
+              GridRange.columnCount([range]) === tableWidth)
```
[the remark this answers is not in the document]
AkshatJawne: Removed - stale check.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
mofojed: Yea in terms of aggregation row we have a bit of a conflict there... it uses `floatingBottomRowCount` when aggregations are pinned to the bottom, and assumes the aggregations are the last rows in the table; but with these pending rows we actually want an "infinite" number of rows/always able to scroll. One of the downsides of the design of floating rows (as opposed to having it as a separate "table" of data for just the floating rows or something).

I think we'll just need to address that with #2156, but I won't hold it up for this (since the aggregations don't appear correctly already anyway).

mofojed (approved):

PR #2155, opened 2024-07-17
fix: add back panel prop to IrisGrid Plugin

### LINKED ISSUE #2093: TablePlugin needs to know current grid selection, table name
As a developer of a TablePlugin, I would like to be able to get the table name and the current selection state from within my plugin.
Previously we could do this with the `panel` prop that was passed in before it was removed: https://github.com/deephaven/web-client-ui/pull/1982
Could use `panel.getTableName()` to get the table name and `panel.irisGrid.current.state` to get the selected range of rows. Would like to still be able to do that.

AkshatJawne:
Closes #2093  (For now, will reopen ticket after merge for better solution to handle deprecated `panel` prop)

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -423,7 +423,7 @@ export class IrisGridPanel extends PureComponent<
         return null;
       }
 
-      // TODO: Find a better way to handle deprecated panel prop
+      // TODO #2093: Find a better way to handle deprecated panel prop
       const deprecatedProps = {
         panel: this,
       };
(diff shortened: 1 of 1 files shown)


packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -423,6 +423,11 @@ export class IrisGridPanel extends PureComponent<
         return null;
       }
 
+      // TODO: Find a better way to handle deprecated panel prop
```
mofojed, line 426: For a TODO like this, we should reference a ticket number and have a ticket created. We can clone the current ticket, or just use the same ticket since this is kind of an interim solution. We want something better.

```suggestion
      // TODO #2096: Find a better way to handle deprecated panel prop
```

mofojed (changes requested):

AkshatJawne:
Added current ticket number in TODO comment, and updated description to indicate that ticket will reopen after this merge to look at a more long-term, proper solution.

mofojed (approved):

PR #2158, opened 2024-07-18
fix: Remove the session wrapper from the embedded widget app

mofojed:
- We creating a new session and getting the session details, even though that was unnecessary for fetching the object itself since we already have a connection
- Just pass `isConsoleAvailable: false` here, since we will never have the console available in the embedded widget scenario
- Should speed up the embed-widget fetch quite a bit


mattrunyon (approved):

mattrunyon:
Shouldn't need a V+ backport. This was added in 0.83.0 https://github.com/deephaven/web-client-ui/commit/b090f200b38a7ecab1056b17f445c2af3ae09a41

PR #2159, opened 2024-07-18
fix: Restrict officially supported browserlist

### LINKED ISSUE #1752: Change officially supported browserlist to be more restrictive
We claim to support browserlist "defaults" as our officially supported browsers (browsers with >0.5% global usage), but that's not actually true. We don't actually support mobile browsers, nor have we ever tested with some of the more obscure browsers. It terms of automated testing, our e2e tests run against chromium, firefox and webkit, so we should treat that as our officially supported list.

We have had this as a question from more than one customer. I propose we narrow our rule to be more specific and not be based on usage but time. In enterprise environments update cadence is generally more on a cycle, like quarterly or every 8 weeks. We should publish a docs page stating what the new rule is that we can refer customers to.

My proposed rule is the last 6 months of any chromium (chrome, edge, opera) or gecko (firefox) based browser, the firefox extended support release (which is every 42 weeks), and the last 1 year of webkit (safari). Safari only does major version bumps annually and the second most used safari version is 7 months old.

```
Chrome > 0 and last 0.5 years,
Edge > 0 and last 0.5 years,
Opera > 0 and last 0.5 years,
Firefox > 0 and last 0.5 years,
Firefox ESR,
Safari > 0 and last 1 year
```

AkshatJawne:
Closes #1752

** The only thing I am not 100% sure about is how we can test this before merging **

I tried experimenting with checking the compatible browsers on their own website: https://browsersl.ist/

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,6 @@
-# Use the default value for browsers we support https://browsersl.ist/#q=defaults
-# Used by post-css only, JS browser targeting is controlled by Vite https://vitejs.dev/config/build-options.html#build-target
-
-defaults
+last 0.5 years and Chrome > 0 and not dead
+last 0.5 years and Edge > 0 and not dead
+last 0.5 years and Opera > 0 and not dead
+last 0.5 years and Firefox > 0 and not dead
+last 1 years and Safari > 0 and not dead
+Firefox ESR and not dead
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,6 +1,6 @@
-last 0.5 years and Chrome > 0 and not dead
-last 0.5 years and Edge > 0 and not dead
-last 0.5 years and Opera > 0 and not dead
-last 0.5 years and Firefox > 0 and not dead
-last 1 years and Safari > 0 and not dead
+last 0.5 years and Chrome > 0
+last 0.5 years and Edge > 0
+last 0.5 years and Opera > 0
+last 0.5 years and Firefox > 0
+last 1 years and Safari > 0
 Firefox ESR and not dead
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "@types/shell-quote": "^1.7.1",
     "@vitejs/plugin-react-swc": "^3.3.0",
     "@vscode/codicons": "0.0.36",
+    "browserslist-to-esbuild": "^2.1.1",
     "chokidar-cli": "^2.1.0",
     "conventional-changelog-conventionalcommits": "^7.0.0",
     "cross-env": "^7.0.2",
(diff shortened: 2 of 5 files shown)


.browserslistrc
```diff
@@ -1,4 +1,6 @@
-# Use the default value for browsers we support https://browsersl.ist/#q=defaults
-# Used by post-css only, JS browser targeting is controlled by Vite https://vitejs.dev/config/build-options.html#build-target
```
mofojed, line 2: Take a look at this comment you removed (about how this is just for post-css...). We need to wire up Vite to use `.browserslistrc` in place of `build.target` using the `browserslist-to-esbuild` module it looks like: https://github.com/vitejs/vite/issues/11489
A blog post about Vite and browserslist may be information as well: https://dev.to/meduzen/when-vite-ignores-your-browserslist-configuration-3hoe

Otherwise this `.browserslistrc` is only going to be used for CSS, and not used when building JS.

We should also be able to compare the package size before/after and see the impact (check the size of the `code-studio/build` folder after doing an `npm run build` before and after).

package.json
```diff
@@ -181,6 +181,7 @@
     "@deephaven/storage": "file:packages/storage",
     "@deephaven/stylelint-config": "file:packages/stylelint-config",
     "@deephaven/tsconfig": "file:packages/tsconfig",
-    "@deephaven/utils": "file:packages/utils"
+    "@deephaven/utils": "file:packages/utils",
+    "browserslist": "^4.23.2"
```
mofojed, line 185: Do we need this package? How is it being used?

mofojed (changes requested):

package.json
AkshatJawne: We do not need this package necessarily , but I thought it could be good for two reasons:
a. When the browser is not compatible, it shows a banner https://github.com/deephaven/deephaven-plugins/issues/510 (3rd tool here)
b. It could give us statistics about browser usage, can be leveraged in the future https://github.com/deephaven/deephaven-plugins/issues/510 (4th tool here)

However, if these benefits are not as valuable, could also be removed, no issues on my end.

AkshatJawne:
@mofojed Size Comparison of `code-studio/build`:

Before changes: 56.4 MB
After changes: 56.9 MB

mofojed (changes requested):
Update the web-client-ui README with the browsers we support/link to this file: https://github.com/deephaven/web-client-ui?tab=readme-ov-file#browser-support
@dsmmcken was there any other testing you wanted to be done for this?

dsmmcken:
No, I had no specific testing requirements. Which browser doesn't support top level await?

AkshatJawne:
@dsmmcken This link outlines which versions of which browsers can use top level await https://caniuse.com/mdn-javascript_operators_await_top_level , it seems that with our new restrictions, the browsers and versions that we have outlined should support top level await

mofojed (approved):

PR #2166, opened 2024-07-22
fix: Propogation of Scroll Events when Scroll Position is at a Boundary

### LINKED ISSUE #2101: IrisGrid - don't stop propagation of scroll events when scroll position is at a boundary
IrisGrid currently handles scroll events when cursor is over the grid. This behaves as expected when there is scrollable data. The user scrolls and sees the table scroll accordingly. In cases where there are too few rows to scroll or when there are no more rows to scroll in the target direction, no UI change occurs when the user scrolls, and the scroll event is swallowed by the component. This can be confusing if the user is scrolling a long page that happens to contain IrisGrid examples (e.g. deephaven.io). The page scrolls until the cursor ends up over an IrisGrid example in which case the page stops scrolling. If the grid is not scrollable, the lack of UI change gives the impression that the UI has locked up somehow.

Proposed solution would be to only stop propagation of scroll events that would result in scroll position changes.
* If table has no scrollable data, propagate the event
* If table has no more rows in the direction the user is attempting to scroll, propagate the event
* If table will scroll in some way in response to the event, stop propagation

AkshatJawne:
Closes #2101

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1909,12 +1909,7 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     // Check if at the top and attempting to scroll up
     // Or at the bottom and attempting to scroll down
-    if (
-      (top === 0 && deltaY < 0) ||
-      (top >= rowCount - scrollableContentHeight / metrics.rowHeight &&
-        deltaY > 0)
-    ) {
-      event.stopPropagation();
+    if ((top === 0 && deltaY < 0) || (top === lastTop && deltaY > 0)) {
       return;
     }
 
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1909,7 +1909,14 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     // Check if at the top and attempting to scroll up
     // Or at the bottom and attempting to scroll down
-    if ((top === 0 && deltaY < 0) || (top === lastTop && deltaY > 0)) {
+    // Or at the left and attempting to scroll left
+    // Or at the right and attempting to scroll right
+    if (
+      (top === 0 && deltaY < 0) ||
+      (top === lastTop && deltaY > 0) ||
+      (left === 0 && deltaX < 0) ||
+      (left === lastLeft && deltaX > 0)
+    ) {
       return;
     }
 
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1896,6 +1896,7 @@ class Grid extends PureComponent<GridProps, GridState> {
       hasVerticalBar,
     } = metrics;
     let { top, left, topOffset, leftOffset } = metrics;
+    const initialMetrics = metrics;
 
     const theme = this.getTheme();
 
@@ -1909,9 +1910,11 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     // Check if we're at the edge of the grid and attempting to scroll
     const willScrollX =
-      (left > 0 && deltaX < 0) || (left < lastLeft && deltaX > 0);
+      ((left > 0 || leftOffset > 0) && deltaX < 0) ||
+      (left < lastLeft && deltaX > 0);
     const willScrollY =
-      (top > 0 && deltaY < 0) || (top < lastTop && deltaY > 0);
+      ((top > 0 || topOffset > 0) && deltaY < 0) ||
+      (top < lastTop && deltaY > 0);
 
     if (!willScrollX && !willScrollY) {
       return;
@@ -2058,10 +2061,16 @@ class Grid extends PureComponent<GridProps, GridState> {
       }
     }
 
-    this.setViewState({ top, left, leftOffset, topOffset });
-
-    event.stopPropagation();
-    event.preventDefault();
+    if (
+      initialMetrics.top !== top ||
+      initialMetrics.left !== left ||
+      initialMetrics.topOffset !== topOffset ||
+      initialMetrics.leftOffset !== leftOffset
+    ) {
+      this.setViewState({ top, left, topOffset, leftOffset });
+      event.stopPropagation();
+      event.preventDefault();
+    }
   }
 
   /**
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1909,7 +1909,11 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     // Check if at the top and attempting to scroll up
     // Or at the bottom and attempting to scroll down
-    if ((top === 0 && deltaY < 0) || (top >= rowCount - 1 && deltaY > 0)) {
+    if (
+      (top === 0 && deltaY < 0) ||
+      (top >= rowCount - scrollableContentHeight / metrics.rowHeight &&
+        deltaY > 0)
+    ) {
       event.stopPropagation();
       return;
     }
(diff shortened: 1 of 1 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -2050,8 +2057,11 @@ class Grid extends PureComponent<GridProps, GridState> {
 
     this.setViewState({ top, left, leftOffset, topOffset });
 
-    event.stopPropagation();
-    event.preventDefault();
+    // Check if we're at the top or bottom of the grid
+    if (top >= 0 && topOffset !== 0) {
```
bmingles, line 2061: Scrolling seems to be broken in `IrisGrid` sample in styleguide. I think something isn't working as expected here. Also, a little confused how this is checking top or bottom? Is this a stale comment, or is there something missing in the code here?

packages/grid/src/Grid.tsx
```diff
@@ -1907,6 +1907,13 @@ class Grid extends PureComponent<GridProps, GridState> {
       metrics.rowHeight
     );
 
+    // Check if at the top and attempting to scroll up
+    // Or at the bottom and attempting to scroll down
+    if ((top === 0 && deltaY < 0) || (top >= rowCount - 1 && deltaY > 0)) {
```
bmingles, line 1912: Something doesn't seem right here that you are introducing a `return` that wasn't there before.

bmingles (changes requested):
IrisGrid example in styleguide is broken

packages/grid/src/Grid.tsx
AkshatJawne: Taking a look rn

packages/grid/src/Grid.tsx
AkshatJawne: Well we return because we are no longer looking to scroll, and thus handle movement of the wheel, if we are attempting to scroll down if we are at the bottom, or scroll up if we are at the top. 

But I am looking into to modify. the logic right now, since as per your observation, the IrisGrid is behaving quite differently then the Grid in terms of the scroll.

AkshatJawne:
Still working on implementing fix, the initial approach of returning early or having it stop propogation at the bottom using the `topOffset` is not working

AkshatJawne:
Ready for re-review

packages/grid/src/Grid.tsx
```diff
@@ -1907,6 +1907,12 @@ class Grid extends PureComponent<GridProps, GridState> {
       metrics.rowHeight
     );
 
+    // Check if at the top and attempting to scroll up
+    // Or at the bottom and attempting to scroll down
+    if ((top === 0 && deltaY < 0) || (top === lastTop && deltaY > 0)) {
```
bmingles, line 1912: This seems to work, but I'm not the most familiar with how Grid was implemented. Would probably be good to check with @mofojed to make sure there aren't any edge cases we are missing here.

bmingles: ah, I think this can break horizontal scrolling when at top / bottom boundary since it's hard to scroll horizontally exactly without at least some deltaY

packages/grid/src/Grid.tsx
```diff
@@ -1907,6 +1907,19 @@ class Grid extends PureComponent<GridProps, GridState> {
       metrics.rowHeight
     );
 
+    // Check if at the top and attempting to scroll up
+    // Or at the bottom and attempting to scroll down
+    // Or at the left and attempting to scroll left
+    // Or at the right and attempting to scroll right
+    if (
```
bmingles, line 1914: I think you need something that only returns if neither vertical or horizontal movement will result from the delta.

e.g. I haven't tested but something like:
```typescript
const willScrollX = left > 0 && deltaX < 0 || left < lastLeft && delta > 0
const willScrollY = top > 0 && deltaY < 0 || top < lastTop && delta > 0

if (!willScrollX && !willScrollY) {
  return
}
```

AkshatJawne: It is behaving the same as my commit, but makes more sense semantically, pushed change.

bmingles (approved):
Looks good to me. I tested scenarios where there were page level horizontal + vertical scrollbars. Grid maintains control while you can still control, and then passes control to page at boundaries. Also verified it no longer stutters.

packages/grid/src/Grid.tsx
```diff
@@ -1907,6 +1907,16 @@ class Grid extends PureComponent<GridProps, GridState> {
       metrics.rowHeight
     );
 
+    // Check if we're at the edge of the grid and attempting to scroll
```
bmingles, line 1910: @mofojed this works for me with all cases I tested. I'm not very familiar with how `lastTop` and `lastLeft` are actually implemented or if there are nuances here we are missing. Any concerns on your part?

mofojed: There is one nuance that's missing. To give a little bit of context:
- The way we handle scrolling is by keeping track of the _cell_ index that is in the top left, and then keeping track of the pixel _offset_ within that cell. We do that so we can support very large grids and grids with variable width/height cells (since the width of a column may change after we've fetched data for it, we resize to match the data).
- `left`/`top` are the index of the cell in the top left
- `leftOffset`/`topOffset` is the pixel offset _within_ that cell that's in the top left
- Current scroll position is the position of the top/left cell + the offset

Here we're just checking `left`, but you're not checking `leftOffset` or `topOffset`. You should check if `(left > 0 || leftOffset > 0) && deltaX < 0` to check if you can scroll left (same with the top).

Actually just to be safe - I'd just track the initial values for all those variables `left`, `top`, `leftOffset`, and `topOffset`, and then just at the end of this method before `this.setViewState` check if they haven't changed, and if they haven't, don't bother setting them or eating the event. That seems like the safest.

mofojed (changes requested):

packages/grid/src/Grid.tsx
bmingles: I think @mofojed suggestion here was to remove this section and just have the check at the end that you added below. I'll let him confirm, but I tested again using only the code at the end, and that seems to work as expected. I don't think we need this code at the beginning anymore.

AkshatJawne: Ahh I might have misinterpreted, I thought just to be safe meant that the check at the bottom would be a failsafe, not the only check. Will test by removing initial code and repush accordingly.

bmingles (approved):
Changed look good. Tested in styleguide. Things seem to work as expected.

packages/grid/src/Grid.tsx
```diff
@@ -1896,6 +1896,7 @@ class Grid extends PureComponent<GridProps, GridState> {
       hasVerticalBar,
     } = metrics;
     let { top, left, topOffset, leftOffset } = metrics;
+    const initialMetrics = metrics;
```
mofojed, line 1899: This is unnecessary - `metrics` will not mutate, and even if it did, you're pointing to the same object so you'd see the same mutations.
```suggestion
```

packages/grid/src/Grid.tsx
```diff
@@ -2048,10 +2049,16 @@ class Grid extends PureComponent<GridProps, GridState> {
...
 
-    this.setViewState({ top, left, leftOffset, topOffset });
-
-    event.stopPropagation();
-    event.preventDefault();
+    if (
+      initialMetrics.top !== top ||
+      initialMetrics.left !== left ||
+      initialMetrics.topOffset !== topOffset ||
+      initialMetrics.leftOffset !== leftOffset
```
mofojed, line 2056:
```suggestion
      metrics.top !== top ||
      metrics.left !== left ||
      metrics.topOffset !== topOffset ||
      metrics.leftOffset !== leftOffset
```

mofojed (changes requested):
Just some cleanup but looks good

mofojed (approved):

PR #2167, opened 2024-07-22
fix: ChartBuilderPlugin fixes for charts built from PPQs in Enterprise

mofojed:
- Don't sanitize the descriptor in `AppDashboards` - the descriptor gets sanitized within the objectFetcher itself
  - By sanitizing too early, we lose metadata needed to load an object
- ChartPanelPlugin uses the `panelFetch` to get the underlying object
  - In cases of a `ParameterizedQuery` on Enterprise, we only have the result of the `ParameterizedQuery` run in the `ParameterizedQueryPanel`
    - Aside, if we're looking at improving `ParameterizedQuery` support, we should complete DH-15760 (which would be breaking an internal API)
- Use correct API when fetching a Chart object

### DIFF
--- a/packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
+++ b/packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
@@ -1,10 +1,11 @@
-import { forwardRef, useMemo } from 'react';
-import { useApi } from '@deephaven/jsapi-bootstrap';
+import { forwardRef, useContext, useMemo } from 'react';
+import { DeferredApiContext } from '@deephaven/jsapi-bootstrap';
 import { ChartModel, ChartModelFactory } from '@deephaven/chart';
 import type { dh as DhType } from '@deephaven/jsapi-types';
 import { IrisGridUtils } from '@deephaven/iris-grid';
 import { getTimeZone, store } from '@deephaven/redux';
 import { WidgetPanelProps } from '@deephaven/plugin';
+import { assertNotNull } from '@deephaven/utils';
 import {
   ChartPanelMetadata,
   GLChartPanelState,
@@ -68,7 +69,7 @@ async function createChartModel(
 
 export const ChartPanelPlugin = forwardRef(
   (props: WidgetPanelProps<DhType.plot.Figure>, ref: React.Ref<ChartPanel>) => {
-    const dh = useApi();
+    const deferredApi = useContext(DeferredApiContext);
 
     const panelState = isChartPanelDehydratedProps(props)
       ? (props as unknown as ChartPanelProps).panelState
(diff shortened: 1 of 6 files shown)


packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
```diff
@@ -23,27 +19,21 @@ import ConnectedChartPanel, {
...
     tableName = metadata.table;
-    figureName = undefined;
     tableSettings = metadata.tableSettings;
   } else {
     settings = {};
     tableName = '';
-    figureName = isChartPanelFigureMetadata(metadata)
-      ? metadata.figure
-      : metadata.name;
     tableSettings = {};
```
mattrunyon, line 37: Could just make these the initial values of the variables

packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
```diff
@@ -64,53 +51,25 @@ async function createChartModel(
...
-    table
-  );
+  return ChartModelFactory.makeModel(dh, settings, figure);
 }
 
 export const ChartPanelPlugin = forwardRef(
   (props: WidgetPanelProps<DhType.plot.Figure>, ref: React.Ref<ChartPanel>) => {
-    const dh = useApi();
-    const fetchObject = useObjectFetcher();
+    const deferredApi = useContext(DeferredApiContext);
```
mattrunyon, line 72: This makes me wonder if we should still have `useApi`. Could `useApi` cause unknown bugs in DHE because we don't have the same complexity in DHC? I guess either way we would have to be cognizant of when to pass the metadata

Also, why not `useDeferredApi(metadata)` here?

mattrunyon (changes requested):
Is there a DHE ticket associated with this?

mattrunyon:
You also probably need to merge main to get updated snapshots

packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
```diff
@@ -69,43 +65,41 @@ async function createChartModel(
...
-        },
-      }),
-      [metadata, localDashboardId, deferredApi, panelFetch, panelState]
+    assertNotNull(metadata);
+    const [dh, error] = useDeferredApi(metadata);
+
+    const makeModel = useCallback(async () => {
+      if (error != null) {
+        throw error;
+      }
```
mattrunyon, line 79: Does this error propagate correctly? Just curious as I've had some issues recently w/ throwing from async or effects. Since it executes outside the react render cycle, it doesn't get caught by the error boundary

packages/jsapi-bootstrap/src/useDeferredApi.ts
```diff
@@ -24,8 +24,8 @@ export const DeferredApiContext = createContext<
  * @returns A tuple with the API instance, and an error if one occurred.
  */
 export function useDeferredApi(
-  widget: DhType.ide.VariableDescriptor
-): [typeof DhType | null, unknown | null] {
+  widget: DhType.ide.VariableDescriptor | null
+): [dh: typeof DhType | null, error: unknown | null] {
```
mattrunyon, line 28: Nice

mattrunyon (changes requested):

packages/dashboard-core-plugins/src/ChartPanelPlugin.tsx
mofojed: Yea it gets caught by ChartPanel, when it calls `makeModel`

mattrunyon (approved):

PR #2168, opened 2024-07-22
fix: Check for the getBaseTable API before calling it

mofojed:
- If it's not available, just fallback to the getKeyTable API (which we were using before)
- Tested against deephaven-core 0.35.1
```

x = new_table({
    "Foo": [0, 1, 2, 0, 1, 2],
    "Bar": [4, 5, 6, 7, 8, 9]}
).partition_by(["Foo"])
```


AkshatJawne (approved):
LGTM

PR #2169, opened 2024-07-23
fix: DH-17454: Wrap Modal in SpectrumThemeProvider

vbabich:
Wrap the `Modal` component attached to `document.body` in `SpectrumThemeProvider`.

Had to add `zIndex: 0` and `position: relative` to the `SpectrumThemeProvider` to fix the issue with `DatePicker` popovers nested in the `Modal` rendering behind it.

Screenshot 2024-07-23 at 4 35 35 PM

### DIFF
--- a/packages/components/scss/BaseStyleSheet.scss
+++ b/packages/components/scss/BaseStyleSheet.scss
@@ -9,12 +9,7 @@
   color: var(--dh-color-text);
 
   .spectrum-theme-provider {
-    // This is important for portals with rounded corners (e.g. Popover) so that
-    // the underlying background color shows.
-    --dh-spectrum-theme-provider-bg: unset;
-
-    background-color: var(--dh-spectrum-theme-provider-bg);
-    color: var(--dh-color-text);
+    display: contents;
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/components/scss/BaseStyleSheet.scss
```diff
@@ -9,12 +9,7 @@
   color: var(--dh-color-text);
 
   .spectrum-theme-provider {
-    // This is important for portals with rounded corners (e.g. Popover) so that
-    // the underlying background color shows.
-    --dh-spectrum-theme-provider-bg: unset;
-
-    background-color: var(--dh-spectrum-theme-provider-bg);
-    color: var(--dh-color-text);
+    display: contents;
```
dsmmcken, line 12: seems like this would probably break something else, as that existing code was serving a purpose.

vbabich: I've reverted the change as it wasn't needed after wrapping the `Modal` in `SpectrumThemeProvider`.

dsmmcken (approved):
Confirmed working.

I checked whether we need z-index for popper so spectrum date picker works there too, but it actually doesn't work at all because the spectrum datepicker popper closes the our popper. That'll be a separate issue if we ever decide to place a picker in a poper. Probably not an issue for dh.ui as we wouldn't ever expose our popper, only the spectrum version.

PR #2170, opened 2024-07-24
feat: Export Internationalized Date Types for DatePicker

dgodinez-dh:


packages/components/src/spectrum/dateAndTime.ts
```diff
@@ -12,3 +18,15 @@ export {
   TimeField,
   type SpectrumTimeFieldProps as TimeFieldProps,
 } from '@adobe/react-spectrum';
+
+export type { CalendarDate, CalendarDateTime, ZonedDateTime };
+
+export type DateValue = CalendarDate | CalendarDateTime | ZonedDateTime;
+
+export type MappedDateValue<T> = T extends ZonedDateTime
```
bmingles, line 26: I would probably add a comment on these types describing why we need them.

dgodinez-dh:

packages/components/src/spectrum/dateAndTime.ts
dgodinez-dh: Pushed an update with some comments.

PR #2173, opened 2024-07-26
fix: DH-17454: Combine modal classes instead of replacing

vbabich:


dsmmcken:
For completion you probably wanna update ModalHeader to be the same format.

vbabich:
> For completion you probably wanna update ModalHeader to be the same format.

Ok, I saw ModalHeader already combines classes, just uses a template literal instead of separate arguments. I'll update it too.

dsmmcken (approved):

PR #2179, opened 2024-08-01
fix: Restrict @adobe/spectrum imports

### LINKED ISSUE #1908: Create an eslint rule to prevent import from `@adobe/spectrum`
We should only use the wrapped or extended version of spectrum components.

We should not allow devs to import the spectrum components directly outside of two specific folders: 
- `packages/components/src/spectrum`
- `packages/components/src/theme`

Everywhere else we should be using our wrapped components.

This can be accomplished I think with a `no-restricted-modules` rule.

AkshatJawne:
Closes #1908

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,55 +5,49 @@ const { packageNames, packageManifest } = buildPackageManifest();
 module.exports = {
   root: true,
   extends: ['@deephaven/eslint-config'],
-  ignorePatterns: ['packages/golden-layout/*', 'jest.config.*'],
+  ignorePatterns: ['packages/golden-layout/*', 'jest.config.js'],
   overrides: [
-    {
-      files: ['**/*.@(ts|tsx)'],
-      parserOptions: {
-        project: ['./tsconfig.eslint.json', './packages/*/tsconfig.json'],
-        tsconfigRootDir: __dirname,
-      },
-    },
-    // For each @deephaven package, forbid importing from itself
     ...packageNames.map(packageName => ({
       files: [`packages/${packageManifest.get(packageName)}/**/*.@(ts|tsx)`],
+      parserOptions: {
+        project: ['./tsconfig.eslint.json', './packages/**/tsconfig.json'],
+        tsconfigRootDir: __dirname,
+      },
       rules: {
         'no-restricted-imports': [
           'error',
           {
-            paths: [
-              {
-                name: packageName,
-                message: 'Forbid importing from owning @deephaven package.',
-              },
-            ],
+            name: packageName,
+            message: 'Forbid importing from owning @deephaven package.',
+          },
+          {
+            name: '@adobe/react-spectrum',
+            message: 'forbid importing from @adobe/react-spectrum.',
           },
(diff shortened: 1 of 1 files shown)


mattrunyon (changes requested):
This ends up removing the no import from deephaven packages rules

I think you need to combine the spectrum and self-package imports. With this config, you can do `import { IrisGrid } from '@deephaven/iris-grid';` inside of `IrisGrid`.

mattrunyon:
This also disables the self import check from the ignored paths even if the other rules are combined. Not sure the best way around that. You might be able to combine another `overrides` inside of the override to turn on the rule for self imports only in those files

AkshatJawne:
Not completely sure what can be done here, I have been experimenting with different configurations, but ultimately, because of the way overrides work, we can only disable the `no-restricted-imports` rule for a file path, not a specific import outlined in the` no-restricted-imports`

Gonna keep experimenting but nothing super promising so far

mattrunyon (changes requested):
It's hard for me to comment on just the certain parts, but the config should be this

- Prevents importing from own package everywhere including spectrum components
- Prevents importing from @adobe/react-spectrum everywhere except spectrum components

```js
const buildPackageManifest = require('./packageManifest');

const { packageNames, packageManifest } = buildPackageManifest();

module.exports = {
  root: true,
  extends: ['@deephaven/eslint-config'],
  ignorePatterns: ['packages/golden-layout/*', 'jest.config.*'],
  overrides: [
    {
      files: ['**/*.@(ts|tsx)'],
      parserOptions: {
        project: ['./tsconfig.eslint.json', './packages/*/tsconfig.json'],
        tsconfigRootDir: __dirname,
      },
    },
    ...packageNames.map(packageName => ({
      files: [`packages/${packageManifest.get(packageName)}/**/*.@(ts|tsx)`],
      rules: {
        'no-restricted-imports': [
          'error',
          {
            name: packageName,
            message: 'Forbid importing from owning @deephaven package.',
          },
          {
            name: '@adobe/react-spectrum',
            message:
              'Import from @deephaven/components instead of @adobe/react-spectrum.',
          },
        ],
      },
      overrides: [
        {
          files: [
            'packages/components/src/spectrum/**/*.@(ts|tsx)',
            'packages/components/src/theme/**/*.@(ts|tsx)',
          ],
          rules: {
            'no-restricted-imports': [
              'error',
              {
                name: packageName,
                message: 'Forbid importing from owning @deephaven package.',
              },
            ],
          },
        },
      ],
    })),
  ],
};
```

AkshatJawne (commented):
@mattrunyon On my end, when I test with the npm test:lint with my code and with the code you suggested, they both behave the same.

However, I will be pushing the change you made if you see it as more valid, I see it as more of a semantic change (the new code is a little clearer for someone to understand)

mattrunyon:
The previous version didn't prevent importing from the owning deephaven package when in the spectrum and theme directories. It was actually preventing `import a from 'packageName';`

mattrunyon (approved):

PR #2181, opened 2024-08-07
fix: TablePlugin needs to know table name and selected range

### LINKED ISSUE #2093: TablePlugin needs to know current grid selection, table name

AkshatJawne:
Resolves #2093

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -423,11 +423,6 @@ export class IrisGridPanel extends PureComponent<
         return null;
       }
 
-      // TODO #2093: Find a better way to handle deprecated panel prop
-      const deprecatedProps = {
-        panel: this,
-      };
-
       return (
         <div className="iris-grid-plugin">
           <Plugin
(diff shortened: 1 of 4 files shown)


AkshatJawne (commented):
@mofojed I just passed the `tableName` and `selectedRanges` props down to the plugin, was there anything specific you were wanting here, in terms of a "better way" to handle the prop?

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -423,11 +423,6 @@ export class IrisGridPanel extends PureComponent<
         return null;
       }
 
-      // TODO #2093: Find a better way to handle deprecated panel prop
-      const deprecatedProps = {
-        panel: this,
-      };
-
```
mofojed, line 430: Let's continue to leave this in for now - now that there's a replacement, they have a version to replace the deprecated properties with the new options we're providing.

mofojed (changes requested):

AkshatJawne (commented):
Ready for re-review @mofojed

mofojed (approved):

PR #2183, opened 2024-08-08
feat: Make rollup group behaviour a setting in the global settings menu

AkshatJawne:
WIP - Cannot seem to figure out why the "Group" column is still being added when the `showExtraGroupColumn` is set to false

### DIFF
--- a/packages/code-studio/src/settings/FormattingSectionContent.tsx
+++ b/packages/code-studio/src/settings/FormattingSectionContent.tsx
@@ -28,6 +28,7 @@ import {
   getTruncateNumbersWithPound,
   getShowEmptyStrings,
   getShowNullStrings,
+  getShowExtraGroupColumn,
   updateSettings as updateSettingsAction,
   RootState,
   WorkspaceSettings,
@@ -56,6 +57,7 @@ interface FormattingSectionContentProps {
   truncateNumbersWithPound: boolean;
   showEmptyStrings: boolean;
   showNullStrings: boolean;
+  showExtraGroupColumn: boolean;
   updateSettings: (settings: Partial<WorkspaceSettings>) => void;
   defaultDecimalFormatOptions: FormatOption;
   defaultIntegerFormatOptions: FormatOption;
@@ -72,6 +74,7 @@ interface FormattingSectionContentState {
   truncateNumbersWithPound: boolean;
   showEmptyStrings: boolean;
   showNullStrings: boolean;
+  showExtraGroupColumn: boolean;
   timestampAtMenuOpen: Date;
 }
 
@@ -113,6 +116,8 @@ export class FormattingSectionContent extends PureComponent<
       this.handleShowEmptyStringsChange.bind(this);
     this.handleShowNullStringsChange =
       this.handleShowNullStringsChange.bind(this);
+    this.handleShowExtraGroupColumnChange =
+      this.handleShowExtraGroupColumnChange.bind(this);
 
     const {
       defaultDateTimeFormat,
@@ -124,6 +129,7 @@ export class FormattingSectionContent extends PureComponent<
       truncateNumbersWithPound,
       showEmptyStrings,
       showNullStrings,
+      showExtraGroupColumn,
     } = props;
 
     this.containerRef = React.createRef();
@@ -139,6 +145,7 @@ export class FormattingSectionContent extends PureComponent<
       truncateNumbersWithPound,
       showEmptyStrings,
       showNullStrings,
+      showExtraGroupColumn,
       timestampAtMenuOpen: new Date(),
     };
   }
@@ -330,6 +337,16 @@ export class FormattingSectionContent extends PureComponent<
     this.queueUpdate(update);
   }
 
+  handleShowExtraGroupColumnChange(): void {
+    const { showExtraGroupColumn } = this.state;
+    const update = {
+      showExtraGroupColumn: !showExtraGroupColumn,
+    };
+    console.log('After updating state constructor', showExtraGroupColumn);
+    this.setState(update);
+    this.queueUpdate(update);
+  }
+
   commitChanges(): void {
     const { updateSettings } = this.props;
     const updates = this.pendingUpdates.reduce(
@@ -356,6 +373,7 @@ export class FormattingSectionContent extends PureComponent<
       truncateNumbersWithPound,
       showEmptyStrings,
       showNullStrings,
+      showExtraGroupColumn,
     } = this.state;
 
     const {
@@ -596,6 +614,23 @@ export class FormattingSectionContent extends PureComponent<
               </Checkbox>
             </div>
           </div>
+
+          <div className="form-row mb-3">
+            <label
+              className="col-form-label col-3"
+              htmlFor="default-integer-format-input"
+            >
+              Rollup
+            </label>
+            <div className="col pr-0 pt-2">
+              <Checkbox
+                checked={showExtraGroupColumn ?? null}
+                onChange={this.handleShowExtraGroupColumnChange}
+              >
+                Show extra &quot;group&quot; column
+              </Checkbox>
+            </div>
+          </div>
         </div>
       </div>
     );
@@ -614,6 +649,7 @@ const mapStateToProps = (
   truncateNumbersWithPound: getTruncateNumbersWithPound(state),
   showEmptyStrings: getShowEmptyStrings(state),
   showNullStrings: getShowNullStrings(state),
+  showExtraGroupColumn: getShowExtraGroupColumn(state),
   timeZone: getTimeZone(state),
   defaults: getDefaultSettings(state),
 });
--- a/packages/iris-grid/src/IrisGridTreeTableModel.ts
+++ b/packages/iris-grid/src/IrisGridTreeTableModel.ts
@@ -44,8 +44,12 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
     inputTable: DhType.InputTable | null = null
   ) {
     super(dh, table, formatter, inputTable);
+    console.log(
+      'IrisGridTreeTableModel constructor',
+      this.formatter.showExtraGroupColumn
+    );
     this.virtualColumns =
-      table.groupedColumns.length > 1
+      table.groupedColumns.length > 1 && this.formatter.showExtraGroupColumn
         ? [
             {
               name: '__DH_UI_GROUP__',
@@ -203,10 +207,20 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
   }
 
   get columns(): DhType.Column[] {
-    return this.getCachedColumns(this.virtualColumns, super.columns);
+    return this.getCachedColumns(
+      // this.formatter.showExtraGroupColumn ? this.virtualColumns : [],
+      this.virtualColumns,
+      super.columns
+    );
   }
 
   get groupedColumns(): readonly DhType.Column[] {
+    // console.log(
+    //   'this.formatter.showExtraGroupColumn',
+    //   this.formatter.showExtraGroupColumn,
+    //   'this.virtualColumns',
+    //   this.virtualColumns
+    // );
     return this.getCachedGroupColumns(
       this.virtualColumns,
       this.table.groupedColumns
@@ -321,14 +335,20 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
     (
       virtualColumns: readonly DhType.Column[],
       tableColumns: readonly DhType.Column[]
-    ) => [...virtualColumns, ...tableColumns]
+    ) => [
+      ...(this.formatter.showExtraGroupColumn ? virtualColumns : []),
+      ...tableColumns,
+    ]
   );
 
   getCachedGroupColumns = memoize(
     (
       virtualColumns: readonly DhType.Column[],
       tableGroupedColumns: readonly DhType.Column[]
-    ) => [...virtualColumns, ...tableGroupedColumns]
+    ) => [
+      ...(this.formatter.showExtraGroupColumn ? virtualColumns : []),
+      ...tableGroupedColumns,
+    ]
   );
 
   getCachedFilterableColumnSet = memoize(
--- a/packages/jsapi-utils/src/FormatterUtils.ts
+++ b/packages/jsapi-utils/src/FormatterUtils.ts
@@ -26,6 +26,7 @@ export function createFormatterFromSettings(
     truncateNumbersWithPound,
     showEmptyStrings,
     showNullStrings,
+    showExtraGroupColumn,
   } = settings ?? {};
 
   return new Formatter(
@@ -36,7 +37,8 @@ export function createFormatterFromSettings(
     defaultIntegerFormatOptions,
     truncateNumbersWithPound,
     showEmptyStrings,
-    showNullStrings
+    showNullStrings,
+    showExtraGroupColumn
   );
 }
 
--- a/packages/app-utils/src/storage/LocalWorkspaceStorage.ts
+++ b/packages/app-utils/src/storage/LocalWorkspaceStorage.ts
@@ -56,6 +56,7 @@ export class LocalWorkspaceStorage implements WorkspaceStorage {
       truncateNumbersWithPound: false,
       showEmptyStrings: true,
       showNullStrings: true,
+      showExtraGroupColumn: true,
       defaultNotebookSettings: {
         isMinimapEnabled: false,
       },
@@ -103,6 +104,7 @@ export class LocalWorkspaceStorage implements WorkspaceStorage {
         serverConfigValues,
         'showNullStrings'
       ),
+      showExtraGroupColumn: true,
       defaultNotebookSettings:
         serverConfigValues?.get('isMinimapEnabled') !== undefined
           ? {
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -523,6 +523,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       truncateNumbersWithPound: false,
       showEmptyStrings: true,
       showNullStrings: true,
+      showExtraGroupColumn: true,
       formatter: EMPTY_ARRAY,
       decimalFormatOptions: PropTypes.shape({
         defaultFormatString: PropTypes.string,
@@ -637,6 +638,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.truncateNumbersWithPound = false;
     this.showEmptyStrings = true;
     this.showNullStrings = true;
+    this.showExtraGroupColumn = true;
 
     // When the loading scrim started/when it should extend to the end of the screen.
     this.tableSaver = null;
@@ -1022,6 +1024,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
   showEmptyStrings: boolean;
 
   showNullStrings: boolean;
+
+  showExtraGroupColumn: boolean;
 
   // When the loading scrim started/when it should extend to the end of the screen.
   loadingScrimStartTime?: number;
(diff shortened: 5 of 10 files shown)


packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -203,19 +254,32 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
   }
 
   get columns(): DhType.Column[] {
-    return this.getCachedColumns(this.virtualColumns, super.columns);
+    return this.getCachedColumns(
+      this.formatter.showExtraGroupColumn ? this.virtualColumns : [],
```
mofojed, line 258: You shouldn't need to check `this.formatter.showExtraGroupColumn` anywhere else except when setting `virtualColumns`.
All these methods are already checking `virtualColumns` and behaving accordingly if there are or are not virtual columns.

Also, you're using a new empty array each time this method is being called, which is bad for memoization (i.e. it will never be memoized, so you're going to get a new column set every time).

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -44,47 +44,98 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
...
+  }
+
+  updateFormatterSetting(newFormatter: Formatter): void {
+    if (
+      this.formatter.showExtraGroupColumn !== newFormatter.showExtraGroupColumn
+    ) {
+      this.formatter.showExtraGroupColumn = newFormatter.showExtraGroupColumn;
+      this.updateVirtualColumns();
+    }
+  }
```
mofojed, line 98: This isn't being called from anywhere.
In `IrisGrid.tsx`, you're already creating a new formatter when the settings change, which will call `set formatter`. Override that method and call your `updateVirtualColumns` from there, e.g.:
```typescript
set formatter(newFormatter: Formatter) {
  const oldFormatter = super.formatter;
  super.formatter = newFormatter;
  if (newFormatter.showExtraGroupColumn !== oldFormatter.showExtraGroupColumn) {
    this.updateVirtualColumns();
  }
}
```

mofojed (changes requested):

AkshatJawne:
@mofojed I tried experimenting with trying to have the `IrisGridModelUpdater` listen for the `COLUMNS_CHANGED` event, and then accordingly set the `modelColumns` in the event handler, but that was not working (it also effectively does the same hack, if it were to work, where we set to the `model.columns` anyways).

I am going to push the "hacky" code that we discussed yesterday, and then we can continue the conversation on finding a cleaner way to do this.

AkshatJawne (commented):
Will investigate unit test failures

AkshatJawne:
For the image mismatches, will need to have new images taken, to replace `the rollup-rows-and-aggregate-columns `images,  where the group column is shown by default (given that the setting will be activated by default)

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -71,6 +71,10 @@ const IrisGridModelUpdater = React.memo(
     columnHeaderGroups,
     partitionConfig,
   }: IrisGridModelUpdaterProps) => {
+    if (model.formatter !== formatter) {
```
mofojed, line 74: I think the "less" hacky way to fix this is to get rid of all the `useEffect`s in here. Those run after the render, when really all we want from this component is to set values when they have changed. And then make sure it checks on every render (i.e. the component shouldn't be memoized at the top), and then we don't need to pass in a `modelColumns` prop. Instead could make a hook to run immediately when something has changed, e.g.

```typescript
function useOnChange(callback: () => void, deps: DependencyList): void {
  // Want to call the callback if the deps have changed from the previous time
  const prevDeps = usePrevious(deps);
  if (prevDeps === undefined || !deps.every((dep, i) => dep === prevDeps[i])) {
    callback();
  }
}
```

Then instead of `useEffect` in `IrisGridModelUpdater`, just use `useOnChange`, and use `model.columns` instead of `modelColumns`, get rid of the `modelColumns` prop, and don't wrap the function in `React.memo`. I think that will work.

mofojed (changes requested):

AkshatJawne:
@mofojed Adjusted logic according to your review, e2e tests may still fail because of the images. How can we update those>

mofojed:
@AkshatJawne Run  `npm run e2e:update-ci-snapshots` to update the snapshots: https://github.com/deephaven/web-client-ui/blob/main/README.md#e2e-tests

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -16,9 +17,21 @@ import {
 
 const COLUMN_BUFFER_PAGES = 1;
 
+/**
+ * Custom hook that triggers a callback function when any of the dependencies change.
+ *
+ * @param callback - The function to be called when the dependencies change.
+ * @param deps - The list of dependencies to watch for changes.
+ */
+function useOnChange(callback: () => void, deps: DependencyList): void {
```
mofojed, line 26: Put this in it's own file, add some tests for it

mofojed (changes requested):

AkshatJawne:
@mofojed Ran command, appears the snapshots were not updated?

AkshatJawne:
Screenshot 2024-08-16 at 10 26 03 AM

Tried running `npm run clean` then `npm i`, and then also restarted my computer and TS server, keep getting the above

packages/iris-grid/src/IrisGridTreeTableModel.ts
```diff
@@ -44,47 +44,104 @@ class IrisGridTreeTableModel extends IrisGridTableModelTemplate<
...
+
+  get formatter(): Formatter {
+    return super.formatter;
+  }
+
+  updateVirtualColumns(): void {
+    this.virtualColumns = this.formatter.showExtraGroupColumn
+      ? [
+          {
+            name: '__DH_UI_GROUP__',
```
mofojed, line 105: We shouldn't duplicate this object between this method and the constructor. Two options:
1. Define the constant at the top, wrap it in an `Object.freeze`, and use it in both places
2. Add a flag to `updateVirtualColumns(sendUpdate = true)` and only send an update when the param is `true`, then call `this.updateVirtualColumns(false)` from the constructor instead of assigning the property directly. Also requires adding a `!` to the `virtualColumns` declaration so that it doesn't complain about it not being initialized, e.g. `private virtualColumns!: DisplayColumn[]

mofojed (changes requested):

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -596,6 +613,23 @@ export class FormattingSectionContent extends PureComponent<
               </Checkbox>
             </div>
           </div>
+
+          <div className="form-row mb-3">
+            <label
+              className="col-form-label col-3"
+              htmlFor="default-integer-format-input"
```
mattrunyon, line 620: This needs to be a different id (it's duplicated from above) and there needs to be a corresponding id on the Checkbox. Also we should just be using `for` probably since `htmlFor` is only needed if you need to modify the label ID w/ JS, but it doesn't really matter here

https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -596,6 +613,23 @@ export class FormattingSectionContent extends PureComponent<
...
+            >
+              Rollup
+            </label>
+            <div className="col pr-0 pt-2">
+              <Checkbox
+                checked={showExtraGroupColumn ?? null}
```
mattrunyon, line 626: Shouldn't need the `?? null` here. This value should have a default value as part of the default workspace and never be null or undefined here

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -45,159 +45,153 @@ interface IrisGridModelUpdaterProps {
...
-          right,
-          movedColumns,
-          hiddenColumns,
-          alwaysFetchColumns,
-          COLUMN_BUFFER_PAGES
-        ),
-      [
-        modelColumns,
+function IrisGridModelUpdater({
+  model,
```
mattrunyon, line 49: Don't remove the `React.memo`. It is the equivalent to pure components in class based react and means if the props don't change then the component does not re-render

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -18,7 +19,6 @@ const COLUMN_BUFFER_PAGES = 1;
 
 interface IrisGridModelUpdaterProps {
   model: IrisGridModel;
-  modelColumns: readonly dh.Column[];
```
mattrunyon, line 21: Why are you removing this and using `model.columns` instead? Not sure anything in this file needs to be modified

tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
mattrunyon, line 1: Not sure why the first row of `String` isn't `null` any more and is displaying as just a blank space. Seems like something is incorrect there, but might not be something from this change

packages/jsapi-utils/src/FormatterUtils.ts
```diff
@@ -73,6 +75,9 @@ export function isCustomColumnFormatDefined(
   columnName: string,
   columnType: string
 ): boolean {
+  if (formatter == null) {
+    return false;
+  }
```
mattrunyon, line 80: Why is this added?

packages/react-hooks/src/useOnChange.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param callback - The function to be called when the dependencies change.
+ * @param deps - The list of dependencies to watch for changes.
+ */
+export function useOnChange(callback: () => void, deps: DependencyList): void {
+  const prevDeps = usePrevious(deps);
+  if (prevDeps === undefined || !deps.every((dep, i) => dep === prevDeps[i])) {
+    callback();
+  }
+}
```
mattrunyon, line 15: How is this different from `useEffect`? It's not immediate since `usePrevious` relies on `useEffect` to change its value. And `useEffect` already tracks the previous value for deps. I don't think this hook is necessary to replace the `useEffect` calls in `IrisGridModelUpdater`.

mattrunyon (changes requested):
Haven't looked through everything yet or tested it

packages/iris-grid/src/IrisGridModelUpdater.tsx
mofojed: I suggested he remove it. Problem is columns can change without the model changing. I think it's fine if this re-renders as it's going to be checking `useOnChange` anyway before applying changes. https://github.com/deephaven/web-client-ui/pull/2183#discussion_r1718831523

packages/iris-grid/src/IrisGridModelUpdater.tsx
mofojed: https://github.com/deephaven/web-client-ui/pull/2183#discussion_r1718831523

packages/code-studio/src/settings/FormattingSectionContent.tsx
AkshatJawne: I tried adding for, but then I have to add an exception since I am getting an ESLint error?

tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
AkshatJawne: Have been trying to investigate this for the last little while, not sure how my code would change the value, will retake snapshots to see if maybe any of of the changes I am adding in the new commit will affect it.

packages/react-hooks/src/useOnChange.ts
AkshatJawne: When we were testing using the existing `useEffect` 's, the code was not working, which I believe was because they run after the render.

mattrunyon: This would call the callback during render, but `usePrevious` will trigger its updated value using `useEffect`. So I guess this works it just wouldn't have an updated `usePrevious` value if it ran twice in a render. So the callback may be called multiple times since `usePrevious` won't change until after `useEffect` has run.

If `useEffect` wasn't working, was there something changing the value mid-render that would cause it to not work?

Also, if you wanted true `onChange` and not just change from the last render if something interrupted and caused a 2nd render pass, then use a ref to track the prev deps and update it at the end of the hook

mattrunyon:
I don't understand why this is going on the formatter and not as a prop to `IrisGrid` or something like that. The value is only used in `IrisGridTreeTableModel`, so why pass it through the formatter? To me it doesn't feel like it belongs on formatter since that's for "how to display a cell based on its value and column type".

packages/iris-grid/src/IrisGridModelUpdater.tsx
mattrunyon: Still seems like a smell to me. We're getting rid of the memoization by React to memoize it based on props ourselves instead? I think the smell is this doesn't belong on formatter which then indirectly changes the visible columns. Formatter prrior to this should only change things when it's directly called. Now it's indirectly changing the visible columns instead of just setting that directly in the model

mattrunyon:
So I think my proposal is to remove this from the formatter settings and instead in `componentDidUpdate` where it checks the settings, add a separate check for if this value changed. If it did, then update the model. That way should work without the worry of `useEffect`

Or you could add `showExtraGroupColumn` to the `IrisGridModelUpdater` as a prop and then you would still need the change from `useEffect` I think. Or columns would need to be updated at the end possibly. Either way shouldn't need `useEffect` and if you remove the `usePrevious` hook and use a ref instead in the `useOnChange` hook (so that the value is always up to date even if the render is discarded so effects don't run), then that should be fine

packages/iris-grid/src/IrisGridModelUpdater.tsx
mattrunyon: Fair. Although the hook with `usePrevious` is not quite right. Previous won't be updated until after effects fire, so if a render is discarded due to something like a setState, then the change callback would fire again. At least I'm 95% sure that's how it works

packages/code-studio/src/settings/FormattingSectionContent.tsx
```diff
@@ -614,16 +614,16 @@ export class FormattingSectionContent extends PureComponent<
...
-          <div className="form-row mb-3">
+          <div className="form-row mb-3" id="show_extra_group_column_div">
             <label
               className="col-form-label col-3"
               htmlFor="default-integer-format-input"
             >
               Rollup
             </label>
-            <div className="col pr-0 pt-2">
+            <div className="col pr-0 pt-2" id="show_extra_grou_column_cb">
```
mattrunyon, line 624: This is still incorrect. The label should have `htmlFor="default-show-group-column"` and then the Checkbox with a matching id.

We're actually misusing this in other places though. Labels are supposed to be 1 to 1 for form elements. In some previous sections we now have multiple inputs, but are using the label for the whole section. This breaks accessibility in those cases because the label corresponds to a specific input, but we actually have multiple inputs in that section. Some without a label.

mattrunyon: So the `Checkbox` component already sets a label. It's actually the use of `label` for String (which also has the wrong ID) that is incorrect because there's no input that label is associated with. The others like Integer are technically fine because they associate with the actual input element and then we have a checkbox that has its own label via the component

AkshatJawne (commented):
Ready for re-review @mofojed @mattrunyon

packages/iris-grid/src/IrisGridModelUpdater.tsx
mattrunyon: Checked w/ Akshat on changing `usePrevious` to use a ref instead of `useEffect` for the previous value and it caused a loading loop for the table, so left it as is.

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -40,164 +41,163 @@ interface IrisGridModelUpdaterProps {
...
-      },
-      [model, top, bottom, columns]
-    );
-    useEffect(
-      function updateRollupCOnfig() {
-        if (model.isRollupAvailable) {
-          model.rollupConfig = rollupConfig;
-        }
-      },
-      [model, model.isRollupAvailable, rollupConfig]
```
mattrunyon, line 145: I'm confused why the eslint rule would want this for `useEffect`, but not for `useOnChange` as a custom exhaustive deps hook. Should add it back so this doesn't accidentally break in some way

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -40,164 +41,163 @@ interface IrisGridModelUpdaterProps {
...
-      },
-      [model, selectDistinctColumns]
-    );
-    useEffect(
-      function updateTotalsConfig() {
-        if (model.isTotalsAvailable) {
-          model.totalsConfig = totalsConfig;
-        }
-      },
-      [model, model.isTotalsAvailable, totalsConfig]
```
mattrunyon, line 161: Add back

packages/react-hooks/src/useOnChange.ts
mattrunyon: Changing to a ref caused infinite loading for the table, so leaving `usePrevious` which uses `useEffect` is fine

tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
mattrunyon, line 1: Still not sure why this one specifically doesn't show `null` for the string value after the change. Might need a different ticket to fix that or might not be an issue. I see in the expanded tests it shows `null` for the expanded rows. Just weird it still shows `null` for the char column in this screenshot

mattrunyon (changes requested):

tests/table-operations.spec.ts-snapshots/rollup-rows-and-aggregrate-columns-1-chromium-linux.png
AkshatJawne: Yeah not exactly sure, I can file a ticket if you would like. I just tried regenerating snapshots and nothing changed, and I don't think anyone else made a change to rollups, so not sure what's going on.

packages/iris-grid/src/IrisGridModelUpdater.tsx
```diff
@@ -143,7 +143,8 @@ function IrisGridModelUpdater({
         model.rollupConfig = rollupConfig;
       }
     },
-    [model, rollupConfig]
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [model, model.rollupConfig, rollupConfig]
```
mattrunyon, line 147: This should be `model.isRollupAvailable`

I found out a way to not need the eslint exception. Just destructure the model to get `isRollupAvailable` as a separate variable and then use that instead of `model.isRollupAvailable`.

Weirdly, with `useEffect` it didn't complain with or without `model.isRollupAvailable`

mattrunyon (changes requested):

mattrunyon (dismissed):

mattrunyon (dismissed):

packages/iris-grid/src/IrisGridTreeTableModel.test.ts
```diff
@@ -6,15 +6,21 @@ const irisGridTestUtils = new IrisGridTestUtils(dh);
...
+    name: '__DH_UI_GROUP__',
+    type: 'string',
   });
   const columns = irisGridTestUtils.makeColumns();
 
   test.each([
-    [0, columns, columns],
-    [1, columns, columns],
+    [1, columns, [...columns]],
+    [1, columns, [...columns]],
```
mofojed, line 23: This shouldn't have been changed

mofojed (changes requested):

mofojed (approved):

mofojed (changes requested):
Need to fix test case

mofojed (approved):

PR #2186, opened 2024-08-12
fix: makeEventFunctions take an array of parameters

mofojed:
- Previously every emit required exactly one parameter, which is not the case
- Allow zero or more parameters

### DIFF
--- a/packages/golden-layout/src/utils/EventUtils.ts
+++ b/packages/golden-layout/src/utils/EventUtils.ts
@@ -1,20 +1,23 @@
 import { useEffect } from 'react';
 import EventEmitter from './EventEmitter';
 
+export type EventHandlerFunction<P = []> = (
+  ...parameters: P extends unknown[] ? P : [P]
+) => void;
 export type EventListenerRemover = () => void;
-export type EventListenFunction<TParameters extends unknown[] = []> = (
+export type EventListenFunction<TParameters = []> = (
   eventEmitter: EventEmitter,
-  handler: (...parameters: TParameters) => void
+  handler: EventHandlerFunction<TParameters>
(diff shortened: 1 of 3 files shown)


mattrunyon (commented):
Not entirely sure where you'd want to put it or how to incorporate it, but you can use conditional types to support a non-array type as a single param, array/tuple

```ts
type MyFunc<P = []> = (...params: P extends unknown[] ? P : [P]) => void;
```

Also an interesting thing I saw was you can label tuples for argument signatures. See [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements) for more info, but the gist is you could do `makeEventFunctions<[count: number, title: string]>(...)` and TS should then show the params as `count` and `title` in autocomplete instead of `arg0` and `arg1`

packages/golden-layout/src/utils/EventUtils.ts
```diff
@@ -1,20 +1,23 @@
 import { useEffect } from 'react';
 import EventEmitter from './EventEmitter';
 
+export type EventHandlerFunction<P = []> = (
+  ...parameters: P extends unknown[] ? P : [P]
+) => void;
```
mattrunyon, line 6: I don't care about changing it since it's only in 2 places, but just something to note for future reference is you could turn the extends logic into a helper type so you don't need to reuse it

```ts
type AsArray<P> = P extends unknown[] ? P : [P];

export type EventHandlerFunction<P = []> = (...parameters: AsArray<P>) => void;
```

mattrunyon (dismissed):

mattrunyon (approved):

PR #2189, opened 2024-08-14
feat: Add shared objects to embed-widget

jnumainville:
Requires https://github.com/deephaven/deephaven-core/pull/5854

Adds ability to fetch shared objects with `embed-widget`. 
`name` takes the shared ticket, then `shared` specifies the type of shared object that needs to be fetched, which is required for `getSharedObject`.

### DIFF
--- a/packages/embed-widget/src/App.tsx
+++ b/packages/embed-widget/src/App.tsx
@@ -63,12 +63,16 @@ const LAYOUT_SETTINGS = {
 function App(): JSX.Element {
   const [error, setError] = useState<string>();
   const [definition, setDefinition] = useState<dh.ide.VariableDefinition>();
+  const [sharedObject, setSharedObject] = useState<any>();
   const searchParams = useMemo(
     () => new URLSearchParams(window.location.search),
     []
   );
   // Get the widget name from the query param `name`.
   const name = searchParams.get('name');
+  // Get the type of shared object from the query param,
+  // which is necessary if the widget is shared.
+  const shared = searchParams.get('shared');
   const api = useApi();
   const connection = useConnection();
   const client = useClient();
@@ -115,13 +119,26 @@ function App(): JSX.Element {
             )
           );
 
-          log.debug(`Loading widget definition for ${name}...`);
+          if (shared != null) {
+            log.debug(`Loading shared object for ${name}...`);
 
-          const newDefinition = await fetchVariableDefinition(connection, name);
+            const obj = await connection.getSharedObject(name, shared);
 
-          setDefinition(newDefinition);
+            setSharedObject(obj);
 
-          log.debug(`Widget definition successfully loaded for ${name}`);
+            log.debug(`Shared object successfully loaded for ${name}`);
+          } else {
+            log.debug(`Loading widget definition for ${name}...`);
+
+            const newDefinition = await fetchVariableDefinition(
+              connection,
+              name
+            );
+
+            setDefinition(newDefinition);
+
+            log.debug(`Widget definition successfully loaded for ${name}`);
+          }
         } catch (e: unknown) {
           log.error(`Unable to load widget definition for ${name}`, e);
           setError(`${e}`);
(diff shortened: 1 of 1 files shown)


packages/embed-widget/src/App.tsx
```diff
@@ -63,12 +63,17 @@ const LAYOUT_SETTINGS = {
 function App(): JSX.Element {
   const [error, setError] = useState<string>();
   const [definition, setDefinition] = useState<dh.ide.VariableDefinition>();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const [sharedObject, setSharedObject] = useState<Promise<any>>();
```
mofojed, line 67: Use `unknown` instead of `any`.
```suggestion
  const [sharedObject, setSharedObject] = useState<Promise<unknown>>();
```

packages/embed-widget/src/App.tsx
```diff
@@ -115,34 +120,57 @@ function App(): JSX.Element {
             )
           );
 
-          log.debug(`Loading widget definition for ${name}...`);
+          if (shared != null) {
+            log.debug(`Loading shared object for ${name}...`);
 
-          const newDefinition = await fetchVariableDefinition(connection, name);
+            const obj = await connection.getSharedObject(name, shared);
+            console.log(typeof obj);
```
mofojed, line 127:
```suggestion
```

packages/embed-widget/src/App.tsx
```diff
@@ -115,34 +120,57 @@ function App(): JSX.Element {
             )
           );
 
-          log.debug(`Loading widget definition for ${name}...`);
+          if (shared != null) {
+            log.debug(`Loading shared object for ${name}...`);
 
-          const newDefinition = await fetchVariableDefinition(connection, name);
+            const obj = await connection.getSharedObject(name, shared);
```
mofojed, line 126: We'll need the types published to make use of this...
One thing you can do so that this doesn't break is create our own type predicate that would do this (like we did with `isPartitionedTableWithBaseTable`), e.g.

```typescript
type ConnectionWithGetSharedObject = dh.IdeConnection & {
  getSharedObject(name: string, ticket: string): Promise<unknown>;
}

function isConnectionWithGetSharedObject(connection: dh.IdeConnection): connection is ConnectionWithGetSharedObject {
  return (
    'getSharedObject' in connection &&
    typeof connection.getSharedObject === 'function'
  );
}
```

Then you can do a check before calling it, or throw if that type doesn't exist.

mofojed (changes requested):
Need to clean up a few things.

packages/embed-widget/src/App.tsx
```diff
@@ -148,29 +149,57 @@ function App(): JSX.Element {
...
+      user,
+      isShared,
+      type,
+    ]
   );
 
   const isLoaded =
-    (definition != null || sharedObject != null) && error == null;
-  const isLoading = definition == null && sharedObject == null && error == null;
+    (definition != null || sharedReady != false) && error == null;
```
mofojed, line 166:
```suggestion
    (definition != null || sharedReady) && error == null;
```

packages/embed-widget/src/App.tsx
```diff
@@ -148,29 +149,57 @@ function App(): JSX.Element {
...
+  const isLoading = definition == null && sharedReady == false && error == null;
```
mofojed, line 167:
```suggestion
  const isLoading = definition == null && !sharedReady && error == null;
```

packages/embed-widget/src/App.tsx
```diff
@@ -148,29 +149,57 @@ function App(): JSX.Element {
...
+  };
+
+  function isConnectionWithGetSharedObject(
+    connection: dh.IdeConnection
+  ): connection is ConnectionWithGetSharedObject {
+    return (
+      'getSharedObject' in connection &&
+      typeof connection.getSharedObject === 'function'
+    );
+  }
```
mofojed, line 180: Move the type definition and typePredicate outside of/above this functional component.

packages/embed-widget/src/App.tsx
```diff
@@ -148,29 +149,57 @@ function App(): JSX.Element {
...
-        return sharedObject;
+      if (name != null && type != null) {
+        if (isConnectionWithGetSharedObject(connection)) {
+          return connection.getSharedObject(name, type);
+        }
+        throw new Error(
+          'Connection does not have getSharedObject method. Cannot fetch shared object.'
+        );
       }
       if (definition != null) {
```
mofojed, line 197: I'd personally put the `definition != null` case first.

mofojed (changes requested):

PR #2190, opened 2024-08-14
feat: Deephaven UI table databar support

mattrunyon:
Changes needed for dh.ui databars

- Support for extracting data in viewport columns not listed in `table.columns`. This allows us to hide the aggregation columns for databars from the user but still get their data. If a column is not in `table.columns` then it will be added to viewport data via its name instead of model index
- Improved gradient rendering to be much more efficient
- Fixed opacity being ignored if a gradient was used
- Fixed some of the databar spacing (was 1px off center)
- Modified `colorValueStyle` to have overrides so if we pass a string we don't get back `string | undefined` when we know we'll get a string
- Modified `resolveCssVariablesInRecord` to always resolve CSS colors regardless of if they contain a variable


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1169,13 +1169,17 @@ class IrisGridTableModelTemplate<
...
   extractViewportRow(row: DhType.Row, columns: DhType.Column[]): R {
-    const data = new Map<ModelIndex, CellData>();
+    const data = new Map<ModelIndex | ColumnName, CellData>();
     for (let c = 0; c < columns.length; c += 1) {
       const column = columns[c];
 
       const index = this.getColumnIndexByName(column.name);
-      assertNotNull(index);
-      data.set(index, {
+      // console.log(column.name, index);
```
AkshatJawne, line 1177: Stale comment?

packages/components/src/theme/ThemeUtils.test.ts
```diff
@@ -558,9 +558,10 @@ describe.each([undefined, document.createElement('div')])(
         bbb: 'bbb',
       };
 
+      jest.spyOn(window.CSS, 'supports').mockReturnValue(false);
```
mofojed, line 561: Test case for when this is true?

packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -103,166 +137,99 @@ class DataBarCellRenderer extends CellRenderer {
       dataBarWidth,
     } = this.getDataBarRenderMetrics(context, state, column, row);
 
-    if (this.heightOfDigits === undefined) {
-      const { actualBoundingBoxAscent, actualBoundingBoxDescent } =
-        context.measureText('1234567890');
-      this.heightOfDigits = actualBoundingBoxAscent + actualBoundingBoxDescent;
-    }
-
     context.save();
```
mofojed, line 140: With the levels of context.saves(), it'd be nice if they were at the start/end of a block, extracting this info out or just enclosing it in blocks. Maybe even have our own function for it?

```typescript
function withContextState(context, callback: () => undefined) {
  context.save();
  callback();
  context.restore();
}
```

Just musing though. Not necessary to change, but I did find it a little confusing seeing two `context.restore()`s close to each other below, and then tracing it back to the matching `context.save()`

packages/iris-grid/src/CommonTypes.tsx
```diff
@@ -59,7 +59,12 @@ export type InputFilter = {
 };
 
 export interface UIRow {
-  data: Map<ModelIndex, CellData>;
+  /**
+   * The data in the row indexed by column number.
+   * If a column is not part of the columns array (i.e. it's hidden by the model/table),
+   * then it will be included by its name instead of index.
+   */
+  data: Map<ModelIndex | ColumnName, CellData>;
```
mofojed, line 67: This shouldn't be necessary? I think it should map to the `dataBarOptionsForCell` already no?

mattrunyon: No. This is part of `model.viewportData`.

The problem is when we subscribe to columns and the UITableModel in plugins adds the auto-agg columns to the subscription, but hides them from `model.columns`. In the `extractViewportRow` method, we expect the column to exist in `model.columns` because we index based on our model index.

For the auto-agg columns which are hidden from `model.columns`, but added to the subscription, these show up in `extractViewportRow`, but don't have a model index associated with the column.

So this change is so subscribing to columns that aren't part of `model.columns` works without throwing.

The value is then read by column name from the model `viewportData` in the plugin

packages/grid/src/DataBarCellRenderer.ts
mattrunyon: Ya that would be neat and I agree it is a bit confusing. It's because you have to clip first and then draw, so 

I added the 2nd context save/restore because I needed it for re-using the gradient across multiple parts of the grid (or even multiple grids since I made the cache static on the renderer)

I think I might be able to simplify some of it and at least put both restores near each other by using the flip/translate for everything based on LTR or RTL. Basically we don't need to calculate special positioning based on if it's LTR or RTL, but instead calculate just based on LTR. Then if it's RTL, the context `scale(-1, 1)` w/ translation makes it draw RTL even though the coordinates are LTR. I already did this w/ the gradient, but could also add it for the markers and axis

packages/iris-grid/src/CommonTypes.tsx
mofojed: Ah yea - `extractViewportRow` should be using `this.table.columns` instead of `this.columns`, then I think it works out.
The viewport data should only be accessed internally from that model, so the access should always match up with the table columns I believe.

mattrunyon: I'll try moving the proxy off of the table and onto the model in the plugin for `columns`. `model.columns` just returns `this.table.columns` by default.

My concern with that was somewhere not using `model.columns` and instead using `table.columns` and causing some breakage

Just glancing through `IrisGridTableModelTemplate`, `get columnMap` accesses `this.table.columns` instead of `this.columns`. It's something that's hard to guard against

mattrunyon: And using `this.table.columns` in `extractViewportRow` might not fix it. It already iterates the columns in the viewport data and then we lookup the index with `this.getColumnIndexByName`. So if the model hides the column from `this.columns`, then it still won't have an index for aggregations when extracting the viewport row.

mattrunyon: So do you want me to change anything here? I can get rid of `get columnMap`

I think the proper way (as a separate ticket) would be to move away from the notion of a model index and just to names or visible index. We already do this for some things (like sorts seem to be serialized by name), so I'm not sure where model index should be used over the column name.

mofojed (approved):

packages/components/src/theme/ThemeUtils.test.ts
mattrunyon: Should be pretty much all the other test cases. We mocked it to true by default

mofojed (approved):

PR #2191, opened 2024-08-15
refactor: Change TabEvent to object literal, add TabEventMap

vbabich:
Add TabEventMap associating TabEvent types to callback signatures.
Change TabEvent to object literal so it could be used to define TabEventMap.

BREAKING CHANGE: Delete unused event types: `openPQObject`, `openControl`, `reload`, `clearAllFilters`.

### DIFF
--- a/packages/dashboard-core-plugins/src/events/TabEvent.ts
+++ b/packages/dashboard-core-plugins/src/events/TabEvent.ts
@@ -1,15 +1,6 @@
-class TabEvent {
-  static focus = 'TabEvent.focus';
-
-  static blur = 'TabEvent.blur';
-
-  static openPQObject = 'TabEvent.openPQObject';
-
-  static openControl = 'TabEvent.openControl';
-
-  static reload = 'TabEvent.reload';
-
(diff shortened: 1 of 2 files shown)


packages/dashboard-core-plugins/src/events/TabEvent.ts
```diff
@@ -1,15 +1,6 @@
...
-  static openControl = 'TabEvent.openControl';
-
-  static reload = 'TabEvent.reload';
-
-  static clearAllFilters = 'TabEvent.clearAllFilters';
-}
+const TabEvent = {
+  focus: 'TabEvent.focus',
+  blur: 'TabEvent.blur',
+} as const;
```
mofojed, line 4: Instead use `Object.freeze`:
```suggestion
const TabEvent = Object.freeze({
  focus: 'TabEvent.focus',
  blur: 'TabEvent.blur',
});
```
image
It's effectively the same thing, and also ensure the object cannot be modified.

mofojed (changes requested):

mofojed (approved):

PR #2198, opened 2024-08-20
refactor: Remove @adobe/react-spectrum imports where prohibited

### LINKED ISSUE #2195: Fix linting errors for importing from @adobe/react-spectrum
When completing #1908, we forgot to run the linter on all files locally since CI only lints changed files. As a result, there are some places that violate the rule and fail linting, but weren't caught because the files weren't changed.

I think we should expand the rule to allow `@adobe/react-spectrum` anywhere inside `packages/components`. It's unclear to me what `components/src/spectrum` actually is. We modify and re-export `ComboBox` for example, but then have `SearchableComboBox` in `components/src`. What makes the modified and re-exported `ComboBox` belong in the spectrum folder? I thought they should all be pure re-exports

mattrunyon:
Fixes #2195 

Had to just ignore some of the errors because we compare spectrum components in the styleguide and I didn't want to mess with that for now.

I still think it's a bit confusing that there are things in `components/src/spectrum` that aren't pure re-exports (e.g. ComboBox). Makes it confusing to know which to use.

### DIFF
--- a/packages/components/src/dialogs/ConfirmationDialog.tsx
+++ b/packages/components/src/dialogs/ConfirmationDialog.tsx
@@ -1,16 +1,16 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import { ReactElement, ReactNode } from 'react';
+import type { SpectrumLabelableProps } from '@react-types/shared';
+import { useFormWithDetachedSubmitButton } from '@deephaven/react-hooks';
 import {
-  Button,
+  SpectrumButton,
   ButtonGroup,
   Content,
   Dialog,
   Divider,
   Form,
-} from '@adobe/react-spectrum';
-import type { SpectrumLabelableProps } from '@react-types/shared';
-import { useFormWithDetachedSubmitButton } from '@deephaven/react-hooks';
-import { Heading } from '../spectrum';
+  Heading,
+} from '../spectrum';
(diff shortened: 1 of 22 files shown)


packages/components/src/dialogs/ConfirmationDialog.tsx
```diff
@@ -1,16 +1,16 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import { ReactElement, ReactNode } from 'react';
+import type { SpectrumLabelableProps } from '@react-types/shared';
```
bmingles, line 3: We may want to include @react-types/shared in the exclusion rule, but I haven't thought through it.

mattrunyon: It creates a circular dependency in some cases. I tried removing it from `@deephaven/react-hooks` and re-exporting from `@deephaven/components`. But `@deephaven/components` already relies on `@deephaven/react-hooks`, so it creates a circular dependency

bmingles: Talked to @mofojed on this. I think this will be fine as-is.

bmingles (approved):
LGTM. I'd just double check ACL Editor works and nothing weird in things like ConfirmationDialog

PR #2204, opened 2024-08-28
feat: Set selected theme via query string param

### LINKED ISSUE #2203: Allow setting theme via query string parameter
Embed widget doesn't expose the theme selector. It would be nice as a consumer to be able to pass the desired theme key via a query string parameter. This will enable things like the vscode extension to pick a default dh theme that matches the vscode theme ([Theme Selection / Configuration #37](https://github.com/deephaven/vscode-deephaven/issues/37))

bmingles:
Allow setting selected theme via a `themeKey` query string param.

Can test by passing `themeKey=default-light` and `themeKey=default-dark` as query string params.

resolves #2203

### DIFF
--- a/packages/components/src/theme/ThemeModel.ts
+++ b/packages/components/src/theme/ThemeModel.ts
@@ -43,7 +43,7 @@ export type ThemeIconsRequiringManualColorChanges =
 
 export const DEFAULT_DARK_THEME_KEY = 'default-dark' satisfies BaseThemeKey;
 export const DEFAULT_LIGHT_THEME_KEY = 'default-light' satisfies BaseThemeKey;
-export const THEME_KEY_OVERRIDE_QUERY_PARAM = 'themeKey';
+export const THEME_KEY_OVERRIDE_QUERY_PARAM = 'theme';
 
 // Hex versions of some of the default dark theme color palette needed for
 // preload defaults.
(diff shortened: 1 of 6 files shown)


packages/components/src/theme/ThemeModel.ts
```diff
@@ -43,6 +43,7 @@ export type ThemeIconsRequiringManualColorChanges =
 
 export const DEFAULT_DARK_THEME_KEY = 'default-dark' satisfies BaseThemeKey;
 export const DEFAULT_LIGHT_THEME_KEY = 'default-light' satisfies BaseThemeKey;
+export const THEME_KEY_OVERRIDE_QUERY_PARAM = 'themeKey';
```
mofojed, line 46: Should we bother with the `Key` part of this param? May as well just call it `theme`.

packages/components/src/theme/ThemeProvider.tsx
```diff
@@ -47,8 +48,13 @@ export function ThemeProvider({
...
   const [value, setValue] = useState<ThemeContextValue | null>(null);
 
+  const themeKeyOverride = useMemo(() => getThemeKeyOverride(), []);
+
   const [selectedThemeKey, setSelectedThemeKey] = useState<string>(
-    () => getThemePreloadData()?.themeKey ?? DEFAULT_DARK_THEME_KEY
+    () =>
+      themeKeyOverride ??
+      getThemePreloadData()?.themeKey ??
+      DEFAULT_DARK_THEME_KEY
```
mofojed, line 57: May as well wrap this all in a function, `getDefaultTheme` or `getDefaultSelectedTheme`. Then this just becomes.
```
... useState<string>(getDefaultSelectedTheme)
```

packages/components/src/theme/ThemeModel.ts
mofojed: Didn't see a response to this - why not just call it `theme`?

bmingles: ah, didn't see this comment. I have renamed it.

mofojed (approved):

PR #2212, opened 2024-09-09
feat: checkbox_group re-export

### LINKED ISSUE #2211: checkbox_group re-export

AkshatJawne:
Resolves #2211 

Needed for changes in plugins PR: https://github.com/deephaven/deephaven-plugins/pull/813


packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children,
+  ...props
+}: SpectrumCheckboxGroupProps): JSX.Element {
+  const wrappedChildren = useMemo(
+    () =>
+      React.Children.map(children, child => {
+        if (typeof child === 'function') {
+          return child;
+        }
+        return <Checkbox>{child}</Checkbox>;
```
bmingles, line 22: This has issues. I tried this in the styleguide:

```typescript
<CheckboxGroup>
       <SpectrumCheckbox>Aaa</SpectrumCheckbox>
</CheckboxGroup>
```

And it shows a nested checkbox:
image

Are there actually cases where the child would be a `function`?

I would recommend putting some examples in the styleguide to prove the different child types work (I did this with ListView and Picker)

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+      }),
+    [children]
+  );
+
+  return (
+    <SpectrumCheckboxGroup
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...props}
+      // eslint-disable-next-line react/no-children-prop
+      children={wrappedChildren}
```
bmingles, line 32: You should be able to get rid of the eslint exclusions and the children prop and do

```typescript
<SpectrumCheckboxGroup {...props}>
{wrappedChildren}
</SpectrumCheckboxGroup>

File: packages/components/src/spectrum/CheckboxGroup.tsx [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+    />
+  );
+}
+
+CheckboxGroup.displayName = 'CheckboxGroup';
```
bmingles, line 37: I could be mistaken, but I don't think we need `displayName` for plain function components. There are cases with forwardRef and probably some others where we do, but I think the function name works as expected.

bmingles (changes requested):
Left some comments

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,34 @@
...
+  children,
+  ...props
+}: SpectrumCheckboxGroupProps): JSX.Element {
+  const wrappedChildren = useMemo(
+    () =>
+      React.Children.map(children, child => {
+        if (isElementOfType(child, Checkbox)) {
+          return child;
+        }
+        return <Checkbox checked={false}>{child}</Checkbox>;
```
bmingles, line 23: I think this needs to be the Spectrum CheckBox

bmingles: Also, I'd still highly recommend trying some examples in the Styleguide. As-is, I don't think it actually accepts primitive children.

bmingles (changes requested):
Left some comments

AkshatJawne:
Cannot verify in styleguide as labels are currently blank due to latest core change @bmingles , will update and add examples as soon as that is fixed

bmingles:
> Cannot verify in styleguide as labels are currently blank due to latest core change @bmingles , will update and add examples as soon as that is fixed

@AkshatJawne I don't think those are related. The labels that are missing are in the ListView right?

AkshatJawne:
Screenshot 2024-09-12 at 9 11 20 AM
@bmingles Getting the above when I am running the following snippet, which I believe is related, since there is no reason for the label's to disappear.

```
from deephaven import ui
my_radio_group_basic_with_value = ui.radio_group(
    ui.radio("Dogs", value="dogs"),
    ui.radio("Cats", value="cats"),
    ui.radio("Lizard", value="lizard"),
    ui.radio("Fish", value="fish"),
    label="What is the best type of pet",
    on_change=handle_on_change
)
```

packages/code-studio/src/styleguide/CheckboxGroups.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { Checkbox, CheckboxGroup, Flex, Text } from '@deephaven/components';
```
bmingles, line 2: I think this is the incorrect `Checkbox` component. Need to use `SpectrumCheckbox`

packages/code-studio/src/styleguide/CheckboxGroups.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <Text>Single Child</Text>
+          <CheckboxGroup aria-label="Single Child">
+            <Checkbox checked={false}>Aaa</Checkbox>
+          </CheckboxGroup>
+        </Flex>
+
+        <Flex direction="column">
+          <Text>Multiple Children</Text>
+          <CheckboxGroup aria-label="Multiple Children">
+            <Checkbox checked={false}>Aaa</Checkbox>
```
bmingles, line 20: If you use the `SpectrumCheckbox`, you shouldn't need to set `checked` for uncontrolled component.

packages/code-studio/src/styleguide/CheckboxGroups.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        <Flex direction="column">
+          <Text>Mixed Children Types</Text>
+          <CheckboxGroup aria-label="Mixed Children Types">
+            {/* eslint-disable react/jsx-curly-brace-presence */}
+            {'String 1'}
+            {'String 2'}
+            {444}
+            {999}
+            {true}
```
bmingles, line 34: Looking at the styleguide, I think there's a bug with handling of `true` / `false`

image

There's probably a stringify that needs to happen somewhere

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,59 @@
+import { isElementOfType } from '@deephaven/react-hooks';
+import React, { ReactNode, useMemo, useState } from 'react';
+import {
+  CheckboxGroup as SpectrumCheckboxGroup,
+  SpectrumCheckboxGroupProps,
+} from '@adobe/react-spectrum';
+import Checkbox from '../Checkbox';
```
bmingles, line 7: I think this needs to be the `SpectrumCheckbox`

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Augmented version of the Spectrum CheckboxGroup component that supports
+ * primitive item children.
+ */
+export function CheckboxGroup({
+  children,
+  ...props
+}: CheckboxGroupProps): JSX.Element {
+  const [checkedState, setCheckedState] = useState<{ [key: number]: boolean }>(
```
bmingles, line 21: Why do we need to store checkbox state?

bmingles (changes requested):
Left some comments.

packages/components/src/spectrum/CheckboxGroup.tsx
AkshatJawne: When I use the DH Checkbox, there is this weird effect where I had to store state, since when I was checking one of the checkbox's in the group, all of the checkboxes were going into a checked state.

packages/code-studio/src/styleguide/CheckboxGroups.tsx
AkshatJawne: When I wrap my children in wrappedChildren, I pass in `String(child) `into the checkbox, which seems to work for the other primitives. Is there any other way I can stringify?

packages/components/src/spectrum/CheckboxGroup.tsx
bmingles: So is this unnecessary if you switch to `SpectrumCheckbox` ?

AkshatJawne:
Ironed out comments, not sure why the String(child) is resulting in booleans resulting in null showing up as the text, @bmingles any ideas?

packages/code-studio/src/styleguide/CheckboxGroups.tsx
bmingles: I think the issue here is that `React.Children.map` converts `true` / `false` to null.

bmingles: You might try `ensureArray` from @deephaven/utils instead

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,61 @@
...
+      ...prevState,
+      [index]: !prevState[index],
+    }));
+  };
+
+  const wrappedChildren = useMemo(
+    () =>
+      React.Children.map(children, (child, index) => {
+        if (isElementOfType(child, Checkbox)) {
+          return React.cloneElement(child, {
```
bmingles, line 36: There shouldn't be a need to clone or set any props if a Checkbox was provided. In cases where the primitive is provided, should only have to set value and children props. The CheckboxGroup does the rest

bmingles (changes requested):
See inline comments

packages/code-studio/src/styleguide/CheckboxGroups.tsx
bmingles: Actually, you may still need React.Children.map to deal with `key` props, but you'll need to convert booleans to strings before passing into it.

bmingles: I think something like this will work:
```
const wrappedChildren = useMemo(
  () =>
    ensureArray(children).map(child =>
      isElementOfType(child, Checkbox) ? (
        child
      ) : (
        <Checkbox key={String(child)} value={String(child)}>
          {String(child)}
        </Checkbox>
      )
    ),
  [children]
);
```

packages/components/src/spectrum/CheckboxGroup.tsx
bmingles: I think something like this will work:
```typescript
const wrappedChildren = useMemo(
  () =>
    ensureArray(children).map(child =>
      isElementOfType(child, Checkbox) ? (
        child
      ) : (
        <Checkbox key={String(child)} value={String(child)}>
          {String(child)}
        </Checkbox>
      )
    ),
  [children]
);
```

AkshatJawne:
Should be good now, I still had to clone element, or else when we passed in Checkbox children, then the value would not be assigned properly, and thus selecting one would select all of them. @bmingles

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -18,38 +20,24 @@ export function CheckboxGroup({
...
-            value: `checkbox-${index}`,
-            onChange: () => handleCheckboxChange(index),
-          });
-        }
-        return (
+      ensureArray(children).map((child, index) =>
+        isElementOfType(child, Checkbox) ? (
+          React.cloneElement(child, {
+            key: `${index}-${String(child)}`,
+            value: `${index}-${String(child)}`,
```
bmingles, line 29: Why does `value` include `index`?

bmingles: Actually, we shouldn't be cloning here either. <Checkbox/> children passed in need to provide their own values (see Spectrum examples)

packages/components/src/spectrum/CheckboxGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+          })
+        ) : (
+          <Checkbox
+            key={`${index}-${String(child)}`}
+            value={`${index}-${String(child)}`}
```
bmingles, line 34: Why add `index` to value? This is going to impact selected values passed to `onChange`

bmingles: Also, I don't think we need `index` for the key since checkbox values would be expected to be unique already. eslint actually complained for me when I tried adding an index in my testing.

bmingles:
I think the unit test failures are at least in part to needing a rebase / merge on latest main.

bmingles:
@AkshatJawne Code looks good. I think e2e tests may need to be updated.

AkshatJawne:
Attempting to update snapshots right now, running into some difficulty, will update them soon hopefully.

bmingles (approved):
LGTM

mofojed (approved):

PR #2217, opened 2024-09-11
fix: Use correct offset in snapshot

mofojed:
- Was using `row.offsetInSnapshot`, which was a "protected" API and recently removed from the JS API
- API was removed in JS API refactoring: https://github.com/deephaven/deephaven-core/pull/5890
- Instead just use the viewport `offset` and add the index of the row in the snapshot, which is there in both versions of the API
  - New API does support `row.index`, but this way is compatible with both
- Updated unit tests
- Tested using a deephaven.ui.list_view:
```python
from deephaven import time_table, ui
import datetime

initial_row_count = 200
column_types = time_table(
    "PT1S",
    start_time=datetime.datetime.now() - datetime.timedelta(seconds=initial_row_count),
).update(
    [
        "Id=new Integer(i)",
        "Display=new String(`Display `+i)",
    ]
)

@ui.component
def ui_list_view_table():
    value, set_value = ui.use_state([])

    lv = ui.list_view(
        column_types,
        aria_label="List View",
        on_change=set_value,
        selected_keys=value,
    )

    text = ui.text("Selection: " + ", ".join(map(str, value)))

    return ui.flex(
        lv,
        text,
        direction="column",
        margin=10,
        gap=10,
        width=500,
        # necessary to avoid overflowing container height
        min_height=0,
    )

lv_table = ui_list_view_table()
```

### DIFF
--- a/packages/jsapi-components/src/useViewportData.test.tsx
+++ b/packages/jsapi-components/src/useViewportData.test.tsx
@@ -3,7 +3,6 @@ import type { dh as DhType } from '@deephaven/jsapi-types';
 import dh from '@deephaven/jsapi-shim';
 import {
   OnTableUpdatedEvent,
-  ViewportRow,
   generateEmptyKeyedItems,
   isClosed,
   ITEM_KEY_PREFIX,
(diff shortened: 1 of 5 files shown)


dgodinez-dh (approved):

mofojed:
e2e tests are failing due to #2218 
Forcing this one to merge as it fixes an issue that is affecting the e2e tests for deephaven-plugins.

PR #2226, opened 2024-09-17
build: Upgrade to TypeScript 5.6.2

### LINKED ISSUE #2222: Upgrade TypeScript version
We should be using the latest (5.6)

bmingles:
Upgraded to TypeScript 5.6.2

resolves #2222


packages/redux/src/reducers/common/replaceByIdReducer.ts
```diff
@@ -12,7 +12,7 @@ export default function replaceByIdReducer<S extends Record<string, unknown>>(
   checkIfChanged = true
 ): Reducer<S> {
   return (state = initialState, action?) => {
-    switch (action.type) {
+    switch (action?.type) {
```
mattrunyon, line 15: Something the TS upgrade caught that wasn't caught before?

mattrunyon (approved):

PR #2230, opened 2024-09-19
refactor: Added consistent-type-imports eslint rule and ran --fix

### LINKED ISSUE #2229: eslint rule: @typescript-eslint/consistent-type-imports
We should configure `@typescript-eslint/consistent-type-imports` to enforce `type` only imports when appropriate.

bmingles:
- Added `@typescript-eslint/consistent-type-imports` linting rule
- Added package script `lint:packages`
- Ran `npm run lint:packages -- --fix`

> Note: despite the crazy large number of files, most of these were changed automatically by eslint --fix, so should be able to skim provided all automated tests pass.

resolves #2229

BREAKING CHANGE: eslint rule will require type only imports where possible


mattrunyon:
Can you update the rule to prefer `inline-type-imports`? It should remove a bunch of the duplicated imports just for types. TS 4.5+ added so you can do `import { a, type b } from 'c';`. Might reduce the number of files too since we do this in a bunch of places

bmingles:
> Can you update the rule to prefer `inline-type-imports`? It should remove a bunch of the duplicated imports just for types. TS 4.5+ added so you can do `import { a, type b } from 'c';`. Might reduce the number of files too since we do this in a bunch of places

Done. Good call

dgodinez-dh (approved):

PR #2241, opened 2024-09-30
fix: Layout shifts when opening and closing panels from a fresh state

### LINKED ISSUE #1268: Layout shift when closing the last column in a row
**Description**

This only appears to happen when you close or move the last item in the right-most column in a row. There is a vertical layout shift. Also, the drop zones can be temporarily wrong. They show correctly, but if there are many items in the left column and you try to drag the item from the right column into the left, it will sometimes drop in the completely wrong place (I think this happens on page reload specifically)

**Steps to reproduce**

1. Create a multi-row layout. Put multiple columns in the last row (I think it works w/ middle row too)
2. Put only 1 item in the last column of the row
3. Close or drag the tab

**Expected results**

No layout shift

**Actual results**

Layout shift. If you drop the item back into its own column where it originally was and re-drag, no further layout shift occurs. Resizing and then dragging again causes layout shift

The shift looks like it's around the height of the header, but not always. Sometimes the shift is up and other times down.

https://user-images.githubusercontent.com/8619789/236583309-a1c4bd42-4b30-4ba6-9f65-967a661e5a03.mov

**Versions**

 - Web UI Version: 0.37.3

mattrunyon:
Fixes #1268

This mostly occurs when you have a relatively fresh state. There's a video in the ticket, but if you have just 2 rows then this is frequently triggered. Using incognito mode should give you a fresh state which is affected by this bug.

This fixes both the vertical shift when you open a new panel type (open a notebook from a fresh state or open a table from a fresh layout state) and when you close the last panel in a stack (like in the ticket)

### DIFF
--- a/packages/dashboard/src/layout/LayoutUtils.test.ts
+++ b/packages/dashboard/src/layout/LayoutUtils.test.ts
@@ -24,11 +24,20 @@ function makeContentItem(type = 'root'): Partial<ContentItem> {
         contentItems.splice(index, 1);
       }
     }),
+    replaceChild: jest.fn((oldChild, newChild) => {
+      const index = contentItems.indexOf(oldChild);
+      if (index >= 0) {
+        contentItems[index] = newChild;
+      }
+    }),
     isComponent: type === 'component',
     isColumn: type === 'column',
     isRow: type === 'row',
     isRoot: type === 'root',
     type,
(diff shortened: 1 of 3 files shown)


bmingles (approved):
Changes look good. I confirmed the layout no longer shifts.

PR #2246, opened 2024-10-03
fix: Change ruff errors to warnings and fix config saving

mattrunyon:
Changed all except syntax errors to warning underlines instead of error underline since they are not actual errors.

Also changed redux to not save if the config matches the default. This should fix an edge case where a user opens the settings editor just to see what the config is, then hits "Save". Previously, this would save the current defaults as their settings and then if the defaults changed they would not get updated settings. This way, if a user was on defaults and the default changes they will get the new default.

Also added a little bit of debug logging for Ruff.


packages/code-studio/src/settings/EditorSectionContent.tsx
```diff
@@ -51,15 +51,24 @@ export function EditorSectionContent(): JSX.Element {
     (newValue: Record<string, unknown>) => {
       dispatch(
         updateNotebookSettings({
-          python: { linter: { ...ruffSettings, config: newValue } },
+          python: {
+            linter: {
+              ...ruffSettings,
+              config:
+                JSON.stringify(newValue) ===
```
bmingles, line 58: Is the argument mismatch in the JSON.stringify calls intentional?

packages/code-studio/src/settings/EditorSectionContent.tsx
```diff
@@ -51,15 +51,24 @@ export function EditorSectionContent(): JSX.Element {
...
+          },
         })
       );
       MonacoProviders.setRuffSettings(newValue);
     },
     [dispatch, ruffSettings]
   );
 
-  const [val, setVal] = useState(JSON.stringify(ruffConfig, null, 2));
+  const val = JSON.stringify(ruffConfig, null, 2);
```
bmingles, line 71: Should we memoize this?

bmingles (changes requested):
Looks good overall. Left 2 suggestion / questions

packages/code-studio/src/settings/EditorSectionContent.tsx
mattrunyon: Yes. I want to check if the value was reset to default here, but the objects won't be the same. The other option was to do this check in the settings modal before onSave is called and then call with null or undefined to indicate the default.

packages/code-studio/src/settings/EditorSectionContent.tsx
mattrunyon: I don't think it matters. The object isn't that big and there's still some overhead to do the string comparison for memoization. Also the component shouldn't re-render often.

packages/code-studio/src/settings/EditorSectionContent.tsx
bmingles: Oh, I misread this. I thought your were passing in `undefined, 2` to one of the JSON.stringify calls

bmingles (approved):
LGTM

PR #2248, opened 2024-10-04
fix: improve color contrast of editor find in dark mode

dsmmcken:
User reported editor find contrast in dark mode was bad. Changed color for semantic editor highlight in dark mode to an orange similar to what monokai uses for find in vs code. Confirmed existing color in light mode was fine.

Before:
image

After:
image


bmingles (approved):
LGTM

Issue #2247, opened 2024-10-04
Markdown formatted autocomplete


wusteven815:
Autocomplete looks messy when the function signature is too long. The docstring should also be formatted, similar to how VS Code does it.

- Core PR: https://github.com/deephaven/deephaven-core/pull/6170
- Web Client UI PR - #2250    
- Server docker PR: https://github.com/deephaven/deephaven-server-docker/pull/122

PR #2250, opened 2024-10-07
feat: add monaco docs CSS override

wusteven815:
- Adds #2247 
- Adds a few CSS overrides to Monaco

### DIFF
--- a/packages/console/src/ConsoleInput.scss
+++ b/packages/console/src/ConsoleInput.scss
@@ -74,16 +74,3 @@ $console-input-margin-color: $foreground;
     }
   }
 }
-
-.docs {
-  min-width: 200px;
-  
-  hr {
-    background-color: $text-muted;
-    height: 1px;
-  }
-
-  blockquote {
-    margin-left: 20px;
-  }
-}
--- a/packages/console/src/ConsoleInput.scss
+++ b/packages/console/src/ConsoleInput.scss
@@ -74,3 +74,16 @@ $console-input-margin-color: $foreground;
     }
   }
 }
+
+.docs {
+  min-width: 200px;
+  
+  hr {
+    background-color: $text-muted;
+    height: 1px;
+  }
+
+  blockquote {
+    margin-left: 20px;
+  }
+}
--- a/packages/console/src/notebook/Editor.scss
+++ b/packages/console/src/notebook/Editor.scss
@@ -16,6 +16,7 @@
   }
 }
 
+// Override Monaco's default markdown popups (i.e signature help) to resemble VS Code closer
 .rendered-markdown.markdown-docs {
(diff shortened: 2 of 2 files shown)


wusteven815:
- `min-width` - it should have a min width to avoid looking thin
- `hr` - make horizontal breaks visible
- `blockquote` - make block quotes have a margin to make them indented

packages/console/src/ConsoleInput.scss
```diff
@@ -74,3 +74,16 @@ $console-input-margin-color: $foreground;
     }
   }
 }
+
+.docs {
```
dsmmcken, line 78: That's a pretty greedy selector, and not scoped. Where is ".docs" as the class defined?

I think Editor.css is probably a better place for this as well.

packages/console/src/ConsoleInput.scss
```diff
@@ -74,3 +74,16 @@ $console-input-margin-color: $foreground;
...
+  min-width: 200px;
```
dsmmcken, line 79:
```suggestion
  min-width: 12rem;
```

packages/console/src/ConsoleInput.scss
```diff
@@ -74,3 +74,16 @@ $console-input-margin-color: $foreground;
...
+  
+  hr {
+    background-color: $text-muted;
+    height: 1px;
+  }
+
+  blockquote {
+    margin-left: 20px;
```
dsmmcken, line 87:
```suggestion
    margin-left: $spacer-3;
```

Or spacer-2 whichever is closer.

dsmmcken (dismissed):

packages/console/src/notebook/Editor.scss
```diff
@@ -15,3 +15,16 @@
     color: $link-color;
   }
 }
+
+.rendered-markdown.markdown-docs {
```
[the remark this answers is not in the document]
wusteven815: Yes, this is the selector for single element with `.rendered-markdown` and `.markdown-docs` classes

dsmmcken (approved):

dsmmcken (changes requested):
Actually, one last thing. Comment what this is for. In future I could see a case of someone searching this codebase and not finding any matches for it so it could accidentally get deleted without knowing it originates elsewhere.

packages/console/src/notebook/Editor.scss
```diff
@@ -16,6 +16,7 @@
   }
 }
 
+// Override Monaco's default markdown popups (i.e signature help) to resemble VS Code closer
```
dsmmcken, line 19: Please specifically mention that this class is defined in core or wherever.

dsmmcken (approved):

PR #2255, opened 2024-10-15
feat: Add render blocking errors to Chart

jnumainville:
Adds new events that enable render-blocking overlays to the chart.
The current error pop-up is not sufficient because it still allows the chart to render. This is problematic in the case of webgl because disabling webgl means we don't want the chart rendering at all in case the computer can't handle it.

This is specifically built with https://github.com/deephaven/deephaven-plugins/issues/612 in mind.

Here is an example of the pop-up as implemented in https://github.com/deephaven/deephaven-plugins/pull/934
image

### DIFF
--- a/packages/chart/src/Chart.tsx
+++ b/packages/chart/src/Chart.tsx
@@ -88,7 +88,7 @@ interface ChartState {
   layout: Partial<Layout>;
   revision: number;
 
-  /** A message that blocks the chart from rendering */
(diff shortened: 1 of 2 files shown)


mattrunyon (commented):
Just leaving as comments instead of request changes since I'll be out tomorrow and Friday

mofojed (approved):

PR #2275, opened 2024-10-30
fix: console scrolls on 1st code block run

### LINKED ISSUE #2207: Console stuck to bottom does not stick to bottom on first code block run
**Description**

If you run a code block that will make the console scroll as the first command, it doesn't scroll to the bottom.

**Steps to reproduce**
1. Run `clear` in your console to clear the console
2. Run a large block of code such that it will cause the console to scroll

**Expected results**
2. Console should be scrolled to the bottom

**Actual results**
2. Scroll bar is at the top

[Screencast from 2024-08-30 03:39:09 PM.webm](https://github.com/user-attachments/assets/9afcf53b-ab71-406b-8b69-5b24c9f27886)

**Versions**
Engine Version: 0.37.0-SNAPSHOT
Web UI Version: 0.91.0
Java Version: 11.0.24
Barrage Version: 0.6.0
Browser Name: Chrome 128
OS Name: Linux
matplotlib: 0.0.0
plotly-express: 0.0.0
auth-keycloak: 0.0.0
ui: 0.0.0
example-theme: 0.0.0

wusteven815:
- Fixes #2207

### DIFF
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -401,6 +401,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
       return;
     }
 
+    this.scrollConsoleHistoryToBottom();
     this.updateHistory(result, newHistoryItem);
     this.updateKnownObjects(newHistoryItem);
     this.updateWorkspaceHistoryItem(


packages/console/src/Console.tsx
```diff
@@ -401,6 +401,7 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
       return;
     }
 
+    this.scrollConsoleHistoryToBottom();
```
mattrunyon, line 404: This will always scroll to the bottom after any command run. It should only scroll if it was already at the bottom I think. Unless we want to change that behavior, but would need Don to approve a new behavior

mattrunyon (changes requested):

packages/console/src/Console.tsx
mattrunyon: Actually it looks like we already scroll to bottom when a command starts, so this is probably fine

bmingles (approved):
Changes look good. I verified the behavior is fixed and it properly scrolls to the bottom when running a long script on clean console.

PR #2279, opened 2024-11-04
refactor: DH-17778: Move exportLogs to the log package

vbabich:
- Move `exportLogs` and `logInit` to the `log` package so they could be reused in Enterprise
- Remove `@deephaven/redux` and `@deephaven/jsapi-shim` dependencies from `LogExport.ts`
- Serialize Maps in redux data
- Unit tests for `getReduxDataString`


packages/code-studio/src/index.tsx
```diff
@@ -5,9 +5,12 @@ import { Provider } from 'react-redux';
 import { LoadingOverlay, preloadTheme } from '@deephaven/components';
 import { ApiBootstrap } from '@deephaven/jsapi-bootstrap';
 import { store } from '@deephaven/redux';
-import logInit from './log/LogInit';
+import { logInit } from '@deephaven/log';
 
-logInit();
+logInit(
+  parseInt(import.meta.env.VITE_LOG_LEVEL ?? '', 10),
```
bmingles, line 11: Is it ok that `''` parses to `NaN`?

packages/log/src/LogExport.ts
```diff
@@ -104,48 +103,61 @@ function makeSafeToStringify(
...
+ * @param blacklist List of JSON paths to blacklist in redux data. A JSON path is a list representing the path to that value (e.g. client.data would be `['client', 'data']`)
+ * @param fileNamePrefix The zip file name without the .zip extension. Ex: test will be saved as test.zip
  * @returns A promise that resolves successfully if the log archive is created and downloaded successfully, rejected if there's an error
  */
 export async function exportLogs(
-  fileNamePrefix = `${dh.i18n.DateTimeFormat.format(
-    FILENAME_DATE_FORMAT,
-    new Date()
-  )}_support_logs`,
+  logHistory: LogHistory,
```
bmingles, line 146: Should this be marked as a breaking change in the PR?

bmingles (changes requested):
Had 1 question + I think this needs to be marked as a breaking change due to `exportLogs` contract change. Otherwise looks good.

packages/code-studio/src/index.tsx
vbabich: This is existing code moved from `LogInit` - https://github.com/deephaven/web-client-ui/pull/2279/files#diff-8f738bb62eda99bdb9b608feb9a011431c9daef91788b6ad56166ff4d950ca9eL15
I don't know if this was the original intent, but `NaN` here results in a warning - 
```
if (!Number.isFinite(level)) {
      console.warn(
        `Expected a number for log level. Received: ${level}. Ignoring`
      );
      return;
    }
```

packages/log/src/LogExport.ts
vbabich: The `exportLogs` wasn't previously exposed. The `@deephaven/code-studio` package didn't export anything from `LogExport.ts`.

bmingles (approved):
Looks good

PR #2280, opened 2024-11-06
feat: Ruff updates for DHE support

mattrunyon:
Support for DH-17923 and also fixes some other edge cases with Ruff.

- Adds a hook for configuring Ruff and keeping it in sync w/ redux settings.
- Fixes a bug where if Ruff was enabled, but there was no user session then quick fixes were shown but linting underlines were not. Changed so Python is always linted if opened since it doesn't require a session to lint.
- Allow passing a default config to the Ruff config editor since DHE can set a custom config as the default via server props
- Added readOnly mode to the Ruff config editor
- Moved some formatting logic into MonacoUtils since DHE implements its own notebook panel. Eventually we should just consume the DHC NotebookPanel in DHE with extensions for PQs and other DHE options

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
@@ -15,7 +15,12 @@ import {
   type DropdownAction,
   LoadingOverlay,
 } from '@deephaven/components';
-import { ScriptEditor, ScriptEditorUtils, SHORTCUTS } from '@deephaven/console';
+import {
+  MonacoUtils,
+  ScriptEditor,
+  ScriptEditorUtils,
+  SHORTCUTS,
+} from '@deephaven/console';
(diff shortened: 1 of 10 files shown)


packages/dashboard-core-plugins/src/panels/NotebookPanel.tsx
```diff
@@ -1322,6 +1323,8 @@ class NotebookPanel extends Component<NotebookPanelProps, NotebookPanelState> {
 
     const portal = tab?.element.find('.lm_title_before').get(0);
 
+    console.log(this.notebook?.editor);
```
vbabich, line 1326: Must be a leftover from debugging

vbabich (changes requested):

vbabich (dismissed):

vbabich (approved):

PR #2281, opened 2024-11-06
fix: open file blocks logout

### LINKED ISSUE #1685: Open file blocks logout

wusteven815:
- Fixes #1685 
- Delays `AuthPluginPsk` remount so the broadcast message is received in time


wusteven815:
There seems to be a race condition where if a file is open, `AuthPluginPsk` gets destroyed before its logout event is received. This delays the state change that causes a remount, so that the logout event is reached in time.

PR #2286, opened 2024-11-08
feat: datetime tooltip to show full value

### LINKED ISSUE #614: Full values of DateTimes should be more accessible
It would be helpful if datetime columns had tooltips that showed the complete date and time:

image

wusteven815:
- Closes #614 
- Adds tooltip for datetime cells showing full value


packages/grid/src/mouse-handlers/index.ts
```diff
@@ -8,5 +8,5 @@ export { default as GridRowTreeMouseHandler } from './GridRowTreeMouseHandler';
 export { default as GridScrollBarCornerMouseHandler } from './GridScrollBarCornerMouseHandler';
 export { default as GridVerticalScrollBarMouseHandler } from './GridVerticalScrollBarMouseHandler';
 export { default as EditMouseHandler } from './EditMouseHandler';
-export { default as GridTokenMouseHandler } from './GridTokenMouseHandler';
+export { default as GridTooltipMouseHandler } from './GridTooltipMouseHandler';
```
mofojed, line 11: Technically this is a breaking change now, since we were exporting `GridTokenMouseHandler`. You could simply re-export `GridTooltipMouseHandler` as `GridTokenMouseHandler` in the `index.ts` so it's not a breaking change, along with a comment. Or just not bother renaming them, and just add a comment to the handler that it also helps with tooltips.
We want to avoid breaking changes when possible - we don't know if somebody is using this class or not.

packages/grid/src/GridModel.ts
```diff
@@ -71,6 +72,10 @@ abstract class GridModel<
    */
   abstract textForCell(column: ModelIndex, row: ModelIndex): string;
 
+  tooltipForCell(column: GridRangeIndex, row: GridRangeIndex): string | null {
+    return null;
+  }
```
mofojed, line 77: Add to `IrisGridModel` instead, this isn't actually used within `Grid` (and it doesn't display tooltips).

packages/iris-grid/src/mousehandlers/IrisGridTooltipMouseHandler.tsx
```diff
@@ -0,0 +1,132 @@
...
+      return { stopPropagation: false, preventDefault: false };
+    }
+    this.cursor = null;
+    return false;
+  }
+
+  onMove(gridPoint: GridPoint, grid: Grid): EventHandlerResult {
+    const { model } = this.irisGrid.props;
+    const isUserHoveringLink = this.isHoveringLink(gridPoint, grid);
+    const dateTooltip = model.tooltipForCell(gridPoint.column, gridPoint.row);
```
mofojed, line 43: Should just call this `tooltip`. Nothing here suggests it would be a date only, that's up to the model implementation details.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3976,13 +3976,13 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
   );
 
-  getLinkHoverTooltip = memoize(
+  getHoverTooltip = memoize(
     (linkHoverTooltipProps: CSSProperties): ReactNode => {
```
mofojed, line 3980: Rename to `hoverTooltipProps`

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3997,11 +3997,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       return (
         <div style={wrapperStyle}>
           <Tooltip options={popperOptions} ref={this.handleTooltipRef}>
-            <div className="link-hover-tooltip">
-              {linkHoverDisplayValue} - Click once to follow.
-              <br />
-              Click and hold to select this cell.
-            </div>
+            <div className="link-hover-tooltip">{hoverDisplayValue}</div>
```
mofojed, line 4000: I wonder if the flakiness is the tooltip appearing as we're moving the mouse around. Maybe moving the mouse somewhere outside of the grid instead of waiting inside the grid before taking a snapshot.

mofojed (dismissed):

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const { hoverTooltipProps } = this.irisGrid.state;
+      if (this.currentLinkBox == null) {
+        return false;
+      }
+      const { x1: left, y1: top, x2: right, y2: bottom } = this.currentLinkBox;
+      const { href } = this.currentLinkBox.token;
+      const width = right - left;
+      const height = bottom - top;
+      const newProps = { left, top: top + 1, width, height };
+      if (!deepEqual(hoverTooltipProps, newProps)) {
```
bmingles, line 60: All of the variables seem to only be used if `!deepEqual(hoverTooltipProps, newProps)` is true. The same seems true for the `else if (tooltip !== null)` block. I would suggest checking `deepEqual(hoverTooltipProps, newProps)` before all of this logic and bailing early if it is true. This should cut down on some nested if / else blocks.

bmingles:
```typescript
if (deepEqual(hoverTooltipProps, newProps)) {
  this.destroyTooltip();
}
else if (isUserHoveringLink &&
      this.currentLinkBox != null &&
      isLinkToken(this.currentLinkBox.token)) {
      ...
}
else if (tooltip !== null) {
   ...
}
```

bmingles (changes requested):
Left a suggestion to simplify nested if / else structure

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.tsx
wusteven815: `newProps` is different depending on whether it's the link or tooltip condition. Would it still be ideal to have some logic like:
```javascript
let tooltip = {};
if (
  isUserHoveringLink &&
  this.currentLinkBox != null &&
   isLinkToken(this.currentLinkBox.token)
) {
  tooltip = { ... };
} else if (tooltip !== null) {
  tooltip = { ... };
}
```

or would it be better to just do:
```javascript
if (deepEqual(hoverTooltipProps, newProps)) continue;
```

bmingles: Ok yeah I missed that before.

packages/iris-grid/src/mousehandlers/IrisGridTokenMouseHandler.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const isUserHoveringLink = this.isHoveringLink(gridPoint, grid);
+    const tooltip = model.tooltipForCell(gridPoint.column, gridPoint.row);
+
+    if (
+      isUserHoveringLink &&
+      this.currentLinkBox != null &&
+      isLinkToken(this.currentLinkBox.token)
+    ) {
+      const { hoverTooltipProps } = this.irisGrid.state;
+      if (this.currentLinkBox == null) {
```
bmingles, line 52: Can this if statement ever be true?

wusteven815: I think yes, in the following spots:

https://github.com/deephaven/web-client-ui/blob/36d910da75a0c628b9e45baf12282fb55e51e7e0/packages/grid/src/mouse-handlers/GridTokenMouseHandler.ts#L34
https://github.com/deephaven/web-client-ui/blob/36d910da75a0c628b9e45baf12282fb55e51e7e0/packages/grid/src/mouse-handlers/GridTokenMouseHandler.ts#L79

wusteven815: Ignore, missed the earlier condition check

PR #2303, opened 2024-11-26
fix: Update react-spectrum packages

mofojed:
- Fixes a bug with Meter not displaying it's variant correctly
- Needed to resolve issue with `ListData` now allowing `undefined` to be returned from `getItem`
  - Just throw an error in this case, which should be the same behaviour as before


dgodinez-dh (approved):

packages/react-hooks/src/useMappedSelection.ts
```diff
@@ -25,6 +25,7 @@ export function useMappedSelection<TItem, TValue>(
     );
 
     return {
+      // TODO: Need to handle this... not sure how yet
```
bmingles, line 28: Not sure what this means? Can you add any more details to the comment?

bmingles (changes requested):
Left 1 request to improve a TODO comment.

packages/react-hooks/src/useMappedSelection.ts
mofojed: Whoops, I had already handled it. Basically handled it by throwing if an `undefined` is returned from the `getItem` call.

PR #2304, opened 2024-11-27
refactor: Move Panel into @deephaven/dashboard

mofojed:
- Move Panel into @deephaven/dashboard, removing functionality depending on core plugins and keeping in CorePanel in the dashboard-core-plugins package
- Should be no breaking changes, as dashboard-core-plugins re-exports the components that were removed
- Allows plugins (such as deephaven.ui) to simply use the `Panel` component without having to depend on other core plugins
- Tests pass, ran a smoke test with latest version of deephaven.ui to ensure functionality still worked
- Tested from Enterprise as well against the latest G+, was able to open table panels and deephaven.ui panels without any changes to Enterprise code.

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/CorePanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/CorePanel.tsx
@@ -14,9 +14,7 @@ export type CorePanelProps = BasePanelProps & {
 };
 
 /**
- * Generic panel component that emits mount/unmount/focus events.
- * Also wires up some triggers for common events:
- * Focus, Resize, Show, Session open/close, client disconnect/reconnect.
+ * Extends the base panel component to handle Session events and input filter events.
  */
 class CorePanel extends PureComponent<CorePanelProps> {
   constructor(props: CorePanelProps) {


packages/dashboard-core-plugins/src/events/index.ts
```diff
@@ -5,4 +5,6 @@ export { default as IrisGridEvent } from './IrisGridEvent';
 export { default as MarkdownEvent } from './MarkdownEvent';
 export { default as NotebookEvent } from './NotebookEvent';
 export { default as PandasEvent } from './PandasEvent';
-export { default as TabEvent } from './TabEvent';
+
+// Deprecated - use TabEvent from @deephaven/dashboard
+export { type TabEvent } from '@deephaven/dashboard';
```
mattrunyon, line 10: You could do something like this so the deprecation actually shows up in vscode

```ts
import { type TabEvent as TabEventImport } from '@deephaven/dashboard';

/**
 * @deprecated Use TabEvent from `@deephaven/dashboard`
 */
export type TabEvent = typeof TabEventImport; // Not sure why typeof is needed here
```

This will give the deprecation strikethrough in VSCode

packages/dashboard-core-plugins/src/panels/index.ts
```diff
@@ -17,7 +17,9 @@ export { default as MarkdownPanel } from './MarkdownPanel';
 export { default as NotebookPanel } from './NotebookPanel';
 export { default as PandasPanel } from './PandasPanel';
 export * from './PandasPanel';
-export { default as Panel } from './Panel';
+export { default as CorePanel } from './CorePanel';
+// Deprecated - use CorePanel instead
+export { default as Panel } from './CorePanel';
```
mattrunyon, line 22: Same comment about importing and renaming for the deprecated export w/ a deprecated tag.

`CorePanel` is what `Panel` used to be right? Why would someone need/want `Panel` from `@deephaven/dashboard`? dh.ui would still need the dashboard-core-components version, right? Or does nothing there need the session open/close events?

packages/dashboard/src/Panel.tsx
```diff
@@ -83,18 +74,15 @@ interface PanelState {
  * Also wires up some triggers for common events:
  * Focus, Resize, Show, Session open/close, client disconnect/reconnect.
```
mattrunyon, line 75: Remove session open/close. And client disconnect/reconnect?

packages/dashboard-core-plugins/src/panels/index.ts
mofojed: deephaven.ui doesn't use the `onSessionClosed` or `onSessionOpened` props, it only needs the BasePanel

mattrunyon (dismissed):
I assume you've tested whatever we needed again since this is an old PR that looks like we forgot about. Looks like e2e and unit tests are passing after you updated from main.

I can't remember if we were holding off this PR for any reason or if it just fell off the radar.

mofojed:
@mattrunyon I retested and rebased this. Found one issue with TabEvent deprecated export in dashboard-core-plugins, should be resolved now with no breaking changes.

packages/dashboard-core-plugins/src/panels/CorePanel.tsx
```diff
@@ -0,0 +1,80 @@
...
+    { language, sessionId }: { language: string; sessionId: string }
+  ) => void;
+};
+
+/**
+ * Generic panel component that emits mount/unmount/focus events.
+ * Also wires up some triggers for common events:
+ * Focus, Resize, Show, Session open/close, client disconnect/reconnect.
+ */
+class CorePanel extends PureComponent<CorePanelProps> {
```
mattrunyon, line 21: ~~Doesn't handle focus, resize, show, or client disconnect/reconnect? Update comment or add those events~~ Nevermind I see this renders `BasePanel` inside of it

Also, just to clarify. "Core" Panel is talking about DH Core?
`CorePanel` is in `dashboard-core-plugins`. Could call `DHCPanel` or just `DHPanel`, but I'm indifferent
`Panel` is in `Dashboard` which seems to handle the other events mentioned in this comment

packages/dashboard/src/Panel.tsx
```diff
@@ -81,20 +72,17 @@ interface PanelState {
 /**
  * Generic panel component that emits mount/unmount/focus events.
  * Also wires up some triggers for common events:
- * Focus, Resize, Show, Session open/close, client disconnect/reconnect.
+ * Focus, Resize, Show
  */
-class Panel extends PureComponent<CorePanelProps, PanelState> {
-  constructor(props: CorePanelProps) {
+class Panel extends PureComponent<BasePanelProps, PanelState> {
```
mattrunyon, line 77: Should we call this class (and file) `BasePanel`?

packages/dashboard/src/index.ts
```diff
@@ -12,9 +12,16 @@ export { default as DashboardUtils } from './DashboardUtils';
 export * from './LazyDashboard';
 export * from './layout';
 export * from './redux';
+export {
+  type BasePanelProps,
+  default as Panel,
+  // Alias for Panel
+  default as BasePanel,
```
mattrunyon, line 19: Should these be swapped and `Panel` is an alias for `BasePanel`?

PR #2306, opened 2024-11-27
chore: Delete unused styles for console-disconnected-info

vbabich:
`.console-disconnected-info` isn't used in DHC. Added missing message color on Enterprise side in DH-17922.


packages/console/src/Console.scss
```diff
@@ -41,13 +41,4 @@
       flex-grow: 1;
     }
   }
-
-  .console-disconnected-info {
-    padding-top: $spacer-1;
-    .console-disconnected-info-message {
-      font-family: $font-family-monospace;
-      color: $danger;
-      font-size: ConsoleVariables.$font-size;
```
mofojed, line 50: Should be able to remove the `@use ConsoleVariables` import as well?

mofojed (approved):

PR #2307, opened 2024-12-05
chore: Update Playwright to v1.51

mofojed:
- Should be no breaking changes
- Fixes some issues when running locally on Ubuntu 24
- Has some other interesting features, such as the `--only-changed` cli option
- Added the `'tnum'` option back for grids (we were overriding that setting in our Playwright css file)


dgodinez-dh (approved):

dgodinez-dh (approved):

tests/context-menu.spec.ts-snapshots/advanced-filters-1-chromium-linux.png
mofojed, line 1: Kerning seems to be enabled, even though we should be disabling it in `font-kerning: none !important`. Is the Playwright.css not being loaded?

PR #2315, opened 2024-12-13
build: Fix e2e tests for builds on v0.78

mofojed:
- The edge version of deephaven-core installs the latest version of deephaven.ui, which is not compatible with this old version of the UI
- Just pin the e2e tests to an older build of core
- Disabled webkit - the tests were flakey, and updating them would require updating playwright which would require updating all snapshots and reverifying them


vbabich (approved):

mofojed:
Not sure why this is failing e2e. From [the logs](https://github.com/deephaven/web-client-ui/actions/runs/12323142537/job/34485171127?pr=2315):
```
[WebServer] @deephaven/code-studio: failed to load config from /home/runner/work/web-client-ui/web-client-ui/packages/code-studio/vite.config.ts
[WebServer] @deephaven/code-studio: error when starting preview server:
[WebServer] @deephaven/code-studio: TypeError [ERR_INVALID_ARG_TYPE]: The "paths[1]" argument must be of type string. Received undefined
```
Which seems to indicate that `VITE_BUILD_PATH` is `undefined`. Looking at the web-build logs attached to the [failed e2e tests](https://github.com/deephaven/web-client-ui/actions/runs/12323142537?pr=2315), it looks like the `.env` file that would set this does not exist, or any other "hidden" file.

vbabich (approved):

PR #2327, opened 2025-01-02
feat: Removed esbuild runtime dependency

bmingles:
DH-18191: Removed esbuild runtime dependency and replaced it with an optional `PostDownloadTransform` function. Files now get downloaded directly to the `serverStorageDir` instead of `target` subdirectory.

BREAKING CHANGES: @deephaven/jsapi-nodejs `loadModules` arguments has changed. Transformation of module types is now the responsibility of the consumer.

### DIFF
--- a/packages/jsapi-nodejs/src/loaderUtils.ts
+++ b/packages/jsapi-nodejs/src/loaderUtils.ts
@@ -1,6 +1,5 @@
 import fs from 'node:fs';
 import path from 'node:path';
-import esbuild, { type BuildOptions } from 'esbuild-wasm';
 
 import { downloadFromURL, urlToDirectoryName } from './serverUtils.js';
 import { polyfillWs } from './polyfillWs.js';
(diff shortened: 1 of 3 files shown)


packages/jsapi-nodejs/src/loaderUtils.ts
```diff
@@ -27,73 +27,54 @@ export type LoadModuleOptions = {
...
+        download(serverPaths[i], content)
+      );
+    }
+
     // Write to disk
-    const downloadPaths = getDownloadPaths(
-      needsTranspile ? sourceDir : targetDir,
-      serverPaths
-    );
+    const downloadPaths = getDownloadPaths(serverStorageDir, serverPaths);
```
mofojed, line 68: Should mention this change in the PR description as well - it's now downloading to the `serverStorageDir` instead of a `target` subdirectory.

mofojed (approved):

PR #2328, opened 2025-01-02
fix: ComboBox show all items on open

bmingles:
DH-18088 & DH-18087: ComboBox now clears search filter when opening the ComboBox unless it is triggered by user input.

### Testing
- Typing in a combobox should open it filtered by the search text
- Clicking the dropdown should open the CombBox unfiltered
- Typing while open should filter results
   > Note that there seems to be a Spectrum bug that can cause some weird scrolling behavior when typing in an opened ComboBox https://github.com/adobe/react-spectrum/issues/7573

#### menu_trigger="focus"
- Clicking search input should open it unfiltered

Here's a script with a few different configurations

```python
import deephaven.ui as ui
import datetime

# Ticking table with initial row count of 200 that adds a row every second
initial_row_count = 200
_table = time_table(
    "PT1S",
    start_time=datetime.datetime.now() - datetime.timedelta(seconds=initial_row_count),
).update(
    [
        "Int=new Integer(i)",
        "Text=new String(`Display `+i)",
    ]
)

item_list = [ui.item(f"Display {i}") for i in range(1, 201)]


# Basic ComboBox
@ui.component
def ui_combo_box_basic():
    value, set_value = ui.use_state("Display 91")

    return ui.combo_box(
        item_list,
        label=f"Basic ({value})",
        selected_key=value,
        on_change=set_value,
        width="size-3000"
    )


# Uncontrolled ComboBox (Table source)
@ui.component
def ui_combo_box_uncontrolled(table):
    value, set_value = ui.use_state("")

    combo1 = ui.combo_box(
        ui.item_table_source(table, key_column="Text", label_column="Text"),
        default_selected_key="Display 92",
        label=f"Uncontrolled Table Source ({value or 'None'})",
        on_change=set_value,
        width="size-3000"
    )

    return combo1


# Controlled ComboBox (Table source)
@ui.component
def ui_combo_box_controlled(table, menu_trigger):
    value, set_value = ui.use_state("Display 93")

    combo1 = ui.combo_box(
        ui.item_table_source(table, key_column="Text", label_column="Text"),
        selected_key=value,
        label=f"Controlled Table Source ({value}) {menu_trigger or ''}",
        menu_trigger=menu_trigger,
        on_change=set_value,
        width="size-3000"
    )

    btn = ui.button("Set Value", on_press=lambda: set_value("Display 104"))

    return combo1, btn


# Controlled input ComboBox (Table source)
@ui.component
def ui_combo_box_input_controlled(table, menu_trigger):
    input_value, set_input_value = ui.use_state("Display 94")
    value, set_value = ui.use_state("Display 94")

    combo1 = ui.combo_box(
        ui.item_table_source(table, key_column="Text", label_column="Text"),
        input_value=input_value,
        on_input_change=set_input_value,
        default_selected_key=value,
        label=f"Controlled Input Table Source ({value}) {menu_trigger or ''}",
        menu_trigger=menu_trigger,
        on_change=set_value,
        width="size-3000"
    )

    btn = ui.button("Set Input", on_press=lambda: set_input_value("Display 104"))

    return combo1, btn

# Layout
@ui.component
def ui_layout():
    return (
        ui_combo_box_basic(),
        ui_combo_box_uncontrolled(_table), 
        ui_combo_box_controlled(_table, None), 
        ui_combo_box_controlled(_table, "focus"),
        ui_combo_box_input_controlled(_table, None),
    )

tests = ui_layout()
```


packages/components/src/spectrum/comboBox/ComboBox.tsx
```diff
@@ -10,7 +10,7 @@ import type { NormalizedItem } from '../utils';
 import { type PickerPropsT, usePickerProps } from '../picker';
 
 export type ComboBoxProps = PickerPropsT<SpectrumComboBoxProps<NormalizedItem>>;
-
+export { type MenuTriggerAction } from '@react-types/combobox';
```
mofojed, line 13: Aside - not our naming, but I was worried this name would conflict with something from the `@react-types/menu` package. But there they have [`MenuTriggerType`](https://github.com/adobe/react-spectrum/blob/82a4de5b01de2549e2706d048def879b8e6f12b0/packages/%40react-types/menu/src/index.d.ts#L17).
I don't know why they didn't call this `ComboBoxTriggerType` or something instead, but whatever.

packages/jsapi-components/src/spectrum/ComboBox.tsx
```diff
@@ -18,19 +19,57 @@ export function ComboBox(props: ComboBoxProps): JSX.Element {
...
+      // When the ComboBox is closed, `onInputChange` may have been called as a
+      // result of user search input, ComboBox selection, or by selected key
+      // prop changes. We can't determine the source here, so we clear the search
+      // text and store the search value so that the list is unfiltered the next
+      // time the ComboBox is opened. We also store the search value so we can
+      // re-apply it in `onOpenChange` if the ComboBox is opened by user search
+      // input.
+      else {
+        onSearchTextChange('');
+        inputValueRef.current = value;
```
mofojed, line 42: Shouldn't this be set in all cases? Everything else looks good.

bmingles: I'm on the fence on this one. I had only put it here since the only consumer should be the `menuTrigger === 'input'` case which should only follow this `else` case. I was trying to conceptually couple them in an attempt to make the code easier to reason about. That said, there's probably not any harm in setting it, and there could be an edge case that hits it I'm unaware of. In general, reasoning about the relationship of this 2 functions is a bit of a pain.

mofojed (approved):

PR #2330, opened 2025-01-06
fix: Update Spectrum Theme for Missing ContextualHelpTrigger Icon

dgodinez-dh:
This is for `ui.menu`
https://deephaven.atlassian.net/browse/DH-18089
The ContexualHelpTrigger icon does not display because incorrect box-sizing.

### DIFF
--- a/packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
+++ b/packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
@@ -9,6 +9,10 @@ label[class*='spectrum-'] {
   margin-bottom: 0;
 }
 
+/*
+* This workaround can be removed when this fix is pulled in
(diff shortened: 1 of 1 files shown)


dsmmcken (dismissed):

packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
```diff
@@ -9,6 +9,12 @@ label[class*='spectrum-'] {
   margin-bottom: 0;
 }
 
+[class*='spectrum-Menu-itemGrid'] svg,
```
dsmmcken, line 12: you could consider linking this to PR that would allow this lines removal

dgodinez-dh: Added

dsmmcken (approved):

PR #2336, opened 2025-01-09
feat: Add global shortcut to export logs

### LINKED ISSUE #1963: No way to export support logs before logging in or after there's a connection error
Right now the only way to export support logs is via the Settings menu. However, there are a few scenarios where the settings menu isn't available, e.g.:
- Before the user is logged in
- On a reconnect error (overlays the full screen)
- Embedded widgets/iframes
In those cases we should still be able to export the support logs somehow. Would also be nice to be able to have the version information accessible as well.

ericlln:
- Added `Ctrl+Alt+Shift+L/Cmd-Option-Shift-L` as a global shortcut for exporting support logs
- When pressed in code-studio, it will download logs with the same contents as logs produced by pressing the 'Export Logs' button in the settings menu
- When pressed in other contexts: when not logged in, on a reconnect error, or in embed-widgets, the logs are exported without the plugin and server info
- Tested by E2E tests that navigate to the various contexts, presses the shortcut, and checks if a file was downloaded. Can further inspect the file downloaded in the tests but unsure if that's necessary

Closes #1963

### DIFF
--- a/packages/app-utils/src/components/AppBootstrap.tsx
+++ b/packages/app-utils/src/components/AppBootstrap.tsx
@@ -5,17 +5,12 @@ import '@deephaven/components/scss/BaseStyleSheet.scss';
 import { ClientBootstrap } from '@deephaven/jsapi-bootstrap';
 import { useBroadcastLoginListener } from '@deephaven/jsapi-components';
 import { type Plugin } from '@deephaven/plugin';
-import { exportLogs, logHistory } from '@deephaven/log';
-import {
-  ContextActions,
-  ContextMenuRoot,
-  GLOBAL_SHORTCUTS,
-} from '@deephaven/components';
+import { ContextActions, ContextMenuRoot } from '@deephaven/components';
 import FontBootstrap from './FontBootstrap';
 import PluginsBootstrap from './PluginsBootstrap';
 import AuthBootstrap from './AuthBootstrap';
 import ConnectionBootstrap from './ConnectionBootstrap';
-import { getConnectOptions } from '../utils';
+import { getConnectOptions, createExportLogsContextAction } from '../utils';
 import FontsLoaded from './FontsLoaded';
 import UserBootstrap from './UserBootstrap';
 import ServerConfigBootstrap from './ServerConfigBootstrap';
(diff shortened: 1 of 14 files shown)


ericlln:

packages/app-utils/src/components/AppBootstrap.tsx
```diff
@@ -19,6 +25,9 @@ export type AppBootstrapProps = {
   /** URL of the server. */
   serverUrl: string;
 
+  /** Version of front-end. */
+  uiVersion: string;
```
mofojed, line 29: This prop should be optional - it's not really required, and by adding it we're making this a breaking change for anybody consuming this package (since they'd need to provide `uiVersion` or TypeScript will fail:
```suggestion
  uiVersion?: string;
```

I'd actually go one further, and just take a `logMetadata?: Record<string, unknown>` prop, in case we want to include any other properties in there.

packages/app-utils/src/components/AppBootstrap.tsx
```diff
@@ -56,6 +66,25 @@ export function AppBootstrap({
...
+            uiVersion,
+            userAgent: navigator.userAgent,
+          },
+          store.getState()
+        );
+      },
+      shortcut: GLOBAL_SHORTCUTS.EXPORT_LOGS,
+      isGlobal: true,
+    },
+  ];
```
mofojed, line 86: I had mentioned on our 1:1 that the action was only in two places, so we didn't need to factor it out again. But now we've got it in a few places, I'd make a method in the `app-utils` package to create the context action.

```typescript
function createExportLogsContextAction(metadata?: ..., reduxData?: ...) {
  ...
}
```

We should also memoize the `contextActions` here using `useMemo` so we're not re-rendering unnecessarily.

tests/shortcuts.spec.ts
```diff
@@ -23,7 +23,17 @@ test('shortcut downloads logs in full screen error', async ({ page }) => {
 });
 
 test('shortcut downloads logs in embeded-widget', async ({ page }) => {
+  test.slow(); // Extend timeout to prevent a failure before page loads
+
+  // The embed-widgets page and the table itself have seperate loading spinners,
```
mofojed, line 28: nitpick
```suggestion
  // The embed-widgets page and the table itself have separate loading spinners,
```

tests/shortcuts.spec.ts
```diff
@@ -23,7 +23,17 @@ test('shortcut downloads logs in full screen error', async ({ page }) => {
 });
 
 test('shortcut downloads logs in embeded-widget', async ({ page }) => {
+  test.slow(); // Extend timeout to prevent a failure before page loads
```
mofojed, line 26: Neat, haven't used this API before. Looks like you can limit it based on browser as well, which is nice.
Instead of having a separate comment, you could include it in the description of the `test.slow`, e.g.
```suggestion
  test.slow(true, "Extend timeout to prevent a failure before page loads");
```

mofojed (approved):
Looks great and resolves the ticket as specified! Going forward we can improve these error situations by having a visible button so the user doesn't need to remember a keyboard shortcut.

mofojed (approved):

PR #2338, opened 2025-01-09
chore: DH-18086: Update jsapi types

bmingles:
DH-18086:
* Upgrade @deephaven/jsapi-types to ^1.0.0-dev0.37.2
* Fixed types


packages/chart/src/Chart.tsx
```diff
@@ -506,7 +508,11 @@ class Chart extends Component<ChartProps, ChartState> {
         });
 
         const { onError } = this.props;
-        onError(new DownsamplingError(downsamplingError));
+        onError(
+          new DownsamplingError(
+            downsamplingError == null ? undefined : `${downsamplingError}`
```
bmingles, line 513: This is technically a runtime change, but the DownsamplingError constructor should be stringifying it's first arg anyway.

packages/jsapi-utils/src/SessionUtils.ts
```diff
@@ -24,18 +24,6 @@ export interface SessionWrapper {
...
-/**
- * @returns New connection to the server
- */
-export function createConnection(
-  dh: typeof DhType,
-  websocketUrl: string
-): DhType.IdeConnection {
-  log.info(`Starting connection to '${websocketUrl}'...`);
-
-  return new dh.IdeConnection(websocketUrl);
```
bmingles, line 36: `IdeConnection` is a protected constructor which now shows a ts error in the 0.37+ types. The wrapping connection doesn't appear to be called anywhere, so deleting it. 

Per @niloc132 this call shouldn't have worked as of early 2023.

dgodinez-dh (approved):

PR #2339, opened 2025-01-09
feat: DH-18086: gRPC transport implementation for nodejs backed by http2

bmingles:
DH-18086: gRPC transport implementation for nodejs backed by http2

Note that this is not consumed anywhere in the web but will be used by the vscode extension and tested on
https://github.com/deephaven/vscode-deephaven/pull/202

### DIFF
--- a/packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
+++ b/packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
@@ -64,7 +64,7 @@ export class NodeHttp2gRPCTransport implements GrpcTransport {
   /**
    * Private constructor to limit instantiation to the static factory method.
    * @param options Transport options.
-   * @param session node:http2 session.
+   * @param session node:http2 session to use for data transport.
    */
   private constructor(
     options: GrpcTransportOptions,
--- a/packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
+++ b/packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
@@ -177,7 +177,6 @@ export class NodeHttp2gRPCTransport implements GrpcTransport {
     logger.debug('cancel');
     assertNotNull(this.request, 'request is required');
     this.request.close();
-    this.request = null;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * End the stream, both notifying the server that no more messages will be
+   * sent nor received, and preventing the client from receiving any more events.
+   */
+  cancel(): void {
+    logger.debug('cancel');
+    assertNotNull(this.request, 'request is required');
+    this.request.close();
```
vbabich, line 179: Should we reset `request` to `null` here? Probably not needed, if `request.write` handles closed requests fine.

bmingles: Good question. @niloc132 do you have any thoughts on this one?

niloc132: It looks safe to me to not null out - but nulling it out does come with the benefit of clearly showing why an error happened if it is reused somehow. I'm certainly not against that style of defensive coding, but I think we'd want to update a few other places too to reflect this.

GC-wise, I don't see any issue with the current implementation.

vbabich (dismissed):

packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
```diff
@@ -177,7 +177,6 @@ export class NodeHttp2gRPCTransport implements GrpcTransport {
     logger.debug('cancel');
     assertNotNull(this.request, 'request is required');
     this.request.close();
-    this.request = null;
```
bmingles: @vbabich I ended up reverting this as Colin pointed out this breaks the check in the `start` method to avoid calling start multiple times. We could make the state logic smarter, but I think it will introduce complexity that can't easily be tested. Probably best to leave simple for now, unless we run into any particular problems.

packages/jsapi-nodejs/src/NodeHttp2gRPCTransport.ts
bmingles: Opting to leave as-is for now to avoid adding complexity. Setting to null will require changes to `start` which assumes this stays set to know if it has already been called.

vbabich (approved):

PR #2341, opened 2025-01-15
fix: No context menu item for paste in an input table

ericlln:
- Added paste as a context menu option in IrisGrid
- Added methods to check for and request clipboard permissions, which are called when clicking paste in context menu
- For browsers that support the `clipboard-read` permission (e.g. Chrome), clicking paste results in the same behaviour as pressing the keyboard shortcut, provided the user has accepted the request for permission 
- For browsers that don't support the `clipboard-read` permission (e.g. Firefox and Safari), a modal is shown to explain the reason and to suggest using the keyboard shortcut instead

Additional related issues fixed
- Show relevant warning when attempting to paste too many columns into input table
- Fixed copy and pasting between Deephaven tables on Firefox. The issue is caused by tabs in pasted content being converted to `&nbsp;&nbsp;&nbsp; ` instead of `&nbsp;&nbsp; &nbsp;` which we were expecting before. Was able to reproduce former behaviour on an older build of Firefox (91.0), meaning this discrepancy is likely caused by an update to Firefox (unsure when exactly)

### DIFF
--- a/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
+++ b/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
@@ -45,6 +45,7 @@ import {
   assertNotNaN,
   assertNotNull,
   copyToClipboard,
+  readFromClipboard,
 } from '@deephaven/utils';
 import {
   DateTimeFormatContextMenu,
@@ -476,6 +477,22 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
         },
       });
     }
+
+    actions.push({
+      title: 'Paste',
+      group: IrisGridContextMenuHandler.GROUP_COPY,
+      order: 50,
+      action: async () => {
+        const text = await readFromClipboard();
+        if (text !== null) {
+          const items = text.split('\n').map(row => row.split('\t'));
+          await grid.pasteValue(items);
+        } else {
+          // todo: Bring up a popup
+          console.error('no permissions');
+        }
+      },
+    });
 
     actions.push({
       title: 'View Cell Contents',
--- a/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
+++ b/packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
@@ -40,6 +40,8 @@ import {
 import Log from '@deephaven/log';
 import type { DebouncedFunc } from 'lodash';
 import {
+  ClipboardPermissionsDeniedError,
+  ClipboardUnavailableError,
   TextUtils,
   assertNotEmpty,
   assertNotNaN,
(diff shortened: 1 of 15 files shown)


packages/utils/src/ClipboardUtils.ts
```diff
@@ -38,3 +40,42 @@ export function copyToClipboardExecCommand(text: string): void {
 
   if (!result) throw new Error('Unable to execute copy command');
 }
+
+/**
+ * Reads text from the clipboard.
+ * @returns Promise that resolves to text from clipboard as string, or null if permissions are not supported/granted.
+ */
+export async function readFromClipboard(): Promise<string | null> {
```
mofojed, line 48: Instead of returning `string | null`, you should throw an error if the permissions are invalid. It's more descriptive and you could throw different kinds of errors. For this case, I'd create a `ClipboardPermissionsDeniedError` and `ClipboardUnavailableError` type (see `PasteError` for an example of creating a permissions type).
Then when calling `readFromClipboard`, wrap it in a `try/catch` (in IrisGridContextMenuHandler) and you can display an appropriate error message based on the type of error or check that it throws as expected (from the tests).

packages/utils/src/PermissionUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+  permission: string
+): Promise<PermissionState | null> {
+  try {
+    // Typescript doesn't recognize certain permissions as a valid PermissionName
+    // https://github.com/microsoft/TypeScript/issues/33923
+    const name = permission as PermissionName;
+    return (await navigator.permissions.query({ name })).state;
+  } catch (error) {
+    // Permission is unsupported by browser
+    return null;
```
mofojed, line 17: As mentioned in my other comment - you should still throw an error here rather than mapping it to `null`.

mofojed (changes requested):
Exceptions should still be thrown/caught, rather than mapping to `null`. Excellent work, great find confirming the old behaviour in Firefox as well!

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -477,6 +478,21 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
     }
 
+    actions.push({
+      title: 'Paste',
+      group: IrisGridContextMenuHandler.GROUP_COPY,
+      order: 50,
+      action: async () => {
+        const text = await readFromClipboard();
+        if (text !== null) {
+          const items = text.split('\n').map(row => row.split('\t'));
```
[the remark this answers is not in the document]
ericlln: I think that only applies to the pasting into a dummy element approach anyway, since pasting with the weird native browser context menu workaround that I showed in Firefox works with this code as well.

packages/utils/src/ClipboardUtils.ts
```diff
@@ -43,39 +48,51 @@ export function copyToClipboardExecCommand(text: string): void {
...
+
+      if (permissionState === 'prompt' || permissionState === 'denied') {
+        // Prompt means user closed out of the previous permission prompt, also treat it as a denial
+        throw new ClipboardPermissionsDeniedError();
+      }
+    }
+
+    if (permissionState === 'denied') {
+      throw new ClipboardPermissionsDeniedError();
+    }
```
mofojed, line 88: You can simplify this this:
```suggestion
    }

    if (permissionState === 'prompt' || permissionState === 'denied') {
        // Prompt means user closed out of the previous permission prompt, also treat it as a denial
      throw new ClipboardPermissionsDeniedError();
    }
```

mofojed (changes requested):
Great work! Just one minor thing that can be simplified.

See how using exceptions instead of returning null gives you better flexibility about how you handle exceptional cases.

mofojed (approved):

PR #2344, opened 2025-01-17
feat: Add indicator theme variables

jnumainville:
Adds theme variables for indicator


mofojed (changes requested):
You'll need to run `npm run e2e:update-ci-snapshots`

mofojed (approved):

PR #2348, opened 2025-01-23
fix: Long dashboard titles overflow the navbar tab

ericlln:
- Moved the title into its own element in the navbar tab so that `text-overflow: ellipsis` would work properly. 
- Tested by creating dashboards with both long and short titles to confirm that long titles are truncated with ellipses, and short titles display in the same way as before.

### DIFF
--- a/packages/components/src/navigation/NavTab.tsx
+++ b/packages/components/src/navigation/NavTab.tsx
@@ -93,7 +93,7 @@ const NavTab = memo(
               }}
             >
               {iconElem}
-              {title}
+              <span className="btn-nav-tab-title">{title}</span>
               {isClosable && (
                 <Button
                   kind="ghost"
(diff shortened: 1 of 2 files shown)


dsmmcken (changes requested):
if we are truncating it we should also tooltip it so users can get the full name.

If the user has a dash:
my_long_dashboard_name_1
my_long_dashboard_name_2

There's no way to tell them apart.

ericlln:
Currently the `Tooltip` component is placed inside of the span for the title to make it appear only when the text is hovered, which prevents it from conflicting with the tooltip on the close button. On enterprise, there is no tooltip for the close button, and the tooltip is triggered by hovering over any part of the tab, rather than only the text.

dsmmcken (approved):

dsmmcken (approved):

PR #2349, opened 2025-01-24
fix: Use virtual-webgl to circumvent browser limits on WebGL contexts

mofojed:
- Browsers often have a restriction on the number of WebGL contexts that can be created
  - e.g. Chrome on MacOS has a limit of 8
- Use virtual-webgl to circumvent that limit. Basically it creates one context that is used for the whole page.


packages/chart/src/index.ts
```diff
@@ -1,3 +1,7 @@
+// Used to show many WebGL plots on the same page
+// https://github.com/plotly/plotly.js/?tab=readme-ov-file#need-to-have-several-webgl-graphs-on-a-page
+import './virtual-webgl';
```
mattrunyon, line 3: Just curious if you tried the webgl2 version. I know the plotly readme says webgl1 specifically, but curious if webgl2 would work. I don't care to find out why it won't if it doesn't, just curious if you tried

mattrunyon (commented):
Should we add an e2e test for this?

packages/chart/src/index.ts
mofojed: WebGL2 did not work and threw errors.

mofojed:
> Should we add an e2e test for this?

e2e test kind of annoying because don't have deephaven.ui in the e2e tests here so can't define the dashboard programmatically, and adding 16 plots and moving them all to be visible cycling through them will be annoyingly slow/difficult to write robustly... Hmmmmm actually I just realized I could define all the plots programmatically and then create a layout with all those plots, and have that in the data. That wouldn't be too bad...

mattrunyon (approved):

PR #2351, opened 2025-01-28
fix: Combobox exported type error

mofojed:
- Was receiving an error on running `npm start`: [ERROR] Failed to resolve entry for package "@react-types/combobox". The package may have incorrect main/module/exports specified in its package.json. [plugin vite:dep-scan]
- Seemed to be an issue with how we exported the type. An issue has been filed with esbuild: https://github.com/evanw/esbuild/issues/4054
- Just change the export so that we avoid this issue. Tested by running `npm start` again


bmingles (approved):
Code looks good to me.

Code coverage failed with the known intermittent error so may need to kick that off again.

PR #2352, opened 2025-01-29
feat: Add chart event for title change

jnumainville:
These are for https://github.com/deephaven/deephaven-plugins/pull/1088

Currently the layout is only set once, but with indicator it may change, although specifically the title, so I added an event for that. Not sure if it's the best approach.


packages/chart/src/Chart.tsx
```diff
@@ -531,6 +531,22 @@ class Chart extends Component<ChartProps, ChartState> {
         this.setState({ shownBlocker: null });
         break;
       }
+      case ChartModel.EVENT_TITLE_CHANGE: {
+        const titleText = `${detail}`;
+        const oldTitle = this.state.layout.title;
```
mofojed, line 536:
```suggestion
        const { title: oldTitle } = this.state.layout;
```

packages/chart/src/Chart.tsx
```diff
@@ -531,6 +531,22 @@ class Chart extends Component<ChartProps, ChartState> {
         this.setState({ shownBlocker: null });
         break;
       }
+      case ChartModel.EVENT_TITLE_CHANGE: {
```
mofojed, line 534: Instead of an `EVENT_TITLE_CHANGE`, go with `EVENT_LAYOUT_UPDATED` and emit the new layout.
`ChartModel` already has the layout (`.getLayout()`) and that should contain the updated title.
Also see in `ChartModel.initData` how we use the chart layout on top of the state layout... so this would be something like:
```
case ChartModel.EVENT_LAYOUT_UPDATED: {
  const newLayout = detail as Partial<Layout>;
  this.setState(({ layout, revision }) => ({ layout: { ...layout, ...newLayout }, revision: revision + 1 }))
}
```

mofojed (changes requested):

mofojed:
@jnumainville also if you could separate the decimal fix into a separate PR that would make cherry-picking changes if needed easier.

jnumainville:
format options added to https://github.com/deephaven/web-client-ui/pull/2353

packages/chart/src/Chart.tsx
```diff
@@ -45,8 +45,8 @@ const log = Log.module('Chart');
 
 type ChartSettings = ColumnFormatSettings &
   DateTimeFormatSettings & {
-    decimalFormatOptions?: DecimalColumnFormatterOptions;
-    integerFormatOptions?: IntegerColumnFormatterOptions;
+    defaultDecimalFormatOptions?: DecimalColumnFormatterOptions;
```
mofojed, line 48: Should merge the latest main so this doesn't show up in this PR.

mofojed (changes requested):
Please merge the latest main into this branch

mofojed (approved):

PR #2355, opened 2025-01-31
fix: Use virtual-webgl to circumvent browser limits on WebGL contexts…

mofojed:
… (#2349)

- Browsers often have a restriction on the number of WebGL contexts that can be created
  - e.g. Chrome on MacOS has a limit of 8
- Use virtual-webgl to circumvent that limit. Basically it creates one context that is used for the whole page.


mofojed:
Performance in ticking plots is pretty bad with virtual-webgl. I've updated the example to use a `time_table`:
```python
from deephaven import empty_table, time_table, ui

_t = time_table("PT1s").update(["x=i", "y=Math.sin(i)"])

panels = [ui.panel(dx.line(_t, "x", "y")) for i in range(5)]
webgl_dash = ui.dashboard(ui.row(*[ui.column(panels) for i in range(4)]))
```
The result looks crappy, and the page gets choppy:

https://github.com/user-attachments/assets/73c336b1-9c69-4b86-b4b4-4f388caf9085

I have less confidence in this change, and am thinking we don't want to pick this back yet without further assessment.

mofojed:
Instead of picking this back, will be fixing the WebGL flag for Deephaven express: https://github.com/deephaven/deephaven-plugins/pull/1103
Then WebGL can be disabled easily to revert to pre-WebGL behaviour.
The performance with this virtual-webgl is unaccpetable.

PR #2356, opened 2025-02-04
revert: "fix: Use virtual-webgl to circumvent browser limits on WebGL contexts (#2349)"

mofojed:
This reverts commit 6a20d6977d354d89704e05ca687df619b8eeb113.


mattrunyon (approved):
Commenting just so the reasoning is documented. Reverting this because the performance is terrible with virtual webgl and ticking plots

PR #2358, opened 2025-02-04
feat: Resize columns option in web UI

### LINKED ISSUE #1486: "Resize all columns" option in context menu on column headers
It would be great to add a "resize all columns" feature (like the swing UI has) to automatically resize columns to fit their data/headers. It's really useful for keeping the UI compact.

For example, in the table below my columns all got way too wide, and it would take forever to make them compact again (I'd have to keep scrolling right just to grab the column separator to drag it):

image

Just for context, this is the button I'm talking about:

image

Relates to #885

ericlln:
Following below draft spec by @dsmmcken 
- **Resize Column**: If the column is currently manually sized, switch it to auto-resize, and reset the max column width seen to only what's in the current viewport (NOT the max ever seen, but the actual max visible in current viewport). If the column is currently auto sized, set the column to manual and size to max column width seen in the current viewport. 
- The assumption that can be made is that whatever the current behaviour is, it’s not what the users wants, so no need to require the user to make a conscious choice between auto and resize, just present one option to resize, and do the opposite of what we are doing currently.
- **Resize All Columns**: Same behaviour as above but apply across all columns, and treat like an indeterminate checkbox. If any are manual, resize all as above, but flip all to manual.  If all are manual, resize and flip all to auto.
- We should try this and see if it feels natural.

Fixed an issue where double clicking the column separator in the header bar resizes the column without changing the column to auto resize.
- There is now a discrepancy between double clicking the column separator in the header bar and clicking "Resize Column", as the former auto-resizes to the largest cached width, whilst the latter resizes to the width of content in the current viewport. 
- I assume double clicking the column separator follows the behaviour of Excel so this discrepancy is intended.

Below is a useful snippet for testing, that creates a table that has two columns that continuously gets bigger
```

tt = time_table("PT0.5s").update(["x=`a`.repeat(i)", "y=`b`.repeat(i)"])
```

Closes #1486


dsmmcken (dismissed):
approved for behaviour only, no code review performed.

vbabich (dismissed):

vbabich (approved):

PR #2359, opened 2025-02-06
fix: Columns not auto sizing correctly

### LINKED ISSUE #2288: Column header names do not truncate based on text measurement
**Description**

Column header names are truncated based on an estimated width in `GridRenderer.drawColumnHeader`. This is problematic since column headers do not use a monospaced font. If there are many wide characters in a name, it can overflow the bounds instead of truncating.

**Steps to reproduce**

```py
t = empty_table(10).update(["X=ii", "MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM=ii"])
```

**Expected results**

The long repeating column name is truncated

**Actual results**

image

**Versions**

Engine Version: 0.37.0-SNAPSHOT
Web UI Version: 0.97.0
Java Version: 17.0.12
Barrage Version: 0.6.0
Browser Name: Chrome 130
OS Name: Linux
@deephaven/js-plugin-ui: 0.22.0
@deephaven/js-plugin-plotly-express: 0.11.2
@deephaven/js-plugin-matplotlib: 0.5.0

ericlln:
- Improved logic for calculating the range to truncate a string, based on the width of the assumed largest character and smallest character for a given font. This fixes the observed issue of wide strings not being truncated.
- Changed column width calculation to calculate text width by iterating through the text and adding up individual character widths which are cached. This approach size columns more precisely than estimating based on a single character's width and the length of the string, and does not appear to have a significant performance impact. 

Test snippet used to see if column width calculation works as expected; resize columns to observe truncation behaviour
```

a = new_table([string_col("allllllllllllllllllllllllllllllllllllllllllllllll", ["a"])])
b = new_table([string_col("ammmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm", ["b"])])
c = new_table([string_col("c", ["allllllllllllllllllllllllllllllllllllllllllllllll"])])
d = new_table([string_col("d", ["ammmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm"])])
```
 
Closes #2288


packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1800,27 +1870,51 @@ export class GridMetricCalculator {
...
+    font: GridFont,
+    context: CanvasRenderingContext2D
+  ): void {
+    if (this.fontWidthsUpper.has(font)) {
+      return;
+    }
+
+    context.font = font;
+    // Assume char `m` is the largest character
+    const textMetrics = context.measureText('m');
```
dgodinez-dh, line 1911: In the ticket, the complaint was about a capital 'M' overflowing the width. Is lowercase 'm' going to have the same width?

packages/grid/src/DataBarCellRenderer.ts
```diff
@@ -103,13 +104,15 @@ class DataBarCellRenderer extends CellRenderer {
       width: textWidth,
     } = GridUtils.getTextRenderMetrics(state, column, row);
 
-    const fontWidth = fontWidths?.get(context.font) ?? DEFAULT_FONT_WIDTH;
```
dgodinez-dh, line 106: I see in a lot of places that we no longer check the default font. What happens if the font is not the map? Do we get a default somehow, or is it guaranteed to be calculated?

ericlln: I saw that this constant was being used in a lot of different places, since `CellRenderer.getCachedTruncatedString` doesn’t accept an undefined value for the font widths. The function `GridRenderer.truncateToWidth` that it calls does though, and already uses that constant for default values, so I made `getCachedTruncatedString` accept undefined for the font widths, and the constant can be removed from other places safely. 

If you did happen to need the font widths for something else in those methods, you would have to add back that undefined check.

packages/grid/src/GridMetricCalculator.ts
ericlln: When I was initially investigating this issue I measured all alphanumeric characters using our canvas context, and found that a lowercase ‘m’ is a wider than an upper case one by about 8%, so I ended up taking that as the largest value. This might not be true for all fonts but it seems to be more common based on my limited research.

[The Width of Lowercase ´m´](https://www.wingframe.com/articles/the-width-of-lowercase-m)

PR #2360, opened 2025-02-06
fix: DH-18542: Remove duplicate and invalid ruff quick fixes

mattrunyon:
Should remove the invalid quick fixes to disable syntax errors (wouldn't that be nice if you could do that)

Also if there is a duplicate code detected (usually only if you have a selection and use the light bulb instead of the hover quick fix link in the modal), then the "diasble RULE for file" should only appear once. There should also be "disable RULE for line X" and "disable RULE for line Y" instead of just "for line".

### DIFF
--- a/packages/console/src/monaco/MonacoProviders.tsx
+++ b/packages/console/src/monaco/MonacoProviders.tsx
@@ -251,15 +251,17 @@ class MonacoProviders extends PureComponent<
       };
     }
 
-    const diagnostics = MonacoProviders.getDiagnostics(model).filter(d => {
-      const diagnosticRange = new monaco.Range(
-        d.location.row,
-        d.location.column,
-        d.end_location.row,
-        d.end_location.column
-      );
-      return diagnosticRange.intersectRanges(range);
-    });
+    const diagnostics = MonacoProviders.getDiagnostics(model)
+      .filter(d => d.code != null) // Syntax errors have no code and can't be fixed/disabled
+      .filter(d => {
+        const diagnosticRange = new monaco.Range(
+          d.location.row,
+          d.location.column,
+          d.end_location.row,
+          d.end_location.column
+        );
+        return diagnosticRange.intersectRanges(range);
+      });
 
     const fixActions: monaco.languages.CodeAction[] = diagnostics
       .filter(({ fix }) => fix != null)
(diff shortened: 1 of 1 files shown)


packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -251,15 +251,17 @@ class MonacoProviders extends PureComponent<
...
-      const diagnosticRange = new monaco.Range(
-        d.location.row,
-        d.location.column,
-        d.end_location.row,
-        d.end_location.column
-      );
-      return diagnosticRange.intersectRanges(range);
-    });
+    const diagnostics = MonacoProviders.getDiagnostics(model)
+      .filter(d => d.code != null) // Syntax errors have no code and can't be fixed/disabled
```
bmingles, line 255: small nitpick. I would probably combine these into 1 callback to avoid 2 passes of the array

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -272,6 +274,7 @@ class MonacoProviders extends PureComponent<
             title = `Fix ${d.code}`;
           }
         }
+        const b = 4;
```
bmingles, line 277: Seems to be left over testing code?

packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -296,8 +299,23 @@ class MonacoProviders extends PureComponent<
         };
       });
 
-    const disableActions: monaco.languages.CodeAction[] = diagnostics
+    const seenCodes = new Set<string>();
+    const duplicateCodes = new Set<string>();
+    diagnostics.forEach(d => {
+      if (d.code == null) {
```
bmingles, line 305: Should already be filtered out right?

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: Oops yes I put that there to see how vscode/ts service handles quick fixes

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: Yes, but TS doesn't know that so I had to add this check

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: Good point

bmingles (approved):
LGTM. Tested the 2 scenarios described in original issue.

PR #2365, opened 2025-02-07
fix: Don't use HTML in panel titles, treat as text

mofojed:
- Golden Layout was setting the title in panel titles with `.html()` method, which allows remote code execution, and potentially could craft a notebook with a malicious name that could run arbitrary JS
- Instead just use `.text` and treat it like text, as we should.
- Tested on DHC - created a notebook with the name `.py`. It now appears correctly as text and does not pop up an alert message. Also verified that title still appears in italics when in "preview" mode.
- Fixes DH-18645


bmingles (changes requested):
Unit tests are failing. Looks like they need to be updated with escaped values.

bmingles (approved):
LGTM

PR #2370, opened 2025-02-19
fix: Fix column width calculation logic in grid

ericlln:
- Noticed this when testing some things on Enterprise. The column header “SPet500” in the stocks table is truncated when it shouldn’t be. For some reason this case didn’t occur on Core, perhaps due to slight differences in canvas configurations. 
- Rounded down text measurement done in `GridRenderer.binaryTruncateToWidth()` as the measured text width with can differ slightly from the passed in estimated column width. This will prevent cases where text is truncated when it shouldn't be.

Stocks table for testing:
```
stocks = dx.data.stocks();
```


dgodinez-dh (approved):

packages/grid/src/GridRenderer.ts
```diff
@@ -71,7 +71,12 @@ export class GridRenderer {
     end = str.length,
     truncationChar?: string
   ): string {
-    if (end >= str.length && context.measureText(str).width <= width) {
+    if (
+      end >= str.length &&
+      // The passed in width is an estimated column width and can differ slightly from the str width measured here
+      // So it is rounded down to prevent cases where text is truncated when it shouldn't be
+      Math.floor(context.measureText(str).width) <= width
+    ) {
```
mattrunyon, line 79: I'm not sure this is the behavior we want. `width` should be known here. If we only have 60px to draw and the text takes up 60.7px then truncation is correct. The estimated width is wrong if it's truncating, not the truncation logic

mattrunyon (changes requested):

packages/grid/src/GridRenderer.ts
mattrunyon: The column doesn't truncate for me on latest core, grizzly core+, or sanluis core+. So it might be an OS/browser issue, or if you're using your local UI with changes from 2359, then the logic in 2359 to change truncation is underestimating in these cases.

Either way I think it's the logic in estimating width and not the actual measurement that is the issue.

@mofojed wanted your thoughts as well since you know the grid rendering and metrics better than I do.

mofojed: The discrepancy between estimation and actual calculation is likely related to kerning: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fontKerning
Perhaps we can get confirmation of that. @ericlln if you disable kerning (set `fontKerning` to `false`) do you still see discrepancies in the estimated vs calculated? Wondering if we should disable kerning.
Ideally we'd want to estimates to be exact as the calculations; next best would be over-estimating the size slightly (which results in a little extra padding) rather than under-estimating (which results in truncation).

ericlln: After taking a closer look, a part of the discrepancy in the calculated and measured column header widths is due to not setting the correct font on the context when measuring the character widths. I might have initially misinterpreted that it was being set on the context in a different place. 

However, the majority of the difference is caused by kerning, and disabling it does in fact make the calculations match the measured value exactly. I think disabling kerning would make sense, and might also be beneficial since https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fontKerning mentions that kerning on small fonts could harm the readability of text. 

We currently ceil the estimated column width so it will now always overestimate. Technically it can be removed such that the column width is even more exact, so I'm wondering if we should keep that threshold there still. 

Observed Data
Screenshot 2025-02-24 at 3 26 48 PM

dsmmcken: Disabling kerning on the grid is not an acceptable solution, and would harm overall readability of string columns. By small fonts they mean really small, like 8px. 

A string like "AVAVAVAVAVAVAVAVAVAVAVAVA" will probably give you something that produces a pronounced difference in length due to kerning, if you need something for testing. I would try to over-estimate.

ericlln: Okay that makes more sense. There are also combinations such as "f?" that increases the width between characters, so it is not feasible to predict how much to overestimate by in extreme cases (those characters repeated many times). I changed the column width calculation logic to take into account font kerning by measuring pairs of characters now.

packages/grid/src/GridMetricCalculator.test.ts
```diff
@@ -128,4 +131,33 @@ describe('getTextWidth', () => {
...
     expect(textWidth).toEqual(50);
   });
+
+  it.each([
+    [''],
+    ['a'],
+    ['AV'],
+    ['AVAVAV'],
+    ['f?'],
+    ['f?f?AVAV'],
```
mofojed, line 141: I'd add a few more cases here, just some lorem ipsum text, all `i`s, all `M`s, etc.
Or something like `AVEAVEAVEAVE` just to break up the pairings a bit.
Also need a case that is past the max width.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1837,36 +1837,70 @@ export class GridMetricCalculator {
...
 
-      if (charWidths.has(char)) {
-        width += getOrThrow(charWidths, char);
-      } else {
-        const textMetrics = context.measureText(char);
-        const { width: charWidth } = textMetrics;
-        charWidths.set(char, charWidth);
-        width += charWidth;
+      if (!charWidths.has(char)) {
+        context.font = font;
```
mofojed, line 1866: We should just set the font at the start of this method... Interesting that we weren't setting the font before this change, I wonder if that caused any issues.

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1837,36 +1837,70 @@ export class GridMetricCalculator {
...
       this.allCharWidths.set(font, new Map());
     }
-
     const charWidths = getOrThrow(this.allCharWidths, font);
 
-    let width = 0;
+    let result = 0;
     for (let i = 0; i < text.length; i += 1) {
       const char = text[i];
+      const nextChar = i < text.length - 1 ? text[i + 1] : undefined;
```
mofojed, line 1863: No need to check against length here, going past the end of the array will return `undefined` anyway.
```suggestion
      const nextChar = text[i + 1];
```

packages/grid/src/GridMetricCalculator.ts
```diff
@@ -1837,36 +1837,70 @@ export class GridMetricCalculator {
...
-      if (maxWidth !== undefined && width > maxWidth) {
-        return maxWidth;
+      if (nextChar !== undefined) {
+        if (!charWidths.has(nextChar)) {
+          context.font = font;
+          charWidths.set(nextChar, context.measureText(nextChar).width);
+        }
+
+        const pair = char + nextChar;
+        const isFirstPair = i === 0;
```
mofojed, line 1877: I think you should continue doing the same pattern here, e.g.
```typescript
if (!charWidths.has(pair)) {
  charWidths.set(pair, context.measureText(pair).width);
}
```

Then it becomes more clear afterwards:
```typescript
result += getOrThrow(charWidths, pair);
if (i > 0) {
  // Need to remove the current character that was already counted as the previous pair
  result -= getOrThrow(charWidths, char);
}
```

mofojed (changes requested):

packages/grid/src/GridMetricCalculator.test.ts
ericlln: I added some more test cases, but I should note that this test roughly checks that the calculation logic is equivalent to the native canvas one. It doesn’t verify that it addresses font kerning properly, since it doesn’t render the text and `context.measureText()` just returns the text length. It would be nice to be able to test text rendering with fonts, but that doesn’t seem feasible. I guess one way is to compare snapshots of a grid drawn using this method and a grid drawn with the native canvas method, not sure how feasible that would be either. 

As for the max width case, I already have a test for it using a mock implementation of `context.measureText()`. Shouldn’t be too different from creating the canvas context considering its behaviour with `context.measureText()` mentioned above.

packages/grid/src/GridMetricCalculator.ts
ericlln: Yup missing that initially definitely caused issues and the header text width was not being calculated correctly as a result. Since the header font and cell font only differ slightly, it wasn’t able to be noticed visually. Again, it would be nice to be able to catch something like this through testing.

mofojed (approved):

mattrunyon (approved):
Curious why we aren't just measuring the text with a cache like we already do in `GridRenderer` to determine if it needs to be truncated. I guess in theory this should reduce the number of `measureText` calls in the metric calculator (we will still measure the whole string in `GridRenderer` and might be faster if the values are constantly different where the cache isn't much help.

We _could_ add a measured text cache to `GridMetrics` which is then accessible by `GridRenderer` and avoid double measuring the whole string. But that would be a separate ticket

PR #2373, opened 2025-02-26
fix: Proxy model appears partition even if original model is not

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-18545

`IrisGridProxyModel` is a `PartitionedGridModel` and always returns `true` for `isPartitionedGridModelProvider` and `isPartitionedGridModel`.

This fix changes the proxy return values so those methods return `false` if they would return `false` for the original model.

### DIFF
--- a/packages/iris-grid/src/IrisGridProxyModel.ts
+++ b/packages/iris-grid/src/IrisGridProxyModel.ts
@@ -17,6 +17,7 @@ import {
   type PartitionConfig,
   type PartitionedGridModel,
   isPartitionedGridModelProvider,
+  isPartitionedGridModel,
 } from './PartitionedGridModel';
 
 const log = Log.module('IrisGridProxyModel');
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGridProxyModel.ts
```diff
@@ -300,6 +301,10 @@ class IrisGridProxyModel extends IrisGridModel implements PartitionedGridModel {
       !isPartitionedGridModelProvider(this.originalModel) ||
       !this.originalModel.isPartitionRequired
     ) {
+      if (!isPartitionedGridModel(this.originalModel)) {
+        // @ts-expect-error If the original model has an undefined partitionConfig return undefined to make the proxy model also return false in isPartitionedGridModel
```
vbabich, line 305: Would it be difficult to change the return type instead of suppressing the error in `partitionConfig` and `isPartitionRequired`?

dgodinez-dh: It could be changed, but I think it is correct as it is. A real `PartitionedGridModel` should never return `undefined` for these values, and I think code using a real `PartitionedGridModel` should not need to worry about checking for `undefined`.

vbabich (approved):

PR #2386, opened 2025-03-18
fix: allow input table to delete row with null key

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-18681

IrisGridTableModel was failing to create a filter on a `null` key. Updated to use the nullable util method.


bmingles (approved):
Changes look good. I confirmed on dev sanluis that rows can be deleted via the context menu as well as programatically.

PR #2387, opened 2025-03-19
fix: allow input table to delete row with null key (#2386)

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-18681

Cherry pick back for Grizzly.


bmingles (approved):
LGTM

PR #2396, opened 2025-03-27
feat: Allow wildcards for logs blacklist

ericlln:
- Wildcards are now accepted for logs blacklist JSON paths (e.g. `dashboardData.*.sessionWrapper`)
- Updated default blacklist to remove the `sessionWrapper` from any dashboard under `dashboardData`
- This technically removes the ability to look for a `*` literal path and an escape could be added for that case, but it seems like an extreme edge case to handle.

### DIFF
--- a/packages/log/src/LogExport.ts
+++ b/packages/log/src/LogExport.ts
@@ -3,6 +3,7 @@ import type LogHistory from './LogHistory';
 
 // List of objects to blacklist
 // '' represents the root object
+// * represents a wildcard key
 export const DEFAULT_PATH_BLACKLIST: string[][] = [
   ['api'],
   ['client'],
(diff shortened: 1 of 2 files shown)


packages/log/src/LogExport.ts
```diff
@@ -3,11 +3,12 @@ import type LogHistory from './LogHistory';
 
 // List of objects to blacklist
 // '' represents the root object
+// * represents a wildcard key
 export const DEFAULT_PATH_BLACKLIST: string[][] = [
   ['api'],
   ['client'],
   ['dashboardData', 'default', 'connection'],
```
vbabich, line 10: Would it be reasonable to replace `default` with `*` here as well?

ericlln: On enterprise `connection` seems to be under the `sessionWrapper` object, which is why I didn't initially change this as well. Perhaps it's better to be safe here anyway, and would remove the need to check for this `default` key specific to Core.

vbabich (approved):

PR #2402, opened 2025-04-02
feat: Add usePersistentState hook

mattrunyon:
Adds the `usePersistentState` hook and `PersistentStateProvider` which used together track calls to `usePersistentState` and calls an `onChange` when the persisted state changes. The state is persisted as an array of the calls to the hook within the provider.

If the tree initially renders in a different order or with fewer/more hook calls, then the rendering may break depending on what state is now swapped. We could easily detect if the number of calls to `usePersistentState` changed on a hydration and throw an error or re-render the component from scratch, but right now the implementation does not do that. I think it would just be checking if `initialState.length === initialStateMap.size()` after the first render. If it does, clear the initial state and set a key on the context to get a clean re-render.

This should handle memoized components, nested components below the provider, and components which immediately re-render via a conditional setState (`useEffect` in the provider is only fired once even though the component renders twice).

It is technically possible to use this with class components via context and mimicking the logic in the hook by calling `addState` in `render` and `trigger` anywhere that should trigger a state change, but I think it would be best to just wrap the few class components that need this in a thin functional wrapper to use the hook.

If there is no provider, `usePersistentState` still functions effectively just as `useState`. This means we should be able to call the hook in widgets that can be used standalone and in dh.ui without issue (like plotly express or IrisGrid)


packages/plugin/src/PersistentStateContext.tsx
```diff
@@ -0,0 +1,144 @@
...
+   * Will re-render the components that have persistent state to capture their current state.
+   */
+  scheduleStateUpdate: () => void;
+
+  /**
+   * Gets the next item from the initial state iterator.
+   * It is expected that each hook will call this once to get its initial state.
+   * @param id A unique ID generated by the caller.
+   *           This is used to prevent one caller from consuming multiple values from the initial state.
+   * @returns The iterator result whih contains the initial state value and a done flag.
```
mofojed, line 32:
```suggestion
   * @returns The iterator result containing the initial state value and a done flag.
```

packages/plugin/src/usePersistentState.ts
```diff
@@ -0,0 +1,156 @@
...
+  return migratedState;
+}
+
+/**
+ * Functions identically to useState except that a PersistentStateProvider can be used to
+ * track all calls to this hook and persist the value for future page loads.
+ * Primarily used in Deephaven UI so we can persist state of multiple components within a panel.
+ *
+ * @param initialState The initial state if there is no previously persisted state.
+ * @param config.type The type of the state. This identifier is used to validate the state being rehydrated.
```
mofojed, line 74:
```suggestion
 * @param config.type The type of the state. This identifier is used to validate the state being rehydrated. Should be unique to your component.
```

packages/plugin/src/usePersistentState.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @param config.version The version of the state. This is used to determine if the state needs to be migrated.
```
mofojed, line 75:
```suggestion
 * @param config.version The version of the state. This is used to determine if the state needs to be migrated. Value should be incremented when data structure changes for this type.
```

packages/plugin/src/usePersistentState.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @param config.migrations An array of migrations to apply to the state if the version of the persisted state is below the current version.
+ * @param config.migrations.from The version to migrate from.
+ * @param config.migrations.to The version to migrate to.
```
mofojed, line 78: Not sure we really _need_ a `to`, though I guess maybe you'd have a migrator that just skips intermediate versions...

mofojed (changes requested):
Just correcting a few comments, looking good though

packages/plugin/src/usePersistentState.ts
mattrunyon: Ya that was my initial thought was a migrator skipping versions.

However, that poses a problem for intermediate versions. Like if you had a migrator `1->3` and 1 client w/ version 1 with another on version 2, the version 2 client wouldn't upgrade and would throw unless you had `1->3` and `2->3`. And the logic right now prevents having `1->3` and `1->2` simultaneously (because I wanted to keep the logic relatively simple).

I think I'll just remove the `to` and add that the versions need to be incrementing ints. Just simplifies things and we don't need to support a more complicated versioning structure for something that should be fairly infrequent.

Another option would be to remove `from` as well and just say each migrator is applied in order and it is expected you start from `1` and each function increments the version by 1.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -984,6 +984,11 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       }
     }
 
+    // if (prevProps.reverse !== this.props.reverse) {
+    //   console.log('changing reverse');
+    //   this.setState({ reverse: this.props.reverse });
+    // }
+
```
mofojed, line 991: Unintentionally commented out? Or add in a different PR?

mofojed (approved):

mofojed (changes requested):
Clean up commented out code

packages/iris-grid/src/IrisGrid.tsx
mattrunyon: Ya something from testing when I was implementing combining client state and server state for `ui.table`

mofojed (approved):

mofojed (approved):

PR #2404, opened 2025-04-02
fix: Cherry-pick DH-17599: Fix table name validation (#2403)

vbabich:
PR #2398 allowed hyphens in namespace and table names. However, I
noticed that server-side validation rejects names starting with a
hyphen, while these names appear valid on the client-side.

`io.deephaven.db.tables.utils.DBNameValidator` uses separate regular
expressions to sanitize and validate names. This change adds the correct
`TABLE_NAME_PATTERN` for table name validation instead of checking if
the name matches its sanitized version.

---------

Co-authored-by: Copilot <175728472+Copilot@users.noreply.github.com>

### DIFF
--- a/packages/utils/src/DbNameValidator.test.ts
+++ b/packages/utils/src/DbNameValidator.test.ts
@@ -4,8 +4,8 @@ const TABLE_PREFIX = 'table_';
 const COLUMN_PREFIX = 'column_';
 
 const VALID_TABLE_NAME = '$+@abc-123_ABC';
-const INVALID_TABLE_NAME = '%^&ab-c';
-const CLEANED_INVALID_TABLE_NAME = 'ab-c';
+const INVALID_TABLE_NAMES = ['%^&ab-c', '-abc', '-'];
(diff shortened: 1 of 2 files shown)


packages/utils/src/DbNameValidator.test.ts
```diff
@@ -4,8 +4,8 @@ const TABLE_PREFIX = 'table_';
 const COLUMN_PREFIX = 'column_';
 
 const VALID_TABLE_NAME = '$+@abc-123_ABC';
-const INVALID_TABLE_NAME = '%^&ab-c';
-const CLEANED_INVALID_TABLE_NAME = 'ab-c';
+const INVALID_TABLE_NAMES = ['%^&ab-c', '-abc', '-'];
+const CLEANED_INVALID_TABLE_NAMES = ['ab-c', '-abc', '-'];
```
bmingles, line 8: Confused that `-abc` gets cleaned as the same value `-abc`

bmingles (changes requested):
Left one question

bmingles (approved):
LGTM

PR #2409, opened 2025-04-08
fix: DH-17599: Change illegal characters to underscores in variable names

vbabich:
- Add `makeVariableName` method to create valid variable names
- Add/update unit tests.


bmingles (approved):
LGTM

PR #2410, opened 2025-04-09
fix: Cherry-pick DH-17599: Change illegal characters to underscores in variable names

vbabich:
Cherry-pick #2409 into v0.85:
- Add `makeVariableName` method to create valid variable names
- Add/update unit tests.


bmingles (approved):
LGTM

PR #2416, opened 2025-04-21
feat: Add IrisGridCacheUtils for memoizing iris grid state

mattrunyon:
Adds `IrisGridCacheUtils` which provides 3 methods to create memoized dehydrators: `makeMemoizedGridStateDehydrator`, `makeMemoizedIrisGridStateDehydrator`, and `makeMemoizedCombinedGridStateDehydrator`. The first 2 are used for `IrisGridPanel` which saves the state under separate keys for `gridState` and `irisGridState`. The 3rd is for newer uses like `UITable` which can just save the combined state and spread it to `IrisGrid` since `IrisGrid` accepts both `GridState` and `IrisGridState` props.

Added tests for `dehydrateIrisGridState`.

Improved the `memoize-one` type override to infer the types of the equality function based on the actual function args.

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -103,11 +103,11 @@ export interface PanelState {
   gridState: {
     isStuckToBottom: boolean;
     isStuckToRight: boolean;
-    movedColumns: {
+    movedColumns: readonly {
       from: string | ModelIndex | [string, string] | [ModelIndex, ModelIndex];
       to: string | ModelIndex;
     }[];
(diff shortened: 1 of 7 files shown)


packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -103,11 +104,11 @@ export interface PanelState {
   gridState: {
     isStuckToBottom: boolean;
     isStuckToRight: boolean;
-    movedColumns: {
+    movedColumns: readonly {
```
mofojed, line 107: Yesssss `readonly` supreme.

mofojed (approved):

PR #2417, opened 2025-04-22
refactor: DH-19267: columns changed filter event takes an id rather than a panel

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-19267

Refactors `InputFilterEvent.COLUMNS_CHANGED` to take an id instead of a panel. This can be a panel id or a widget id.  This will allow `deephaven.ui` components to send this event. Note that more work will need to be done on Linker. This will be done later when we address https://deephaven.atlassian.net/browse/DH-18840

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
+++ b/packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
@@ -1,4 +1,5 @@
 import { type ReactNode } from 'react';
+import { type Brand } from '@deephaven/utils';
 
 export type WidgetPanelDescriptor = {
   /** Type of the widget. */
@@ -25,4 +26,4 @@ export type WidgetPanelTooltipProps = {
   children?: ReactNode;
 };
 
-export type WidgetId = string;
+export type WidgetId = Brand<'WidgetId'>;
--- a/packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
+++ b/packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
@@ -24,3 +24,5 @@ export type WidgetPanelTooltipProps = {
   /** Children to render within this tooltip */
   children?: ReactNode;
 };
+
+export type WidgetId = string;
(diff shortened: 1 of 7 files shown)


packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
```diff
@@ -24,3 +24,5 @@ export type WidgetPanelTooltipProps = {
   /** Children to render within this tooltip */
   children?: ReactNode;
 };
+
+export type WidgetId = string;
```
bmingles, line 28: In case you haven't seen it, we now have a `Brand` type (`packages/utils/src/TypeUtils.ts`). It's useful for ensuring things like Ids don't get assigned to the wrong a field expecting a different id type. Usage is:
`type WidgetId = Brand<'WidgetId'>`

That said, it does require some type assertions in certain cases when creating ids, so there are tradeoffs. Just wanted to call it out in case it is a good fit here.

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -29,6 +29,8 @@ const log = Log.module('LayoutUtils');
 
 type LayoutConfig = { id?: string; component?: string };
 
+export type PanelId = string | string[] | null | undefined;
```
bmingles, line 32: Some suggestion here to consider the `Brand` type.

```suggestion
export type PanelId = Brand<'PanelId', string | string[] | null | undefined>;
```

bmingles (changes requested):
Left suggestions for using `Brand` type helper.

packages/dashboard-core-plugins/src/panels/WidgetPanelTypes.ts
dgodinez-dh: Updated

packages/dashboard/src/layout/LayoutUtils.ts
dgodinez-dh: Updated

bmingles (approved):
LGTM

PR #2418, opened 2025-04-23
fix: DH-13244: remove trailing whitespace from console log messages

dgodinez-dh:
https://deephaven.atlassian.net/browse/DH-13244

There was a fix for single line console messages here: https://github.com/deephaven/web-client-ui/pull/350
However, this fix did not account for trailing newline characters in the message. This fix trims whitespace from the end of the message.


bmingles (approved):
LGTM

PR #2424, opened 2025-04-29
fix: DH-18832: propagate min and max to plotly figure axis

jnumainville:
I went with `autorangeoptions` instead of range as it provides more intuitive behavior.

```from deephaven import read_csv
stocks = dx.data.stocks().where("Sym = `FISH`")

# plot the data
plot_single = (
    Figure()
    .plot_xy(
        series_name="Price",
        t=stocks,
        x="Timestamp",
        y="Price",
    )
    .y_axis(min=0.0, max=170.0)
    .x_twin()
    .plot_xy(
        series_name="SPet500",
        t=stocks,
        x="Timestamp",
        y="SPet500",
    )
    .y_axis(log=True, max=300.0)
    .show()
)
```

### DIFF
--- a/packages/chart/src/ChartUtils.ts
+++ b/packages/chart/src/ChartUtils.ts
@@ -1539,6 +1539,22 @@ class ChartUtils {
       );
       layoutAxis.domain = [bottom, top];
     }
+
+    // autorangeoptions exists in plotly.js but not in the types
+    // axis.range only applies to the initial axis range, can't handle specifying min or max independently,
+    // and cannot be returned to if the user changes the range via interaction
+    // minallowed and maxallowed can be set independently and can be returned to if the user changes
+    // the range via interaction
+    const { minRange, maxRange, log } = axis;
+    if (!Number.isNaN(minRange) || !Number.isNaN(maxRange)) {
+      (layoutAxis as any).autorangeoptions = {};
+      if (!Number.isNaN(minRange)) {
+        (layoutAxis as any).autorangeoptions.minallowed = log ? Math.log10(minRange) : minRange;
+      }
+      if (!Number.isNaN(maxRange)) {
+        (layoutAxis as any).autorangeoptions.maxallowed = log ? Math.log10(maxRange) : maxRange;;
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


jnumainville:
@mofojed 

Yes it's possible to do a log axis, but there is a bug here. 
```
import math

dx_log = dx.line(stocks, "Timestamp", ["Price", "SPet500"], range_y=[[0, 170], [math.log10(1), math.log10(300)]], yaxis_sequence=[1, 2], log_y=[False, True])
```

Note that I'm manually converting to the log. In `px` this is calculated and should be the same in dx. 
```
import plotly.express as px
import deephaven.pandas as dhpd

px_log = px.line(dhpd.to_pandas(stocks), "Timestamp", "SPet500", range_y=[1, 300], log_y=True)
```

That said, `px` uses range directly, as does `dx`. We should consider adding a `min` and `max` option that hooks directly into `autorangeoptions` in the same way. You could do an unsafe update for this, but I think it's much more intuitive behavior that we should expose directly. `range` is more like `initial_range`

jnumainville:
https://deephaven.atlassian.net/browse/DH-19344 for log axis calculation bug 
https://deephaven.atlassian.net/browse/DH-19345 for new min and max options

packages/chart/src/ChartUtils.ts
```diff
@@ -1545,14 +1545,21 @@ class ChartUtils {
     // and cannot be returned to if the user changes the range via interaction
     // minallowed and maxallowed can be set independently and can be returned to if the user changes
     // the range via interaction
-    const { minRange, maxRange, log } = axis;
+    const { minRange, maxRange, log: logAxis } = axis;
     if (!Number.isNaN(minRange) || !Number.isNaN(maxRange)) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       (layoutAxis as any).autorangeoptions = {};
```
mofojed, line 1551: Looks like `autorangeoptions` was added in `@types/plotly.js v2.33.x`. Can we just update our plotly packages instead of casting to `any`?
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70014/files#diff-2749af52d42facedd35dc9652faa609282f1d528fe532bd40dbcb680e511b8c0R645

mofojed (changes requested):
It is good, but would like to update plotly packages as well to remove the casting.

packages/chart/src/ChartUtils.ts
jnumainville: done

mofojed (approved):

PR #2425, opened 2025-04-30
feat: DH-19307: External theme support

bmingles:
DH-19307: External theme support via `postMessage` apis
- Moved common MessageUtils into @deephaven/utils
- External themes can be enabled by setting `theme=[EXTERNAL_THEME_KEY]` query string param
- ThemeBootstrap now checks if external themes are enabled. If so, the theme is requested from the parent Window via `MSG_REQUEST_GET_THEME` postMessage
- Current or parent Window can explicitly set external theme via `MSG_REQUEST_SET_THEME` postMessage

### Testing
You can download this html file: https://gist.github.com/bmingles/02451e09d218cdbe61d8c506e556562b . You should be able to open it in browser from local filesystem, no need for server.
1. Start DH locally in this branch (assumes port 4000)
1. Open the html file
1. Initial load should show red background color showing DH requesting theme from the parent
1. Clicking the "Update Theme" button in bottom left should change to random bg color. Demonstrates explicit setting of theme from parent Window

### DIFF
--- a/packages/app-utils/src/components/ThemeBootstrap.tsx
+++ b/packages/app-utils/src/components/ThemeBootstrap.tsx
@@ -1,7 +1,7 @@
 import { useContext, useMemo } from 'react';
 import { ChartThemeProvider } from '@deephaven/chart';
 import { MonacoThemeProvider } from '@deephaven/console';
-import { ThemeProvider } from '@deephaven/components';
+import { ThemeProvider, useParentWindowTheme } from '@deephaven/components';
 import { IrisGridThemeProvider } from '@deephaven/iris-grid';
 import { getThemeDataFromPlugins, PluginsContext } from '@deephaven/plugin';
 import { getSettings } from '@deephaven/redux';
@@ -14,16 +14,37 @@ export interface ThemeBootstrapProps {
 export function ThemeBootstrap({
   children,
 }: ThemeBootstrapProps): JSX.Element | null {
+  const {
+    isEnabled: isParentThemeEnabled,
+    isPending: isParentThemePending,
+    themeData: parentThemeData,
+  } = useParentWindowTheme();
+
   // The `usePlugins` hook throws if the context value is null. Since this is
   // the state while plugins load asynchronously, we are using `useContext`
   // directly to avoid the exception.
   const pluginModules = useContext(PluginsContext);
 
-  const themes = useMemo(
-    () =>
-      pluginModules == null ? null : getThemeDataFromPlugins(pluginModules),
-    [pluginModules]
-  );
+  const themes = useMemo(() => {
+    // Get theme from parent window via `postMessage` apis
+    if (isParentThemeEnabled) {
+      if (isParentThemePending) {
+        return null;
+      }
+
+      return parentThemeData ? [parentThemeData] : [];
+    }
+
+    // Get themes from plugins
+    return pluginModules == null
+      ? null
+      : getThemeDataFromPlugins(pluginModules);
+  }, [
+    isParentThemeEnabled,
+    isParentThemePending,
+    parentThemeData,
+    pluginModules,
+  ]);
 
   const settings = useAppSelector(getSettings);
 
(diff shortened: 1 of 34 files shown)


packages/utils/src/MessageUtils.ts
bmingles, line 1: Most of the functions in this module were pre-existing and moved from @deephaven/app-util. The only new function is `sendMessageToParent`

packages/components/src/theme/ThemeProvider.tsx
```diff
@@ -36,12 +36,14 @@ export interface ThemeProviderProps {
    * tell the provider to activate the base themes.
    */
   themes: ThemeData[] | null;
+  waitForActivation?: boolean;
```
mofojed, line 39: As mentioned in my other comment - do we really need this prop on ThemeProvider? If we do, add a comment about it and what the component does if this is true.

packages/components/src/theme/useExternalTheme.ts
```diff
@@ -0,0 +1,90 @@
...
+      .then(handleExternalThemeData)
+      .catch(err => {
+        logger.error(err);
+        setResult({ isEnabled: true, isPending: false });
+      });
+
+    return () => {
+      window.removeEventListener('message', onMessage);
+    };
+  }, [result.isEnabled]);
```
mofojed, line 85: I'd probably refactor this effect to make it more concise... the `handleExternalThemeData` doesn't need to be within it...
```typescript
const handleExternalThemeData = useCallback((externalThemeData: ExternalThemeData) => ...);

useEffect(function startListeningWindowMessages() {
   function onMessage(...) { ... }
  window.addEventListener('message', onMessage);
  return () => { window.removeEventListener( .... );
}, [handleExternalThemeData]);

useEffect(function initializeExternalTheme() {
  ...
  request.ExternalThemeData().then(handleExternalThemeData).catch(...)
}, [handleExternalThemeData]);
```

packages/jsapi-utils/src/MessageUtils.ts
```diff
@@ -25,34 +21,14 @@ export const BROADCAST_LOGIN_MESSAGE = 'io.deephaven.broadcast.Login';
...
-
-export function isMessage(obj: unknown): obj is Message {
-  const message = obj as Message;
-  return (
-    message != null &&
-    typeof message.id === 'string' &&
-    typeof message.message === 'string'
-  );
-}
-
```
mofojed, line 55: Removing all this stuff is a breaking change. Deprecate it instead (and re-export from utils where possible).

mofojed (changes requested):

packages/app-utils/src/components/ThemeBootstrap.tsx
```diff
@@ -1,32 +1,30 @@
...
-    [pluginModules]
-  );
-
-  const settings = useAppSelector(getSettings);
+  const themes = useCustomThemes(pluginModules);
+  const waitForActivation = isExternalThemeEnabled();
 
   return (
-    <ThemeProvider themes={themes}>
+    <ThemeProvider themes={themes} waitForActivation={waitForActivation}>
```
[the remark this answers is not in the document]
bmingles: @mofojed The reason we need a way to specify is for plugin themes. Since they get loaded after login, we have to be able to render children to show the login form while they are still null. For external themes, we can use the null check and only render children if they are there since they don't depend on login.

I think you are right in that we don't have to pass it down, but as far as I can tell, we do need some mechanism to differentiate.

I'll update accordingly and add more comments, unless you are seeing a better way.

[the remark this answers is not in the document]
bmingles: Actually, looks like we do need `waitForActivation` in ThemeProvider since the condition can't be determined from the passed in `themes` prop. I've added a more detailed comment to the `waitForActivation` prop.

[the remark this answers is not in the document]
bmingles: See notes on other comment.

packages/components/src/theme/useExternalTheme.ts
bmingles: Fixed

mofojed (approved):

PR #2429, opened 2025-05-01
fix: DH-19350: Ruff tries to lint before initializing when disabled

mattrunyon:
Looks like it was an issue caused if Ruff was disabled when we changed settings (we set settings on initial load). This was causing a check to pass and then trying to use Ruff without initializing it

Also added some log suppression because Ruff internals call `console.log` when the workspace is created and it's not useful to us. It's something along the lines of `glob 25 created ...` and I just found it unnecessary noise

### DIFF
--- a/packages/console/src/monaco/MonacoProviders.tsx
+++ b/packages/console/src/monaco/MonacoProviders.tsx
@@ -71,9 +71,6 @@ class MonacoProviders extends PureComponent<
       MonacoProviders.ruffWorkspace = new Workspace(
         MonacoProviders.ruffSettings
       );
-    } catch (e) {
-      console.log = prevLog; // Restore console.log in case the re-throw isn't caught and finally doesn't run
-      throw e;
     } finally {
       console.log = prevLog; // Restore console.log
     }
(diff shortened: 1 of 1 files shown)


packages/console/src/monaco/MonacoProviders.tsx
```diff
@@ -62,17 +83,12 @@ class MonacoProviders extends PureComponent<
   ): Promise<void> {
     MonacoProviders.ruffSettings = settings;
 
-    // Ruff has not been initialized yet
-    if (
-      MonacoProviders.ruffWorkspace == null &&
-      MonacoProviders.isRuffEnabled
-    ) {
+    if (!MonacoProviders.isRuffInitialized) {
       MonacoProviders.initRuff();
```
bmingles, line 87: Should we be awaiting this Promise? Otherwise not suer why setRuffSettings is marked as `async`

bmingles (changes requested):
Left a couple of comments

packages/console/src/monaco/MonacoProviders.tsx
mattrunyon: Nope doesn't look like we await it where it's called either (confirmed in enterprise as well).

Probably originally I wrote it with something async and just didn't catch that it was unnecessary in a refactor. Looks like there's an eslint rule if we wanted it [here](https://typescript-eslint.io/rules/require-await/)

Instead this now initializes ruff if needed which will apply the linter after initialization finishes, or if already initialized update the workspace which will do the linting. Awaiting the promise and removing the return would run the linter twice in quick succession.

bmingles (approved):
LGTM

PR #2433, opened 2025-05-06
fix: DH-18562: Deephaven picker hover blends with non-hovered styling in dark mode

gzh2003:
DH-18562: Fixes issue with picker styling when using dark mode. Previously the background did not serve as a good indicator when hovering over items.

I checked in with @dsmmcken to confirm the styling changes. I also tested the picker component in both the style guide and theme selector under settings.

**Hovered Styling (Before)**
Before Screenshot

**Hovered Styling (After)**
After Screenshot

### DIFF
--- a/packages/components/src/theme/theme-dark/theme-dark-semantic.css
+++ b/packages/components/src/theme/theme-dark/theme-dark-semantic.css
@@ -148,7 +148,7 @@
   );
   --dh-color-overlay-hover-bg: color-mix(
     in srgb,
-    var(--dh-color-black) 8%,
+    var(--dh-color-white) 8%,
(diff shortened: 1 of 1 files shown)


gzh2003:

mofojed (approved):
I assume you checked with @dsmmcken about which colour variable to use.

In the PR description, I encourage including details about how you tested a particular ticket (e.g. "Checked using the theme selector under settings") so it's clear what has been tested (for both the current reviewer, and yourself or any other developer who may come across this ticket in the future); and for a ticket like this which is a small visual change, a screenshot of before/after is also good.

Looks great, great work!

PR #2434, opened 2025-05-06
feat: Add keepMounted param to TabPanels for persisting panels when they are not active

mattrunyon:
Part of DH-18349. Will need a small update in dh.ui to enable support there which is what the ticket wants.

Does not change default behavior of `TabPanels`, just adds an extra param to keep static panels mounted (in React, but not actually mounted in the DOM).

There are unit tests, but can also test with local plugins from this PR https://github.com/deephaven/deephaven-plugins/pull/1177

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24996,6 +24996,15 @@
       "version": "17.0.2",
       "license": "MIT"
     },
+    "node_modules/react-reverse-portal": {
+      "version": "2.3.0",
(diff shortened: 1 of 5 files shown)


dgodinez-dh (approved):

mattrunyon:
Found a potential (although probably extremely unlikely to happen) leak if using different keys to add new panels.

Added a test to ensure components unmount if they aren't in the actual panel any more.

dgodinez-dh (approved):

PR #2437, opened 2025-05-07
fix: Simple Pivot prerequisites

vbabich:
- Reset `movedColumns` in `IrisGrid` and re-calculate metrics on `TABLE_CHANGED` event
- Fix the issue with grid metrics calculations based on stale state in `Grid.componentDidUpdate`
- Disable advanced filters for non-filterable columns
- Expose `DisplayColumn` type export
- Trigger `COLUMNS_CHANGED` instead of `TABLE_CHANGED` in `updateFrozenColumns`


mofojed (dismissed):

mofojed:
@vbabich Might need to update snapshots... for some reason I can't download the playwright report right now.

packages/iris-grid/src/IrisGridTableModel.ts
```diff
@@ -196,7 +196,7 @@ class IrisGridTableModel
   updateFrozenColumns(columns: ColumnName[]): void {
     this.userFrozenColumns = columns;
     this.dispatchEvent(
-      new EventShimCustomEvent(IrisGridModel.EVENT.TABLE_CHANGED, {
+      new EventShimCustomEvent(IrisGridModel.EVENT.COLUMNS_CHANGED, {
         detail: this.table,
       })
```
mofojed, line 201: This isn't right - it should be emitting the table columns, not the table itself. 
```suggestion
      new EventShimCustomEvent(IrisGridModel.EVENT.COLUMNS_CHANGED, {
        detail: this.columns,
      })
```

mofojed (changes requested):

packages/iris-grid/src/IrisGridTableModel.ts
vbabich: Updated, thanks!

mofojed (approved):

PR #2438, opened 2025-05-08
feat: Add input filter support to GridWidgetPlugin

mattrunyon:
This adds a few new hooks
- `useDhId` which will be used to route to a specific component in dh.ui (via prop), or the panel ID (via context) as a fallback. This uses `useFiber` which is not exported because it's an escape hatch that should be used very rarely. `useDhId` takes no props and can read the `__dhId` prop of the component via the `useFiber` hook.
- `useDashboardId` gives the ID of the current dashboard
- `useDashboardColumnFilters` is used to subscribe to column filter changes and also advertise the columns and table (if provided) available for filtering. This should allow us to hook up the existing filtering/linking to dh.ui and probably also `ChartBuilder` in dh.ui tables.

The contexts for `useDhId` and `useDashboardId` are added to `Dashboard` so there shouldn't be any changes for enterprise since it uses the `Dashboard` component.

Also reworked the events around this to have stronger types using the helper methods we added.

Test with dh.ui single normal table in a panel (ui.table support PR will follow in plugins repo)

Run this. Then add an input filter via the UI and the columns from the table should be filterable in the table
```py
from deephaven import ui

_stocks = dx.data.stocks()

t = ui.panel(_stocks, title="UI Panel")
```


packages/golden-layout/src/utils/EventUtils.ts
```diff
@@ -1,13 +1,12 @@
...
 import EventEmitter from './EventEmitter';
 
 type AsArray<P> = P extends unknown[] ? P : [P];
 
-export type EventHandlerFunction<P = []> = (...parameters: AsArray<P>) => void;
 export type EventListenerRemover = () => void;
 export type EventListenFunction<TParameters = []> = (
   eventEmitter: EventEmitter,
-  handler: EventHandlerFunction<TParameters>
+  handler: (...params: AsArray<TParameters>) => void
```
mattrunyon, line 9: Just going to comment on this now. I changed this so the type errors are more friendly. With the type alias, you end up with something like `Argument '(id: int) => void' is not assignable to EventHandlerFunction<[id: string]>`.

By removing that type alias, you get a much cleaner error like `Argument '(id: int) => void' is not assignable to '(id: string) => void'`

packages/iris-grid/src/IrisGridUtils.ts
```diff
@@ -418,7 +418,7 @@ class IrisGridUtils {
   }
 
   static getInputFiltersForColumns(
-    columns: readonly DhType.Column[],
+    columns: readonly { name: string; type: string }[],
```
mofojed, line 421: Hrm. Another one of these types, though we don't want `iris-grid` depending on `dashboard-core-plugins`.
We've got out `@deephaven/filters` package - maybe makes sense to pull the input filter type definitions down and use them in both `iris-grid`/`dashboard-core-plugins`?

packages/golden-layout/src/utils/EventUtils.ts
mattrunyon: I find the first half of the error clearer, but I guess the last half with the specific incompatible param is the same. First half is just clear it's a function and what the params are. Not that `EventHandlerFunction` isn't fairly obvious it's a function, just less so than `(params) => void`

What difference would this make for defining the handler functions? You still just call `makeEventFunctions` the same way as previously.

I can put the alias back, I just found the error more direct without it

packages/dashboard-core-plugins/src/GridWidgetPlugin.tsx
```diff
@@ -82,6 +87,25 @@ export function GridWidgetPlugin({
     [fetchResult, setState, dehydrateIrisGridState]
   );
 
+  const inputFilters = useDashboardColumnFilters(
+    fetchResult.status === 'success' ? fetchResult.model.columns : EMPTY_ARRAY
+  );
+
+  const [irisGrid, setIrisGrid] = useState<IrisGridType | null>(null);
```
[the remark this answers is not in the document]
mattrunyon: Ya I think I pulled the logic I used in `UITable` which needs to know that `irisGrid` exists to create mouse handlers

I see there's a caveat on the React docs about callback refs mentioning inline functions will get called w/ null and then the value on updates. I should probably update the `UITable` version of this to avoid that (although it's possible the multiple set state calls in the render will result in a no-op)

packages/dashboard-core-plugins/src/GridWidgetPlugin.tsx
```diff
@@ -82,6 +87,25 @@ export function GridWidgetPlugin({
     [fetchResult, setState, dehydrateIrisGridState]
   );
 
+  const inputFilters = useDashboardColumnFilters(
```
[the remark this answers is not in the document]
mattrunyon: Ya I can swap it to a warn.

The issue is we mount all the panels under the `DashboardPlugin` for dh.ui which is above the `Dashboard` component in the tree (I think it's a sibling)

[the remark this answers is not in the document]
mattrunyon: Looks like the `PortalPanel` does get the context, but `ReactPanel` does not because it's in a different part of the tree from the portal panel

The structure ends up like this

```tsx
<Dashboard>
  <DashboardLayout>
    {goldenLayoutPanls} // Includes PortalPanels. These are all wrapped in DhIdProvider
    <DhUIDashboardPlugin> // Not a panel, so no wrapper
      {DhUIReactPanels} // Where useDhId is called from
    </DhUIDashboardPlugin>
  </DashboardLayout>
</Dashboard>
```

packages/dashboard-core-plugins/src/FilterEvents.ts
```diff
@@ -0,0 +1,37 @@
...
+} from './FilterPlugin';
+
+export const {
+  listen: listenForFilterColumnsChanged,
+  emit: emitFilterColumnsChanged,
+  useListener: useFilterColumnsChangedListener,
+} = makeEventFunctions<
+  [
+    sourceId: FilterColumnSourceId,
+    columns: readonly { name: string; type: string }[] | null,
```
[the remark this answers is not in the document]
mattrunyon: `null` clears the entry in the respective maps when unmounting of a component.

For panels, that's handled by waiting for the panel to unmount, but that won't always be the case in dh.ui. Also, the ID of the panel will be different than the ID of the components inside of it for dh.ui, so panel unmount would do nothing useful for dh.ui case.

`null` is table/component went away. Empty array would be the table/component exists, but has no columns that can be filtered.

packages/iris-grid/src/IrisGridUtils.ts
mattrunyon: This util doesn't even need to be in `IrisGridUtils`. It's only used by the panel and the new hook which are both in `dashboard-core-plugins`. I don't see it used anywhere in enterprise

For relatively simple types like this, I'm not opposed to just having it as an inline literal. The util doesn't actually care that it's an input filter column or a table column or a dropdown filter column. It cares that it has those 2 props and all those other types satisfy that constraint

Originally I had all of these as just `dh.Column[]`, but that is not the case when using it within charts. Chart has a `FilterColumnMap` type that has values `{ name: string; type: string }`

packages/dashboard-core-plugins/src/FilterEvents.ts
```diff
@@ -0,0 +1,37 @@
+import { makeEventFunctions } from '@deephaven/golden-layout';
+import { type dh } from '@deephaven/jsapi-types';
+import { InputFilterEvent } from './events';
+import {
+  type FilterChangeEvent,
+  type FilterColumnSourceId,
+} from './FilterPlugin';
```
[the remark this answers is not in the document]
mattrunyon: I'm not sure how I feel about the branded type for the ID in the first place. I'm not sure if we're really making the mistake of passing other IDs or not. The one place I think a branded type would be useful is grid/iris-grid stuff around `ModelIndex` and `VisibleIndex` just because those are quite easy to mix up and we use both in the same functions a lot.

The separate type imports was just how eslint auto-fixed the issue. Probably so you can still easily add a non-type import. But I can change it if you want

packages/dashboard-core-plugins/src/GridWidgetPlugin.test.tsx
```diff
@@ -22,7 +31,7 @@ jest.mock('@deephaven/iris-grid', () => {
 });
 
 it('mounts without crashing', async () => {
-  const table = TestUtils.createMockProxy<DhType.Table>();
+  const table = TestUtils.createMockProxy<DhType.Table>({ columns: [] });
```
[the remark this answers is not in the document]
mattrunyon: There's a test in `useDhId.test.ts` which tests with the provider.

For this test I just mocked the `useDhId` hook because I needed several contexts or wrapping with `Dashboard` which causes other issues (like you then need to add a layout with this component). The other option was expose `useFiber` from the dashboard package so I could use the context outside of `Dashboard`, but I didn't want to encourage using it outside of its 1 usage

packages/iris-grid/src/IrisGridUtils.ts
mofojed: Yea it probably doesn't really make sense in IrisGridUtils. But no need to make that a breaking change here.

packages/dashboard/src/useDhId.ts
```diff
@@ -32,9 +35,10 @@ export function useDhId(): string {
   }
 
   if (dhId == null) {
-    throw new Error(
-      `useDhId must be used within a DhIdContext provider if there is no ${DH_ID_PROP} prop`
+    log.warn(
+      `useDhId must be used within a DhIdContext provider if there is no ${DH_ID_PROP} prop. Defaulting to empty string.`
     );
+    return '';
```
mofojed, line 41: Ehhh I don't like returning an empty string either - should return a `string | null`, or even two signatures with a param e.g. `useDhId(throwOnNull: false): string | null` and `useDhId(throwOnNull: true): string`

In `useDashboardColumnFilters` we shouldn't be emitting events with a blank or null `panelId` I don't think.

mofojed (changes requested):

packages/dashboard/src/useDhId.ts
mattrunyon: The original intent was to throw always on no ID, but that breaks dh.ui without the update to dh.ui to pass through the panel ID

I've just switched to `string | null` return for now. Although it's annoying using the branded type for `FilterColumnSourceId` because I have to cast when calling `useDhId` and changing the return type to `string | null` didn't cause the original cast to fail.

I am in favor of removing the branded type for `FilterColumnSourceId`. I don't think it's somewhere we will commonly mix up ID types and they're both strings at the end of the day. The filter doesn't care if it's a panel or widget ID, it just needs a unique identifier

packages/golden-layout/src/utils/EventUtils.ts
mattrunyon: Ehh it's in 3 places and I think `EventUtils` will be read/used much less frequently than adding new listeners and calling them will be. But I can change it back

packages/dashboard-core-plugins/src/useDashboardColumnFilters.ts
```diff
@@ -0,0 +1,81 @@
...
+  );
+
+  useEffect(function tableChanged() {
+    if (table == null || panelId == null) {
+      return;
+    }
+    emitFilterTableChanged(eventHub, panelId, table);
+
+    return () => emitFilterTableChanged(eventHub, panelId, null);
+  });
```
mofojed, line 54:
```suggestion
  }, [eventHub, panelId, table]);
```

mofojed (changes requested):

mofojed (approved):

PR #2441, opened 2025-05-14
fix: DH-19392: resolve interaction issue caused by nested radio content

gzh2003:
DH-19392: Previously, form controls were nested as children of the Spectrum `Radio` component, which made them not clickable. The fix moves the form controls outside the `Radio` component.

### DIFF
--- a/packages/iris-grid/src/sidebar/TableCsvExporter.scss
+++ b/packages/iris-grid/src/sidebar/TableCsvExporter.scss
@@ -10,9 +10,9 @@
   .radio-input-row {
     display: flex;
     align-items: center;
-    margin-left: 10px;
-    margin-top: 7px;
-    margin-right: 8px;
+    margin-left: var(--spectrum-global-dimension-size-125);;
+    margin-top: var(--spectrum-global-dimension-size-85);
+    margin-right: $spacer-2;
 
     .form-control {
       min-width: 0;
--- a/packages/iris-grid/src/sidebar/TableCsvExporter.scss
+++ b/packages/iris-grid/src/sidebar/TableCsvExporter.scss
@@ -10,6 +10,9 @@
   .radio-input-row {
     display: flex;
     align-items: center;
+    margin-left: 10px;
+    margin-top: 7px;
+    margin-right: 8px;
 
     .form-control {
       min-width: 0;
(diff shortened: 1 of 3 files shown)


packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -388,10 +388,16 @@ class TableCsvExporter extends Component<
...
+            <div className="d-flex">
+              <Radio
+                value={TableCsvExporter.DOWNLOAD_ROW_OPTIONS.CUSTOMIZED_ROWS}
+                data-testid="radio-csv-exporter-customized-rows"
+                UNSAFE_className="mr-2 pr-1"
+                // Empty aria-label workaround is needed to satisfy accessibility requirement since the
+                // Radio's label is rendered separately. This prevents accessibility errors that
+                // can trigger race conditions with Service Worker updates.
+                aria-label=" "
+              />
```
mofojed, line 400: Below, use a `Picker` instead of a `Select`, and a `NumberField` instead of the `input` below.

mofojed (changes requested):

packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -388,10 +388,16 @@ class TableCsvExporter extends Component<
...
-            >
+            <div className="d-flex">
+              <Radio
+                value={TableCsvExporter.DOWNLOAD_ROW_OPTIONS.CUSTOMIZED_ROWS}
+                data-testid="radio-csv-exporter-customized-rows"
+                UNSAFE_className="mr-2 pr-1"
+                // Empty aria-label workaround is needed to satisfy accessibility requirement since the
+                // Radio's label is rendered separately. This prevents accessibility errors that
+                // can trigger race conditions with Service Worker updates.
+                aria-label=" "
```
mofojed, line 399: Should not have a blank aria-label
```suggestion
                aria-label="First or last number of rows"
```

packages/iris-grid/src/sidebar/TableCsvExporter.scss
gzh2003, line 1: The explicit CSS declarations come from the computed margin values of the Radio's generated label. Since this isn't Spectrum there wasn't any variable I could draw from.

mofojed: @dsmmcken is there something Germain should be pulling from?

dsmmcken: var(--spectrum-global-dimension-size-85);
and
var(--spectrum-global-dimension-size-125);

for right, just use our $spacer or $spacer-2 not sure which works better.

mofojed (approved):

PR #2447, opened 2025-05-15
refactor: Change iris-grid reverseType to reverse and deprecate reverseType (#2278)

mattrunyon:
Cherry pick 3 of 5 (I think) for usePersistentState. Wasn't a completely clean pick, but verified against the original PR that the expected changes are all that changed.

This shouldn't be breaking. `reverseType` prop is deprecated, but still accepted. `reverseType` in saved states will be changed into `reverse` and then dropped from the saved state.

Tested this by exporting a layout with a table with reverse from the current DHC release. Then imported the layout to my dev UI and checked the reverse persisted.

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -46,10 +46,8 @@ import {
   PartitionConfig,
 } from '@deephaven/iris-grid';
 import {
-  AdvancedFilterOptions,
-  FormattingRule,
-  ReverseType,
-  TableUtils,
+  type AdvancedFilterOptions,
+  type FormattingRule,
 } from '@deephaven/jsapi-utils';
 import Log from '@deephaven/log';
 import {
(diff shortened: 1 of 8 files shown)


mofojed (approved):

mofojed:
@mattrunyon just a note we screwed up a bit here - we said we deprecated the `reverseType` type, but we just removed it from `IrisGridState` which is exported without deprecating it. Caused a build break when updating Enterprise:
```
> Task :web-client-ui:ideClient
[types] src/querymonitor/tabs/QueryLogsTab.tsx(137,3): error TS2344: Type '"customColumns" | "sorts" | "reverseType" | "columnHeaderGroups" | "pendingDataMap" | "advancedFilters" | "isFilterBarShown" | "conditionalFormats" | "customColumnFormatMap" | ... 9 more ... | "frozenColumns"' does not satisfy the constraint 'keyof IrisGridState'.
[types]   Type '"reverseType"' is not assignable to type 'keyof IrisGridState'.
[types] src/querymonitor/tabs/QueryLogsTab.tsx(414,11): error TS2345: Argument of type 'HydratedIrisGridState' is not assignable to parameter of type 'import("/home/jenkins/agent/workspace/iris/web/client-ui/ci_build/node_modules/@deephaven/iris-grid/dist/IrisGridUtils").HydratedIrisGridState'.
[types]   Property 'reverse' is missing in type 'HydratedIrisGridState' but required in type 'import("/home/jenkins/agent/workspace/iris/web/client-ui/ci_build/node_modules/@deephaven/iris-grid/dist/IrisGridUtils").HydratedIrisGridState'.
```
Not a big enough issue for me to want to have a patch in this branch, but just another example that we should be more careful.

PR #2456, opened 2025-05-30
feat: DH-18840: Add dx linker support

jnumainville:
This requires https://github.com/deephaven/deephaven-plugins/pull/1185 for full support, but this is all that's needed to support linker directly with those changes.


mattrunyon:
Yes I'm removing link type/validity based on panel component and instead it's emitted by the corresponding component when selection occurs. So this will go away after my changes

PR #2457, opened 2025-06-05
fix: DH-18976: Add proper bounds to pendingRow logic

gzh2003:
Part of DH-18976. Fixes an issue where the pendingRow check didn't account for totals rows, causing aggregations to be incorrectly flagged as pending in the color logic.

### DIFF
--- a/packages/iris-grid/src/IrisGridTableModelTemplate.ts
+++ b/packages/iris-grid/src/IrisGridTableModelTemplate.ts
@@ -1648,6 +1648,23 @@ class IrisGridTableModelTemplate<
     return false;
   }
 
+  isAppendableRange(range: GridRange): boolean {
+    // Make sure we have an input table and a valid range
+    if (
+      this.inputTable == null ||
+      range.startRow == null ||
+      range.endRow == null
+    ) {
+      return false;
+    }
+
+    // Check that the edit is in the appendable area
+    const pendingAreaStart = this.floatingTopRowCount + this.table.size;
+    return (
+      range.startRow >= pendingAreaStart && range.endRow >= pendingAreaStart
+    );
+  }
+
   isEditableRange(range: GridRange): boolean {
     // Make sure we have an input table and a valid range
     if (
@@ -1896,32 +1913,15 @@ class IrisGridTableModelTemplate<
   async setValues(edits: readonly EditOperation[] = []): Promise<void> {
     log.debug('setValues(', edits, ')');
 
-    const invalidEdits = edits.filter(edit => {
-      const column = edit.column ?? edit.x;
-      const row = edit.row ?? edit.y;
-      const range = GridRange.makeCell(column, row);
-
-      if (this.isEditableRange(range)) {
-        return false;
-      }
-
-      if (
-        this.inputTable != null &&
-        range.startRow != null &&
-        range.startColumn != null
-      ) {
-        const pendingAreaStart = this.floatingTopRowCount + this.table.size;
-        if (range.startRow >= pendingAreaStart) {
-          const isKeyColumn = this.isKeyColumn(range.startColumn);
-          const hasKeyColumns = this.keyColumnSet.size > 0;
-          return hasKeyColumns && isKeyColumn;
-        }
-      }
-
-      return true;
-    });
-
-    if (invalidEdits.length > 0) {
+    if (
+      !edits.every(edit => {
+        const cell = GridRange.makeCell(
+          edit.column ?? edit.x,
+          edit.row ?? edit.y
+        );
+        return this.isEditableRange(cell) || this.isAppendableRange(cell);
+      })
+    ) {
       throw new Error(`Uneditable ranges ${edits}`);
     }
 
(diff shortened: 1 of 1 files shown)


packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1893,17 +1895,59 @@ class IrisGridTableModelTemplate<
...
+
+      return true;
+    });
+
+    if (invalidEdits.length > 0) {
       throw new Error(`Uneditable ranges ${edits}`);
     }
 
     try {
+      // Expand the pending area to fit all edits
```
gzh2003, line 1929: This was previously handled implicitly because this.pendingRow(y) would return true for any y above pendingRowStart, even if it wasn’t a valid pending row. With the new bounds check, this no longer applies, so the count needs to be updated.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1893,17 +1895,59 @@ class IrisGridTableModelTemplate<
...
+          return hasKeyColumns && isKeyColumn;
+        }
+      }
+
+      return true;
+    });
+
+    if (invalidEdits.length > 0) {
       throw new Error(`Uneditable ranges ${edits}`);
     }
```
mofojed, line 1926: Before the change we would throw on the first uneditable range, without needing to iterate through the remaining ranges. Now we're always iterating through every one, even though we throw if any of them are invalid. That might make sense if we included the list of invalid ranges in the thrown error, but we don't - we just list all the ranges.
This logic should just be incorporated into the `isEditableRange` method anyway, and this should just be calling that method (as it was before). Anything else that is checking for if something is an editable range should have the same logic.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1893,17 +1895,59 @@ class IrisGridTableModelTemplate<
...
+      if (
+        this.inputTable != null &&
+        range.startRow != null &&
+        range.startColumn != null
+      ) {
+        const pendingAreaStart = this.floatingTopRowCount + this.table.size;
+        if (range.startRow >= pendingAreaStart) {
+          const isKeyColumn = this.isKeyColumn(range.startColumn);
+          const hasKeyColumns = this.keyColumnSet.size > 0;
+          return hasKeyColumns && isKeyColumn;
```
mofojed, line 1917: Why are we doing a key column check here?

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1893,17 +1895,59 @@ class IrisGridTableModelTemplate<
 
   async setValues(edits: readonly EditOperation[] = []): Promise<void> {
```
mofojed, line 1896: Why is setValues being changed at all?

mofojed (changes requested):
I think the only thing you need for this fix would be the `pendingRow` change, no? The `setValues` changes seem to be fixing a different issue.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
gzh2003: The main issue was that there are certain paste operations that would attempt to set values in ranges that technically aren't editable yet, but are still valid paste operations. 

For example, if a user were to paste 30 rows into an input table where:
- Main table content: 20 rows
- Current pending area: 2 rows
- Total editable area: 22 rows

only the first 22 rows would be processed (filling the existing editable area), while the remaining 8 rows would be silently ignored because they fell outside the current pending row range. 

This behaviour didn't occur before as `isEditableRange` incorrectly returned true for these out-of-bounds rows. This allowed paste operations to proceed without restriction. This also ties into the issues with aggregations, as those extra rows were mistakenly included in calculations due to isEditableRange returning true when it shouldn't have.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1646,6 +1648,23 @@ class IrisGridTableModelTemplate<
...
+      return false;
+    }
+
+    // Check that the edit is in the appendable area
+    const pendingAreaStart = this.floatingTopRowCount + this.table.size;
+    return (
+      range.startRow >= pendingAreaStart && range.endRow >= pendingAreaStart
+    );
+  }
+
```
gzh2003, line 1667: Following the huddle yesterday, I figured we should probably still treat it as a separate case, but move the logic to its own method i.e. `isAppendableRange`.

I've also removed the keyed column checks from `isAppendableRange`. While this was the prior behavior, it actually went against the spec.

mofojed (changes requested):
Please update the Test Plan on the ticket and include what snippets you're testing with.
Now we're not just testing the aggregation totals rows are the correct colours, we need to also test the paste/pending logic on a large input table with aggregations applied as well.

mofojed (changes requested):
I don't like this change because it's addressing too much... I don't like the `isAppendableRange` logic added, and it only halfway addresses the issue where we apply aggregations to an input table.
File a ticket for issues with adding aggregations to an input table screws things up (e.g. the aggregations are appearing the "Pending rows", both with this fix and without it). For this change, let's simply address the aggregations getting flagged as pending rows.
For that I think all we need to do is check `this.isEditable` first - if the table isn't editable, then there are no pending rows, so we can return `null` right away.

For the new ticket where aggregations are screwy on an editable/input table... I think an appropriate fix there would be to simply limit the aggregations to only be pinned to the top, and not the bottom. That way we have the bottom of the table for pending rows, aggregations at the top, and we don't need to worry about conflicts or anything like that.

mofojed (approved):

PR #2458, opened 2025-06-10
feat: DH-19632: Add shortcuts to cycle stacks and panels

gzh2003:
Part of DH-19632. Introduces keyboard navigation to cycle between and within stacks. Navigating through shortcuts activates content items and respects any existing active selections.

`Ctrl + '` / `Ctrl + ;`: to cycle forward/backward between different stacks in the layout
`Ctrl + Shift + '` / `Ctrl + Shift + ;`: to cycle forward/backward between tabs within the currently focused stack

Changes:
- Create`NavigationEvent` system
- Implement stack and tab cycling handlers in `PanelManager`
- Add `getFocusedStack()` and `getFocusedStackIndex()` helper methods to `LayoutUtils` for focus detection
- Add `forceFocus` parameter that bypasses the normal `isFocusonShow` check

### DIFF
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -40,6 +40,10 @@ import {
   setDashboardData as setDashboardDataAction,
   setDashboardPluginData as setDashboardPluginDataAction,
   updateDashboardData as updateDashboardDataAction,
+  emitCycleToNextStack,
+  emitCycleToPreviousStack,
+  emitCycleToNextTab,
+  emitCycleToPreviousTab,
 } from '@deephaven/dashboard';
 import {
   ConsolePlugin,
@@ -441,19 +445,19 @@ export class AppMainContainer extends Component<
   }
 
   sendCycleStackForward(): void {
-    this.emitLayoutEvent(PanelEvent.CYCLE_TO_NEXT_STACK);
+    emitCycleToNextStack(this.getActiveEventHub());
   }
 
   sendCycleStackBackward(): void {
-    this.emitLayoutEvent(PanelEvent.CYCLE_TO_PREVIOUS_STACK);
+    emitCycleToPreviousStack(this.getActiveEventHub());
   }
 
   sendCycleTabForward(): void {
-    this.emitLayoutEvent(PanelEvent.CYCLE_TO_NEXT_TAB);
+    emitCycleToNextTab(this.getActiveEventHub());
   }
 
   sendCycleTabBackward(): void {
-    this.emitLayoutEvent(PanelEvent.CYCLE_TO_PREVIOUS_TAB);
+    emitCycleToPreviousTab(this.getActiveEventHub());
   }
 
   sendReopenLast(): void {
--- a/packages/code-studio/src/main/AppMainContainer.tsx
+++ b/packages/code-studio/src/main/AppMainContainer.tsx
@@ -14,6 +14,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import {
   ContextActions,
   GLOBAL_SHORTCUTS,
+  NAVIGATION_SHORTCUTS,
   Popper,
   type ContextAction,
   Button,
@@ -237,6 +238,34 @@ export class AppMainContainer extends Component<
         },
         {
           action: () => {
+            this.sendCycleStackForward();
+          },
+          shortcut: NAVIGATION_SHORTCUTS.CYCLE_TO_NEXT_STACK,
+          isGlobal: true,
+        },
+        {
+          action: () => {
+            this.sendCycleStackBackward();
+          },
+          shortcut: NAVIGATION_SHORTCUTS.CYCLE_TO_PREVIOUS_STACK,
+          isGlobal: true,
+        },
+        {
+          action: () => {
+            this.sendCycleTabForward();
+          },
+          shortcut: NAVIGATION_SHORTCUTS.CYCLE_TO_NEXT_TAB,
+          isGlobal: true,
+        },
+        {
+          action: () => {
+            this.sendCycleTabBackward();
+          },
+          shortcut: NAVIGATION_SHORTCUTS.CYCLE_TO_PREVIOUS_TAB,
+          isGlobal: true,
+        },
+        {
+          action: () => {
             this.sendReopenLast();
           },
           shortcut: GLOBAL_SHORTCUTS.REOPEN_CLOSED_PANEL,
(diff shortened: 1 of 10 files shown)


packages/components/src/shortcuts/NavigationShortcuts.ts
```diff
@@ -0,0 +1,35 @@
...
+    macShortcut: [MODIFIER.CMD, KEY.SINGLE_QUOTE],
+    isEditable: true,
+  }),
+  CYCLE_TO_PREVIOUS_STACK: ShortcutRegistry.createAndAdd({
+    id: 'NAVIGATION.CYCLE_TO_PREVIOUS_STACK',
+    name: 'Cycle To Previous Stack',
+    shortcut: [MODIFIER.CTRL, KEY.SEMICOLON],
+    macShortcut: [MODIFIER.CMD, KEY.SEMICOLON],
+    isEditable: true,
+  }),
```
gzh2003, line 18: On Mac, `Command + ;` is a shortcut to "Check Document Now" when grammar and spelling checks are enabled.

This isn't a frequently used shortcut for most users (especially within the context of the application), so overriding it is unlikely to cause any significant issues.

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -164,6 +164,56 @@ class LayoutUtils {
...
+    // will have to rely on this internal CSS class.
+    const focusedStack = allStacks.find(stack =>
+      stack.header.tabs.some(tab =>
+        tab.element[0].classList.contains('lm_focusin')
+      )
+    );
+
+    return focusedStack;
+  }
+
```
mofojed, line 216: I don't like rewriting the same logic/comment in here, but we're just using `findIndex` instead of `find`.

One way to rewrite this would be to make the `getFocusStackIndex` just take the `allStacks` as a parameter. The only other place `getFocusedStackIndex` is used is when you already have `allStacks` so it saves fetching all the stacks again as well.

E.g. something like:
```suggestion
  /**
   * Get the index of the stack that is currently focused
   * @param allStacks All the stacks
   * @returns The focused stack's index or -1 if not found
   */
  static getFocusedStackIndex(allStacks: Stack[]): number {
    // NOTE: We target the 'lm_focusin' class because GoldenLayout automatically applies this class
    // to tab elements when they receive focus. Until we enhance focus tracking in GoldenLayout, we
    // will have to rely on this internal CSS class.
    const foundIndex = allStacks.findIndex(stack =>
      stack.header.tabs.some(tab =>
        tab.element[0].classList.contains('lm_focusin')
      )
    );

    return foundIndex;
  }

  /**
   * Get the stack that is currently focused
   * @param layout GoldenLayout instance
   * @returns The focused stack or undefined if none found
   */
  static getFocusedStack(layout: GoldenLayout): Stack | undefined {
    const allStacks = LayoutUtils.getAllStackContainers(layout);
    const focusedStackIndex = getFocusedStackIndex(allStacks);
    return allStacks[focusedStackIndex];
  }
```

packages/dashboard/src/PanelManager.ts
```diff
@@ -271,6 +301,79 @@ class PanelManager {
     this.sendUpdate();
   }
 
+  handleCycleStack(direction: CycleDirection): void {
```
mofojed, line 304: This function isn't being used as an event handler (e.g. it's not being passed directly to `eventHub.on` or another event emitter, and you don't need to `.bind` it).
Just call this function `cycleStack` instead. We use the `prefix` handle specifically for event handlers, to match with React's guide: https://react.dev/learn/responding-to-events#adding-event-handlers
> Event handler functions:    
> - Are usually defined inside your components.
> - Have names that start with handle, followed by the name of the event.

```suggestion
  cycleStack(direction: CycleDirection): void {
```

packages/dashboard/src/PanelEvent.ts
```diff
@@ -70,6 +70,18 @@ export const PanelEvent = Object.freeze({
...
+  CYCLE_TO_NEXT_STACK: 'PanelEvent.CYCLE_TO_NEXT_STACK',
+
+  // Event to change focus to previous stack
+  CYCLE_TO_PREVIOUS_STACK: 'PanelEvent.CYCLE_TO_PREVIOUS_STACK',
+
+  // Event to change focus to next tab
+  CYCLE_TO_NEXT_TAB: 'PanelEvent.CYCLE_TO_NEXT_TAB',
+
+  // Event to change focus to previous tab
+  CYCLE_TO_PREVIOUS_TAB: 'PanelEvent.CYCLE_TO_PREVIOUS_TAB',
```
mofojed, line 84: I'd prefer you create a new type of event, a `NavigationEvent`. And structure it like `DashboardEvents` instead of how `PanelEvent` is set up, then you can use the methods to wire it up.
`PanelEvent` is for an event occurring on a single panel; `DashboardEvent` is for something related to creating/opening dashboards; `NavigationEvent` seems like it should be separate.

packages/components/src/shortcuts/NavigationShortcuts.ts
mofojed: Nice, yea that's fine to override that keyword.

packages/code-studio/src/main/AppMainContainer.tsx
```diff
@@ -411,6 +440,22 @@ export class AppMainContainer extends Component<
     this.emitLayoutEvent(InputFilterEvent.CLEAR_ALL_FILTERS);
   }
 
+  sendCycleStackForward(): void {
```
mofojed, line 443: See my other change about refactoring to NavigationEvent, then instead of calling `this.emitLayoutEvent(...)` it'll just be `emitCycleToNextStack(this.layout)`

mofojed (changes requested):

packages/components/src/shortcuts/NavigationShortcuts.ts
```diff
@@ -0,0 +1,35 @@
...
+    id: 'NAVIGATION.CYCLE_TO_NEXT_TAB',
+    name: 'Cycle To Next Tab',
+    shortcut: [MODIFIER.CTRL, MODIFIER.SHIFT, KEY.SINGLE_QUOTE],
+    macShortcut: [MODIFIER.CMD, MODIFIER.SHIFT, KEY.SINGLE_QUOTE],
+    isEditable: true,
+  }),
+  CYCLE_TO_PREVIOUS_TAB: ShortcutRegistry.createAndAdd({
+    id: 'NAVIGATION.CYCLE_TO_PREVIOUS_TAB',
+    name: 'Cycle To Previous TAB',
+    shortcut: [MODIFIER.CTRL, MODIFIER.SHIFT, KEY.SEMICOLON],
```
mofojed, line 29: Hmm these Shifted keyboard shortcuts don't work for me on Linux. It needs to reference the double quotes and the colon instead (what's actually typed with the Shift). Those default shortcuts work for you on mac? Wonder if that's a bug in our code or it's a difference in how the OS passes us key events.

```suggestion
    shortcut: [MODIFIER.CTRL, MODIFIER.SHIFT, KEY.DOUBLE_QUOTES],
    macShortcut: [MODIFIER.CMD, MODIFIER.SHIFT, KEY.SINGLE_QUOTE],
    isEditable: true,
  }),
  CYCLE_TO_PREVIOUS_TAB: ShortcutRegistry.createAndAdd({
    id: 'NAVIGATION.CYCLE_TO_PREVIOUS_TAB',
    name: 'Cycle To Previous TAB',
    shortcut: [MODIFIER.CTRL, MODIFIER.SHIFT, KEY.COLON],
```

We'll want to get this tested on Windows as well.

mofojed (changes requested):

packages/dashboard/src/NavigationEvent.ts
```diff
@@ -0,0 +1,34 @@
+import { makeEventFunctions } from '@deephaven/golden-layout';
+
+const NavigationEvent = Object.freeze({
```
mofojed, line 3: I would've done these as separate consts, but I guess this makes sense to namespace them because we have the `index.ts` importing all the consts from a bunch of other files. Good work.

mofojed (approved):

PR #2459, opened 2025-06-10
feat: Add linker support to GridWidgetPlugin

mattrunyon:
This adds support for linker to `GridWidgetPlugin` so if you have tables within a dh.ui component (like 2 tables in a flex), you can link with them.

Cannot use ui.table yet (the hook from this should make it easy to implement). Also will not draw the link to the tab when hidden without a PR for dh.ui to add the panelId context.

Not going to add chart in this PR because there is a lot of code in `ChartPanel` that actually handles all of the linker interactions.


packages/dashboard/src/useDhId.ts
```diff
@@ -1,13 +1,11 @@
-import { createContext, useContext } from 'react';
 import Log from '@deephaven/log';
 import { useFiber } from './useFiber';
+import { usePanelId } from './usePanelId';
 
 const log = Log.module('useDhId');
 
 export const DH_ID_PROP = '__dhId';
 
-export const DhIdContext = createContext<string | null>(null);
```
mofojed, line 9: As discussed, don't rename in this PR.

PR #2461, opened 2025-06-11
fix: DH-19614: Use constituentType for tree table leaf node formatting 

gzh2003:
Part of DH-19614. Fixes an issue with rollups where tree table aggregated constituents were not being rendered based on the original column type. This change ensures that leaf nodes in tree tables use the appropriate column type for formatting and alignment.

Changes:
- Updated `colorForCell` to use `column.constituentType` instead of `column,type` for leaf node values
- Updated `textAlignForCell` to use `column.constituentType` for leaf node values
- Added type guard `isUITreeRow(row: unknown): row is UITreeRow` function in `IrisGridTreeTableModel.ts`


mofojed:
We'll want to add unit tests/e2e tests for that

gzh2003:
> We'll want to add unit tests/e2e tests for that

While working on the e2e tests, I discovered a separate formatting issue and created a new ticket for it: DH-19693. 

Since this issue isn’t directly related to the color and alignment changes in the original ticket, I’m planning to wrap it up without the e2e tests for now. Then, once DH-19693 is resolved, I'll complete the e2e tests and submit them in a separate PR.

PR #2464, opened 2025-06-18
fix: DH-19693: Add null/empty formatting to constituent strings

gzh2003:
Part of DH-19693. Anything that renders as "empty" for empty strings, and "null" for null strings, should still render the same when rolledup. This adds the missing null and empty string formatting logic for tree table constituent values.

Changes:
- Checks if the result should show "null" or "empty" based on formatter


mofojed (dismissed):
Good work. Ideally we'd also add a unit test for this.

PR #2467, opened 2025-06-18
fix: Enable golden-layout jest test and fix broken EventUtils test

mattrunyon:
Noticed we had migrated a test into golden-layout, but it never ran. At a future point I then broke the test (not cleaning up event handlers on unmount), but it wasn't running so it wasn't caught.

This enables the test and fixes the broken logic so the test passes.


bmingles (approved):
LGTM

PR #2471, opened 2025-06-23
fix: Grid making unnecessary onSelectionChanged calls

mattrunyon:
Needed as part of DH-19292 to ensure we don't emit duplicate selection events. Even with debouncing, we might emit the event, and then emit the same selection (in a different array reference) on mouse up. This fixes that so we don't emit the unnecessary events if the mouse up did not remove selections

There are still some quirks around the selection events like when using ctrl+click+drag to deselect part of a range, `onSelectionChanged` is emitted with the subtracted portion as an extra range (e.g., `[[0, 5], [1, 2]]`) until mouse up which commits the selection and removes the overlapping ranges (`[[0, 0], [3, 5]]`). On the dh.ui side I'm just consolidating the ranges, but would probably be good to fix that at some point.

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -1218,8 +1218,9 @@ class Grid extends PureComponent<GridProps, GridState> {
       return {
         cursorRow: newCursorRow,
         cursorColumn: newCursorColumn,
-        // If the selection ranges are non-overlapping, then selectedRanges already includes everything
-        // and the onSelectionChanged callback has already been called. No need to change to the same ranges in a new array.
+        // The onSelectionChanged callback has already been called with the selectedRanges at this point.
+        // If the selection is not changed (e.g., the user is adding via ctrl+click and not removing),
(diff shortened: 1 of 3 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -1206,10 +1209,18 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+        newSelectedRanges.length !== selectedRanges.length ||
+        newSelectedRanges.some(
+          (range, index) => !range.equals(selectedRanges[index])
+        );
+
       return {
         cursorRow: newCursorRow,
         cursorColumn: newCursorColumn,
-        selectedRanges: newSelectedRanges,
+        // If the selection ranges are non-overlapping, then selectedRanges already includes everything
```
bmingles, line 1221: What does "non-overlapping" mean here? Is that synonymous with all ranges are unchanged?

bmingles (approved):
LGTM

PR #2478, opened 2025-06-28
feat: Table plugin support for GridWidgetPlugin

mattrunyon:
Also fixes selection and panelState reactivity (for GridWidget, not IrisGridPanel). Could fix for IrisGridPanel here as well if desired.

Passes the panel name as `tableName` since tables do not actually have names. Mimics the `panel` prop passed to plugins to give `irisGrid` and `getTableName` which might have been previously used. Should keep most things backwards compatible unless there were other items from the panel being used.

One concern I have is the filters from plugins are completely invisible to users once they are applied. They don't go to quick filters or anywhere else, just applied with no visual. This probably makes sense as a separate ticket though.

## Testing

You can checkout my plugins branch https://github.com/mattrunyon/deephaven-plugins/tree/table-plugin-example and use the local JS plugin server.

Then use the following code to test (`stocks` will not update the counter state since it uses the `IrisGridPanel` version which doesn't update state prop to the component)

```py
import deephaven.ui as ui

stocks = dx.data.stocks().with_attributes({'PluginName' : 'table-example'})
t = ui.panel(stocks)
```

### DIFF
--- a/packages/dashboard-core-plugins/src/TablePluginWrapper.tsx
+++ b/packages/dashboard-core-plugins/src/TablePluginWrapper.tsx
@@ -54,6 +54,14 @@ export const TablePluginWrapper = forwardRef(
       return panelItem?.config.title ?? 'unknown';
     }, [layoutManager.root, panelId]);
 
+    const panel = useMemo(
+      () => ({
+        irisGrid: irisGridRef,
+        getTableName: () => panelName,
+      }),
+      [irisGridRef, panelName]
+    );
+
     return (
       <div className="iris-grid-plugin">
         <Plugin
@@ -70,10 +78,7 @@ export const TablePluginWrapper = forwardRef(
           // since we don't have an IrisGridPanel to use here.
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           // @ts-ignore
-          panel={{
-            irisGrid: irisGridRef,
-            getTableName: () => panelName,
-          }}
+          panel={panel}
         />
       </div>
     );
--- a/packages/dashboard-core-plugins/src/useIrisGridModel.ts
+++ b/packages/dashboard-core-plugins/src/useIrisGridModel.ts
@@ -8,12 +8,10 @@ export type IrisGridModelFetch = () => Promise<dh.Table>;
 export type IrisGridModelFetchErrorResult = {
   error: NonNullable<unknown>;
   status: 'error';
-  model: undefined;
 };
 
 export type IrisGridModelFetchLoadingResult = {
   status: 'loading';
-  model: undefined;
 };
 
 export type IrisGridModelFetchSuccessResult = {
--- a/packages/dashboard-core-plugins/src/useTablePlugin.tsx
+++ b/packages/dashboard-core-plugins/src/useTablePlugin.tsx
@@ -13,12 +13,35 @@ import { type GridRange } from '@deephaven/grid';
 import { TablePluginWrapper } from './TablePluginWrapper';
 
 interface UseTablePluginProps {
+  /**
+   * The IrisGrid model for this plugin.
+   * Currently only IrisGridTableModelTemplate types are supported.
+   * Other IrisGrid model types will be ignored for now.
+   */
   model: IrisGridModel | undefined;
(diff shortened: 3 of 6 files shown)


packages/dashboard-core-plugins/src/TablePluginWrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+          onStateChange={onStateChange}
+          pluginState={pluginState}
+          // Mimic the panel containing `irisGrid.current` for backwards compatibility
+          // since we don't have an IrisGridPanel to use here.
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          panel={{
+            irisGrid: irisGridRef,
+            getTableName: () => panelName,
+          }}
```
mofojed, line 76: Memoize.

packages/dashboard-core-plugins/src/useTablePlugin.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { TablePluginWrapper } from './TablePluginWrapper';
+
+interface UseTablePluginProps {
+  model: IrisGridModel | undefined;
+  irisGridRef: React.MutableRefObject<IrisGridType | null>;
+  irisGridUtils: IrisGridUtils;
+  selectedRanges: readonly GridRange[] | undefined;
+}
+
+export function useTablePlugin({
```
mofojed, line 22: Add some docs for this hook.

packages/dashboard-core-plugins/src/useIrisGridModel.ts
```diff
@@ -8,10 +8,12 @@ export type IrisGridModelFetch = () => Promise<dh.Table>;
 export type IrisGridModelFetchErrorResult = {
   error: NonNullable<unknown>;
   status: 'error';
+  model: undefined;
 };
 
 export type IrisGridModelFetchLoadingResult = {
   status: 'loading';
+  model: undefined;
```
mofojed, line 16: I don't like this - now if you use either of these specific types, it will suggest `.model` even though it's always `undefined` for this specific type.

image

Instead, I'd narrow the type first where you were doing this, e.g.
```typescript
// Currently:
const { model } = fetchResult;

// Instead do 
  const model =
    fetchResult.status === 'success' ? fetchResult.model : undefined;
```

I think that's more clear. You could also make it a type predicate, or a function to get model from the result, but adding the `model: undefined` here pollutes the types IMO

packages/dashboard/src/layout/LayoutUtils.ts
```diff
@@ -240,6 +240,36 @@ class LayoutUtils {
     return null;
   }
 
+  /**
+   * Gets a content item by its ID
+   * @param item Golden layout content item to search for the content item
+   * @param searchId the ID
+   */
+  static getContentItemById(
```
mofojed, line 248: I'm surprised this didn't already exist.

packages/plugin/src/TablePlugin.ts
```diff
@@ -29,7 +29,7 @@ export interface TablePluginProps<S = unknown> {
   /**
    * The model for the table this plugin is associated with.
    */
-  model: IrisGridModel;
+  model: IrisGridTableModelTemplate;
```
mofojed, line 32: Mmm at some point wish we could clean up our GridModel types. Should all be defined as types/interfaces instead of classes.

mofojed (changes requested):
Good overall, I haven't actually tested it with a plugin. If you update the Test Plan on the ticket attach a built plugin to make it easy for testers.

packages/dashboard/src/layout/LayoutUtils.ts
mattrunyon: It looks like we have `PanelManager.getContainerByPanelId` which is close. I'm going to leave this util I think because the closest we can do is the logic from that which gets the stack for the item with config containing the id, then gets the item from the stack

This also lets you get a row or column which wouldn't work with the other logic (since they're not in a stack), but I can't think of a useful reason for that.

PR #2481, opened 2025-07-09
feat: DH-19722: Better handling for large numbers of open dashboards

gzh2003:
For DH-19722. This PR adds a popper for searching and filtering through dashboards. It supports keyboard navigation and includes a global shortcut (`Ctrl+Shift+D` / `Cmd+Shift+D`) to toggle the quick filter popper.

Changes:
- Dashboard search: Filter dashboards by typing in the search input
- Global shortcut: Ctrl+Shift+D/Cmd+Shift+D opens the dashboard search popper (only when 2+ dashboards exist)
- Keyboard interactions: arrow keys navigate between dashboard items; enter sets the active dashboard to the currently focused item in the list (if no dashboard in the list is focused it will default to the first dashboard in the search)
- Adds 'endPlaceholder' to `SearchInput` to display the associated shortcut

### DIFF
--- a/packages/components/src/SearchInput.tsx
+++ b/packages/components/src/SearchInput.tsx
@@ -10,13 +10,15 @@ import { ContextActions } from './context-actions';
 interface SearchInputProps {
   value: string;
   placeholder: string;
-  endPlaceholder?: string; // Optional placeholder text shown on the right side of the input when empty
+  /** Placeholder text shown on the right side of the input when empty */
+  endPlaceholder?: string;
   onBlur?: React.FocusEventHandler<HTMLInputElement>;
   onChange: React.ChangeEventHandler<HTMLInputElement>;
   onKeyDown: React.KeyboardEventHandler<HTMLInputElement>;
   className: string;
   disabled?: boolean;
-  matchCount: number; // Number of search matches
+  /** Number of search matches, displayed at the end of the search field. */
+  matchCount: number;
   id: string;
   'data-testid'?: string;
   cursor?: {


packages/components/src/shortcuts/GlobalShortcuts.ts
```diff
@@ -65,6 +65,13 @@ const GLOBAL_SHORTCUTS = {
     macShortcut: [MODIFIER.CMD, MODIFIER.OPTION, MODIFIER.SHIFT, KEY.L],
     isEditable: true,
   }),
+  OPEN_DASHBOARD_SEARCH_MENU: ShortcutRegistry.createAndAdd({
+    id: 'GLOBAL.OPEN_DASHBOARD_SEARCH_MENU',
+    name: 'Open Dashboard Search Menu',
+    shortcut: [MODIFIER.CTRL, KEY.SLASH],
+    macShortcut: [MODIFIER.CMD, KEY.SLASH],
```
gzh2003, line 72: I found that `Ctrl + /` and `Cmd + /` fits in pretty well with the other dashboard shortcuts. Don also suggested using `Cmd + shift + D`

packages/components/src/shortcuts/GlobalShortcuts.ts
```diff
@@ -65,6 +65,13 @@ const GLOBAL_SHORTCUTS = {
     macShortcut: [MODIFIER.CMD, MODIFIER.OPTION, MODIFIER.SHIFT, KEY.L],
     isEditable: true,
   }),
+  OPEN_DASHBOARD_SEARCH_MENU: ShortcutRegistry.createAndAdd({
+    id: 'GLOBAL.OPEN_DASHBOARD_SEARCH_MENU',
+    name: 'Open Dashboard Search Menu',
+    shortcut: [MODIFIER.CTRL, MODIFIER.SHIFT, KEY.D],
+    macShortcut: [MODIFIER.CMD, MODIFIER.SHIFT, KEY.D],
```
gzh2003, line 72: Went with Don's suggestion to use `Cmd + Shift + D.` I also found that `Cmd + /` fits in pretty well with the other dashboard related shortcuts, considering that period and comma are right beside slash on the keyboard. (it's also used as a quick filter/search shortcut in other programs)

packages/components/src/navigation/DashboardList.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const handleTabSelect = useCallback(
+    (tab: NavTabItem) => {
+      onSelect(tab);
+    },
+    [onSelect]
+  );
+
+  const handleMouseDown = useCallback((event: React.MouseEvent) => {
+    // Prevent mousedown from taking focus away from the search input
```
gzh2003, line 65: I'm not a fan of mousedown stealing focus from the search input. The list items already have CSS styling for this, so I think retaining focus on the search input keeps things cleaner.

packages/components/src/SearchInput.tsx
```diff
@@ -10,12 +10,13 @@ import { ContextActions } from './context-actions';
 interface SearchInputProps {
   value: string;
   placeholder: string;
+  endPlaceholder?: string; // Optional placeholder text shown on the right side of the input when empty
```
mofojed, line 13: Two things:
1. The Doc string needs to go above the prop rather than in line, and also use `/**  */` instead of `//`, or else it does not appear in the autocomplete/docs. Right now if you try and use this prop:
image
vs. when you put the comment above and use `/**  */` style comment:
image
2. "Optional" wording is redundant here; it's optional as indicated by the `?` already. I'd just get rid of that. (This is a bit of a nitpick and really doesn't matter).
```suggestion
  /** Placeholder text shown on the right side of the input when empty */
  endPlaceholder?: string;
```

packages/components/src/SearchInput.tsx
```diff
@@ -10,12 +10,13 @@ import { ContextActions } from './context-actions';
...
   onBlur?: React.FocusEventHandler<HTMLInputElement>;
   onChange: React.ChangeEventHandler<HTMLInputElement>;
   onKeyDown: React.KeyboardEventHandler<HTMLInputElement>;
   className: string;
   disabled?: boolean;
-  matchCount: number;
+  matchCount: number; // Number of search matches
```
mofojed, line 19:
```suggestion
  /** Number of search matches, displayed at the end of the search field. */
  matchCount: number;
```

mofojed (changes requested):
Couple minor things, looks good

mofojed (approved):

PR #2482, opened 2025-07-09
fix: DH-19864 Close tables used by Picker and ComboBox

mattrunyon:
DH-19864 seems to be caused by not closing these tables making the JS API client unhappy and it refuses to send data after so many re-renders with pickers in dh.ui

Not reproducible in DHC, so I will branch this off v0.85, publish an alpha, and deploy a grizzly BHS to confirm. This fixed for me with `npm run start-community` in DHE

Colin and I are still going to track down why exactly the JS API is unhappy about these tables not being closed

### DIFF
--- a/packages/jsapi-components/src/spectrum/utils/usePickerProps.ts
+++ b/packages/jsapi-components/src/spectrum/utils/usePickerProps.ts
@@ -77,6 +77,9 @@ export function usePickerProps<TProps>({
     [tableSource]
   );
 
+  useEffect(() => () => tableSource?.close(), [tableSource]);
+  useEffect(() => () => tableCopy?.close(), [tableCopy]);
+
   const keyColumn = useMemo(
     () =>
       tableCopy == null ? null : getItemKeyColumn(tableCopy, keyColumnName),


mattrunyon:
@bmingles I figured you would know best if there are possibly other table uses in the picker/combo_box code that I missed

packages/jsapi-components/src/spectrum/utils/usePickerProps.ts
```diff
@@ -77,6 +77,9 @@ export function usePickerProps<TProps>({
     [tableSource]
   );
 
+  useEffect(() => () => tableSource?.close(), [tableSource]);
```
bmingles, line 80: I would recommend using `useTableClose` hook here. It's mostly the same except it checks if table is already closed before closing it. IIR, this avoids a console warning.

bmingles (changes requested):
~~I would use `useTableClose` hook.~~

I don't think it is safe to close the source table inside of this hook. The hook should only be responsible for the copy. It is the callers responsibility for managing the lifecycle of the table

packages/jsapi-components/src/spectrum/utils/usePickerProps.ts
bmingles: Oh actually, this is probably not a good thing to do. The tableSource is passed in and presumably managed from the outside. I don't think we want the hook to close the source.

mattrunyon: I moved it up to the components. Probably wouldn't have been a big deal since the hook is only used by these components, but makes sense to close it in the owner instead

packages/jsapi-components/src/spectrum/Picker.tsx
```diff
@@ -1,10 +1,13 @@
 import { PickerNormalized } from '@deephaven/components';
 import { type PickerProps } from './PickerProps';
 import { usePickerProps } from './utils';
+import useTableClose from '../useTableClose';
 
 export function Picker(props: PickerProps): JSX.Element {
   const pickerProps = usePickerProps<PickerProps>(props);
 
+  useTableClose(props.table);
```
bmingles, line 9: I think this is the same issue. The parent component using Picker may not want you to close the table. It's passed in as a reference that should be managed elsewhere.

packages/jsapi-components/src/spectrum/ComboBox.tsx
```diff
@@ -19,6 +20,8 @@ export function ComboBox(props: ComboBoxProps): JSX.Element {
     ...pickerProps
   } = usePickerProps<ComboBoxProps>(props);
 
+  useTableClose(props.table);
```
bmingles, line 23: Same here, closing the table should be owned by the creator of the table.

bmingles (changes requested):
I think closing tables still needs to be in the scope that creates it. Not safe to close a passed in prop.

packages/jsapi-components/src/spectrum/ComboBox.tsx
```diff
@@ -4,7 +4,7 @@ import {
   type NormalizedItem,
   type SpectrumComboBoxProps,
 } from '@deephaven/components';
-import { useCallback, useRef } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
```
bmingles, line 7: Looks like a remaining unused import.

mattrunyon:
It looks like we already close the table where dh.ui uses these components (I missed that initially). So shouldn't need a dh.ui update on top of this

PR #2483, opened 2025-07-10
fix: Re-export isMessage and deprecate it

mofojed:
- We broke this in https://github.com/deephaven/web-client-ui/pull/2425/files#diff-538c1b9fba1be844a9515eb2f507e4d14d56318d6be38bb96d166acafd45c810
  - It wasn't re-exporting isMessage, even though it imported it
- Just re-export isMessage, now #2425 is not a breaking change


bmingles (approved):
LGTM

PR #2489, opened 2025-07-14
fix: DH-19911: Fix initial isLoading state in usePromiseFactory when autoload is true

vbabich:
usePromiseFactory returns `loading = false` on first pass, even though it starts loading immediately


bmingles (approved):
LGTM

PR #2498, opened 2025-07-18
fix: Fixing ComponentType props

jnumainville:
Part of work for DH-19416 and a followup from #2477

If no props are specified for `React.ComponentType` it defaults to `{}`, allowing no props. We need to specify that we can have any.


mofojed (approved):
As long as you've tried this in your template and it's working then I think we're good.

jnumainville:
Yep I don't see any errors with the generated template

PR #2501, opened 2025-07-21
feat: DH-19897: Handle missing jsapi/dh-internal.js files on core workers

bmingles:
DH-19897: Handle missing jsapi/dh-internal.js files on core workers

I added a flag to resolve 404s as empty content instead of throwing an error. 404s were already differentiated from other errors that get retry attempts, so it was a simple change.

Testing
I tested locally by renaming `jsapi/dh-internal.js` in the `loadModules` call to `jsapi/dh-internal2.js` to mimick a 404. Confirmed empty content file gets created in VS Code extension.

I also tested stopping the server which produces a different error that invokes the retry mechanism instead of a 404. Worst case scenario, if we ever get out of sorts, restarting VS Code will delete the tmp download folder and retry everything again.

### DIFF
--- a/packages/jsapi-nodejs/src/loaderUtils.ts
+++ b/packages/jsapi-nodejs/src/loaderUtils.ts
@@ -5,13 +5,18 @@ import type { dh as DhType } from '@deephaven/jsapi-types';
 import { downloadFromURL, urlToDirectoryName } from './serverUtils.js';
 import { polyfillWs } from './polyfillWs.js';
 import { ensureDirectoriesExist, getDownloadPaths } from './fsUtils.js';
+import { HttpError } from './errorUtils.js';
 
 type NonEmptyArray<T> = [T, ...T[]];
+
+const DH_CORE_MODULE = 'jsapi/dh-core.js' as const;
+const DH_INTERNAL_MODULE = 'jsapi/dh-internal.js' as const;
 
 /** Transform downloaded content */
 export type PostDownloadTransform = (
   serverPath: string,
-  content: string
+  content: string,
+  error?: unknown
 ) => string;
 
 export type LoadModuleOptions = {
@@ -20,7 +25,7 @@ export type LoadModuleOptions = {
   download: boolean | PostDownloadTransform;
   storageDir: string;
   targetModuleType: 'esm' | 'cjs';
-  treat404asEmptyContent?: boolean;
+  handleErrorsInPostDownload?: boolean;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/jsapi-nodejs/src/loaderUtils.ts
```diff
@@ -118,6 +121,10 @@ export async function loadDhModules({
     serverPaths: ['jsapi/dh-core.js', 'jsapi/dh-internal.js'],
     storageDir,
     targetModuleType,
+    // jsapi/dh-internal.js will be removed from Core workers in the not too
+    // distant future. Treating 404s as empty content makes this compatible with
+    // both configurations.
+    treat404asEmptyContent: true,
```
mofojed, line 127: This parameter seems very specific to this situation. Pretty hacky solution.

mofojed (changes requested):
This is a pretty hacky solution; my only real issue is making this a breaking change, when we really should try our best not to be breaking APIs unnecessarily.
We could convert the function to take an object, but also still accept the old parameter list for example; or somehow do this so it doesn't require a breaking change... perhaps the post download transform could be used, to handle an error scenario and return content instead.

bmingles:
> This is a pretty hacky solution; my only real issue is making this a breaking change, when we really should try our best not to be breaking APIs unnecessarily. We could convert the function to take an object, but also still accept the old parameter list for example; or somehow do this so it doesn't require a breaking change... perhaps the post download transform could be used, to handle an error scenario and return content instead.

@mofojed I think I can make it such that the api contract doesn't have to change `loadModules` at all and that it doesn't have to be responsible for providing the empty content. I should be able to isolate changes to `loadDhModules`. 

The 1 piece in `loadDhModules` where I think we need to change the behavior is that we want the default behavior to be write empty content so that consumers going forward don't always have to configure this. This wouldn't change any of the function calls, but it does slightly change when errors are thrown. Not sure if this is considered a breaking change or not.

packages/jsapi-nodejs/src/loaderUtils.ts
```diff
@@ -133,7 +166,19 @@ export async function loadDhModules({
...
 
-            if (serverPath === 'jsapi/dh-internal.js') {
+            if (serverPath === DH_INTERNAL_MODULE) {
+              if (error != null) {
+                // `dh-internal.js` module is being removed from future versions
+                // of DH core, but there's not a great way for this library to
+                // know whether it's present or not. Treat 404s as empty content
+                // to make things compatible with both configurations.
+                if (error instanceof HttpError && error.statusCode === 404) {
+                  return '';
```
bmingles, line 176: This should really be the only behavior change from the consumers perspective, and I don't think there's any real scenarios where this file would 404 when a server is up and running except our future scenario when we remove the file.

mofojed (approved):

PR #2502, opened 2025-07-21
fix: DH-19601: Pin aggregations to the top in input tables

gzh2003:
For DH-19601. Input tables with aggregations in the footer were experiencing conflicts between aggregated data and insertions to the pending area. This pr pins aggregations to the top of input tables, that way we have the bottom of the table reserved for pending rows, and we don't need to worry about conflicts.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -864,6 +864,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       rollupConfig,
       rollupSelectedColumns: [],
       aggregationSettings:
+        // Fix aggregations to the top if the grid is editable, so that the bottom is reserved for pending rows
         isEditableGridModel(model) && model.isEditable
           ? { ...aggregationSettings, showOnTop: true }
           : aggregationSettings,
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/sidebar/aggregations/Aggregations.tsx
```diff
@@ -40,6 +40,7 @@ export type AggregationSettings = {
 
 export type AggregationsProps = {
   isRollup: boolean;
+  isEditable: boolean;
```
mofojed, line 43: Prop name should probably be more specific - as you can still edit the aggregations, you just can't edit the position.

```suggestion
  isPositionEditable: boolean;
```

Or even having something like `availablePositions` and passing in `top` or `bottom` and just disabling the ones that are not available, as even `isPositionEditable` does not convey that it will be pinned to the top.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4833,6 +4837,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
             <Aggregations
               settings={aggregationSettings}
               isRollup={isRollup}
+              isEditable={isEditableGridModel(model) && model.isEditable}
```
mofojed, line 4840: Ahh I see, you're passing in `isEditable`, referring to the grid being editable. From within the context of the `Aggregations` component, that doesn't really make sense; see my suggestion above about changing it to `availablePlacements` or `isPlacementDisabled`, or `disabledPlacements`, something like that.

mofojed (changes requested):

mofojed (approved):
Great work!

PR #2506, opened 2025-07-22
fix: DH-19917: reduce contrast of light theme disabled color

dsmmcken:
Changes light themes disabled text color to be an even lighter color. Changed to show more difference between enabled and disabled states.

Color change didn't seem to trigger enough difference to change e2e tests.


mofojed:
Re-ran the test because it looked like a Codecov issue, it has passed now.

PR #2509, opened 2025-07-23
feat: Add URI support and api associated with widget to useWidget

mattrunyon:
Part of DH-19001. Tested with my VM `mattrunyon-gplus` which has the updated dh.ui and DHE web UI as well. Used the notebook w/ the ticket number as several examples, and the `Demo` and `DemoDashboard` queries w/ different users with different permissions to view those queries. Tested using tables/figures/pickers and restarting the queries to ensure they restarted and could continue being used.

This is mostly type changes to augment the object fetcher stuff to accept a URI.

Also adds the associated `api` to the `useWidget` hook needed for PQ URIs in DHE.

Renamed `useTableClose` to `useWidgetClose` as a more generic util and deprecated `useTableClose`. They are the same, just more permissive types since this can apply to non-tables


mofojed (changes requested):
Ticket # / how did you test this?

mattrunyon:
Added info to the PR description. I'll update the Jira ticket with a detailed test plan

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -19,27 +22,36 @@ type WidgetTypes =
 /**
  * Wrapper object for a widget and error status. Both widget and error will be `null` if it is still loading.
  */
-type WidgetWrapper<T extends WidgetTypes = dh.Widget> = {
+export type WidgetWrapper<T extends WidgetTypes = dh.Widget> = {
   /** Widget object to retrieve */
   widget: T | null;
 
+  /** Deephaven JS API to use for this widget */
+  api: typeof dh | null;
```
mofojed, line 30: Good addition

packages/jsapi-bootstrap/src/useWidget.ts
```diff
@@ -97,7 +109,17 @@ export function useWidget<T extends WidgetTypes = dh.Widget>(
     [descriptor, objectFetch]
   );
 
-  return wrapper;
+  const wrapperWithApi = useMemo(
+    () => ({
+      ...wrapper,
+      error:
+        wrapper.error ?? (descriptorApiError as NonNullable<unknown> | null),
```
mofojed, line 116: Maybe move this to `useDeferredApi` where the error is defined there? Would be a little more correct I think, either the error is a non nullable unknown or it's null.

mofojed (approved):
Looks good, one minor suggestion

PR #2512, opened 2025-07-28
feat: Added displayName to React contexts

bmingles:
Added displayName to React contexts so that they can be identified in React dev tools.

Part of DH-19529

### Testing
- Run DH web locally via `npm start`
- Open Components tab from React dev tools
- Verify that context providers now have names in the component tree

### DIFF
--- a/packages/app-utils/src/components/ConnectionContext.ts
+++ b/packages/app-utils/src/components/ConnectionContext.ts
@@ -2,5 +2,6 @@ import { createContext } from 'react';
 import type { dh } from '@deephaven/jsapi-types';
 
 export const ConnectionContext = createContext<dh.IdeConnection | null>(null);
+ConnectionContext.displayName = 'ConnectionContext';
 
 export default ConnectionContext;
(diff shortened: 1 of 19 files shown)


mattrunyon (commented):
The branch has a different ticket number from the PR title. I'm also not sure how this change applies to the ticket in branch name or PR title.

We should try to formalize this, but at least what I've been doing if I needed multiple PRs for a fix is not putting the ticket name in the title of the first PR. So web-client-ui doesn't get the ticket number, the plugins repo does so the ticket doesn't get moved to RTT based on just a web-client-ui update.

Also, would we want to enable an eslint rule for this? We already have it for other components. Looks like it's just a boolean option we can enable [here](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/display-name.md#checkcontextobjects-default-false)

mattrunyon (approved):
Asked Bender and he said let's remove the ticket number from the PR title so it doesn't look like the ticket should be moved to RTT when the package update happens in DHE

I put the ticket number in the body though so it has a reference still, but only the title goes in the changelog on release

PR #2517, opened 2025-08-05
fix: Make ListViewWrapper icon CSS target only icons in rows

mattrunyon:
For `ui.resolve`, we're going to use the `renderEmptyState` prop to show the error message. This has an icon and was being shrunk to the row icon height. Instead this should make the CSS more specific to only affect icons within rows


bmingles (approved):
LGTM

PR #2524, opened 2025-08-18
feat: DH-19413: Support expandable columns in grids v0.85

vbabich:
Merge this after #2537

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -272,6 +272,11 @@ export interface IrisGridContextMenuData {
   modelColumn: GridRangeIndex;
 }
 
+export type MouseHandlersProp = readonly (
+  | GridMouseHandler
+  | ((irisGrid: IrisGrid) => GridMouseHandler)
+)[];
+
 export interface IrisGridProps {
   children: React.ReactNode;
   advancedFilters: ReadonlyAdvancedFilterMap;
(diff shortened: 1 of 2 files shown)


packages/grid/src/ExpandableColumnGridModel.ts
```diff
@@ -0,0 +1,62 @@
+import GridModel from './GridModel';
+import { ModelIndex } from './GridMetrics';
+
+export function isExpandableColumnGridModel(
+  model: GridModel
+): model is ExpandableColumnGridModel {
+  return (model as ExpandableColumnGridModel)?.hasExpandableColumns ?? false;
```
mofojed, line 7: This should just check if `hasExpandableColumns !== undefined` - it's just checking if the model is an `ExpandableColumnGridModel` type, which it is even if the value for `hasExpandableColumns` is `false`. When using it, you should be checking `isExpandableColumnGridModel` and then `model.hasExpandableColumns`. I guess we could consolidate that into a function, but it shouldn't be in the type predicate function that narrows the type (IMHO).

Although I guess we've been bitten by this before with `PartitionedGridModel`.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2483,6 +2490,36 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     });
   }
 
+  toggleExpandColumn(modelIndex: ModelIndex): void {
+    log.debug2('Toggle expand column', modelIndex);
+    const { model } = this.props;
+    if (isExpandableColumnGridModel(model)) {
```
mofojed, line 2496:
```suggestion
    if (isExpandableColumnGridModel(model) && model.hasExpandableColumns) {
```

mofojed (approved):

PR #2526, opened 2025-08-27
feat: DH-19382: Add actions and timing info on console history hover

jnumainville:
Adds new actions when hovering over console history items: copy, rerun, and a contextual help with timing (from the same source as the extant time in our command history but formatted differently to meet spec) as well as new server timing.

Full support is dependent on adding server timing to console history from core. That PR is not up yet but just running this PR will just not show that time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -300,7 +300,6 @@
       "version": "7.22.20",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.22.20.tgz",
       "integrity": "sha512-Y6jd1ahLubuYweD/zJH+vvOY141v4f9igNQAQ+MBgq9JlHS2iTsZKn1aMsb3vGccZsXI16VzTBw52Xx0DWmtnA==",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.22.13",
@@ -864,7 +863,6 @@
       "version": "7.27.1",
       "resolved": "https://registry.npmjs.org/@babel/plugin-syntax-flow/-/plugin-syntax-flow-7.27.1.tgz",
       "integrity": "sha512-p9OkPbZ5G7UT1MofwYFigGebnrzGJacoBSQM0/6bi/PUMVE+qlWDD/OalvQKbwgQzU6dl0xAv6r4X7Jme0RYxA==",
-      "peer": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.27.1"
       },
@@ -1676,7 +1674,6 @@
       "version": "7.27.1",
       "resolved": "https://registry.npmjs.org/@babel/plugin-transform-react-jsx/-/plugin-transform-react-jsx-7.27.1.tgz",
       "integrity": "sha512-2KH4LWGSrJIkVf5tSiBFYuXDAoWRq2MMwgivCf+93dd0GQi8RXLjKA/0EvRnVV5G0hrHczsquXuD01L8s6dmBw==",
-      "peer": true,
       "dependencies": {
         "@babel/helper-annotate-as-pure": "^7.27.1",
         "@babel/helper-module-imports": "^7.27.1",
@@ -2322,7 +2319,6 @@
       "version": "6.3.1",
       "resolved": "https://registry.npmjs.org/@dnd-kit/core/-/core-6.3.1.tgz",
       "integrity": "sha512-xkGBRQQab4RLwgXxoqETICr6S5JlogafbhNsidmrkVv2YRs5MLwpjoF2qpiGjQt8S9AoxtIV603s0GIUpY5eYQ==",
-      "peer": true,
       "dependencies": {
         "@dnd-kit/accessibility": "^3.1.1",
         "@dnd-kit/utilities": "^3.2.2",
@@ -2375,6 +2371,7 @@
       "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.4.5.tgz",
       "integrity": "sha512-XsLw1dEOpkSX/WucdqUhPWP7hDxSvZiY+fsUC14h+FtQ2Ifni4znbBt8punRX+Uj2JG/uDb8nEHVKvrVlvdZ5Q==",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "@emnapi/wasi-threads": "1.0.4",
         "tslib": "^2.4.0"
@@ -2385,6 +2382,7 @@
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.4.5.tgz",
       "integrity": "sha512-++LApOtY0pEEz1zrd9vy1/zXVaVJJ/EbAF3u0fXIzPJEDtnITsBGbbK0EkM72amhl/R5b+5xx0Y/QhcVOpuulg==",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "tslib": "^2.4.0"
       }
@@ -2394,6 +2392,7 @@
       "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.0.4.tgz",
       "integrity": "sha512-PJR+bOmMOPH8AtcTGAyYNiuJ3/Fcoj2XN/gBEWzDIKh254XO+mM9XoXHk5GNEhodxeMznbg7BlRojVbKN+gC6g==",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "tslib": "^2.4.0"
       }
@@ -2889,7 +2888,6 @@
       "version": "6.7.2",
       "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-svg-core/-/fontawesome-svg-core-6.7.2.tgz",
       "integrity": "sha512-yxtOBWDrdi5DD5o1pmVdq3WMCvnobT0LU6R8RyyVXPvFRd2o79/0NCuQoCjNTeZz9EzA9xS3JxNWfv54RIHFEA==",
-      "peer": true,
       "dependencies": {
         "@fortawesome/fontawesome-common-types": "6.7.2"
       },
@@ -2901,7 +2899,6 @@
       "version": "0.2.5",
       "resolved": "https://registry.npmjs.org/@fortawesome/react-fontawesome/-/react-fontawesome-0.2.5.tgz",
       "integrity": "sha512-55xAmidU/xWPVUK2IlG0c4LK+7pk75Qnwh7PcmjxlUTUibbI0+txjAB2ucF3TQZOXHOcjBcpoF8R75b59OxNmg==",
-      "peer": true,
       "dependencies": {
         "prop-types": "^15.8.1"
       },
@@ -2957,8 +2954,7 @@
     "node_modules/@hello-pangea/dnd/node_modules/redux": {
       "version": "5.0.1",
       "resolved": "https://registry.npmjs.org/redux/-/redux-5.0.1.tgz",
-      "integrity": "sha512-M9/ELqF6fy8FwmkpnF0S3YKOqMyoWJ4+CS5Efg2ct3oY9daQvd/Pc71FpGZsVsbl3Cpb+IIcjBDUnnyBdQbq4w==",
-      "peer": true
+      "integrity": "sha512-M9/ELqF6fy8FwmkpnF0S3YKOqMyoWJ4+CS5Efg2ct3oY9daQvd/Pc71FpGZsVsbl3Cpb+IIcjBDUnnyBdQbq4w=="
     },
     "node_modules/@humanwhocodes/config-array": {
       "version": "0.13.0",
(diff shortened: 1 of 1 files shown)


packages/console/src/console-history/ConsoleHistoryItem.scss
```diff
@@ -12,9 +12,31 @@ $button-vert-margin: ConsoleVariables.$button-vert-margin;
 }
 
 .console-history-item-command {
+  position: relative;
   white-space: pre-line;
 }
 
+.console-history-actions {
+  z-index: 1000;
```
mofojed, line 20: We shouldn't have the z-index so high. In bootstraps `_variables.scss`, they define a bunch of `$zindex-*` variables, and `$zindex-dropdown` is `1000`, so this could in theory conflict with that. Probably should only set it to 100 at most.
That being said - why do we need to set this anyway?

packages/console/src/console-history/ConsoleHistoryItem.tsx
```diff
@@ -56,18 +71,48 @@ class ConsoleHistoryItem extends PureComponent<
...
     const hasCommand = item.command != null && item.command !== '';
-
     let commandElement = null;
     if (hasCommand) {
       commandElement = (
-        <div className="console-history-item-command">
+        <div
+          className={this.consoleHistoryItemClasses()}
+          onMouseOver={() => this.setState({ isHovered: true })}
+          onMouseOut={() => this.setState({ isHovered: false })}
```
mofojed, line 102: You shouldn't need this `isHovered` state at all - instead just key off the `:hover` CSS class.

packages/console/src/console-history/ConsoleHistoryItem.scss
```diff
@@ -12,9 +12,31 @@ $button-vert-margin: ConsoleVariables.$button-vert-margin;
...
+  top: 0;
+  right: $spacer-2;
+  display: none;
+  align-items: center;
+  background-color: var(--dh-color-neutral-hover-bg);
+  border-radius: $border-radius;
+  padding: $spacer-1;
+}
+
+.console-history-item-command-hovered {
```
mofojed, line 32: You shouldn't need to do this yourself, you can just use the `:hover` pseudo-class:
```suggestion
.console-history-item-command:hover {
```

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+      // last single items are pushed up to prevent layout shifts
+      classes.push('console-history-last-single-line');
+    } else {
+      classes.push('console-history-single-line');
+    }
+  } else if (lineCount === 2) {
+    classes.push('console-history-two-lines');
+  }
+  return classes.join(' ');
+};
```
mofojed, line 42: I don't think you need the `firstItem` or `lastItem` properties at all - you can do this all with CSS!
Take a look at the `:first-child` and `:last-child` pseudo-classes: https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child

Then in CSS instead of doing `.console-history-last-single-line`, you would do `.console-command-result:first-child` (or even `.console-command-result:first-child .console-history-item-command` if you want to select an inner element within that child).

Then we don't need to manually track the classes.

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+    firstItem,
+    lastItem,
+    handleTooltipVisible,
+  } = props;
+
+  const actionBarClasses = getActionBarClasses(
+    item.command,
+    firstItem ?? false,
+    lastItem ?? false
+  );
```
mofojed, line 57: If possible I'd rather write this using the `classnames` package, e.g.

```

...
return <div className={classNames('console-history-actions', `console-history-actions-line-${lineCount}`)}> ...
```
Also then you can just do CSS for `console-history-actions-line-1` and `console-history-actions-line-2` etc.

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+  firstItem?: boolean;
+  handleTooltipVisible: (isVisible: boolean) => void;
+}
+
+const getActionBarClasses = (
+  command: string | undefined = '',
+  firstItem = false,
+  lastItem = false
+): string => {
+  const lineCount = command.split('\n').length;
```
mofojed, line 24: I think I'd memoize this. If we get a lot of console history items re-rendering it may be costly to re-render all of these items.

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+      classes.push('console-history-single-line');
+    }
+  } else if (lineCount === 2) {
+    classes.push('console-history-two-lines');
+  }
+  return classes.join(' ');
+};
+
+function ConsoleHistoryItem(props: ConsoleHistoryItemProps): ReactElement {
```
mofojed, line 44: Should memoize this whole component actually using `React.memo`

packages/console/src/console-history/ConsoleHistoryItem.scss
```diff
@@ -95,3 +117,37 @@ $button-vert-margin: ConsoleVariables.$button-vert-margin;
...
+  top: -0.125rem;
+}
+
+.console-history-last-single-line {
+  top: -1.5rem;
+}
+
+.console-history-first-single-line {
+  top: 0;
+}
```
mofojed, line 135: Any reason we're using these specific values and not `$spacer-*` or some other variable?

packages/console/src/console-history/ConsoleHistoryItem.scss
```diff
@@ -95,3 +117,37 @@ $button-vert-margin: ConsoleVariables.$button-vert-margin;
...
+  display: grid;
+  grid-template-columns: auto auto;
+  gap: $spacer-1 $spacer-3;
+  text-align: left;
+  margin-bottom: $spacer-2;
+  white-space: nowrap;
+}
+
+.console-history-item-contextual-help {
+  section[class*='spectrum-ContextualHelp-dialog'] {
```
mofojed, line 147: I reckon we just put this in `SpectrumComponent.module.scss` so it applies to all instead of just console history items, since we really do want it to fit the content it all cases.
That can be a separate ticket I guess though.

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Converts a time difference in seconds to a human-readable string.
+ * @param diff The time difference in seconds.
+ * @returns A string representing the time difference.
+ */
+export const convertedDiff = (diff: number): string => {
```
mofojed, line 6: Write as a normal function

```suggestion
export function convertedDiff(diff: number): string {
```

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param startTime The start time.
+ * @param endTime The end time.
+ * @param conversion The conversion type ('ms' or 'ns')
+ * @returns A string representing the time difference, or null if invalid.
+ */
+export const getTimeString = (
+  startTime: string | number | undefined,
+  endTime: string | number | undefined,
+  conversion = 'ms'
+): string | null => {
```
mofojed, line 27:
```suggestion
export function getTimeString(
  startTime: string | number | undefined,
  endTime: string | number | undefined,
  conversion = 'ms'
): string | null {
```

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+/**
+ * Gets a human-readable time string for the difference between two times.
+ * @param startTime The start time.
+ * @param endTime The end time.
+ * @param conversion The conversion type ('ms' or 'ns')
+ * @returns A string representing the time difference, or null if invalid.
+ */
+export const getTimeString = (
```
mofojed, line 23: We should use this function from `CommandHistoryItemTooltip.tsx` as well for consistency.

packages/console/src/console-history/ConsoleHistoryItemTooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface ConsoleHistoryActionItemTooltipProps {
+  item: ConsoleHistoryActionItem;
+}
+
+function CommandHistoryItemTooltip(
+  props: ConsoleHistoryActionItemTooltipProps
+): ReactElement {
+  const { item } = props;
+  const { startTime, serverStartTime, serverEndTime } = item;
+  const endTime = item.endTime ?? Date.now();
```
mofojed, line 15: Should memoize `endTime`, otherwise if this re-renders the date will update.
Could memoize the component as well.

packages/console/src/Console.tsx
```diff
@@ -659,7 +671,11 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
    * @param workspaceItemPromise The workspace data row promise for the workspace item to be updated
```
mofojed, line 671: Update docs

packages/console/src/console-history/ConsoleHistory.tsx
```diff
@@ -15,6 +15,7 @@ interface ConsoleHistoryProps {
   disabled?: boolean;
   supportsType: (type: string) => boolean;
   iconForType: (type: string) => ReactElement;
+  handleCommandSubmit: (command: string) => void;
```
mofojed, line 18: The prop event should be named `on*`, whereas the functions that handle the events should be named `handle*`.
See https://react.dev/learn/responding-to-events
> By convention, it is common to name event handlers as handle followed by the event name. You’ll often see onClick={handleClick}, onMouseEnter={handleMouseEnter}, and so on.

mofojed (changes requested):

packages/console/src/console-history/ConsoleHistoryItem.scss
jnumainville: I originally had it with the pseudo-class, but the behavior is buggy. When I click on the contextual help button, the hover stops, and the contextual help popup moves to the top left corner of the screen (seemingly because the contextual help button is gone). Is there a good way to use the pseudo class that keeps it hovered after clicking on the contextual help?

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
jnumainville: It seems like that has a couple of differences in the `CommandHistoryItemTooltip.tsx` one:
1. If the delta is less than 1 second it just returns "<1s"
2. Possibly has hours, minutes, and seconds, like "1h 3m 23s"
I don't want to remove that formatting completely without verifying how we want to unify this, as the formats are different enough where one isn't inherently better than the other I think.

packages/console/src/Console.tsx
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItemUtils.ts
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
mofojed: `.console-command-result:first-child .console-history-actions-line-1`

packages/console/src/console-history/ConsoleHistoryItem.scss
mofojed: Target the `:hover`, then when the tooltip is shown add another class that can also be targeted

packages/console/src/console-history/ConsoleHistoryItemActions.tsx
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItem.scss
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItem.scss
jnumainville: done

packages/console/src/console-history/ConsoleHistoryItem.tsx
jnumainville: done

mofojed (changes requested):
Whoops, meant to request changes for the classNames suggestion

mofojed (approved):

PR #2529, opened 2025-09-04
ci: Change distTag from 'v0.99' to 'release-v0.99'

mofojed:
From https://docs.npmjs.com/adding-dist-tags-to-packages
> We recommend avoiding dist-tags that start with a number or the letter "v".

This is now in line with what we do with v0.85 as well.


bmingles (approved):
LGTM

PR #2530, opened 2025-09-05
fix: reduce spectrum popover excessive padding

dsmmcken:
Reduces from ~40px to ~16px. Targets popover dialog only.

Added to styleguide just to be able to preview.


packages/components/src/theme/theme-spectrum/theme-spectrum-overrides.css
```diff
@@ -131,3 +131,9 @@ span[class*='spectrum-ToggleSwitch-switch'] {
   /* increase contrast of switch in off position */
   background: var(--dh-toggle-switch-bg);
 }
+
+[class*='spectrum-Popover'] {
+  /* Sspectrum popovers have excessive space, tighten up the dialog padding to match DH style, */
+  /* Originally --spectrum-global-dimension-static-size-500) */
```
bmingles, line 137: There's a stylelint error here. Also typo 'Sspectrum'

bmingles (changes requested):
Unit tests failed due to stylelint error.

Also, looking at the styleguide, I see the CSS variable got updated, but I'm not seeing it impact the padding of the dialog.

image

dsmmcken:
> Looks like --spectrum-dialog-padding may be the variable that needs to be set

ugh, I think copilot "helped" when I re-targeted the class from dialog to popover, it change the variable too for me...

bmingles (approved):
I verified the padding in the styleguide. Looks good.

Issue #1165, opened 2023-03-21
Exit maximized panels with escape


mattrunyon:
If a user has maximized a panel, pressing escape should exit the maximized state if the event bubbles to the container

PR #2533, opened 2025-09-10
feat: close maximized panels with escape key

dsmmcken:
fixes: #1165

Was annoyed with this while working on something, add handling to escape close maximize fields.

Exclude if inputs are focused (ex. monaco uses escape to deselect, but doesn't consume it, native searches clear field but don't consume it either).


mattrunyon:
Actually looks like linker already uses capture phase, so might just be able to consume the event in Grid?

PR #2534, opened 2025-09-15
fix: Add explicit refs to react-transition-group for React 19 support

mattrunyon:
Tested with Settings menu slide out, sections collapse/expand. On a grid the copy handler tests the slide up and then the message fades out

Can also test the styleguide modal and toasts (the toasts must be dismissed by clicking the button, they don't auto-dismiss)

Also added a type override that should throw a type error if you remove `nodeRef` from any of the `CSSTransition` components. That way we don't accidentally introduce it in the future until we move off of `react-transition-group`.

### DIFF
--- a/packages/components/src/transitions/FadeTransition.tsx
+++ b/packages/components/src/transitions/FadeTransition.tsx
@@ -1,32 +1,46 @@
+import { useRef } from 'react';
 import { CSSTransition } from 'react-transition-group';
 import type { CSSTransitionProps } from 'react-transition-group/CSSTransition';
 import type { EndHandler } from 'react-transition-group/Transition';
 import classNames from 'classnames';
+import type { RemoveIndexSignature } from '@deephaven/utils';
 import ThemeExport from '../ThemeExport';
 
-type FadeTransitionProps<Ref extends undefined | HTMLElement = undefined> =
+type FadeTransitionProps =
   // We default the timeout, so user doesn't need to provide it
   // However, CSSTransitionProps get confused if you don't provide a timeout, it requires an endHandler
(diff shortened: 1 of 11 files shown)


packages/components/src/transitions/FadeTransition.tsx
```diff
@@ -1,32 +1,68 @@
...
+      nodeRef={nodeRef}
       timeout={timeout}
-      classNames={classNames('fade', className)}
+      classNames={classNames('fade', classNamesProp)}
+      onEnter={handleOnEnter}
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...props}
-    />
+    >
+      <div ref={setRef} style={{ display: 'contents' }}>
```
bmingles, line 62: Should probably assign `{ display: 'contents' }` to a const outside of the component to avoid unnecessary renders

packages/utils/src/TypeUtils.ts
```diff
@@ -65,3 +65,17 @@ export type UndoPartial<T> = T extends Partial<infer U> ? U : never;
...
  */
 export type ValueOf<T> = T[keyof T];
+
+/**
+ * Removes the index signature from a type.
+ * This can be useful if you want to Omit a key from a type,
+ * but the type has an index signature. Otherwise, the index signature
+ * is the only thing that remains after Omit.
+ */
+export type RemoveIndexSignature<T> = {
```
bmingles, line 75: I'd find it helpful to see an example in the comment block showing what this solves.

bmingles (changes requested):
Changes look good overall. Tested the functionality as well. Left comments for some style objects and comment example.

PR #2537, opened 2025-09-18
feat: DH-13515: Expandable columns support

vbabich:
- Add new `ExpandableColumnGridModel` interface for models with expandable columns
- Export `ColumnHeaderGroup` and `IrisGridCellRendererUtils` from `@deephaven/iris-grid`
- Add `IrisGrid` methods to expand/collapse columns 
- Update metrics on columns change in `IrisGrid`
- `IrisGridMetricCalculator` now stores user column widths by name
- `parseColumnHeaderGroups` in `IrisGridUtils` has an optional argument allowing to preserve the correct header group class  
- `IrisGrid` context menu shows `Expand All` and `Collapse All` if supported in the model
- Update e2e snapshots, add unit tests


packages/iris-grid/src/IrisGridMetricCalculator.ts
```diff
@@ -1,31 +1,101 @@
...
+      !deepEqual(
+        this.getCachedCurrentModelColumnNames(model.columns),
+        this.cachedModelColumnNames
+      )
+    ) {
+      // Preserve column widths when possible to minimize visual shifts in the grid layout
+      this.updateCalculatedColumnWidths(model);
+      this.updateUserColumnWidths(model);
+    }
+    this.cachedModelColumnNames = model.columns.map(col => col.name);
```
mofojed, line 93: This could just use the result from `this.getCachedCurrentModelColumnNames`... I think I'd just rewrite this a bit:
```typescript
const modelColumnNames = this.getCachedCurrentModelColumnNames(model.columns);
if (this.cachedModelColumnNames != null && this.cachedModelColumnNames !== modelColumnNames && !deepEqual(modelColumnNames, cachedModelColumnNames) {
  ...
}
this.cachedModelColumnNames = modelColumnNames;
```

Then you're only doing the `model.columns.map` at most once when this is called (instead of twice), and you can also skip doing a `deepEqual` if the columns haven't actually changed.

packages/iris-grid/src/IrisGridMetricCalculator.ts
```diff
@@ -1,31 +1,101 @@
...
+  }
+
   getGridY(state: IrisGridMetricState): number {
+    // The state here seems to be a GridMetricState with stateOverrides passed from IrisGrid in the props,
+    // not guaranteed to be IrisGridMetricState
```
mofojed, line 98: Mmmm the intent is that the `GridMetricState` combined with the `stateOverrides` _should_ produce an `IrisGridMetricState`, so it _should_ always be an `IrisGridMetricState` here... I don't know if we can do generic typing in Grid to make that enforceable. 
I guess we use the same `stateOverride` object for both `IrisGridRenderState` and `IrisGridMetricState`...
I think what I'd do is define `IrisGridStateOverride` with all the properties we set there, then `IrisGridMetricState = GridMetricState & IrisGridStateOverride`, and `IrisGridRenderState = GridRenderState & IrisGridStateOverride`. Then it should all be typed correctly?

mofojed (changes requested):
Looking good overall. A few small things/suggestions.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1492,6 +1494,27 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     return this.getCachedKeyHandlers(keyHandlers);
   }
 
+  getMetricState(): IrisGridMetricState | undefined {
```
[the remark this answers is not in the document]
vbabich: Changed the values to a `getCachedStateOverride` call

PR #2540, opened 2025-09-30
fix: Pass optional expandDescendants arg to setColumnExpanded

vbabich:
- Pass optional `expandDescendants` arg to `setColumnExpanded`
- Pass mouse handlers and renderer from `IrisGridPanel` to `IrisGrid`


bmingles (changes requested):
Looks good overall. Left comment about 1 place you could use your new `MouseHandlersProp` type

bmingles (approved):
LGTM

PR #2544, opened 2025-10-06
fix: DH-20255 Limit maximum depth to serialize redux log data

ericlln:
Follow up PR to https://github.com/deephaven/web-client-ui/pull/2531 which failed testing, with an error from `JSON.stringify()` of `RangeError: Invalid string length`. 

These errors are caused by linked structures such as a Java LinkedHashMap that's included in the Redux data in various API related keys. We use the blacklist to filter out keys containing these structures, but these keys are not stable and will need to be constantly updated to prevent the log export from outright failing. 

To preemptively deal with this, the maximum depth to serialize to can be limited in the [safe-stable-stringify](https://www.npmjs.com/package/safe-stable-stringify) library to avoid linked structures from blowing up the log size. With no blacklist entries, the exported log size is only `1.2MB`.

With an updated blacklist, exporting the logs with or without the limit both yield log files of almost the exact same size, meaning that no useful data has been lost by limiting the depth. However, I'm open to suggestions to the default depth to serialize to, or whether it should be optional to have a limit on the depth instead of mandating a limit (`Infinity` is not accepted by the library).

### DIFF
--- a/packages/log/src/LogExport.ts
+++ b/packages/log/src/LogExport.ts
@@ -1,5 +1,5 @@
 import JSZip from 'jszip';
-import stringify from 'safe-stable-stringify';
+import { configure } from 'safe-stable-stringify';
 import type LogHistory from './LogHistory';
 
 // List of objects to blacklist
@@ -23,6 +23,9 @@ export const DEFAULT_PATH_BLACKLIST: string[][] = [
   ['corePlusManager', 'dheClient'],
   ['dheClient'],
 ];
+
+// The default maximum depth to serialize to in the redux data
+const DEFAULT_MAXIMUM_DEPTH = 10;
 
 function stringifyReplacer(blacklist: string[][]) {
   // modified from:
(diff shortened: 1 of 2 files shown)


packages/log/src/LogExport.ts
```diff
@@ -109,6 +125,7 @@ export async function exportLogs(
   metadata?: Record<string, unknown>,
   reduxData?: Record<string, unknown>,
   blacklist: string[][] = DEFAULT_PATH_BLACKLIST,
+  maximumDepth: number = DEFAULT_MAXIMUM_DEPTH,
   fileNamePrefix = `${formatDate(new Date())}_support_logs`
```
mofojed, line 129: Changing the order of the parameters is a breaking change, put it on the end.
```suggestion
fileNamePrefix = `${formatDate(new Date())}_support_logs`,
  maximumDepth: number = DEFAULT_MAXIMUM_DEPTH
```

ericlln: In most cases it seems to be the `configs` object under `client` or `dheClient`, e.g. `schemaService.clientUtils.client.configs`, with each key value pair in that map being serialized

Screenshot 2025-10-07 at 8 58 57 AM

packages/log/src/LogExport.ts
```diff
@@ -14,11 +14,19 @@ export const DEFAULT_PATH_BLACKLIST: string[][] = [
   ['dashboardData', '*', 'openedMap'],
   ['draftManager', 'draftStorage'],
   ['layoutStorage'],
+  ['storage'],
+
+  // Below are confirmed enterprise specific keys, and will be moved in DH-20410
```
mofojed, line 19: `draftManager` is also an enterprise specific key.
Okay to have this here for now, but on Enterprise side we should just add to the list at https://github.com/deephaven-ent/iris/blob/rc/sanluis/web/client-ui/src/utils/exportEnterpriseLogs.ts#L18

mofojed (approved):

PR #2613, opened 2026-01-22
feat: Extensible table options

vbabich:


mofojed (commented):
This all looks pretty good... one thing I've been mulling over the last couple days (which I'm sure you're thinking about as well) is how we'll get this to switch to a pivot table view from a default view... or do we just do a `setModel` or something to update the model? How would that work from re-hydration?

PR #2635, opened 2026-03-17
fix: DH-21869: Fix Conditional Formatting on Timestamp

dgodinez-dh:
- the time stamps need to be single quoted as date-time liters per our docs: https://deephaven.io/core/docs/how-to-guides/date-time-literals/
- The backend parsing fails on timezone abbreviations ending with ST or DT. If you have EST or EDT, you need to pass in ET. I have updated the validation to prevent passing these to the backend.


bmingles (approved):
LGTM

PR #2638, opened 2026-03-20
feat: DH-22048: Add optional button to dismiss the ErrorView component

dgodinez-dh:
- add dismiss button to ErrorView
- button is only shown if an optional callback is provided
- parent component must implement logic of callback to dismiss
- show in styleguide


bmingles (approved):
LGTM

PR #2658, opened 2026-04-09
fix: peerDependencies in iris-grid should just use react >=16

mofojed:
- This change was introduced with PR https://github.com/deephaven/web-client-ui/pull/2555/changes#diff-f952417e448401752610b573dafc9c136ac4ae7b0734afad22382babe6608dff
- The only other changes to `package.json` were to @dnd-kit/modifiers and @dnd-kit/sortable, both of which also have a peer dep of react >=16
- Setting it to >= 18 is unnecessarily restrictive

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30574,8 +30574,8 @@
         "node": ">=10"
       },
       "peerDependencies": {
-        "react": ">=18.0.0",
-        "react-dom": ">=18.0.0"
+        "react": ">=16.8.0",
+        "react-dom": ">=16.8.0"
       }
     },
     "packages/jsapi-bootstrap": {
--- a/packages/iris-grid/package.json
+++ b/packages/iris-grid/package.json
@@ -62,8 +62,8 @@
     "react-transition-group": "^4.4.2"
   },
   "peerDependencies": {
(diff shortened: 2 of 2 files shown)


mattrunyon:
It looks like the reason I bumped it to 18 is because in `VisibilityOrderingBuilder.tsx` I needed to add `flushSync` which is an API only in React 18. If you try to build w/ 17  you will get an error b/c it's an unknown import.

Given that React 18 came out 4 years ago I'm not sure how big of a deal it is. Did we have an issue caused by this?

mattrunyon (changes requested):
I left it in a separate comment, but there is a reason for the dep on React 18

mofojed:
@mattrunyon I'm running into some fun issues right now with types and package dependencies in the deephaven-plugins repo, where the different plugins are depending on different versions of the `@deephaven/*` packages: https://github.com/deephaven/deephaven-plugins/pull/1326
I thought this might fix it, but it does not actually fix the issue. Package dependencies, so fun right now.

mofojed:
I just updated deephaven-plugins to React v18: https://github.com/deephaven/deephaven-plugins/pull/1335

PR #2659, opened 2026-04-09
fix: DH-22054: GridMetrics type update

ethanalvizo:


packages/iris-grid/src/IrisGridCacheUtils.ts
```diff
@@ -69,12 +69,10 @@ function areIrisGridStatesEqual(
 
   return (
     irisGridStateA === irisGridStateB ||
-    (irisGridStateA.metrics != null &&
-      irisGridStateB.metrics != null &&
-      irisGridStateA.metrics.userColumnWidths ===
-        irisGridStateB.metrics.userColumnWidths &&
-      irisGridStateA.metrics.userRowHeights ===
-        irisGridStateB.metrics.userRowHeights &&
+    (irisGridStateA.metrics?.userColumnWidths ===
```
mofojed: This is not the same ... now `irisGridState.metrics = undefined` will match with `irisGridState.metrics = { userColumnWidths: undefined, ... }`. I'm not sure that it matters, but we shouldn't be changing logic in these type changes anyways.

mofojed (changes requested):

mofojed (approved):

PR #2660, opened 2026-04-13
feat: DH-21476: Middleware plugin infrastructure for widget chaining

vbabich:
# Middleware Plugin Infrastructure for Widget Chaining

## Summary

Adds middleware plugin support so plugins can wrap existing widgets without replacing them, cross-plugin dependency loading via manifest `package`/`dependencies` fields, and shared plugin-loading utilities in `@deephaven/plugin`.

## What Changed

**Middleware chaining** — New `WidgetMiddlewarePlugin` type, identified by its own `PluginType.MIDDLEWARE_PLUGIN` discriminator. Middleware receives a `Component` prop and wraps the next layer. Multiple middleware compose in registration order. Applied in both `WidgetLoaderPlugin` (dashboard panels) and `WidgetView` (inline widgets) via `createChainedComponent`/`createChainedPanelComponent`. The chaining functions automatically filter middleware by `supportedTypes` at render time, so middleware only activates for matching widget types.

**Cross-plugin dependencies** — Manifest entries can declare `package` (makes the plugin's exports available to other plugins) and `dependencies` (topological sort so deps load first). Plugins load sequentially so each plugin's exports are available to subsequent plugins via standard `import` statements.

**Shared loading utilities** — `getPluginModuleValue`, `registerPlugin`, `processLoadedModule`, `sortPluginsByDependency`, and manifest types moved from `@deephaven/app-utils` into `@deephaven/plugin`. Both web-client-ui and gplus consume these instead of duplicating the logic. ~80 lines removed from app-utils.

**Behavior change — duplicate plugin names.** When two plugins in the manifest are loaded under the same name, the shared `registerPlugin` now **skips the duplicate** and keeps the first registration (with a warning). Previously the duplicate **replaced** the existing entry. This applies to plugin-name collisions in the plugin map; widget-type resolution in `WidgetLoaderPlugin` is unaffected (last base plugin still wins for a given widget type).

## Key Files

| File | Change |
|---|---|
| `packages/plugin/src/PluginTypes.ts` | Middleware types, type guard |
| `packages/plugin/src/PluginUtils.tsx` | Chaining functions, loader utils, manifest types |
| `packages/plugin/src/sortPluginsByDependency.ts` | Topological sort for plugin dependencies |
| `packages/plugin/src/WidgetView.tsx` | Middleware-aware inline rendering |
| `packages/dashboard-core-plugins/src/WidgetLoaderPlugin.tsx` | Middleware-aware panel loading |
| `packages/app-utils/src/plugins/PluginUtils.tsx` | Simplified to use shared utils |
| `packages/plugin/docs/middleware-architecture.md` | Architecture reference |


packages/plugin/src/PluginUtils.tsx
```diff
@@ -97,3 +101,104 @@ export function getPluginsElementMap(pluginMap: PluginModuleMap): ElementMap {
...
+        const MiddlewareComponent = middlewarePlugin.component;
+
+        function ChainedComponent(props: WidgetComponentProps<T>) {
+          return (
+            // eslint-disable-next-line react/jsx-props-no-spreading
+            <MiddlewareComponent {...props} Component={WrappedComponent} />
+          );
+        }
+        ChainedComponent.displayName = `${middlewarePlugin.name}(${
+          (WrappedComponent as React.ComponentType).displayName ??
```
mofojed: We have this same name logic in a few places, might make sense to put it in a utility function.

packages/plugin/src/PluginTypes.ts
```diff
@@ -144,6 +144,77 @@ export interface WidgetPanelProps<T = unknown> extends WidgetComponentProps<T> {
...
+}
+
+/**
+ * Type guard to check if a plugin is a middleware plugin.
+ */
+export function isWidgetMiddlewarePlugin(
+  plugin: PluginModule
+): plugin is WidgetMiddlewarePlugin {
+  return (
+    isWidgetPlugin(plugin) &&
```
mofojed: Hmmm do we want it to be a Widget Plugin with a flag? Or should it be a different plugin type?
Since it relies on there being an existing WidgetPlugin, what benefits do we get from extending WidgetPlugin?

mofojed (changes requested):
Looks good overall, just a couple comments/questions

packages/plugin/src/PluginTypes.ts
vbabich: A different plugin type would make more sense, good point. Updated.

vbabich:
Our e2e tests fail in `main` on `deephaven:edge`.
`deephaven:latest` works fine; the issue must be introduced on the deephaven-core side.

packages/plugin/src/sortPluginsByDependency.ts
```diff
@@ -0,0 +1,110 @@
+import Log from '@deephaven/log';
+import { type PluginManifestPluginInfo } from './PluginUtils';
```
mofojed: Any reason `sortPluginsByDependency` is in it's own file? Circular reference here between PluginUtils/this, though it's just types... but I'd probably just move this to `PluginUtils`, or have a `utils` folder for other utils, or move the type into `PluginTypes` or something.

mofojed (changes requested):

packages/plugin/src/sortPluginsByDependency.ts
vbabich: I originally had this function in the utils file, then decided to move it out to a separate file, but didn't check for circular dependencies. Moved it back. Ok if I keep the tests in a separate file?

mofojed (commented):
Looks good, will merge after next ui release as this has breaking changes.

mofojed (approved):

packages/app-utils/src/plugins/PluginUtils.ts
```diff
@@ -9,23 +9,15 @@ import {
...
+  processLoadedModule,
+  sortPluginsByDependency,
+  type PluginManifest,
 } from '@deephaven/plugin';
 import loadRemoteModule from './loadRemoteModule';
+import { resolve } from './remote-component.config';
 
 const log = Log.module('@deephaven/app-utils.PluginUtils');
 
-export type PluginManifestPluginInfo = {
```
mofojed, line 21: Instead of making this a breaking change and removing these, mark them deprecated and re-export the types from `@deephaven/plugin`.

mofojed (changes requested):
Deprecate instead of break

packages/app-utils/src/plugins/PluginUtils.ts
vbabich: Updated

mofojed (approved):

PR #2661, opened 2026-04-15
fix: DH-19210: Add plaintext paste key handling

gzh2003:
For DH-19210. Adds handling for `text/plain` from the clipboard(this is necessary to handle copy/paste workflows from excel as `text/html` contains only '#' symbols when the column is truncated)

### DIFF
--- a/packages/grid/src/key-handlers/PasteKeyHandler.ts
+++ b/packages/grid/src/key-handlers/PasteKeyHandler.ts
@@ -45,11 +45,15 @@ export function parseValueFromNodes(nodes: NodeListOf<ChildNode>): string[][] {
   return result;
 }
 
-export function parseValueFromText(text: string): string[][] {
-  return text
+export function parseValueFromText(text: string): string | string[][] {
+  const rows = text
     .trim()
     .split('\n')
     .map(row => row.split('\t'));
+  if (rows.length === 1 && rows[0].length === 1) {
+    return rows[0][0];
+  }
+  return rows;
 }
 
 export function parseValueFromElement(
@@ -131,8 +135,8 @@ class PasteKeyHandler extends KeyHandler {
           const inputListener = (): void => {
             cleanup();
             const value =
-              parseValueFromElement(dummyInput) ??
-              (plainText.length > 0 ? parseValueFromText(plainText) : null);
+              (plainText.length > 0 ? parseValueFromText(plainText) : null) ??
+              parseValueFromElement(dummyInput);
             if (value != null) {
               grid.pasteValue(value);
             }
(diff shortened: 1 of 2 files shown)


packages/grid/src/key-handlers/PasteKeyHandler.ts
```diff
@@ -104,21 +115,35 @@ class PasteKeyHandler extends KeyHandler {
...
+            const clipboardEvent = e as ClipboardEvent;
+            plainText =
+              clipboardEvent.clipboardData?.getData('text/plain') ?? '';
+          };
+
+          const inputListener = (): void => {
+            cleanup();
+            const value =
+              (plainText.length > 0 ? parseValueFromText(plainText) : null) ??
+              parseValueFromElement(dummyInput);
```
gzh2003, line 139: The main issue here is that there is no way to distinguish "good data" from "bad data", unless we want to explicitly detect the '###' pattern from excel (not a fan... a bit hacky).

However, prioritizing text/plain is consistent with how the context menu handles paste, and there aren't any issues with dh-table -> dh-table (copy and paste) since text/plain is always populated.

packages/grid/src/key-handlers/PasteKeyHandler.ts
```diff
@@ -45,6 +45,17 @@ export function parseValueFromNodes(nodes: NodeListOf<ChildNode>): string[][] {
   return result;
 }
 
+export function parseValueFromText(text: string): string | string[][] {
```
bmingles, line 48: Would be good to have unit tests for this one since it's pure util logic

packages/grid/src/key-handlers/PasteKeyHandler.ts
```diff
@@ -45,6 +45,17 @@ export function parseValueFromNodes(nodes: NodeListOf<ChildNode>): string[][] {
...
+    .trim()
+    .split('\n')
+    .map(row => row.split('\t'));
+  if (rows.length === 1 && rows[0].length === 1) {
+    return rows[0][0];
+  }
+  return rows;
+}
+
 export function parseValueFromElement(
```
bmingles, line 59: Looks like we have existing tests for this function. Anything that could be added to test the new functionality?

bmingles (changes requested):
could use some tests

packages/grid/src/key-handlers/PasteKeyHandler.ts
gzh2003: Good call. I added tests for whitespace handling and some tabular input to make sure we don’t introduce regressions.

bmingles (dismissed):
Code changes look good. e2e tests are failing. Also worth mentioning I don't have Excel so no way to actually verify the behavior.

mofojed (approved):
I tested on my Mac with Excel, this now pastes dates correctly even if copied while it's showing `###`.
e2e tests were failing because of an issue in the core edge image, which is now resolved. I re-ran the tests and now they passed.

PR #2663, opened 2026-04-17
fix: DH-21869: Fix Timezone for Conditional Date Formatting

dgodinez-dh:
- Use dh api to get a timezone compatible with the backend
- The UI now shows when a value is invalid


bmingles (changes requested):
Only question I have is whether there should be any additional unit tests added for the validation changes?

dgodinez-dh:
> Only question I have is whether there should be any additional unit tests added for the validation changes?

Per our discussion on Slack, the fundamental change is an extra api change, and it doesn't seem worth testing a mocked api call.

bmingles (approved):
LGTM

PR #2664, opened 2026-04-29
fix(golden-layout): DH-22431: preserve focus indicator on nested dashboard tab

mofojed:
- When a dashboard is nested inside another dashboard, focusin events bubble from the inner content up to the outer container's contentElement. The outer Tab's focusin handler would clear the inner tab's lm_focusin class and claim focus for itself, leaving the active tab indicator in the wrong place.
- _onTabContentFocusIn now ignores events whose target lives inside a nested LayoutManager. _onTabClick treats focus inside a nested layout as 'not in this container' so clicking the outer tab properly transfers the indicator back out.
- Tested with the following deeply nested layout:
```
from deephaven import ui

@ui.component
def deeply_nested_dashboard_component():
    """A dashboard nested inside a panel, which is inside another dashboard."""
    return ui.panel(
        ui.dashboard(
            ui.row(
                ui.panel(ui.text("Content Level 1"), title="Level 1"),
                ui.panel(
                    ui.dashboard(
                        ui.row(
                            ui.panel(ui.text("Content Level 2"), title="Level 2"),
                            ui.panel(ui.text("Deepest Content"), title="Deepest Panel"),
                        )
                    ),
                    title="Nested Dashboard Container",
                ),
            )
        ),
        title="Outer Dashboard",
    )
ui_deeply_nested_dashboard = deeply_nested_dashboard_component()
```

### DIFF
--- a/packages/golden-layout/src/__tests__/tab.test.ts
+++ b/packages/golden-layout/src/__tests__/tab.test.ts
@@ -168,6 +168,8 @@ describe('tabs apply their configuration', () => {
     expect(innerTab.element.hasClass('lm_focusin')).toBe(true);
     expect(outerTab.element.hasClass('lm_focusin')).toBe(false);
 
+    // jsdom doesn't implement scrollIntoView, which Tab._onTabClick calls.
+    Element.prototype.scrollIntoView = jest.fn();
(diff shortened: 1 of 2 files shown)


dsmmcken (approved):
Tested with example provided, confirmed working.

PR #2669, opened 2026-05-05
fix: DH-22319: HierarchicalCheckboxValueMap type should be recursive

ethanalvizo:


ethanalvizo:
Not needed, HierarchicalCheckbox only goes 2 layers deep

PR #2670, opened 2026-05-05
fix: DH-20403: Persist user column widths by name in IrisGrid

vbabich:
Persist user column widths by name end-to-end so hidden/resized columns survive layouts where the underlying model doesn't currently expose them (e.g. while a rollup is active and non-aggregated columns are hidden).

`IrisGridMetricCalculator` already tracked widths by name internally, but state was saved and restored only via the by-index map projected against the current model — entries for columns not present in the model were silently dropped on every dehydrate. `dehydrateIrisGridState` now prefers `metricCalculator.getUserColumnWidthsByName()` when available, falling back to the previous by-index projection only for callers that don't pass a metric calculator. `hydrateIrisGridState` returns a `userColumnWidthsByName` map alongside the back-compat by-index map, and `IrisGridPanel` / `GridWidgetPlugin` thread it through to `IrisGrid`, which seeds the metric calculator's by-name map at construction.

Unit tests cover the round-trip: a width entry for a column absent from the rolled-up model survives dehydrate → hydrate → dehydrate, and is re-applied to the by-index map once the original (un-rolled-up) model is hydrated against.

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
@@ -202,6 +202,7 @@ interface IrisGridPanelState {
   quickFilters: ReadonlyQuickFilterMap;
   sorts: readonly SortDescriptor[];
   userColumnWidths: ModelSizeMap;
+  userColumnWidthsByName: Map<ColumnName, number>;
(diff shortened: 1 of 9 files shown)


dgodinez-dh (approved):
Looks good.

PR #2672, opened 2026-05-06
fix: DH-21843: Fix unexpected code studio unmount

dgodinez-dh:
- added unit test that mimics the race condition of the bug
- added unit test for expected behavior when a layout should change
- change DashboardLayout lastConfig to a ref


mofojed (approved):
Looks good, thanks.

PR #2678, opened 2026-05-13
fix: DH-22548: Command Selection Overrides Select File Button

SimonVutov:
[DH-22548](https://deephaven.atlassian.net/browse/DH-22548)

When the CSV upload overlay was open, hovering over the console history would trigger the command action toolbar (`.console-history-actions`), which rendered on top of the overlay and blocked the "Select File..." button from being clickable.

The overlay was rendered before the scroll-pane in the DOM, so the history items — being positioned elements later in the tree — painted on top of it. Fixed by moving the `CsvOverlay` to render after the scroll-pane, so natural DOM stacking order puts it on top without requiring any z-index hacks.

[DH-22548]: https://deephaven.atlassian.net/browse/DH-22548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/console/src/Console.tsx
+++ b/packages/console/src/Console.tsx
@@ -1096,18 +1096,6 @@ export class Console extends PureComponent<ConsoleProps, ConsoleState> {
             onDragEnter={this.handleDragEnter}
             onDragLeave={this.handleDragLeave}
           >
-            {showCsvOverlay && (
-              <CsvOverlay
-                onCancel={this.handleCsvFileCanceled}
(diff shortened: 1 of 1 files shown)


SimonVutov:
The Codecov complaint is irrelevant. The uncovered line is the same that existed before, just at a different position in the file.

mofojed (approved):

PR #2679, opened 2026-05-13
feat: DH-21757: Allow widget plugins to register types as dashboard types

mofojed:
- Allows widget plugins to register a type as a dashboard type, and create the dashboard payload
  - Avoids the current hack where the ui DashboardPlugin intercepts the panel open and then emits a dashboard creation, is more explicit about how it's handled
  - Also allows other plugins to potential register dashboard types
- Used by https://github.com/deephaven/deephaven-plugins/pull/1341, Enterprise draft at https://github.com/deephaven-ent/iris/pull/4653
- Leaves the door open for if we want to have an option for allowing dashboards to open as a panel by default (for now, you can just wrap a dashboard in a `ui.panel`)

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
+++ b/packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
@@ -223,6 +223,10 @@ export class ConsolePanel extends PureComponent<
     return id;
   }
 
+  /**
+   * Removes the item ID mapping from state by its ID value.
+   * @param id The item ID to delete from the itemIds map.
+   */
   private deleteItemId(id: string): void {
     this.setState(({ itemIds }) => {
       log.debug('Delete item', id, itemIds);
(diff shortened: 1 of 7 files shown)


packages/dashboard-core-plugins/src/panels/ConsolePanel.tsx
```diff
@@ -220,6 +223,19 @@ export class ConsolePanel extends PureComponent<
     return id;
   }
 
+  private deleteItemId(id: string): void {
```
dgodinez-dh, line 230: The new methods are pretty straight forward, but I think it would still be good to have a jsdoc comment for deleteItemId and handlePanelClosed.

packages/plugin/src/PluginUtils.test.tsx
```diff
@@ -14,22 +14,31 @@ import {
   PluginType,
   type ThemePlugin,
   type WidgetPlugin,
+<<<<<<< HEAD
```
dgodinez-dh: Looks like a merge got check in?

packages/plugin/src/PluginUtils.tsx
```diff
@@ -490,3 +509,22 @@ export function sortPluginsByDependency<
 
   return sorted;
 }
+
+/**
+ * Get the WidgetPlugin that supports opening a widget of the given type as a dashboard.
+ * @param pluginMap Map of available plugins
+ * @param type Type of widget to check
+ * @returns The WidgetPlugin that supports opening this widget as a dashboard
```
dgodinez-dh: Maybe add "returns null no plugin supports this type" or something.

dgodinez-dh (changes requested):

packages/plugin/src/PluginUtils.test.tsx
mofojed: Eww

dgodinez-dh (commented):
Changes look good. But looks like a few test failures.

dgodinez-dh (approved):

PR #2680, opened 2026-05-14
fix: DH-22318: incorrect typing in ConsoleContainer props and ConsolePanel handler signatures

ethanalvizo:


mofojed (changes requested):
It looks good, but before we merge we should test this with a corresponding Enterprise change cleaning up the types. I've published an alpha package with your changes here: `@deephaven/dashboard-core-plugins@1.19.1-console-types-fixes.0+6580c8a`
Use that on the Enterprise side and get a PR ready that cleans up to use these new types and remove any workarounds.

mofojed (approved):

PR #2682, opened 2026-05-14
fix: DH-17831: Fix partition table losing quick filters

dgodinez-dh:
- partition tables were losing there quick filters
- needed to rebuild the filters on table change for partition tables

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -3389,7 +3389,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     const { model } = this.props;
     // movedColumns reset triggers metricCalculator update in the Grid component
     this.setState({ movedColumns: model.initialMovedColumns });
-    // For partitioned tables, we want to rebuild filters on table change to insure filters are applied to the new partition
+    // For partitioned tables, we want to rebuild filters on table change to ensure filters are applied to the new partition
     const { partitionConfig } = this.state;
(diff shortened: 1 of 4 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3389,6 +3389,11 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     const { model } = this.props;
     // movedColumns reset triggers metricCalculator update in the Grid component
     this.setState({ movedColumns: model.initialMovedColumns });
+    // For partitioned tables, we want to rebuild filters on table change to insure filters are applied to the new partition
```
mofojed:
```suggestion
    // For partitioned tables, we want to rebuild filters on table change to ensure filters are applied to the new partition
```

mofojed (changes requested):
Why is the table changing in the first place? It looks like when we flip back, the table flickers to the partitions table and then to the filtered table, it doesn't look good. I don't think the table should be changing in this scenario.

dgodinez-dh:
> Why is the table changing in the first place? It looks like when we flip back, the table flickers to the partitions table and then to the filtered table, it doesn't look good. I don't think the table should be changing in this scenario.

There are three cases that cause this bug:
- Switch tabs
- Select a different partition key
- Click merge button

In the first case, I think you are right. There is no reason for the table to change. But for the other two cases, I think it makes sense.

Update, added a check to not update partition config if it hasn't changed. This was causing set model to be called and fire table changed.

mofojed (approved):

PR #2683, opened 2026-05-14
fix: DH-22547: ESC cancels background row position

SimonVutov:
**fix: ESC closes cell overflow modal without clearing row selection**

Focuses the modal’s editor when opened so ESC is handled by the Modal’s window key handler.
Restores grid focus after the modal closes for keyboard navigation.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -4184,6 +4184,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.setState({
       showOverflowModal: false,
     });
+    this.grid?.focus(); // Focus on the grid after closing the overflow modal
   }
 
   handleOpenNoPastePermissionModal(errorMessage: string): void {
--- a/packages/iris-grid/src/IrisGridCellOverflowModal.tsx
+++ b/packages/iris-grid/src/IrisGridCellOverflowModal.tsx
@@ -117,6 +117,7 @@ export default function IrisGridCellOverflowModal({
       setHeight(contentHeight)
     );
     autoSetLineNumbers();
+    (document.activeElement as HTMLElement)?.blur();
   }
 
   if (!isOpen && !isOpened) {
--- a/packages/iris-grid/src/mousehandlers/IrisGridCellOverflowMouseHandler.ts
+++ b/packages/iris-grid/src/mousehandlers/IrisGridCellOverflowMouseHandler.ts
@@ -131,6 +131,7 @@ class IrisGridCellOverflowMouseHandler extends GridMouseHandler {
         showOverflowModal: true,
         overflowText: this.irisGrid.getValueForCell(column, row) as string,
       });
+      return true; // Consume the event so GridSelectionMouseHandler doesn't clear the row selection
     }
 
     return false;


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -4186,6 +4188,16 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+  handleOverflowEscapeCapture(
+    event: React.KeyboardEvent<HTMLDivElement>
+  ): void {
+    const { showOverflowModal } = this.state;
+    if (event.key === 'Escape' && showOverflowModal) {
+      event.stopPropagation();
+      this.handleOverflowClose();
+    }
+  }
+
```
mofojed: I don't think this should be necessary. The `Modal` that is shown when in this state should already have a key handler that is used: 
- `IrisGridCellOverflowModal` passes `keyboard` in to the `Modal` component: https://github.com/deephaven/web-client-ui/blob/0b5774ed38609ccda28aba28445fcc7e5243b6c4/packages/iris-grid/src/IrisGridCellOverflowModal.tsx#L130
- The `Modal` should then have handling for that escape key: https://github.com/deephaven/web-client-ui/blob/0b5774ed38609ccda28aba28445fcc7e5243b6c4/packages/components/src/modal/Modal.tsx#L51

So something else must be going on. Perhaps the modal is not getting focus when it is shown or something?

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -5154,7 +5166,11 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     });
 
     return (
-      <div className="iris-grid" role="presentation">
+      <div
+        className="iris-grid"
+        role="presentation"
+        onKeyDownCapture={this.handleOverflowEscapeCapture}
```
mofojed: Shouldn't need to add a key handler here.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
SimonVutov: Yeah, this handler isn't needed. The real issue is the focus. I've pushed the changes.

packages/iris-grid/src/IrisGridCellOverflowModal.tsx
```diff
@@ -117,6 +117,7 @@ export default function IrisGridCellOverflowModal({
       setHeight(contentHeight)
     );
     autoSetLineNumbers();
+    editor.focus(); // Focus the editor so escape key will trigger modal close
```
mofojed: Hmm going to be picky about some small details here... this puts a cursor at the beginning of the editor, which we probably don't want either.
Of note, we're using our own custom `Modal` component, which we generally don't use anymore (we should use the [`Dialog` component](https://react-spectrum.adobe.com/v3/Dialog.html) provided from Spectrum instead).
Converting to use that component instead is an option, but bigger change than I want and introduces risk of breaking something. So I think the easiest options would be to just blur the active element instead.

mofojed (changes requested):

packages/iris-grid/src/IrisGridCellOverflowModal.tsx
SimonVutov: I've changed it to use `(document.activeElement as HTMLElement)?.blur();`, and now that cursor doesn't show up.

mofojed (approved):

PR #2689, opened 2026-05-26
fix: Spectrum version bump and pin

jnumainville:
React Spectrum [recently switched](https://github.com/adobe/react-spectrum/blob/main/rfcs/2025-dependencies.md) from independent packages to packages bound to a specific `@adobe/react-spectrum` version (in minor version bumps across packages). This has led to some broken testing in https://github.com/deephaven/deephaven-plugins/pull/1349 and was flagged as an issue in an enterprise version bump by Copilot.
As an example, note the dependency switch from [5.9.0](https://www.npmjs.com/package/@react-spectrum/overlays/v/5.9.0?activeTab=code) to [5.10.0](https://www.npmjs.com/package/@react-spectrum/overlays/v/5.10.0?activeTab=code) of `@react-spectrum/overlays`.
This PR pins specific and consistent versions for all react spectrum packages, as well as related ones, even if they aren't in the spectrum namespace, to prevent breakages.


packages/components/package.json
```diff
@@ -35,20 +35,20 @@
...
+    "@react-spectrum/theme-default": "3.6.0",
+    "@react-spectrum/toast": "3.2.0",
+    "@react-spectrum/utils": "3.13.0",
+    "@react-stately/overlays": "3.7.0",
+    "@react-stately/utils": "3.12.0",
+    "@react-types/combobox": "3.15.0",
+    "@react-types/radio": "3.9.0",
+    "@react-types/shared": "3.34.0",
+    "@react-types/textfield": "3.12.4",
+    "@spectrum-icons/ui": "3.7.0",
```
bmingles, line 51: What determines which versions you pinned here?

bmingles (changes requested):
Just had a question about where version numbers came from. Looks good otherwise

packages/components/package.json
jnumainville: It was the packages that were released with `"@adobe/react-spectrum": "3.47.0"`, or at least should have been... two were off but have been fixed.

bmingles (approved):
Changes look good to me.

PR #2690, opened 2026-05-27
fix: DH-22737: Group plugins by dependency level for parallel loading

vbabich:
Follow-up change for the middleware infrastructure PR.

Pre-requisite for configurable Table Options ticket, DH-21476.


packages/app-utils/src/plugins/PluginUtils.ts
```diff
@@ -85,29 +86,41 @@ export async function loadModulePlugins(
...
+
+      // eslint-disable-next-line no-await-in-loop
+      const results = await Promise.allSettled(
+        level.map(async plugin => {
+          const { name, main, version, package: packageName } = plugin;
+          const pluginMainUrl = `${modulePluginsUrl}/${name}/${main}`;
+          const pluginExports = await loadModulePlugin(pluginMainUrl);
+          return { pluginExports, name, packageName, version };
+        })
+      );
```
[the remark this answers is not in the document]
vbabich: This suggestion adds a lot of complexity for an edge case that should be fixed by the plugin developer. No load order lets circular deps import each other successfully anyway.

packages/plugin/src/PluginUtils.tsx
```diff
@@ -405,12 +405,102 @@ export function processLoadedModule(
...
 /**
  * Topologically sort plugins so that dependencies are loaded before the
  * plugins that depend on them. Plugins without dependencies or whose
  * dependencies are not in the manifest keep their original relative order
  * (stable sort). Throws if a dependency cycle is detected.
  *
+ * @deprecated Use {@link groupByDependencyLevel} instead, which supports
+ * parallel loading within each level and handles cycles gracefully.
+ *
  * @param plugins The plugin list from the manifest
```
dgodinez-dh: Is there a reason this is deprecated rather than removed? It looks like the only use is in in PluginUtils.ts, which was relpaced with groupByDependencyLevel.

dgodinez-dh (dismissed):
Question about deprecated method, but otherwise everything looks good.

packages/plugin/src/PluginUtils.tsx
vbabich: The other use is in the DHE PR, which is currently in draft and will be updated to use the new method. Technically, deleting a util method is a breaking change, so I deprecated it, but you're right, this is safe to delete. Updated.

dgodinez-dh (approved):

PR #2692, opened 2026-05-29
fix: DH-16187: send imploded value to AutoResizeTextarea parent

jnumainville:
Previously, when editing an `AutoResizeTextarea`, the possibly exploded value would be sent as the value to the `onChange`, and the imploded value was sent `onBlur`. In cases that bypassed blur, such as saving with a keyboard shortcut, this meant the value was kept as the exploded one in the parent. In an `AutoResizeTextarea` that relies on the `onChange` prop to manage history, this meant that a field would show as updated upon next blur, even though it wasn't updated by the user, only by the implode. 
This change makes it so the imploded value is the one sent to the parent as the source of truth, and the exploded values are kept local to the `AutoResizeTextarea`.

### DIFF
--- a/packages/components/src/AutoResizeTextarea.tsx
+++ b/packages/components/src/AutoResizeTextarea.tsx
@@ -41,7 +41,7 @@ function AutoResizeTextarea({
     function syncStateWithProp() {
       // keep state value in sync with prop changes
       // but not while focused, as the display value is exploded
-      // and the prop value is often imploded
+      // and the prop value might not be
       if (!isFocused.current) {
         setValue(propsValue);
       }
--- a/packages/components/src/AutoResizeTextarea.tsx
+++ b/packages/components/src/AutoResizeTextarea.tsx
@@ -36,7 +36,7 @@ function explode(input: string, delimiter: string): string {
 
 /**
  * Makes a textarea that auto resizes based on contents, its height grows with new lines.
- * If a delimeter is set, such as " -" or " ", as used by jvm args or env vars
+ * If a delimiter is set, such as " -" or " ", as used by jvm args or env vars
  * then the field will also "explode" the value by the delimiter over new lines
  * on focus, and implode on blur. By default, it doesn't word wrap.
  */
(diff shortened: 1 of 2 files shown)


packages/components/src/AutoResizeTextarea.tsx
```diff
@@ -35,11 +35,16 @@ function AutoResizeTextarea({
...
+  const isFocused = useRef(false);
 
   useEffect(
     function syncStateWithProp() {
       // keep state value in sync with prop changes
-      setValue(propsValue);
+      // but not while focused, as the display value is exploded
+      // and the prop value might not be
+      if (!isFocused.current) {
+        setValue(propsValue);
```
bmingles, line 65: This seems to completely disable setting the value from props when focused. Can it ever be the case that a parent would provide a different value while the textarea had focus that didn't originate from an onChange from AutoResizeTextarea?

bmingles (changes requested):
Changes look good. I left 1 inline question about disabling value setting when focused.

Also, there is a failing CI check. Seems likely unrelated to your changes, but might be good to re-run

packages/components/src/AutoResizeTextarea.tsx
jnumainville: Not sure if it could happen with the current usages but if nothing else it could theoretically happen. My thought was that overriding a value when the user is editing isn't necessarilly something that needs to happen and it was currently happening just to keep the component in sync, but I recognize that's technically a different behavior than currently expected. We could check the exploded prop value and override if different if we need to maintain that.

bmingles: I guess the decision to make here is what should the behavior be if a prop value were provided that isn't related to the in edit value at all. I'm on the fence.

jnumainville: I think it's reasonable to say that the parent should be the one deciding if it wants to swap out the value in that case, and that's the current behavior, so I will switch it to that. This do not override when editing logic is reasonable as an option rather than the default, but getting beyond the scope of the PR.

jnumainville: Reworked to check for a different (imploded) value if focused with a delimiter.

bmingles (approved):
LGTM

PR #2693, opened 2026-06-01
fix: DH-15773: Coerce string to numeric for input filter

jnumainville:
Input filters always return a string. When passed to the one click back end, this comparison does not work when a partition is on a numeric column.
Note that the linker tool works (at least for the first filter, see https://github.com/deephaven/deephaven-core/pull/8100) as it properly passes through a number.

This code now works with an input filter:
```
from deephaven.plot.selectable_dataset import one_click
source = empty_table(10).update(["X = i", "Y = Math.random()", "Z = i % 2"])
oc = one_click(t=source, by=["Z"])
plot = Figure().plot_xy(series_name="Plot", t=oc, x="X", y="Y").show()
```

### DIFF
--- a/packages/chart/src/FigureChartModel.ts
+++ b/packages/chart/src/FigureChartModel.ts
@@ -12,9 +12,10 @@ import type {
   XAxisName,
   YAxisName,
 } from 'plotly.js';
-import type {
-  DateTimeColumnFormatter,
-  Formatter,
+import {
+  TableUtils,
+  type DateTimeColumnFormatter,
+  type Formatter,
 } from '@deephaven/jsapi-utils';
 import ChartModel, {
   type ChartEvent,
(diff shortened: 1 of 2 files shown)


packages/chart/src/FigureChartModel.ts
```diff
@@ -788,6 +789,30 @@ class FigureChartModel extends ChartModel {
     );
   }
 
+  /**
+   * Coerce a filter value to the appropriate type based on column type.
+   * Input filters always provide string values, but OneClick expects the
+   * native type (e.g. number for numeric columns) to match partition keys.
+   */
+  private coerceFilterValue(value: unknown, columnType: string): unknown {
```
mofojed, line 798: Can we use `TableUtils.makeValue` if it's a string? Need to make sure it works with longs as well.
That being said - you mention that input filters are always a string, but the value is `unknown` here. I don't think input value is always a string, as it could be a raw value (e.g. double clicking a row value that's linked to another table).

mofojed (changes requested):
Just a couple small questions. Unsure if there needs to be changes.

packages/chart/src/FigureChartModel.ts
jnumainville: With this comment I meant using the input filter and typing in a value, not using the linker, which definitely uses the raw value, but also does work with these changes. I updated this comment to be more explicit. Looks like `makeValue` goes the opposite direction. In the backend we'd need to use str as well for the comparison.

mofojed (approved):

PR #2694, opened 2026-06-02
feat: DH-19720: Make Web Advanced Filter Dialog Maximizable

SimonVutov:
Adds a maximize/restore control to the Web Advanced Filter dialog and applies the corresponding layout updates for the maximized state.

Tested in Chrome, Safari, and Firefox.

Has tooltips with "Maximize" and "Restore Size" text.

# Demo:

The state is in the IrisGrid React class component (Field: maximizedAdvancedFilter). IrisGrid is the owner component that renders the popper and holds shownAdvancedFilter and maximizedAdvancedFilter. Maximize button is inside AdvancedFilterCreator, AdvancedFilterCreator is rendered by IrisGrid as Popper content, Popper component is rendered by IrisGrid (IrisGrid (state owner) → Popper → AdvancedFilterCreator → maximize Button). IrisGrid computes whether this specific menu is maximized, then passes class advanced-filter-menu-popper-maximized into Popper containerClassName. The AdvancedFilterCreator.scss is then applied to maximize that modal.

# How this works:

1. IrisGrid owns which advanced filter popper is open and which one is maximized in IrisGrid.tsx.
2. IrisGrid passes isMaximized to Popper (for container layout) in IrisGrid.tsx.
3. IrisGrid passes isMaximized and onToggleMaximize to AdvancedFilterCreator (for button UI + user intent) in IrisGrid.tsx.
4. AdvancedFilterCreator is a controlled child in AdvancedFilterCreator.tsx.

### DIFF
--- a/packages/components/src/popper/Popper.tsx
+++ b/packages/components/src/popper/Popper.tsx
@@ -166,7 +166,7 @@ class Popper extends Component<PopperProps, PopperState> {
   // This is the request animation frame handle number
   rAF: number;
 
-  updateContainerClassName(
+  private updateContainerClassName(
     containerClassName: string,
     isMaximized: boolean
   ): void {
--- a/packages/components/src/popper/Popper.tsx
+++ b/packages/components/src/popper/Popper.tsx
@@ -60,6 +60,7 @@ interface PopperProps {
   children: React.ReactNode;
   options: PopperOptions;
   className: string;
+  containerClassName: string;
   timeout: number;
   onEntered: () => void;
   onExited: () => void;
@@ -81,6 +82,7 @@ class Popper extends Component<PopperProps, PopperState> {
   static defaultProps = {
     options: {},
     className: '',
+    containerClassName: '',
     timeout: ThemeExport.transitionMs,
     onEntered(): void {
       // no-op
@@ -106,7 +108,7 @@ class Popper extends Component<PopperProps, PopperState> {
     this.handleExit = this.handleExit.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
     this.element = document.createElement('div');
-    this.element.className = 'popper-container';
+    this.updateContainerClassName(props.containerClassName);
     this.container = React.createRef<HTMLDivElement>();
 
     // cancelAnimationFrame does nothing if the handle isn't recognized
@@ -122,8 +124,12 @@ class Popper extends Component<PopperProps, PopperState> {
   }
 
   componentDidUpdate(prevProps: PopperProps): void {
-    const { isShown } = this.props;
+    const { isShown, containerClassName } = this.props;
     const { popper } = this.state;
+
+    if (prevProps.containerClassName !== containerClassName) {
+      this.updateContainerClassName(containerClassName);
+    }
 
     if (prevProps.isShown !== isShown) {
       cancelAnimationFrame(this.rAF);
@@ -153,6 +159,10 @@ class Popper extends Component<PopperProps, PopperState> {
 
   // This is the request animation frame handle number
   rAF: number;
+
+  updateContainerClassName(containerClassName: string): void {
+    this.element.className = classNames('popper-container', containerClassName);
+  }
 
   /** Goes through an element and it's parents until the first visible element is found */
   getVisibleElement(element: HTMLElement | null): HTMLElement | null {
--- a/packages/components/src/popper/Popper.tsx
+++ b/packages/components/src/popper/Popper.tsx
@@ -25,6 +25,7 @@ import './Popper.scss';
 import { SpectrumThemeProvider } from '../theme/SpectrumThemeProvider';
 
 const POPPER_CLASS_NAME = 'popper';
+const POPPER_CONTAINER_CLASS_NAME = 'popper-container';
 
 const KEEP_IN_PARENT_OPTIONS: PopperOptions = {
   placement: 'bottom-end',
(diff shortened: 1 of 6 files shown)


packages/iris-grid/src/AdvancedFilterCreator.scss
```diff
@@ -122,3 +122,66 @@
...
+// !important is required because popper.js sets position/top/left/transform as inline styles.
+.popper-container.advanced-filter-menu-popper-maximized {
+  position: fixed !important;
+  top: 50% !important;
+  left: 50% !important;
+  transform: translate(-50%, -50%) !important;
+
+  .advanced-filter-creator {
+    width: min(94vw, 2200px) !important;
+    height: 85vh !important;
```
dsmmcken: sus hardcoded values.

packages/iris-grid/src/AdvancedFilterCreator.scss
```diff
@@ -122,3 +122,66 @@
...
+      // Pin the scroll pane to fill its wrapper. Without this, the fixed
+      // height: 120px rule or the inner virtual list's intrinsic height
+      // (itemCount × rowHeight) would size the pane and defeat virtualization.
+      .select-value-list-scroll-pane {
+        position: absolute;
+        inset: 0;
+        height: auto;
+        max-height: none;
+        min-height: 0;
+      }
```
dsmmcken: some of this maybe could be just the default behaviour and doesn't necessarily need to be nested under maximize, like the idea that the select list growing can apply to the base style and not specific to maximized state (even if thats the only state in which it applies).

dsmmcken (changes requested):

packages/iris-grid/src/AdvancedFilterCreator.scss
SimonVutov: I've named those values at the top of the file now.

packages/iris-grid/src/AdvancedFilterCreator.scss
SimonVutov: The generic layout rules have been moved out of maximize-only scope and now only maximize-specific behavior is kept under .popper-container.advanced-filter-menu-popper-maximized.

packages/components/src/popper/Popper.tsx
```diff
@@ -154,6 +160,10 @@ class Popper extends Component<PopperProps, PopperState> {
   // This is the request animation frame handle number
   rAF: number;
 
+  updateContainerClassName(containerClassName: string): void {
+    this.element.className = classNames('popper-container', containerClassName);
```
mofojed: May as well throw `popper-container` into a constant similar to `POPPER_CLASS_NAME`, e.g. `POPPER_CONTAINER_CLASS_NAME`.

packages/iris-grid/src/AdvancedFilterCreator.scss
```diff
@@ -121,4 +124,67 @@
...
+  }
+}
+
+// Maximized state: override popper.js inline position so the dialog fills the viewport.
+// !important is required because popper.js sets position/top/left/transform as inline styles.
+.popper-container.advanced-filter-menu-popper-maximized {
+  position: fixed !important;
+  top: 50% !important;
+  left: 50% !important;
+  transform: translate(-50%, -50%) !important;
```
mofojed: The only downside of this approach is it's specific to the advanced filter; what if we get a request for another popper to be maximizable?
I think instead this should be built into our Popper wrapper class. E.g. add a prop `isMaximized` to `Popper` that `AdvancedFilterCreator` can just set to `true`.

packages/iris-grid/src/AdvancedFilterCreator.tsx
```diff
@@ -58,6 +63,8 @@ interface AdvancedFilterCreatorProps {
   sortDirection: SortDirection;
   formatter: Formatter;
   tableUtils: TableUtils;
+  isMaximized: boolean;
```
mofojed, line 66: I don't think we need to take this in as a prop, just keep the state internal to AdvancedFilterCreator instead of requiring IrisGrid to keep track.

mofojed (changes requested):
A couple of little things to clean up.

packages/iris-grid/src/AdvancedFilterCreator.tsx
SimonVutov: Right now the setup is:
- IrisGrid owns which advanced filter popper is open and which one is maximized in IrisGrid.tsx.
- IrisGrid passes isMaximized to Popper (for container layout) in IrisGrid.tsx.
- IrisGrid passes isMaximized and onToggleMaximize to AdvancedFilterCreator (for button UI + user intent) in IrisGrid.tsx.
- AdvancedFilterCreator is a controlled child in AdvancedFilterCreator.tsx.

In order to keep the state internal to AdvancedFilterCreator instead of requiring IrisGrid to keep track, we need to move Popper ownership into AdvancedFilterCreator (or a shared wrapper), so the same component controls both maximize state and Popper layout. It might be cleaner to leave it as is. Let me know what you think.

packages/components/src/popper/Popper.tsx
```diff
@@ -160,8 +166,17 @@ class Popper extends Component<PopperProps, PopperState> {
   // This is the request animation frame handle number
   rAF: number;
 
-  updateContainerClassName(containerClassName: string): void {
-    this.element.className = classNames('popper-container', containerClassName);
+  updateContainerClassName(
```
mofojed: Make this method private

mofojed (dismissed):
Looks good. Make sure Don's good with it as well.

mofojed (approved):

dsmmcken (approved):

PR #2695, opened 2026-06-03
fix: DH-16187: Cherry-pick send imploded value to AutoResizeTextarea parent (#2692)

jnumainville:
Cherry-pick of #2692


bmingles (approved):
LGTM

PR #2696, opened 2026-06-04
fix: DH-22803: Update AutoResizeTextarea to handle quoted values

dgodinez-dh:
- add a splitOnDelimiter function to AutoResizeTextarea that handles quoted values correctly (doesn't split inside quotes)
- added unit tests for existing use cases with and without quotes

### DIFF
--- a/packages/components/src/AutoResizeTextarea.test.tsx
+++ b/packages/components/src/AutoResizeTextarea.test.tsx
@@ -482,6 +482,50 @@ describe('AutoResizeTextarea', () => {
       expect(onChange).toHaveBeenCalledWith('FOO="hello world" BAZ=qux');
     });
 
+    it('keeps a double-quoted span containing a single quote on one line', () => {
+      // "it's" contains a single quote inside double quotes — must not break the span
+      renderTextarea({
+        value: `FOO="it's fine" BAZ=qux`,
+        delimiter: DELIMITER,
+      });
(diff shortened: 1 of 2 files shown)


packages/components/src/AutoResizeTextarea.test.tsx
```diff
@@ -335,4 +335,280 @@ describe('AutoResizeTextarea', () => {
...
+      const onChange = jest.fn();
+      renderTextarea({ value: '"/bad /c/d', delimiter: DELIMITER, onChange });
+      fireEvent.focus(getTextarea());
+      // Unbalanced quote is treated as a literal character; split proceeds normally
+      expect(getTextarea()).toHaveValue('"/bad\n/c/d');
+      fireEvent.blur(getTextarea());
+      // Round-trip: original value is preserved
+      expect(onChange).toHaveBeenCalledWith('"/bad /c/d');
+    });
+  });
```
jnumainville, line 657: I don't see test cases for double quotes within single quotes and vice versa. The logic seems correct but those would be good to have.

jnumainville (changes requested):

packages/components/src/AutoResizeTextarea.test.tsx
dgodinez-dh: Added more tests.

jnumainville (approved):

PR #2697, opened 2026-06-11
fix: DH-13522: exclude some aggs for arrays

jnumainville:
CountDistinct, Distinct, Unique aggs now are empty for array columns instead of throwing an error

### DIFF
--- a/packages/iris-grid/src/sidebar/aggregations/AggregationUtils.test.ts
+++ b/packages/iris-grid/src/sidebar/aggregations/AggregationUtils.test.ts
@@ -15,10 +15,11 @@ describe('isValidOperation', () => {
     ];
 
     it.each(ops)('%s returns true for any column type', op => {
-      expect(isValidOperation(op, 'int')).toBe(true);
-      expect(isValidOperation(op, 'java.lang.String')).toBe(true);
-      expect(isValidOperation(op, 'boolean')).toBe(true);
-      expect(isValidOperation(op, 'int[]')).toBe(true);
+      [...numberTypes, ...dateTypes, ...textTypes, ...arrayTypes].forEach(
+        type => {
+          expect(isValidOperation(op, type)).toBe(true);
+        }
+      );
     });
   });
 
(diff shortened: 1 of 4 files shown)


packages/iris-grid/src/sidebar/aggregations/AggregationUtils.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  describe('COUNT, FIRST, LAST', () => {
+    const ops = [
+      AggregationOperation.COUNT,
+      AggregationOperation.FIRST,
+      AggregationOperation.LAST,
+    ];
+
+    it.each(ops)('%s returns true for any column type', op => {
+      expect(isValidOperation(op, 'int')).toBe(true);
```
bmingles: any reason these are single type checks instead of using the type arrays defined at top of test?

packages/iris-grid/src/sidebar/aggregations/AggregationUtils.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  describe('COUNT_DISTINCT, DISTINCT, UNIQUE', () => {
+    const ops = [
+      AggregationOperation.COUNT_DISTINCT,
+      AggregationOperation.DISTINCT,
+      AggregationOperation.UNIQUE,
+    ];
+
+    it.each(ops)('%s returns true for non-array types', op => {
+      expect(isValidOperation(op, 'int')).toBe(true);
```
bmingles: any reason these are single type checks instead of using the type arrays defined at top of test?

bmingles: Same general question for any other tests that only target a single value in cases where we expect all in an array to match.

bmingles (dismissed):
Looks good to me overall. Left a few questions on unit test.

packages/iris-grid/src/sidebar/aggregations/AggregationUtils.test.ts
jnumainville: fixed in all cases

mofojed (approved):

PR #2701, opened 2026-06-15
feat: DH-22800: Support Intersection 2D Window Resizing

SimonVutov:
### Changes
- Adds 2D intersection resizing in Golden Layout by introducing an invisible intersection handle at splitter crossings.
- Uses the existing horizontal and vertical splitter drag logic together: one drag moves both splitters, then applies both size updates atomically on drag stop.
- Reuses existing splitter highlight styling so both crossing lines stay highlighted on hover and during drag (VS Code-like affordance), with proper z-index layering.
- Fixes handle positioning by computing crossing points from bounding-rect centers (offset-parent safe), so nested intersections are reliably grabbable.
- Supports deeply nested crossings by recursively discovering perpendicular splitters that reach the shared edge, not just immediate children.
- Fixes nested drag artifacts by stretching the stem line with CSS transform scale (paint-only) instead of mutating layout width/height, preventing content/tab reflow.
#### Validation
- intersection-drag tests: 9/9 passing: tests in `packages/golden-layout/src/__tests__/intersection-drag.test.ts`

### Demo (23s mov video):
https://github.com/user-attachments/assets/c01af6a0-f739-4aeb-8002-aa008442ea75

### Other Information
The splitter lines are width 10px, and this intersection handle is a square of side length 14px. `this._hitAreaSize = Math.max(this._grabSize, 14);`
Each crossing gets an invisible square handle element and it is centered exactly on the crossing point: It uses `_getIntersectionPosition` to get the intersection position and places the handle centered there with slightly higher z index than splitter lines so that it has priority over them.


packages/golden-layout/scss/goldenlayout-base.scss
```diff
@@ -39,6 +39,11 @@ $height6: 15px; // Appears 1 time
   user-select: none;
 }
 
+.lm_intersection_dragging,
+.lm_intersection_dragging * {
+  cursor: move !important;
+}
```
dsmmcken: Doesn't actually do anything, pointer-events none means cursor wont trigger, and the pointer-events none is needed to prevent interaction with other elements while dragging.

packages/golden-layout/scss/goldenlayout-base.scss
```diff
@@ -79,6 +84,32 @@ $height6: 15px; // Appears 1 time
...
+  }
+}
+
+// Intersection splitter: an invisible grab area at grid crossing points that
+// enables simultaneous (2D) resizing of rows and columns. The visible drag
+// affordance is the two perpendicular splitter lines themselves, which are
+// highlighted (via .lm_dragging) on hover and while dragging.
+.lm_intersection_splitter {
+  position: absolute;
+  z-index: 60;
```
dsmmcken: these z-indexes look sus.

packages/golden-layout/scss/goldenlayout-dark-theme.scss
```diff
@@ -133,6 +133,13 @@ body:not(.lm_dragging) .lm_header .lm_tab:hover .lm_close_tab {
   }
 }
 
+// Intersection splitter (2D grid resize). The grab area is invisible; the drag
+// affordance is the two perpendicular splitter lines, which reuse the standard
+// .lm_splitter.lm_dragging highlight above.
+.lm_intersection_splitter {
+  background: transparent;
+}
```
dsmmcken: Does nothing, transparent by default.

dsmmcken (changes requested):
2d drag, over another splitter triggers it and it gets stuck hovered;
Image


I wonder if we could detect 4-way corners, with a bit of fuzziness (like if the 4th corner is +/- 5px or 10px or whatever), and make them a 4-way drag (that leaves it snapped perfectly aligned on drop)?

Image


Didn't review JS, will leave to someone else to review that.

packages/golden-layout/src/items/RowOrColumn.ts
```diff
@@ -604,9 +648,741 @@ export default class RowOrColumn extends AbstractContentItem {
...
+   * so descend until the edge is no longer shared.
+   *
+   * @param nearEdge true when the bar sits at the start of `item` along the bar
+   * main axis (junction at the near end), false when at the end.
+   */
+  private _collectEdgeStemSplitters(
+    item: AbstractContentItem | undefined,
+    nearEdge: boolean,
+    path: string,
+    out: {
```
mofojed: I don't like using an `out` parameter like this. Why not just return an array of the splitters?

packages/golden-layout/src/items/RowOrColumn.ts
```diff
@@ -604,9 +648,741 @@ export default class RowOrColumn extends AbstractContentItem {
...
+    return ensuredKeys;
+  }
+
+  /**
+   * Collect every perpendicular splitter line within `item`'s subtree that
+   * reaches the shared edge with one of this row/column's splitter bars, so a
+   * crossing handle can be created for it. Lines can be nested arbitrarily deep,
+   * so descend until the edge is no longer shared.
+   *
+   * @param nearEdge true when the bar sits at the start of `item` along the bar
```
mofojed, line 754: JSDocs need to be updated

mofojed (changes requested):
Need to make sure the JSDocs on these functions include all parameters and are up to date. Should also add a couple e2e tests 

There's still some odd behaviours in some scenarios.
1. When dragging from a 1D splitter and then dragging over a intersection point, it changes lights up the whole cross for a moment and then all the highlights disappear as you drag away; however when you drop the change still occurs
https://github.com/user-attachments/assets/f65b9284-077c-4b10-97d0-c845c4d3cd0a

2. When dragging a column in the grid, then dragging over an intersection point, the cursor changes and the splitters highlight unnecessarily. (I would have thought this wouldn't be possible because I thought the grid turns off pointer events for the document when a column drag is occurring).
https://github.com/user-attachments/assets/f0397b7f-412d-44c8-b757-b037131e7e7b

I like the feature though, great work overall.

packages/golden-layout/src/items/RowOrColumn.ts
```diff
@@ -857,9 +865,39 @@ export default class RowOrColumn extends AbstractContentItem {
...
+   * the intersection handle must not light up its cross or swap the cursor as
+   * the pointer passes over a crossing.
+   *
+   * @returns True when a foreign drag is active.
+   */
+  private static _isAnyDragInProgress(): boolean {
+    return (
+      $(document.body).hasClass('lm_dragging') ||
+      $(document.body).hasClass('lm_intersection_dragging') ||
+      document.documentElement.classList.contains('grid-block-events')
```
mofojed: golden-layout should not depend on grid, or any grid specific class.
I don't think this is actually needed for anything? Just remove this grid-block-events check.

mofojed (changes requested):

mofojed (approved):
JS looks good, works fine. z-indexs could be variables but that applies to goldenlayout-base as a whole, not just this change.
@dsmmcken to give approval then we'll merge

dsmmcken (approved):

PR #2702, opened 2026-06-16
chore: DH-22912: Re-export dnd-kit and font-awesome from web-client-ui

vbabich:


packages/iris-grid/src/RemoteComponentModules.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Re-exports third-party module namespaces so plugins share the host app's
+ * single bundled copy via the remote component resolve map.
+ *
+ * `@deephaven/iris-grid` owns the `@dnd-kit/*` packages, so re-exporting them
+ * here lets grid-related plugins share the same dnd-kit instances.
+ */
+export * as DndKitCore from '@dnd-kit/core';
+export * as DndKitModifiers from '@dnd-kit/modifiers';
+export * as DndKitSortable from '@dnd-kit/sortable';
+export * as DndKitUtilities from '@dnd-kit/utilities';
```
vbabich, line 11: Should re-export DndKit from `@deephaven/components` instead of `iris-grid`?

mofojed: It's only used by iris-grid, let's leave it in iris-grid. If we want to change it later, iris-grid can deprecate and re-export from `@deephaven/components`.

mofojed (approved):

PR #2704, opened 2026-06-17
feat: DH-21344: Dropdown for input table enums

dgodinez-dh:
Consumes the new Column Restriction JS API to add an dropdown cell renderer for input tables. The renderer is pass to Grid via a function in a registry in a context around IrisGrid. This will allow Enterprise to register more input table cell renderers.


packages/grid/src/Grid.tsx
```diff
@@ -2280,16 +2291,30 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+    let cellBackgroundColor: string | undefined;
+    if (rowBackgroundColors) {
+      const colorSets = GridRenderer.getCachedBackgroundColors(
+        rowBackgroundColors,
+        maxDepth
+      );
+      const depth = isExpandableGridModel(model) ? model.depthForRow(row) : 0;
+      const colorSet = colorSets[row % colorSets.length];
+      cellBackgroundColor = colorSet[Math.min(depth, colorSet.length - 1)];
+    }
```
mofojed: Hmm this doesn't take into account if the cell background has been set via formatting. I don't particularly like how it's duplicating a bunch of the logic in GridRenderer either, but I don't have a good answer for that.

packages/grid/src/Grid.tsx
```diff
@@ -2311,19 +2336,45 @@ class Grid extends PureComponent<GridProps, GridState> {
         ? model.isValidForCell(modelColumn, modelRow, value)
         : false;
 
+    const { cellInputRendererRegistry } = this.props;
+    const columnRestrictions =
+      modelColumn != null ? model.getColumnRestriction(modelColumn) : [];
```
mofojed: Prefer to use `EMPTY_ARRAY`
```suggestion
      modelColumn != null ? model.getColumnRestriction(modelColumn) : EMPTY_ARRAY;
```

packages/grid/src/Grid.tsx
```diff
@@ -2311,19 +2336,45 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+        ) : (
+          <CellInputField
+            selectionRange={selectionRange}
+            className={classNames({ error: !isValid })}
+            onCancel={this.handleEditCellCancel}
+            onChange={this.handleEditCellChange}
+            onDone={this.handleEditCellCommit}
+            isQuickEdit={isQuickEdit}
+            style={inputStyle}
+            value={value}
```
mofojed:
```suggestion
            {...cellInputProps}
```

packages/grid/src/Grid.tsx
```diff
@@ -2311,19 +2336,45 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+      isQuickEdit,
+      style: inputStyle,
+      value,
+      columnRestrictions,
+    };
+
+    const renderer =
+      columnRestrictions.length === 1
+        ? cellInputRendererRegistry?.get(columnRestrictions[0].type)
+        : undefined;
```
mofojed, line 2376: So if column restrictions has more than one element, we just use undefined? Then what's the point of having an array for column restrictions?

packages/grid/src/GridRendererTypes.ts
```diff
@@ -1,14 +1,45 @@
...
 
+/**
+ * A renderer for a single cell input field based on column restriction type.
+ *
+ * Set `preservesExistingValue = true` on the function to signal that
+ * keystroke-initiated edits should open the editor with the existing cell
+ * value rather than replacing it with the typed character.
+ */
+export type CellInputRendererFn = ((
+  props: CellInputFieldProps & { columnRestrictions: ColumnRestriction[] }
```
mofojed: I think just `restrictions` makes more sense here? Is there a reason to specify the restrictions are for the whole column?
From the perspective of the cell input, it should only matter what the restrictions are for that cell. Doesn't matter if those restrictions are applied to the whole column or just that row for some reason.
Also, should be a `readonly` array provided.
Also strange how it's tacked on to the type here, instead of this type being defined on it's own. Feel like it should just be on `CellInputProps` and that's defined in this file, and then `CellInputField` should reference this for it's props (just ignoring the restrictions in that case).
Also weird that we have an `onContextMenu` prop that is never used. I wonder why that is. Maybe we should just remove that prop.

packages/grid/src/GridRendererTypes.ts
```diff
@@ -1,14 +1,45 @@
...
+) => React.ReactNode) & {
+  /**
+   * When true, keystroke-initiated edits preserve the existing cell value
+   * instead of replacing it with the typed character. Intended for renderers
+   * like dropdowns where the typed character has no meaning as a new value.
+   */
+  preservesExistingValue?: boolean;
```
mofojed, line 38: Is there another PR where this is used?

packages/iris-grid/src/key-handlers/RestrictedEditKeyHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+      return false;
+    }
+
+    const { model } = grid.props;
+    const { cursorColumn, cursorRow } = grid.state;
+    if (
+      cursorColumn == null ||
+      cursorRow == null ||
+      !isEditableGridModel(model) ||
+      !model.isEditableRange(GridRange.makeCell(cursorColumn, cursorRow))
```
mofojed: Hmm I see this is already in the EditKeyHandler, but it's incorrect that we're passing the visible column/row down instead of the model index. I think this might cause problems if you move a key column and then try to edit.

packages/iris-grid/src/key-handlers/RestrictedEditKeyHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+  isEditableGridModel,
+  GridRange,
+} from '@deephaven/grid';
+import type IrisGrid from '../IrisGrid';
+import { DEFAULT_REGISTRY } from '../CellInputRendererContext';
+
+/**
+ * Key handler that intercepts printable-character keystrokes on cells whose
+ * column restriction type has a registered custom renderer (e.g. a dropdown).
+ *
```
mofojed: Why isn't this just baked into the EditKeyHandler? I don't think any of this is IrisGrid specific.

packages/iris-grid/src/CellDropdownField.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { useCallback, useRef, useState } from 'react';
+import classNames from 'classnames';
+import { EMPTY_FUNCTION } from '@deephaven/utils';
+import { Item, Picker, type ItemKey } from '@deephaven/components';
+import { SELECTION_DIRECTION } from '@deephaven/grid';
+import './CellDropdownField.scss';
+
+export type CellDropdownFieldProps = {
```
mofojed: Should extend the `CellInputProps` as mentioned in my other comment.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -534,10 +536,10 @@ export interface IrisGridState {
 }
 
 class IrisGrid extends Component<IrisGridProps, IrisGridState> {
-  static contextType = IrisGridThemeContext;
+  static contextType = IrisGridContext;
```
mofojed, line 537: This is a breaking change now; existing users of `IrisGridThemeContext` (including `ThemeBootstrap`) will not be wired up correctly; they need to add the `IrisGridContextProvider` as a child to the `ThemeBootstrap` (since `IrisGridContextProvider` still uses the `IrisGridThemeContext`). Though having `IrisGridContextProvider` makes more sense when it's providing more things than just the theme.
Hmm. Obviously functional component would make this a little bit easier to handle both contexts, but that's a huge change... 
I think instead, I'd just rename `IrisGridThemeProvider` and `IrisGridThemeContext` to `IrisGridContextProvider` and `IrisGridContext`, have all the functionality in there, and re-export as `IrisGridThemeProvider`/`IrisGridThemeContext` with a deprecated tag. Then we don't break existing users/UIs, and we can build on top of it.

packages/iris-grid/src/IrisGridModel.ts
```diff
@@ -690,6 +691,25 @@ abstract class IrisGridModel<
   tooltipForCell(column: ModelIndex, row: ModelIndex): string | null {
     return null;
   }
+
+  getColumnRestriction(column: ModelIndex): ColumnRestriction[] {
```
mofojed: `getColumnRestriction` - but it returns an array? Should it be `getColumnRestrictions`, or are we just sticking to one restriction for each column? We only seem to use one restriction in the renderer anyways, so why do we allow an array?

mofojed (changes requested):

mofojed:
How should I go about testing/using this?

dgodinez-dh:
> How should I go about testing/using this?

I've been testing by running the branch with latest `deephaven-core`. There is a script in the test plan for the ticket:
https://deephaven.atlassian.net/browse/DH-21344

packages/grid/src/Grid.tsx
dgodinez-dh: Formatting is now taken into account. I tried to refactor for common code, but the GridRenderer does row stripes and formats and two different passes, so I ended up with something that seemed less readable (which I did not commit).

packages/grid/src/Grid.tsx
dgodinez-dh: We need to discuss this. It is an array of restrictions because a user can add as many as they want. I should have a comment on this code, but I think this code is essentially saying, "The UI doesn't know how to render multiple column restrictions, so if you have more than one then we use the default cell renderer."

That being said, maybe there is a discussion to be had about how we render columns with multiple restrictions. But I feel like it would be difficult to create an algorithm or hierarchy that would handle all cases correctly given that a user can combine an restrictions even if they make no sense.

packages/grid/src/GridRendererTypes.ts
dgodinez-dh: Updated

packages/grid/src/GridRendererTypes.ts
dgodinez-dh: No. This is only used here for pulldowns. For input table cells, the current behavior is is to overwrite the value of the cell if the user hits a key on the keyboard. For a pulldown, we want to start editing, but don't overwrite the value. The flag is so that future cell renderers can handle this differently.

packages/iris-grid/src/key-handlers/RestrictedEditKeyHandler.ts
dgodinez-dh: Updated.

packages/iris-grid/src/key-handlers/RestrictedEditKeyHandler.ts
dgodinez-dh: Updated

packages/iris-grid/src/CellDropdownField.tsx
dgodinez-dh: Updated

packages/iris-grid/src/IrisGridModel.ts
dgodinez-dh: Updated to be plural, but per my earlier comment, we should discuss what plural restrictions means.

packages/iris-grid/src/IrisGrid.tsx
dgodinez-dh: Updated

packages/grid/src/Grid.tsx
dgodinez-dh: Added comment.

packages/iris-grid/src/CellDropdownField.scss
```diff
@@ -0,0 +1,60 @@
+.cell-dropdown-field {
```
mofojed, line 3: Hmm I wonder if we can use the `grid-cell-input-field` styling. There's a couple of (minor but) annoying styling issues, such as the border-radius not being set. See the little jump here:

https://github.com/user-attachments/assets/c709357a-a9f6-4541-bd88-3ee660e070b6

Contrast this with a regular cell input, where the transition is pretty seamless:
https://github.com/user-attachments/assets/f5725d90-c659-47b2-bdf1-cd175e780cce

There's also a bit of a mismatch when the Compact theme is selected, it looks like the font size changes. I don't know why that is, and it's also on the regular cell input field. https://deephaven.atlassian.net/browse/DH-21507

mofojed (changes requested):
Looks good, about ready to merge, just some minor styling stuff with the dropdown. Take a couple minutes to look into the compact styling as well for both this and the regular input field, see if that's a quick fix.

packages/iris-grid/src/CellDropdownField.scss
dgodinez-dh: Applied changes to use the grid cell input field styling and fix the compact font size.

packages/iris-grid/src/IrisGrid.scss
```diff
@@ -153,6 +153,17 @@ $cell-invalid-box-shadow:
         }
       }
     }
+
+    // Apply the same focus-ring box-shadow to CellDropdownField that
+    // CellInputField (the textarea) receives, so the cell outline is
+    // visible when a dropdown editor is open.
+    .cell-dropdown-field {
```
mofojed: This should be in `CellDropdownField.scss` instead. You may need to move some variables to `IrisGridTheme.module.scss` (e.g. `$cell-box-shadow`, `$cell-invalid-box-shadow`

packages/iris-grid/src/CellDropdownField.scss
```diff
@@ -0,0 +1,56 @@
...
+  align-items: center;
+
+  // Override Spectrum Picker sizing to match grid cell dimensions.
+  // --spectrum-dropdown-height drives the button block-size and internal padding.
+  --spectrum-dropdown-height: var(--grid-row-height, 20px);
+  // Remove Spectrum's minimum width so the Picker can be as narrow as the cell.
+  --spectrum-dropdown-quiet-min-width: 0;
+  --spectrum-dropdown-min-width: 0;
+  // Match the grid font size (Spectrum reads this variable for the label text).
+  --spectrum-dropdown-text-size: var(--grid-font-size, 12px);
```
mofojed, line 27: I think we need to set the font-size in CellInputField as well. It's currently growing, and showing the scrollbar when in compact mode.

mofojed (changes requested):
The styling looks good, but the left edge is still moving inward a couple pixels. The border should stay in the same spot.

https://github.com/user-attachments/assets/46ebfb83-59b3-44f8-bc00-5d403ac5490e

packages/iris-grid/src/IrisGrid.scss
dgodinez-dh: Updated

packages/iris-grid/src/CellDropdownField.scss
dgodinez-dh: Changing the font size did not fix the scrollbar, so add a line-height 1. Additionally, I noticed that the spacious theme aligned text to the top of the cell, so calculated padding to fix.

mofojed (approved):
@dgodinez-dh There was still an issue with the text and border shifting when going into Edit mode. I got Claude to fix it in Compact, Regular, and Spacious for both the input field and the dropdown, and now it looks good across all the cases. Might need you to approve so this can go in.

PR #2707, opened 2026-06-24
feat: DH-22956: Worker variables context and store

vbabich:
This PR adds subscription-based infrastructure for discovering a worker's variables (for example, whether a worker exposes a PivotService). Plugins subscribe and react as variables appear, including ones that show up only after the worker finishes initializing, rather than relying on a single point-in-time lookup that could run too early.

### Changes

- **`jsapi-bootstrap`**: Adds `WorkerVariablesContext` and the `useWorkerVariables` hook, which subscribes to a worker's field updates and exposes the current set of variables, updating reactively as variables are added or removed.
- **`jsapi-utils`**: Adds `WorkerVariablesStore` to back the subscription and hold the latest variable state.
- **`app-utils`**: Wires the provider into `ConnectionBootstrap` so the variables are available to Core consumers.

### Notes

This PR contains the shared infrastructure only. The DHE client and the pivot-builder plugin adopt `useWorkerVariables` in their respective PRs, which depend on this one being published first.

### DIFF
--- a/packages/app-utils/src/components/ConnectionBootstrap.tsx
+++ b/packages/app-utils/src/components/ConnectionBootstrap.tsx
@@ -14,9 +14,14 @@ import {
   type UriVariableDescriptor,
   useApi,
   useClient,
+  WorkerVariablesContext,
 } from '@deephaven/jsapi-bootstrap';
 import type { dh } from '@deephaven/jsapi-types';
 import Log from '@deephaven/log';
+import {
+  createWorkerVariablesStore,
+  type WorkerVariablesStore,
+} from '@deephaven/jsapi-utils';
 import { assertNotNull } from '@deephaven/utils';
 import { vsDebugDisconnect } from '@deephaven/icons';
 import ConnectionContext from './ConnectionContext';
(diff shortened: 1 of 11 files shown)


packages/jsapi-utils/src/WorkerVariablesStore.ts
```diff
@@ -0,0 +1,231 @@
...
+        if (gen !== entry.generation) return;
+        const removedIds = new Set(changes.removed.map(v => v.id));
+        const updatedIds = new Set(changes.updated.map(v => v.id));
+        const next = (entry.list ?? []).filter(
+          v => !removedIds.has(v.id) && !updatedIds.has(v.id)
+        );
+        next.push(...changes.updated, ...changes.created);
+        entry.list = next;
+        notify(entry);
+      });
```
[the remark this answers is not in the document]
vbabich: Fixed

packages/app-utils/src/components/ConnectionBootstrap.tsx
```diff
@@ -179,6 +184,25 @@ export function ConnectionBootstrap({
...
+  // connection, so the resolver ignores the worker key.
+  const [workerVariablesStore, setWorkerVariablesStore] =
+    useState<WorkerVariablesStore | null>(null);
+  useEffect(
+    function manageWorkerVariablesStore() {
+      if (connection == null) {
+        setWorkerVariablesStore(null);
+        return undefined;
+      }
+      const store = createWorkerVariablesStore(async () => connection);
```
mofojed, line 197: I assume Enterprise will have a similar `createWorkerVariablesStore` but actually pass an async function - do you have a draft PR with that?

mofojed (approved):

packages/app-utils/src/components/ConnectionBootstrap.tsx
vbabich: DHE draft here - https://github.com/deephaven-ent/iris/pull/4525/changes#diff-6619702e724472c37d7aec3ee8ddad0e7f552e00f4aaffb1b6455eb43388c106
I'm going to clean up and split out `WorkerVariablesContext` and `ObjectFetcherManager` into a separate PR.

PR #2709, opened 2026-06-26
feat: DH-22990: improve clear command behaviour in console input

dsmmcken:
Typing `clear`/`cls` in the console was unreliable — Monaco autocomplete would hijack Enter with a fuzzy match instead of running the command. This makes clearing the console reliable and discoverable.

- a single enter now always runs `clear`/`cls` even with the suggest widget open
- `clear`/`cls` now show in autocomplete (without suppressing word-based suggestions).
- added "Clear" to the console overflow menu
- added cmd/ctrl+l shortcut (updates live when rebound in settings), replacing previous shortcut that was supposed to clear the input but didn't actually do anything because monaco ate it

Not tested on mac. Please test.

Not sure how much gets reused in enterprise of this.

### DIFF
--- a/packages/console/src/monaco/MonacoUtils.ts
+++ b/packages/console/src/monaco/MonacoUtils.ts
@@ -56,6 +56,13 @@ class MonacoUtils {
 
     registerLanguages([DbLang, PyLang, GroovyLang, LogLang, ScalaLang]);
 
+    // Register the special console commands (clear/cls) as completions. Uses the
+    // '*' selector so it shares Monaco's low-priority word-based completion group
+    // and coexists with it, instead of suppressing word-based suggestions.
+    monaco.languages.registerCompletionItemProvider('*', {
+      provideCompletionItems: MonacoProviders.provideConsoleCommandCompletions,
+    });
+
     monaco.languages.onLanguage('python', () => {
       monaco.languages.registerCodeActionProvider(
         'python',
--- a/packages/console/src/ClearCommands.ts
+++ b/packages/console/src/ClearCommands.ts
@@ -1,0 +1,16 @@
+/**
+ * Special console commands that clear the console output/history. These are
+ * handled entirely client-side (see Console.handleCommandSubmit) and never sent
+ * to the server.
+ */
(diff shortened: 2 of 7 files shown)


mofojed:
@dsmmcken not sure this is the "right" fix. As I commented on the ticket, native Python REPL in Python just sends the command when you press Enter: https://deephaven.atlassian.net/browse/DH-22990?focusedCommentId=73600
You press Tab to autocomplete.
We should probably do the same, as it's annoying to hit Esc then Enter to just enter stuff.

packages/console/src/monaco/MonacoUtils.ts
```diff
@@ -56,6 +56,13 @@ class MonacoUtils {
 
     registerLanguages([DbLang, PyLang, GroovyLang, LogLang, ScalaLang]);
 
+    // Register the special console commands (clear/cls) as completions. Uses the
+    // '*' selector so it shares Monaco's low-priority word-based completion group
+    // and coexists with it, instead of suppressing word-based suggestions.
+    monaco.languages.registerCompletionItemProvider('*', {
```
mofojed, line 62: This doesn't seem to work. I think it just most specific completion provider available, as I just see the regular suggestions. If I comment out the other registration (e.g. in `MonacoProviders`), I then see these suggestions.

mofojed (changes requested):
The registered suggestions don't appear.

I really don't like this change. It's unexpected behaviour - if I have a variable named `clear_color` or `cls_name`, and then type `cls` and press "Enter", it will just clear my console instead of picking that variable name (which is different behaviour than before, and different behaviour than if you had typed literally _anything_ else).

As mentioned, I really think we should just take "Enter" to run the current command, "Tab" to take the suggestion, as other Python REPLs do.
_OR_ correctly add the `clear` and `cls` as commands in the autosuggestion, so you need to press Enter to get the suggestion, then Enter again to send it.

Note also looked at Chrome console and Firefox Console.
Chrome console just runs the command on Enter. It doesn't autohighlight the first suggestion, and shows `tab` in it to get the suggestion:
Image
If you press Down on the keyboard, then it highlights it:
Image

Firefox shows the item highlighted, and when you press Enter, it selects that item.
Image

PR #2711, opened 2026-06-30
fix: DH-23004: fall back to TablePluginLoaderContext in useLoadTablePlugin

mofojed:
- To support loading legacy Enterprise plugins from Enterprise (e.g. plugins built with this template: https://github.com/deephaven/js-plugin-template)
- Used in Enterprise with https://github.com/deephaven-ent/iris/pull/4911
- Deployed to https://bender-dh-23004.int.illumon.com:8123/iriside/, just run a Legacy Groovy studio and run:
```
t=db.t("LearnDeephaven", "StockTrades").where("Date=`2017-08-21`")
t.setAttribute("PluginName", "ExampleTablePlugin")
```

### DIFF
--- a/packages/dashboard-core-plugins/src/TablePluginLoaderContext.ts
+++ b/packages/dashboard-core-plugins/src/TablePluginLoaderContext.ts
@@ -1,0 +1,14 @@
+import { createContext } from 'react';
+import { type TablePluginComponent } from '@deephaven/plugin';
+
+/**
+ * Context providing an optional function to load a table plugin by name.
+ * When provided, `useLoadTablePlugin` will call this function before falling
+ * back to the built-in plugin registry lookup.
+ */
+export const TablePluginLoaderContext = createContext<
+  ((name: string) => TablePluginComponent) | null
+>(null);
+TablePluginLoaderContext.displayName = 'TablePluginLoaderContext';
+
+export default TablePluginLoaderContext;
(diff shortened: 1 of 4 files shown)


packages/dashboard-core-plugins/src/TablePluginLoaderContext.ts
```diff
@@ -0,0 +1,14 @@
...
+import { type TablePluginComponent } from '@deephaven/plugin';
+
+/**
+ * Context providing an optional function to load a table plugin by name.
+ * When provided, `useLoadTablePlugin` will call this function before falling
+ * back to the built-in plugin registry lookup.
+ */
+export const TablePluginLoaderContext = createContext<
+  ((name: string) => TablePluginComponent) | null
+>(null);
```
mattrunyon, line 11: Is this sync in enterprise? Will it work (or do we care) with lazy loading plugins?

mattrunyon (commented):
It's also worth considering (although it is probably enterprise specific) if we want to load the plugins in the same format as that old example. It looked like it was just a default export of a component for the table plugin. I guess we could keep supporting for just table plugins, but other plugins we'd need to know what type (and I think we had just the named exports there before?)

packages/dashboard-core-plugins/src/TablePluginLoaderContext.ts
mofojed: It's sync in Enterprise, the lazy loading is contained within the component returned: https://github.com/deephaven-ent/iris/blob/654f61458a705a3b58eacf7048411edadb3a8a05/web/client-ui/src/plugins/PluginUtils.tsx#L54

mofojed:
@mattrunyon Previously it was just exporting default, because we knew you were trying to load a TablePlugin when we tried to load the plugin. I don't want to force them to migrate all their old plugins to change the export when it's simple enough to continue support for the existing format.

vbabich (approved):
LGTM

PR #2712, opened 2026-06-30
fix: DH-22976: re-apply the quickFilters prop on change

SimonVutov:
Make IrisGrid re-apply the quickFilters prop on change, so the parent can own/control it. Same as with sorts.

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -1001,6 +1001,7 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       settings,
       model,
       customFilters,
+      quickFilters,
       sorts,
       getMetricCalculator,
     } = this.props;
@@ -1057,6 +1058,9 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     }
     if (sorts !== prevProps.sorts) {
       this.updateSorts(sorts);
+    }
+    if (quickFilters !== prevProps.quickFilters) {
+      this.updateQuickFilters(quickFilters);
     }
     const { loadingScrimStartTime, loadingScrimFinishTime } = this;
     if (loadingScrimStartTime != null && loadingScrimFinishTime != null) {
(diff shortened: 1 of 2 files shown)


packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3016,6 +3020,14 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     this.grid?.forceUpdate();
   }
 
+  updateQuickFilters(quickFilters: ReadonlyQuickFilterMap | null): void {
+    this.startLoading('Filtering...', { resetRanges: true });
+    this.setState({
+      quickFilters: quickFilters != null ? new Map(quickFilters) : new Map(),
+    });
```
mofojed: As Copilot mentions, we shouldn't be cloning this map. We should just use the map as is.
```suggestion
    this.setState({
      quickFilters: quickFilters ?? EMPTY_MAP,
    });
```

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1058,6 +1059,9 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     if (sorts !== prevProps.sorts) {
       this.updateSorts(sorts);
     }
+    if (quickFilters !== prevProps.quickFilters) {
```
mofojed: We should add some unit tests.
Another thing of note - if we apply input filter changes (checked above), those update the quick filters; if quick filters change as well, then we just replace those changes that were just applied from the input filters. I think that's probably okay, and we're unlikely to get a conflict like that.

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1056,10 +1057,13 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
     }
-    if (sorts !== prevProps.sorts) {
+    if (sorts !== prevProps.sorts && !deepEqualEs6(sorts, prevProps.sorts)) {
       this.updateSorts(sorts);
     }
-    if (quickFilters !== prevProps.quickFilters) {
+    if (
+      quickFilters !== prevProps.quickFilters &&
+      !deepEqualEs6(quickFilters, prevProps.quickFilters)
+    ) {
```
mofojed: We shouldn't need `deepEquals` checks here - if the object is different, then we should apply those changes.

mofojed (changes requested):

mofojed (approved):

PR #2713, opened 2026-07-01
fix: DH-23018: Update PluginUtils to use the new manifest schema

vbabich:
Update `PluginUtils` to get the package name and dependencies for the remote component loader from the `loader` namespace in the manifest.


bmingles (approved):
LGTM

PR #2715, opened 2026-07-06
fix: DH-22961: Content toolbar buttons overflow

SimonVutov:
Toolbar buttons overflow outside panel when narrow

When a script editor panel is resized to a small width, the toolbar buttons (Run, Run Selected, Save) now hide and move into the overflow (⋮) menu instead of rendering outside the panel boundary. A `ResizeObserver` tracks the toolbar width and collapses the buttons when it drops below 160px.

Demo Video:

https://github.com/user-attachments/assets/354f8bc6-ce6f-41e2-b706-f5569c93409b

### DIFF
--- a/packages/dashboard-core-plugins/src/panels/NotebookPanel.scss
+++ b/packages/dashboard-core-plugins/src/panels/NotebookPanel.scss
@@ -19,12 +19,22 @@ $toolbar-bg: $content-bg;
 
   .notebook-toolbar {
     display: flex;
-    justify-content: center;
+    justify-content: flex-start;
     flex-direction: row;
     background: $toolbar-bg;
     border-bottom: 2px solid $background;
     padding: $spacer-1 $spacer-2;
     gap: $spacer-1;
+    container-type: inline-size;
+    overflow: hidden;
+
+    .notebook-toolbar-actions {
+      display: contents;
+    }
+
+    .btn-overflow {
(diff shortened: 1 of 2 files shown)


mofojed (changes requested):
@dsmmcken will want to take a look at this first. 
We should probably use container queries to handle this better: https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries

From a code standpoint, you would always put the buttons in the toolbar and in the overflow menu, with a specific class on the items we want to hide from the overflow menu; then CSS would hide those items if the media query matches (e.g. min-width >= 160)

dsmmcken:
Yeah, I would do that with a container query. Also I think if this collapses, the menu button still escapes the panel, so that toolbar still needs to have it's overflow constrained, as dots themselves can probably still escape similar to how the existing how row can:

Screenshot 2026-07-06 at 5 22 38 PM

I would probably just hide with a container query, we could just always have those options in the menu though? (with the shortcuts shown). Would almost want to hide 1 at a time priorizing from left to right + overflow menu?

SimonVutov:
Thanks for the feedback. It's now using the css container queries and hides them one by one. (PR description video shows updated behavior) Run, Run Selected, and Save are always available in the ⋮ menu.

mofojed (approved):

PR #2717, opened 2026-07-09
fix: DH-22991: Command history is disappearing

SimonVutov:
Preserve scroll position across Golden Layout `replaceChild` DOM swap
### Fix:

1. **A global passive scroll listener** (`handleDocumentScroll`) fires whenever *anything* scrolls. It maintains the `scrollOffsetRegistry` map:
   - Scrolled to a non-zero offset → **add/update** the element with its `{ scrollTop, scrollLeft }`.
   - Scrolled back to `0,0` → **remove** it from the map.

2. **On every `replaceChild`**, `capturePreservedScrollOffsets` walks that map (not the DOM), and while doing so:
   - **Prunes** entries whose element is no longer in the DOM (`!element.isConnected`)
   - **Keeps** the entries whose element is inside the subtree being moved (`rootElement.contains(element)`).

3. After the DOM swap, `restorePreservedScrollOffsets` writes those saved offsets back onto the (re-attached) elements.

So the map is a small, self-maintaining list of "things currently scrolled anywhere," and each `replaceChild` just restores the relevant subset. No per-replace tree walk, and it works for every panel/window generically.

### Summary / Findings

Scroll is normal before `parentNode?.replaceChild(newChild.element[0], oldChild.element[0]);` (line 288 of `packages/golden-layout/src/items/AbstractContentItem.ts`), but `0` after this line when reproducing the bug.

Investigation showed that the first observed scroll position reset occurs immediately after the DOM replacement inside `packages/golden-layout/src/items/AbstractContentItem.ts`, before the `destroy`, `init`, `setSize`, and `ResizeObserver` paths execute.
* The first observed `scrollTop` transition immediately after `parentNode.replaceChild(...)`.
* `ResizeObserver`, rendering, and viewport updates all occur later and observe an already-reset scroll position.
* No application-initiated `scrollTop`, `scrollTo()`, or `scrollBy()` writes were observed immediately before the transition.


packages/components/src/ItemList.tsx
```diff
@@ -537,6 +537,9 @@ export class ItemList<T> extends PureComponent<
   }
 
   handleResize({ height }: Size): void {
+    if (height == null || height <= 0) {
```
mofojed: Hmm... so this raises a couple questions:
- _Why_ is AutoSizer reporting a size of 0? In the example, the container just gets bigger, is there something we need to fix upstream? Having logic just in the ItemList fixes this case, but will this break other things?
- I think when the command history gets backgrounded (e.g. you switch to the "File Explorer" tab and "Command History" is backgrounded), the ItemList gets a height of 0 in that case, which is legit - we shouldn't need to keep the viewport open when the command history isn't visible anyways. Check what the behaviour is when the ItemList gets backgrounded, and then comes back (compare to previous behaviour). Also if you scroll to the bottom, background it, then enter new commands (which appends new commands), then come back - is it still pinned to the bottom? Etc.

SimonVutov: The blank frame had `scrollOffset: 8418` (scrolled to bottom) while the loaded window was `[0, 1)`, which can only happen if `AutoSizer` reported height 0 and pushed a `setViewport(0,0)` that collapsed the loaded rows. So the three guards (ignore 0-height resize, don't emit a 0-height viewport, keep last-good height in render) each block one step of that chain and fix this without touching normal resize behavior. On *why* AutoSizer reports 0: it measures via `getBoundingClientRect().height` and fires `onResize` on every intermediate value during a reflow, so when a docked widget is closed and Golden Layout reflows the column, the container's height passes through 0 for a frame before settling larger. I don't think there's an upstream fix worth making here, since reporting intermediate sizes during a reflow is expected AutoSizer behavior. A height of 0 should never cause us to change what data we've loaded; at most it means it shouldnt render right now, so the right fix is for the list to ignore it rather than turn it into a `setViewport(0,0)` that throws away the rows we already have.

mofojed: What I'm saying is that there's no where in the reflow that this container _should_ be height 0 - it should only be getting larger, so why is it reporting height 0?

And when the panel is actually backgrounded, are we still unnecessarily subscribing to the table? (Which we shouldn't be doing, because the viewport size in that case _should_ be 0 and we shouldn't need to subscribe to get any updates).

SimonVutov: When you close the last widget in the bottom stack, `RowOrColumn.removeChild` collapses the row via `replaceChild`, which calls `parentNode.replaceChild(newChild.element[0], oldChild.element[0])` (a DOM re-parenting). A detached/re-attached element measures `getBoundingClientRect().height === 0`. So the container never actually shrinks to 0; its subtree is just briefly detached during the move, and a detached node measures `getBoundingClientRect().height === 0`. AutoSizer reads that transient 0 before Golden Layout's subsequent `callDownwards('setSize')` applies the new larger height. I also added a unit test confirming the updater goes inactive and explicitly unsubscribes (setViewport(undefined)) when the panel is backgrounded.

packages/console/src/command-history/CommandHistoryStorage.ts
```diff
@@ -21,6 +22,8 @@ export interface CommandHistoryStorageItem extends StorageItem {
 export interface CommandHistoryTable
   extends StorageTable<CommandHistoryStorageItem> {
   setSearch: (search: string) => void;
+  // Allow clearing the viewport (unsubscribe) when the panel is backgrounded.
+  setViewport: (viewport: StorageTableViewport | undefined) => void;
```
mofojed: This is a breaking change. If we do this and deployed these packages, we'd get an error when updating Enterprise because the `PQCommandHistoryTable` implementation doesn't handle an `undefined` viewport: https://github.com/deephaven-ent/iris/blob/c46ac3ecdd3646401c88a9c96ede58fcd088c39b/web/client-ui/src/storage/pq-storage/PQCommandHistoryTable.ts#L48
We shouldn't do a breaking change for this.

I guess previously when hidden, it wouldn't unsubscribe completely, it would just subscribe to one row. I do like the unsubscribe completely, but that doesn't seem to be built into the API - narrowing it down to one row is likely good enough.

packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx
```diff
@@ -109,6 +113,11 @@ class CommandHistoryPanel extends Component<
   handleShow(): void {
     // virtual list requires a forced reset to scroll position when picked up and dropped in same place
     this.container.current?.restoreScrollPosition();
+    this.setState({ isPanelActive: true });
+  }
+
+  handleHide(): void {
```
mofojed: This will work for a panel in this dashboard; you could still get hidden if you open another dashboard entirely (backgrounding this dashboard). So we're not really handling that case.

packages/components/src/ItemList.tsx
```diff
@@ -783,13 +786,18 @@ export class ItemList<T> extends PureComponent<
...
+      height: lastGoodHeight,
+    } = this.state;
+
     return (
       <AutoSizer className="item-list-auto-sizer" onResize={this.handleResize}>
         {({ width, height }) => (
           <List
             className="item-list-scroll-pane"
-            height={height}
+            height={height > 0 ? height : lastGoodHeight ?? height}
```
mofojed: So to be clear with the bug - we're getting a height of 0 and then _not_ getting a good height from AutoSizer? This seems like a bug in auto sizer then, since the container is indeed getting larger than 0. So why is an update getting missed?
Maybe check if there's an updated AutoSizer? Or perhaps we should just use a ResizeObserver instead (see `useResizeObserver` hook and usages). It's possible that auto-sizer is simply out of date, we used it before ResizeObserver existed.

mofojed (changes requested):

packages/components/src/ItemList.tsx
SimonVutov: I did some debugging and updated the PR description with what I found. The logs show the first scroll reset right after `parentNode.replaceChild()`. The current fix just preserves and restores the scroll positions across that replacement. It fixes the bug, but I wasn't able to figure out why the scroll gets reset in the first place.

packages/golden-layout/src/items/AbstractContentItem.ts
```diff
@@ -8,6 +8,51 @@ import type Stack from './Stack';
...
+      offsets.push({
+        element,
+        scrollTop: element.scrollTop,
+        scrollLeft: element.scrollLeft,
+      });
+    }
+  };
+
+  captureOffset(rootElement);
+  const descendants = rootElement.querySelectorAll<HTMLElement>('*');
```
mofojed: We only want to select the direct descendants of the root element. We shouldn't have to query the entire DOM tree.

mofojed (changes requested):

packages/golden-layout/src/items/AbstractContentItem.ts
mofojed: Actually the scroll views are not a direct child so this wouldn't work.
Selecting all nodes could be thousands of nodes, so ideally we don't want to do that.
The scroll seems to get reset when we replaceChild, e.g. removing the node and then adding it back to the DOM. see if there's any other precedent or examples for keeping scroll state in those situations.

mofojed (changes requested):
I still don't like this approach - now we're adding a scroll registry that contains any element that's scrolled, even outside of golden layout; then when doing a replaceChild we still need to iterate through all the elements in the registry and check if they're a child of this content item.

Another thing to consider - this happens with any scrollable element that is scrolled, removed from the DOM, then added to another element with a different height (though behaviour in different browsers does vary a bit). So there's a couple things to consider:
1. Do we want golden-layout to reset the scroll position in all cases when panels are moved around? Note this could cause issues, if a component internally tries to keep track of it's scroll position
2. Do we want CommandHistory/ItemList to keep track of this state, so if it's removed/added to the DOM anywhere it retains it's scroll position?
  - I'd argue _this_ is where the real issue is. Not only is the scroll getting reset, but ItemList isn't reacting to that reset position correctly
  - We have something in [CommandHistoryPanel to restore the scroll position on a re-show](https://github.com/deephaven/web-client-ui/blob/fc7dc25cc945906ab25192803d2de0c68b0f9f85/packages/dashboard-core-plugins/src/panels/CommandHistoryPanel.tsx#L111), which just calls a [`restoreScrollPosition` method in `ItemList`](https://github.com/deephaven/web-client-ui/blob/fc7dc25cc945906ab25192803d2de0c68b0f9f85/packages/components/src/ItemList.tsx#L337). So there is precedent for ItemList doing this already.
  - Note that I would probably take this one step further - if we're in the `isStuckToBottom` state, we should probably put the scroll to the bottom, rather than just the last state (e.g. if the ItemList moved into a smaller panel, it should stay stuck to the bottom, rather than just the same scrollTop which won't be at the bottom).

I think we want to go with 2, where the CommandHistory or the ItemList is reacting to the scroll reset appropriately.

We should also add an e2e test for this. I've pushed a branch with an e2e test that tests the behaviour (currently failing): https://github.com/mofojed/web-client-ui/commit/f4bac86a89aa8851d09f8d8a4ebe94ac849167ad

Build off of those and build a fix that fixes that test, and open that as a PR.

PR #2718, opened 2026-07-09
feat: DH-22030: Add event handler prereqs to chart

jnumainville:
Adds some chart and chart model improvements for https://github.com/deephaven/deephaven-plugins/pull/1373

### DIFF
--- a/packages/chart/src/Chart.tsx
+++ b/packages/chart/src/Chart.tsx
@@ -340,17 +340,27 @@ class Chart extends Component<ChartProps, ChartState> {
         ({ type }) => type != null && type.includes('3d')
       );
 
-      const buttons2D = [
-        'zoomIn2d',
-        'zoomOut2d',
-        'autoScale2d',
-        'resetScale2d',
-      ] as const;
-      const buttons3D = [
-        'orbitRotation',
-        'tableRotation',
-        'resetCameraDefault3d',
-      ] as const;
+      // zoom2d and pan2d work for both 2D and 3D charts
+      const interactionButtons: ModeBarButtonAny[] = [
+        'zoom2d',
+        'pan2d',
+        ...(hasSelectionCallbacks && has2D ? ['select2d', 'lasso2d'] : []),
+      ] as ModeBarButtonAny[];
+
+      const dimensionButtons: ModeBarButtonAny[] = [
+        ...(has2D
+          ? ['zoomIn2d', 'zoomOut2d', 'autoScale2d', 'resetScale2d']
+          : []),
+        ...(has3D
+          ? ['orbitRotation', 'tableRotation', 'resetCameraDefault3d']
+          : []),
+      ] as ModeBarButtonAny[];
+
+      // Display the mode bar persistently if there's an error or downsampling so user can see progress
+      // Yes, the value is a boolean or the string 'hover': https://github.com/plotly/plotly.js/blob/master/src/plot_api/plot_config.js#L249
+      const alwaysShowModeBar =
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+        isDownsampleInProgress || hasDownsampleError || hasError;
 
       return {
         displaylogo: false,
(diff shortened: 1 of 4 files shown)


packages/chart/src/ChartModel.ts
```diff
@@ -89,6 +96,102 @@ class ChartModel {
...
+  /**
+   * Called by the Chart component when Plotly fires a legend item click.
+   * Override to handle legend clicks. Return false to prevent Plotly's default
+   * visibility toggle.
+   * @param event The Plotly legend click event
+   * @returns False to prevent the default, true (or void) to allow it
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onLegendClick(event: LegendClickEvent): boolean {
+    return true;
```
mofojed: I thought returning `true` would block the default behaviour (e.g. consuming the event), but returning `false` seems to be in line with plotly.js API. They've got it backwards IMO but oh well.

packages/chart/src/ChartModel.ts
```diff
@@ -89,6 +96,102 @@ class ChartModel {
...
+  onLegendDoubleClick(event: LegendClickEvent): boolean {
+    return true;
+  }
+
+  /**
+   * Called by the Chart component when Plotly fires a WebGL context lost event.
+   * Override to handle WebGL context loss.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  onWebGlContextLost(): void {}
```
mofojed: This is interesting... we could display something to say "Show SVG" in this case?
I didn't think we got a proper/reliable event when the WebGL context was lost.

packages/chart/src/ChartModel.ts
```diff
@@ -89,6 +96,102 @@ class ChartModel {
...
+
+  /**
+   * Called by the Chart component once Plotly has initialized, providing the
+   * actual Plotly graph div element. Models can use this to attach imperative
+   * event listeners if needed
+   * @param plotElement The Plotly graph div element
+   */
+  // eslint-disable-next-line @typescript-eslint/no-empty-function, @typescript-eslint/no-unused-vars
+  setPlotElement(plotElement: HTMLElement | null): void {}
```
mofojed: This kind of sucks that we have to pass the plot element into the model directly; wish it could be a little more abstract...
Also, for something like Grid, the model doesn't handle clicks and such directly, those are handled by event handlers passed into the Grid component. I'm not sure I like the events being passed through the model here, couldn't we register them from the PlotlyExpressChart instead?

mofojed (commented):
It kind of sucks we need to pass the plot element into the model directly.

packages/chart/src/ChartModel.ts
jnumainville: Yeah we could pass in the handlers to chart instead. 
This plotElement is unfortunate but rather minimal in use so we can really hijack the plotly behavior.

packages/chart/src/ChartModel.ts
jnumainville: I experimented with it a bit and we definitely can use this event for better handling of the WebGL handling issues. Created https://deephaven.atlassian.net/browse/DH-23070 for this.

packages/dashboard-core-plugins/src/panels/ChartPanel.tsx
```diff
@@ -1147,6 +1178,16 @@ export class ChartPanel extends Component<ChartPanelProps, ChartPanelState> {
                 onSettingsChanged={this.handleSettingsChanged}
                 Plotly={Plotly}
                 containerRef={containerRef}
+                onPlotlyRelayout={onPlotlyRelayout}
+                onPlotlyClick={onPlotlyClick}
+                onPlotlyDoubleClick={onPlotlyDoubleClick}
+                onPlotlySelected={onPlotlySelected}
+                onPlotlyDeselect={onPlotlyDeselect}
```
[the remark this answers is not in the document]
jnumainville: done

packages/chart/src/Chart.tsx
```diff
@@ -571,6 +619,50 @@ class Chart extends Component<ChartProps, ChartState> {
     }
 
     this.updateModelDimensions();
+
+    const { onPlotlyRelayout } = this.props;
+    onPlotlyRelayout?.(changes);
+  }
+
+  handleClick(data: Readonly<PlotMouseEvent>): void {
+    console.log('handleClick', data, this.props.onPlotlyClick);
```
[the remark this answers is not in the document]
jnumainville: done

mofojed (approved):

PR #2720, opened 2026-07-16
fix: DH-22991: Command history is disappearing

SimonVutov:
### Problem
Removing a sibling panel next to Command History caused it to go blank and lose its scroll position. Golden Layout re-parents the Command History DOM node (via `replaceChild`), which resets the scroll container to the top while react-window keeps its internal offset — so no items render in the viewport.

### Fix
`ItemList` now re-asserts its scroll position when it's resized (the reliable in-component signal for the reparent):

- **`restoreScrollPosition()`** is now purely geometric — it clamps the last scroll offset to the current bounds (`clamp(scrollOffset, 0, itemCount * rowHeight - height)`). This restores the correct position and, for a bottom-anchored list moving into a smaller/larger panel, keeps it at the new bottom without any special-casing.
- **`handleResize()`** calls `restoreScrollPosition()` after applying the new height.

No Golden Layout changes.

### Test
Adds an e2e test (`tests/command-history.spec.ts`) that scrolls Command History to the middle, removes a sibling panel, and asserts the scroll position and rendered items are preserved.

### DIFF
--- a/packages/components/src/ItemList.tsx
+++ b/packages/components/src/ItemList.tsx
@@ -335,12 +335,27 @@ export class ItemList<T> extends PureComponent<
   }
 
   restoreScrollPosition(): void {
-    const { scrollOffset } = this.state;
-    if (scrollOffset != null) {
-      // manually restore the scroll containers offset
-      // virtual list doesn't restore scrolloffset in a re-render if it's the same
-      this.listContainer.current?.scrollTo(0, scrollOffset);
+    const container = this.listContainer.current;
+    if (container == null) {
+      return;
     }
+
+    const { scrollOffset } = this.state;
+    if (scrollOffset == null) {
+      return;
+    }
+
+    // Re-parenting the panel (e.g. removing a sibling panel) resets the scroll
+    // container to the top while react-window keeps its internal offset, which
+    // leaves the list blank. Re-assert the scroll position purely from the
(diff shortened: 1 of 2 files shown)


packages/components/src/ItemList.tsx
```diff
@@ -335,12 +335,27 @@ export class ItemList<T> extends PureComponent<
...
+    // container to the top while react-window keeps its internal offset, which
+    // leaves the list blank. Re-assert the scroll position purely from the
+    // current geometry: clamp the last real offset to the current bounds. On a
+    // grow this naturally keeps a bottom-anchored list at the (new) bottom, so
+    // restore never needs to reason about sticky state.
+    const { itemCount, rowHeight } = this.props;
+    const { height } = this.state;
+    const maxOffset = Math.max(0, itemCount * rowHeight - (height ?? 0));
+    const newOffset = Math.min(Math.max(scrollOffset, 0), maxOffset);
+    container.scrollTo(0, newOffset);
```
mofojed, line 360: We should be checking the `isStuckToBottom` and if that's true, scroll to maxOffset. Otherwise we can just scroll to the newOffset like we do here.

mofojed (changes requested):

mofojed (approved):
Great work

PR #2722, opened 2026-07-20
fix: DH-23126: Use Subscription for Update Event with Viewport Options

dgodinez-dh:
https://github.com/deephaven/deephaven-core/commit/de3420a8e6dcc3986597429059be719958d20fa7#diff-162aa86fd5cbdd89292447104bd0ce09e4d45907c433bd9d10e41924e66bf4a3R848
Due to the above change, table subscriptions are severed from the table meaning that the TABLE_UPDATE event no longer fires on the table if you have a subscription.

This fix does the following:
- return the subscription ref from useSetPaddedViewportCallback
- in useViewportData, subscribe to events if the subscription is not null

### DIFF
--- a/packages/jsapi-components/src/useSetPaddedViewportCallback.ts
+++ b/packages/jsapi-components/src/useSetPaddedViewportCallback.ts
@@ -1,10 +1,22 @@
-import { useCallback, useEffect, useRef } from 'react';
+import { type MutableRefObject, useCallback, useEffect, useRef } from 'react';
 import type { dh } from '@deephaven/jsapi-types';
 import {
   getSize,
   padFirstAndLastRow,
   TableUtils,
 } from '@deephaven/jsapi-utils';
+
+/**
+ * The return type of `useSetPaddedViewportCallback`. It is callable as a
+ * function and also exposes `subscriptionRef` so that callers can register
+ * event listeners directly on the underlying `TableViewportSubscription` when
+ * one is in use (required since DH-20800 severed the table→subscription event
+ * propagation).
+ */
+export type UseSetPaddedViewportCallbackResult = {
+  (firstRow: number): void;
+  subscriptionRef: MutableRefObject<dh.TableViewportSubscription | null>;
+};
 
 /**
  * Creates a callback function that will set a Table viewport. The callback has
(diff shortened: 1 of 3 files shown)


packages/jsapi-components/src/useViewportData.ts
mofojed, line 143: I don't think we need to bother with this listener anymore.

EDIT: Looks like it may be needed in the case where we're currently not creating a subscription (e.g. no `viewportSubscriptionOptions` passed in).

packages/jsapi-components/src/useSetPaddedViewportCallback.ts
mofojed, line 93: So we still have cases where the subscription may _not_ be created if the `viewportSubscriptionOptions` are `null`, in which case we should still be listening on the table. This seems kind of messy...

packages/jsapi-components/src/useSetPaddedViewportCallback.ts
```diff
@@ -17,14 +29,15 @@ import {
...
+ * @returns A callable result with a `subscriptionRef` property exposing the
+ * active `TableViewportSubscription` (if any) so callers can attach listeners.
  */
 export function useSetPaddedViewportCallback(
   table: dh.Table | dh.TreeTable | null,
   viewportSize: number,
   viewportPadding: number,
   viewportSubscriptionOptions: Partial<dh.ViewportSubscriptionOptions> | null = null
-): (firstRow: number) => void {
+): UseSetPaddedViewportCallbackResult {
```
mofojed, line 40: This is kind of gross... what if we pass in an `onUpdate` callback as a prop instead? Then the `useSetPaddedViewportCallback` can manage the subscription and call the callback as necessary.
Passing back the subscription ref seems like it'll just cause hard to solve problems down the road.

mofojed (changes requested):

packages/jsapi-components/src/useViewportData.ts
dgodinez-dh: Yeah, I don't think this is a good fix. It's really just some quick code to prove that this was really the issue. I don't like that we still have the table listener (especially in the case when we don't need it). But think the really relevant issue is that JsTable.setViewport is marked deprecated.

@deprecated Use {@link #createViewportSubscription(Object)} instead.

So maybe we need to rewrite the hook to always use a subscription (although it looks like we make an exception for Tree Tables, so that bears looking into).

I'll start on a new fix in the morning. But do talk to Ryan about this. If we have pressure to get a release out, maybe we should wait so we to rush this. I think useViewportData is a pretty instrumental hook in our code.

mofojed: We want to cherry pick this fix into this release certification.
`useViewportData` is used in many places, and I believe is the only place that actually uses the `useSetPaddedViewportCallback`. If we pass in an `onUpdate` listener into `useSetPaddedViewportCallback`, then `useSetPaddedViewportCallback` can set up the listener appropriately (whether on the table or the subscription itself), and `useViewportData` doesn't need to be concerned with those details anymore.
It is annoying that JsTreeTable and JsTable have different APIs for setting viewport/subscribing, keeping that contained within `useSetPaddedViewportCallback` makes sense to me.

PR #2725, opened 2026-07-23
feat: DH-22220: Allow Conditional Formatting to Compare Columns

dgodinez-dh:
Update to Conditional Formatting:
- right hand value can now be a value or a column
- user determines value or column by selecting a value or column condition
- optionally can select multiple columns to apply the format
- commented in and updated e2e tests


packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.test.ts
```diff
@@ -221,6 +229,190 @@ describe('getFormatColumns', () => {
...
+        `${lhv} != ${rhvName}`,
+      ],
+      [
+        StringCondition.CONTAINS,
+        `${lhv} != null && ${lhv}.contains("${val}")`,
+        `${lhv} != null && ${lhv}.contains(${rhvName})`,
+      ],
+      [
+        StringCondition.DOES_NOT_CONTAIN,
+        `${lhv} != null && !${lhv}.contains("${val}")`,
```
bmingles, line 281: Just to confirm, if lhv == null, this returns true. Is that the expected result for does not contain?

packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
```diff
@@ -202,75 +218,144 @@ export function getStyleDBString(config: BaseFormatConfig): string | undefined {
...
+  const rhv = formatRHV(rightHandValue, '"');
+  switch (config.condition as StringCondition) {
+    case StringCondition.IS_EXACTLY:
+      return `${columnName} == ${rhv}`;
+    case StringCondition.IS_NOT_EXACTLY:
+      return `${columnName} != ${rhv}`;
+    case StringCondition.CONTAINS:
+      return `${columnName} != null && ${columnName}.contains(${rhv})`;
+    case StringCondition.DOES_NOT_CONTAIN:
+      return `${columnName} != null && !${columnName}.contains(${rhv})`;
```
bmingles, line 276: same question here. Is columnName == null expected to be false here?

packages/iris-grid/src/sidebar/conditional-formatting/MultiColumnFormatEditor.tsx
```diff
@@ -0,0 +1,261 @@
...
+          <StyleEditor config={selectedStyle} onChange={handleStyleChange} />
+        </>
+      )}
+
+      <div className="mb-2">
+        <label className="mb-0 mt-1">Apply to Columns</label>
+        <MultiSelect
+          aria-label="Select columns to apply formatting to"
+          width="100%"
+          selectedKeys={selectedFormattedColumns.map(c => c.name)}
```
bmingles: We might want to memoize this

bmingles (changes requested):
Left a question and a suggestion

packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.test.ts
dgodinez-dh: If lhv is null, I think this expression should return `false`. Which I think is the correct result for does not contain.

packages/iris-grid/src/sidebar/conditional-formatting/ConditionalFormattingUtils.ts
dgodinez-dh: Same answer as above.

packages/iris-grid/src/sidebar/conditional-formatting/MultiColumnFormatEditor.tsx
dgodinez-dh: Updated

bmingles (approved):
Changes look good

PR #2726, opened 2026-07-23
fix: DH-23165: Fix spectrum Picker closing Popper dialogs on open

vbabich:
Add optional `containPortals` prop to the Popper dialog to allow nested spectrum Pickers and not dismiss the dialog on Picker open.

Verified manually on the http://localhost:4220/ide/styleguide#popover-dialog page and the PivotBuilder plugin aggregations config section (will post a separate PR in the deephaven-plugins repo).

Added unit and e2e tests.


tests/styleguide.spec.ts
```diff
@@ -216,3 +216,26 @@ test('Inputs regression test', async ({ page }) => {
...
+  // Open the Picker's own overlay
+  const picker = page.getByRole('button', { name: 'Fruit' });
+  await expect(picker).toBeVisible();
+  await picker.click();
+
+  // The Picker listbox renders in a portal, so screenshot the whole page
+  const listbox = page.getByRole('listbox');
+  await expect(listbox).toBeVisible();
+
+  await expect(page).toHaveScreenshot('dialog-with-picker-open.png');
```
mofojed, line 243: Should select a new option in the picker to make sure you can actually change options.

mofojed (changes requested):

mofojed (dismissed):

mofojed (approved):

PR #2727, opened 2026-07-24
feat: DH-21538: Add reset button to broken table

dgodinez-dh:
- Adds a reset button to Loading Overlay
- It is only display if there is an error and a handler
- For IrisGrid, calls rollback


bmingles (approved):
changes look good

PR #2730, opened 2026-08-03
fix: DH-23079: Fix Console Errors after log out and log in

dgodinez-dh:
After log out and log in, redux has a stale sessionWrapper. `closeDocument` and `changeDocuemnt` would fire against the wrong session causing a `KeyError`.  The fix reinitializes the editor on session change so all notifications stay on the same session.


vbabich (approved):
LGTM

PR #2731, opened 2026-08-05
fix: DH-23176: Better Error Handling for Conditional Formatting

dgodinez-dh:
- adds conditional formats and preview to IrisGrid rollback
- displays and error message in the Conditional Format UI for failed formats


bmingles (approved):
Changes look good.

PR #2733, opened 2026-08-10
fix: DH-20535: Fix missing key data warning in IrisGridTreeTableModel

vbabich:
## Summary

Fixes two issues with the rollup **Group** column in `IrisGridTreeTableModel` when a tree/rollup has 2+ grouping columns.

- **Removed console warning spam** (`Missing key data for virtual column`): `extractViewportRow` no longer tries to populate the Group column for the root/total row (depth 1), which has no grouping key. It now only copies from real grouped source columns.
- **Fixed a crash on hover** (`Cannot read properties of undefined (reading 'name')` in `textAlignForCell`): the Group column's source-column lookup now clamps to the deepest grouped column, so leaf rows (whose depth extends past the last grouped column) no longer index past the end of the columns array. Extracted the shared logic into `virtualSourceColumnIndex`.

## Testing

- Added unit tests covering the Group column mapping at each depth, root-row emptiness, and leaf-row clamping.
- Verified live in the browser: no warnings, no crash when hovering the Group column across all depths.


bmingles (dismissed):
LGTM

bmingles (approved):
re-approved after e2e snapshot reverts

PR #2734, opened 2026-08-12
ci: Create Jira release ticket when a release is published

mofojed:
Posts the release version, notes, and referenced DH- issue keys to a Jira Automation incoming webhook, which creates the "Update web version on G+" task and links the related issues.

Requires the Jira webhook setup, JIRA_AUTOMATION_WEBHOOK_URL and JIRA_AUTOMATION_WEBHOOK_SECRET repo secrets (already set).

Manually tested locally, see an example ticket created: https://deephaven.atlassian.net/browse/DH-23365

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -58,3 +58,36 @@ jobs:
           -H "Authorization: Bearer ${{ secrets.DEEPHAVENINTERNAL_PUBLIC_TOKEN }}" \
           -H "X-GitHub-Api-Version: 2022-11-28" \
           https://api.github.com/repos/deephaven/deephaven-core/actions/workflows/update-web.yml/dispatches --data '{"ref": "main"}'
+      - name: Create Jira release ticket
+        # Posts to a Jira Automation "Incoming webhook" rule that creates the
+        # "Update web version on G+" task and links the DH- issues in the notes
+        if: ${{ github.event_name == 'release' }}
+        continue-on-error: true
+        env:
+          RELEASE_VERSION: ${{ github.event.release.tag_name }}
+          RELEASE_URL: ${{ github.event.release.html_url }}
+          RELEASE_NOTES: ${{ github.event.release.body }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -58,3 +58,36 @@ jobs:
           -H "Authorization: Bearer ${{ secrets.DEEPHAVENINTERNAL_PUBLIC_TOKEN }}" \
           -H "X-GitHub-Api-Version: 2022-11-28" \
           https://api.github.com/repos/deephaven/deephaven-core/actions/workflows/update-web.yml/dispatches --data '{"ref": "main"}'
+      - name: Create Jira release ticket
+        # Posts to a Jira Automation "Incoming webhook" rule that creates the
+        # "Update web version on G+" task and links the DH- issues in the notes
+        if: ${{ github.event_name == 'release' }}
+        continue-on-error: true
```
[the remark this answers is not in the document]
mofojed: I don't want the action to fail in this case, just log that there's an error.

.github/workflows/publish-packages.yml
```diff
@@ -58,3 +58,36 @@ jobs:
...
+        env:
+          RELEASE_VERSION: ${{ github.event.release.tag_name }}
+          RELEASE_URL: ${{ github.event.release.html_url }}
+          RELEASE_NOTES: ${{ github.event.release.body }}
+          # Names the thing being updated in the ticket summary
+          JIRA_COMPONENT: 'web'
+          # Bump when the target enterprise release changes
```
dgodinez-dh, line 72: It would be nice if there was some way to make this dynamic based on the web version.

dgodinez-dh (approved):
Looks fine, but I

.github/workflows/publish-packages.yml
mofojed: Well, it will be - if we have a long running branch that we'll maintain (like `release/v0.85`) then we can update the enterprise release in that branch.
For now it's just G+.

PR #2735, opened 2026-08-14
feat: DH-23429: Create Deephaven TableView wrapper

SimonVutov:
This should be merged before merging this change on iris: https://github.com/deephaven-ent/iris/pull/5120


packages/iris-grid/src/IrisGrid.tsx
mofojed, line 1: Why does this PR have IrisGrid changes? I don't think IrisGrid is being used at all from the Dashboards screen.

mofojed (changes requested):
We should also have a TableViews example page in the styleguide, similar to the ListViews.tsx page.

packages/iris-grid/src/IrisGrid.tsx
SimonVutov: Thanks for catching that. It looks like I had some changes stashed and accidentally included them in this PR. I’ve removed those changes and added a TableView page to the styleguide.

mofojed:
@SimonVutov I created a subtask on the original Jira ticket specifically for this and updated the PR description.

packages/code-studio/src/styleguide/StyleGuide.scss
```diff
@@ -128,6 +128,14 @@ pre {
   padding: 1rem;
 }
 
+.style-guide-table-view {
```
mofojed: We shouldn't need to add any new styles here.

packages/code-studio/src/styleguide/TableViews.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+        <Flex direction="column" gap={8}>
+          <Text>Windowed data at offset 3 of 12 rows</Text>
+          <div className="style-guide-table-view style-guide-table-view-windowed">
+            <TableView
+              aria-label="Windowed data table"
+              columns={columns}
+              items={rows.slice(3, 9)}
+              itemCount={12}
+              offset={3}
```
mofojed: Use LabeledFlexContainers instead, check how ListViews lays out it's examples. Right now it looks like a mess.

packages/components/src/spectrum/tableView/TableView.tsx
```diff
@@ -0,0 +1,91 @@
...
+  sortDescriptor?: SortDescriptor;
+  onAction?: (item: T) => void;
+  onSortChange?: (descriptor: SortDescriptor) => void;
+  onViewportChange?: (top: number, bottom: number) => void;
+  renderCell: (item: T, columnKey: Key) => ReactNode;
+  getTextValue?: (item: T) => string;
+  renderEmptyState?: () => JSX.Element;
+  'aria-label'?: string;
+  UNSAFE_className?: string;
+}
```
mofojed: We shouldn't be redefining these Props definitions, we should just extend the `SpectrumTableViewProps`, and use `Omit` to drop the props we don't want (like `ListView` does, or just use `Pick` to pick the props we want.

packages/components/src/spectrum/tableView/TableViewNormalized.tsx
```diff
@@ -0,0 +1,111 @@
...
+            key={key}
+            textValue={item == null ? '' : getTextValue?.(item) ?? ''}
+          >
+            {columnKey => (
+              <Cell>
+                {item == null ? null : (
+                  <div className="dh-table-view-cell" data-table-view-key={key}>
+                    {renderCell(item, columnKey)}
+                  </div>
+                )}
```
mofojed: I  don't think we need this wrapper div:

```suggestion
                  {item == null ? null : renderCell(item, columnKey)}
```

mofojed (changes requested):
I've not convinced we need these styles, but also things look really messed up in the styleguide right now: 

Image

Look at how ListViews does it and link it up style it similarly so each example is contained correctly instead of overlapping.

mofojed:
@SimonVutov you also need to add the missing e2e screenshots

packages/code-studio/src/styleguide/TableViews.tsx
```diff
@@ -1,18 +1,45 @@
...
+      direction="column"
+      gap={10}
+      minWidth={0}
+    >
+      <Text>{label}</Text>
+      {children}
+    </Flex>
+  );
+}
+
```
mofojed: We should put this into its own file that exports the component and is used by both ListViews and TableViews, rather than duplicating in both files.

tests/styleguide.spec.ts-snapshots/table-views-chromium-linux.png
mofojed, line 1: We should have an example in the styleguide that triggers the onAction as well (which could just display an alert or something).

mofojed (changes requested):

packages/code-studio/src/styleguide/TableViews.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <LabeledFlexContainer label="Windowed rows" height="100%" minWidth={0}>
+          <TableView
+            aria-label="Windowed rows"
+            columns={columns}
+            items={rows.slice(3)}
+            itemCount={12}
+            offset={3}
+            renderCell={renderCell}
+          />
+        </LabeledFlexContainer>
```
mofojed: This example is still very weird. We should just generate a list of a bunch of items (like ListView does with the icons list).
Also we should show `onAction` usage (like we do in ListViews).

mofojed (changes requested):

packages/components/src/spectrum/tableView/TableView.tsx
```diff
@@ -18,16 +18,19 @@ export interface TableViewColumn {
...
+  extends Omit<
+    SpectrumTableProps<T>,
+    | 'children'
+    | 'items'
+    | 'onAction'
+    | 'selectionMode'
+    | 'selectedKeys'
+    | 'defaultSelectedKeys'
+    | 'disabledKeys'
+    | 'onSelectionChange'
```
mofojed: Hmm. I think we want an example with selection and onAction as well in the styleguide.
Ultimately we want this TableView to be hooked up as the new dashboard list, which includes selecting multiple items and then choosing an action (like import/export) or just hitting an action item in the table row.
Also I don't think we should override `onAction` to return `T` instead of just `Key` like `SpectrumTableView`. Just use the same prop (like we do in ListView).

mofojed (changes requested):

packages/components/src/spectrum/tableView/TableView.tsx
dgodinez-dh: updated

packages/components/src/spectrum/tableView/TableViewNormalized.tsx
```diff
@@ -0,0 +1,87 @@
...
+          >
+            {column.name}
+          </Column>
+        )}
+      </TableHeader>
+      <TableBody items={normalizedItems}>
+        {({ key, item }) => (
+          <Row
+            key={key}
+            textValue={item == null ? '' : getTextValue?.(item) ?? ''}
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/components/src/spectrum/tableView/TableViewWrapper.tsx
```diff
@@ -0,0 +1,130 @@
...
+    (element: HTMLElement) => {
+      if (
+        onViewportChangeRef.current == null ||
+        rowHeight == null ||
+        rowHeight <= 0 ||
+        itemCount <= 0
+      ) {
+        return;
+      }
+      const top = Math.max(0, Math.floor(element.scrollTop / rowHeight));
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/components/src/spectrum/tableView/TableView.tsx
```diff
@@ -0,0 +1,80 @@
...
+  hideHeader?: boolean;
+}
+
+export interface TableViewProps<T>
+  extends Omit<SpectrumTableProps<T>, 'children' | 'items' | 'overflowMode'> {
+  columns: TableViewColumn[];
+  items: readonly T[];
+  itemCount: number;
+  offset?: number;
+  onViewportChange?: (top: number, bottom: number) => void;
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

mofojed (approved):

PR #2736, opened 2026-08-17
feat: DH-9378: Web Keyed Selection

dgodinez-dh:
Implements keyed row selection for iris-grid.
- selection is now an object with interface Selection
- selection is either RangedSelection or KeyedSelection based on table attributes
- render logic updated to support keys selecting multiple rows
- Copy, Filter By Value, and Download CSV updated to snapshot with keys
See ticket for test code in the test plan.


packages/grid/src/Grid.tsx
```diff
@@ -495,8 +521,9 @@ class Grid extends PureComponent<GridProps, GridState> {
       // Currently selected ranges and previously selected ranges
       // Store the previously selected ranges to determine if the new selection should
       // deselect again (if it's the same range)
-      selectedRanges: EMPTY_ARRAY,
-      lastSelectedRanges: EMPTY_ARRAY,
+      selection: props.createEmptySelection(this.getModel),
+      lastSelection: props.createEmptySelection(this.getModel),
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,337 @@
...
+    if (overlayRanges.length === 0) {
+      this.gestureKeys = new Set();
+    } else {
+      const keys = new Set<string>();
+      for (let i = 0; i < overlayRanges.length; i += 1) {
+        const { startRow, endRow } = overlayRanges[i];
+        if (startRow == null) continue; // eslint-disable-line no-continue
+        const last = endRow ?? startRow;
+        for (let r = startRow; r <= last; r += 1) {
+          keys.add(this.getRowKeyData(r).key);
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed by bounding the range to the viewport.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1542,6 +1567,169 @@ class IrisGridTableModelTemplate<
...
+    if (keyValues.size === 0) return null;
+
+    // Create an AND filter for each set of key values
+    const keyFilters: DhType.FilterCondition[] = [];
+    keyValues.forEach(values => {
+      const colFilters = values.map((val, i) => {
+        const col = keyColumns[i];
+        const filterVal = this.tableUtils.makeFilterRawValue(col.type, val);
+        return col.filter().eq(filterVal);
+      });
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -142,32 +150,64 @@ class TableCsvExporter extends Component<
...
+                  return a.startRow - b.startRow;
+                }
+                return 0;
+              })
+          );
+        } else if (selection instanceof KeyedSelection) {
+          // keyed: the frozenTable is already filtered; snapshot all its rows
+          snapshotRanges.push(
+            new GridRange(0, 0, columnCount - 1, keyedTableSize - 1)
+          );
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,337 @@
...
+      for (let r = startRow; r <= last; r += 1) {
+        rows.push(r);
+      }
+    }
+    if (rows.length === 0) return this;
+    const rowKeyData = rows.map(r => this.getRowKeyData(r));
+    const next = new Set(this.selectedKeys);
+    const nextKeyValues = new Map(this.selectedKeyValues);
+    // Toggle: if every row is already selected, remove them; otherwise add all.
+    if (rowKeyData.every(({ key }) => next.has(key))) {
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -619,94 +624,68 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
+      model,
+      movedColumns,
+      userColumnWidths
+    );
+
+    // For full-row snapshots, computeModelRanges removes hidden columns so the
+    // model-index sourceColumn must be offset by the count of hidden columns before it.
+    const hiddenColumns = IrisGridUtils.getHiddenColumns(userColumnWidths);
+    const snapshotColumnIndex =
+      sourceColumn - hiddenColumns.filter(h => h < sourceColumn).length;
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/plugin/src/TablePlugin.ts
```diff
@@ -43,9 +43,15 @@ export interface TablePluginProps<S = unknown> {
...
   /**
    * The currently selected ranges in the table.
+   * @deprecated Use `selection` instead.
    */
   selectedRanges: readonly GridRange[] | undefined;
 
+  /**
+   * The current grid selection, including keyed selections for tables with key columns.
+   */
+  selection: Selection | null | undefined;
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/grid/src/key-handlers/TreeKeyHandler.ts
```diff
@@ -18,9 +19,11 @@ class TreeKeyHandler extends KeyHandler {
   }
 
   handleExpandKey(event: KeyboardEvent, grid: Grid): boolean {
-    const { selectedRanges } = grid.state;
-    if (selectedRanges.length === 1) {
-      const range = selectedRanges[0] as GridRange;
+    // toRanges() is only available on RangedSelection; keyed tables have no expandable rows
+    assertIsRangedSelection(grid.state.selection);
+    const ranges = grid.state.selection.toRanges();
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1542,6 +1567,168 @@ class IrisGridTableModelTemplate<
...
+      );
+    }
+    const keyColumns = this.selectionKeyColumnIndices.map(i => this.columns[i]);
+    const keyFilter = this.buildKeyFilter(keyValues, keyColumns);
+    const copy = await (this.table as DhType.Table).copy();
+    // Skip applyFilter when no filter is needed — an empty filter array on a fresh copy
+    // produces no filterchanged event, causing applyFilter to time out.
+    if (keyFilter != null) {
+      const filter = invertedSelection ? [keyFilter.not()] : [keyFilter];
+      await this.tableUtils.applyFilter(copy, filter);
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/Grid.tsx
```diff
@@ -1164,9 +1246,14 @@ class Grid extends PureComponent<GridProps, GridState> {
...
         const newRanges = [...selectedRanges];
-        newRanges[newRanges.length - 1] = selectedRange;
+        if (newRanges.length > 0) {
+          newRanges[newRanges.length - 1] = selectedRange;
+        } else {
+          newRanges.push(selectedRange);
+        }
         return {
-          selectedRanges: newRanges,
+          selection: selection.withMouseGestureRanges(newRanges),
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/GridRendererTypes.ts
```diff
@@ -90,8 +90,8 @@ export type GridRenderState = {
   cursorColumn: VisibleIndex | null;
   cursorRow: VisibleIndex | null;
 
-  // Currently selected ranges
-  selectedRanges: readonly GridRange[];
+  // Current selection
+  selection: Selection;
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,337 @@
...
+  /** Returns both the serialized key and the raw values for a visible row. */
+  private getRowKeyData(row: VisibleIndex): {
+    key: string;
+    values: readonly unknown[];
+  } {
+    const model = this.getModel();
+    const values = model.selectionKeyColumnIndices.map((col: ModelIndex) =>
+      model.valueForCell(col, row)
+    );
+    return { key: JSON.stringify(values), values };
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3642,26 +3683,48 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
     if (copyOperation != null) {
       this.setState({ copyOperation: null });
     }
+    const singleRow = selection.getLastSingleSelectedRow();
+    if (singleRow != null) {
+      this.setState({ gotoRow: `${singleRow + 1}` });
+    }
+    onSelectionChange?.(selection);
+    if (selection instanceof KeyedSelection && selection.pendingRows != null) {
+      this.resolveKeyedSelection(selection);
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/Grid.tsx
```diff
@@ -1164,9 +1246,14 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+          selectedRanges: selectionToRanges(selection),
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -325,7 +331,9 @@ export interface IrisGridProps {
    */
   userColumnWidthsByName?: ReadonlyMap<ColumnName, number>;
   userRowHeights: ReadonlyMap<ModelIndex, number>;
-  onSelectionChanged: (gridRanges: readonly GridRange[]) => void;
+  /** @deprecated Use onSelectionChange instead. */
+  onSelectionChanged?: (gridRanges: readonly GridRange[]) => void;
+  onSelectionChange?: (selection: Selection) => void;
```
mofojed, line 343: Hm. This is slightly confusing... add some JSDocs for `onSelectionChange` at least.
Note that other props we have the d (like `onColumnSelected`, `onDataSelected`) and others we don't (`onStateChange`, `onAdvancedSettingsChange`). Seems a little inconsistent, and I don't have a strong belief one way or another, just something I'd like to note.

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,353 @@
...
+    );
+  }
+
+  clear(): KeyedSelection {
+    return new KeyedSelection(this.getModel, new Set());
+  }
+
+  // Shift+click needs a clean slate so the range replaces rather than extends the old keys.
+  trimmed(): KeyedSelection {
+    return new KeyedSelection(this.getModel, new Set());
```
mofojed: Hmmm so this is interesting... this shift+click behaviour is currently incorrect, I think because it extends from where the last click was the _row_ index before this shift-click, not where the row currently is. So if you have rows tick in above, the shift-click will give you the wrong range.

https://github.com/user-attachments/assets/e2805e4b-c964-49d4-a6b1-fbb5dfdb1085

I think we want to return where the selection currently is if it's still known? If it's outside the viewport, then a new selection?

packages/dashboard-core-plugins/src/panels/IrisGridPanel.tsx
```diff
@@ -485,6 +486,10 @@ export class IrisGridPanel extends PureComponent<
         panel: this,
       };
 
+      const gridSel = this.irisGrid.current?.state.gridSelection;
+      const selectedRanges =
+        gridSel != null && isRangedSelection(gridSel) ? gridSel.toRanges() : [];
```
mofojed: Use `EMPTY_ARRAY` so it's a stable empty object instead of recreating it every render.
```suggestion
        gridSel != null && isRangedSelection(gridSel) ? gridSel.toRanges() : EMPTY_ARRAY;
```

packages/grid/src/Grid.tsx
```diff
@@ -116,16 +122,22 @@ export type GridProps = typeof Grid.defaultProps & {
   keyHandlers?: readonly KeyHandler[];
   mouseHandlers?: readonly GridMouseHandler[];
 
+  // Factory that creates an empty Selection; defaults to an empty RangedSelection.
+  createEmptySelection?: (getModel: () => GridModel) => Selection;
```
mofojed: Hmm. Why is this a function that can get the model passed in instead of just the model? That would be something worth commenting about. Or just pass through `model` instead?

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,353 @@
...
+
+export class KeyedSelection implements Selection {
+  static empty(getModel: GetKeyedModel): KeyedSelection {
+    return new KeyedSelection(getModel, new Set());
+  }
+
+  private readonly gestureKeys: ReadonlySet<string>;
+
+  constructor(
+    private readonly getModel: GetKeyedModel,
```
mofojed: Put some JSDocs on these, one thing I wonder is why we pass in a `getModel` instead of just a `model`.

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,353 @@
...
+    readonly selectedKeys: ReadonlySet<string>,
+    private readonly overlayRanges: readonly GridRange[] = EMPTY_ARRAY,
```
mofojed: This is a lot of parameters - wondering if it should be like a config object that's passed in instead of ordered params, as you have to fill in all the intermediate parameters if you want to specify a later parameter even if you just want the defaults for the intermediate parameter.

mofojed (changes requested):
Not done yet, looking good so far but a few comments on findings.

packages/grid/src/Grid.tsx
```diff
@@ -230,6 +247,11 @@ export type GridState = {
   updateRevision: number;
 };
 
+/** Returns `selection.toRanges()` for a RangedSelection, otherwise `[]`. */
+function selectionToRanges(selection: Selection): readonly GridRange[] {
+  return isRangedSelection(selection) ? selection.toRanges() : EMPTY_ARRAY;
+}
```
mofojed: Could probably put this in a utils function to share it in a couple places.

packages/grid/src/Grid.tsx
```diff
@@ -279,6 +302,8 @@ class Grid extends PureComponent<GridProps, GridState> {
       autoSelectColumn: false,
       autoSelectRow: false,
     } as Partial<GridThemeType>,
+    createEmptySelection: (getModel: () => GridModel): Selection =>
+      RangedSelection.empty(getModel),
```
mofojed:
```suggestion
    createEmptySelection: RangedSelection.empty,
```

packages/grid/src/Grid.tsx
mofojed: Should we create this once and then re-use it for the `lastSelection`?

packages/grid/src/Grid.tsx
```diff
@@ -631,6 +659,13 @@ class Grid extends PureComponent<GridProps, GridState> {
       this.metricCalculator = metricCalculator ?? new GridMetricCalculator();
     }
 
+    if (prevProps.createEmptySelection !== createEmptySelection) {
+      const empty = createEmptySelection(this.getModel);
+      stateUpdates.selection = empty;
+      stateUpdates.lastSelection = empty;
```
mofojed, line 678: See here we re-use the same object

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,52 @@
...
+  withUpdatedRanges: (ranges: readonly GridRange[]) => Selection;
+  /** Applies mouse gesture ranges to this selection. */
+  withMouseGestureRanges: (ranges: readonly GridRange[]) => Selection;
+  /**
+   * Commits the current mouse gesture and returns the settled selection.
+   * Returns this (identity) when there is nothing to commit.
+   */
+  commitMouseGesture: (
+    lastCommitted: Selection,
+    autoSelectRow: boolean
```
mofojed: I think instead of an `autoSelectRow` here, we should have `options`; leaves possibility of adding new selection options (e.g. `autoSelectColumn` ? Who knows).

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,52 @@
...
+  /** Returns column [start, end] pairs for scrollbar tick rendering. */
+  getColumnTickRanges: () => readonly BoundedAxisRange[];
+  /** Returns row [start, end] pairs for scrollbar tick rendering. */
+  getRowTickRanges: () => readonly BoundedAxisRange[];
+  /** Returns a new Selection with selection cleared. */
+  clear: () => Selection;
+  /** Returns a new Selection keeping only the last range. */
+  trimmed: () => Selection;
+  /** Returns a new Selection generated from the supplied ranges */
+  withUpdatedRanges: (ranges: readonly GridRange[]) => Selection;
```
mofojed: Why not just name this `withRanges`? What makes it "updated"?

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,52 @@
...
+  /** Applies mouse gesture ranges to this selection. */
+  withMouseGestureRanges: (ranges: readonly GridRange[]) => Selection;
```
mofojed: We should probably clarify what `withRanges` and `withMouseGestureRanges` entails - why are they different methods? Should it instead be one method with an option indicating it's from a mouse gesture?

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { BoundedAxisRange } from './GridAxisRange';
+
+/** Provides current model data to Selection instances without holding a stale reference. */
+export type GetModel = () => GridModel;
+
+/**
+ * Immutable value object representing the current selection state of the grid.
+ * Mutations return new instances; Grid stores the result in React state.
+ */
+export interface Selection {
```
mofojed: Might be nice to clean up the interface a bit, particularly the operations that return a new `Selection` should be in a separate interface. E.g.

```
export interface SelectionOperations {
  trimmed()
  clear()
  selectAll()
  withRanges()
  truncate()
}

export interface SelectionProperties {
  isEmpty
  isCellSelected
  isValid
  toRanges
  getColumnTickRanges
  getRowTickRanges
  getLastSingleSelectedRow
}

export interface Selection extends SelectionOperations, SelectionProperties {}
```

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,52 @@
...
+  /** Returns true if no cells are selected. */
+  isEmpty: () => boolean;
+  /** Returns true if the specified cell is part of the selection. */
+  isCellSelected: (row: VisibleIndex, column: VisibleIndex) => boolean;
+  /** Returns true if the entire row is part of the selection. */
+  isRowSelected: (row: VisibleIndex) => boolean;
+  /** Returns false if any selected range exceeds the given column/row bounds. */
+  isValid: (columnCount: number, rowCount: number) => boolean;
+  /** Returns the ranges Grid uses for cursor positioning, extend-selection, and keyboard navigation. */
+  toActiveRanges: () => readonly GridRange[];
```
mofojed: Should this only be on `RangedSelection`?

mofojed (changes requested):

packages/iris-grid/src/IrisGrid.tsx
dgodinez-dh: I went with onSelectionChange because we already had onSelectionChanged. And I wanted to keep the old one deprecated for backwards compatibility until San Luis.

packages/iris-grid/src/KeyedSelection.ts
dgodinez-dh: Per our discussion, I have implemented shift click that works in viewport and degrades when the key row ticks out of viewport. Will write tickets for api to shift click out of viewport.

packages/grid/src/Selection.ts
dgodinez-dh: Updated

packages/grid/src/Selection.ts
dgodinez-dh: renamed to withCommittedRanges for clarity.

packages/grid/src/Selection.ts
dgodinez-dh: Updated jsdocs

packages/grid/src/Selection.ts
dgodinez-dh: Updated

packages/grid/src/Selection.ts
dgodinez-dh: This is needed in KeyedSelection for cursor selection in Grid. I think it is clarified now with recent jsdocs updates.

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,628 @@
...
+    }
+    if (rows.length === 0) {
+      return new KeyedSelection({ getModel: this.getModel });
+    }
+    const next = new Set<string>();
+    const nextKeyValues = new Map<string, readonly unknown[]>();
+    rows.forEach(r => {
+      const { key: k, values } = this.getRowKeyData(r);
+      next.add(k);
+      nextKeyValues.set(k, values);
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/grid/src/Grid.tsx
```diff
@@ -1020,12 +1060,13 @@ class Grid extends PureComponent<GridProps, GridState> {
...
-    if (selectedRanges !== oldSelectedRanges) {
-      const { onSelectionChanged } = this.props;
-      onSelectionChanged(selectedRanges);
+    const { selection } = this.state;
+    if (selection !== prevState.selection) {
+      const { onSelectionChanged, onSelectionChange } = this.props;
+      if (isRangedSelection(selection)) {
+        onSelectionChanged(selection.toRanges());
+      }
+      onSelectionChange(selection);
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/sidebar/TableCsvExporter.tsx
```diff
@@ -221,46 +264,78 @@ class TableCsvExporter extends Component<
...
-        this.setState({
-          errorMessage: (
-            <p>
-              <FontAwesomeIcon icon={vsWarning} /> {`${error}`}
-            </p>
-          ),
-        });
+        // freeze to static snapshot; TableSaver closes frozenTable in finishDownload/cancelDownload
+        frozenTable = await filteredTable.freeze();
+        filteredTable.close();
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -1542,6 +1571,198 @@ class IrisGridTableModelTemplate<
...
+      await this.tableUtils.applyFilter(copy, [neverMatch]);
+      return copy;
+    }
+    // Skip applyFilter when no filter is needed — an empty filter array on a fresh copy
+    // produces no filterchanged event, causing applyFilter to time out.
+    if (keyFilter != null) {
+      const filter = invertedSelection ? [keyFilter.not()] : [keyFilter];
+      await this.tableUtils.applyFilter(copy, filter);
+    }
+    return copy;
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/grid/src/Grid.tsx
```diff
@@ -1056,27 +1093,51 @@ class Grid extends PureComponent<GridProps, GridState> {
...
+      selectedRanges: selectionToRanges(trimmed),
+    });
+  }
+
+  /** Sets the selection directly, bypassing mouse/keyboard gesture state. */
+  setSelection(selection: Selection): void {
+    this.setState({
+      selection,
+      selectedRanges: selectionToRanges(selection),
+    });
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/KeyedGridModel.ts
```diff
@@ -0,0 +1,86 @@
...
+    columns: readonly DhType.Column[],
+    keyValues: ReadonlyMap<string, readonly unknown[]>,
+    invertedSelection: boolean,
+    includeHeaders?: boolean,
+    formatValue?: (value: unknown, column: DhType.Column) => unknown,
+    maxRows?: number | null
+  ) => Promise<unknown[][]>;
+
+  /**
+   * Text version of snapshotByKeys: rows tab-separated, columns newline-separated.
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/GridRenderer.ts
```diff
@@ -2091,51 +2079,97 @@ export class GridRenderer {
...
+        flushRowRun();
+        // Partial selection — check each cell and coalesce consecutive selected columns.
+        let runStart: VisibleIndex | null = null;
+        for (let c = left; c <= right; c += 1) {
+          if (selection.isCellSelected(r, c)) {
+            if (runStart === null) runStart = c;
+          } else if (runStart !== null) {
+            const x = Math.max(
+              Math.round(getOrThrow(allColumnXs, runStart)) + 0.5,
+              minX
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,610 @@
...
+      });
+    }
+
+    // Slow path: any row outside the viewport would resolve to a phantom
+    // all-null key via undefined valueForCell reads. Defer to async key fetch;
+    // IrisGrid's onSelectionChange handler picks up pendingRows.
+    return new KeyedSelection({
+      getModel: this.getModel,
+      pendingRows: new GridRange(null, first, null, last),
+    });
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/Grid.tsx
```diff
@@ -1428,6 +1484,8 @@ class Grid extends PureComponent<GridProps, GridState> {
     }
 
     this.moveSelection(column, row, extendSelection, maximizePreviousRange);
+    // Commit after every keyboard move so KeyedSelection resolves keys immediately.
+    this.commitSelection();
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/GridRenderer.ts
```diff
@@ -2091,51 +2080,97 @@ export class GridRenderer {
...
+      if (selection.isRowSelected(r)) {
+        // Extend or start the full-row run.
+        if (rowRunStartY == null) rowRunStartY = y;
+        rowRunEndY = endY;
+      } else {
+        flushRowRun();
+        // Partial selection — check each cell and coalesce consecutive selected columns.
+        let runStart: VisibleIndex | null = null;
+        for (let c = left; c <= right; c += 1) {
+          if (selection.isCellSelected(r, c)) {
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/Grid.tsx
```diff
@@ -302,8 +297,9 @@ class Grid extends PureComponent<GridProps, GridState> {
       autoSelectColumn: false,
       autoSelectRow: false,
     } as Partial<GridThemeType>,
-    createEmptySelection: (getModel: () => GridModel): Selection =>
-      RangedSelection.empty(getModel),
+    createEmptySelection: RangedSelection.empty as (
+      getModel: GetModel
+    ) => Selection,
```
mofojed, line 316: I don't think you need to cast this

packages/grid/src/GridRenderer.ts
```diff
@@ -2111,65 +2112,113 @@ export class GridRenderer {
...
+            partialRunStartY,
+            endX - x,
+            partialRunEndY - partialRunStartY
+          );
+        }
+      }
+      partialRunStartY = null;
+      partialRuns = [];
+    };
+
```
mofojed: Mmmm are we certain this logic needs to change? I would have thought the selection was already consolidated by this point.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -3718,16 +3762,51 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
...
+    if (minStart === null || maxEnd === null) return;
+
     try {
-      const keyValues = await model.fetchKeyValuesForRowRange(
-        pendingRows.startRow,
-        pendingRows.endRow
+      const rowsByIndex = await model.fetchKeyValuesForRowRange(
+        minStart,
+        maxEnd
       );
```
mofojed: This logic should probably be handled by the model, and the model interface should handle multiple ranges, e.g.
```
model.fetchKeyValuesForRowRanges(pendingRanges)
```
Then the model can decide how it wants to delegate it (one call to the JSAPI for a snapshotted viewport, or perhaps there's a new JSAPI added at some point that we can do it with multiple ranges or something, either way should be part of the model implementation).

mofojed (changes requested):
Just a question about GridRenderer changes, and where we should have the logic for getting the keys from the row ranges

packages/grid/src/Grid.tsx
dgodinez-dh: Removing the cast produces this error:
```
packages/iris-grid/src/IrisGrid.tsx:5680:13 - error TS2322: Type '((getModel: () => GridModel<Record<string, Event<string>>, "standard">) => Selection) | undefined' is not assignable to type '(((getModel: GetModel) => RangedSelection) & ((getModel: GetModel) => Selection)) | undefined'.
  Type '(getModel: () => GridModel) => Selection' is not assignable to type '((getModel: GetModel) => RangedSelection) & ((getModel: GetModel) => Selection)'.
    Type '(getModel: () => GridModel) => Selection' is not assignable to type '(getModel: GetModel) => RangedSelection'.
      Type 'Selection' is missing the following properties from type 'RangedSelection': ranges, getModel, gestureStartRow, gestureStartColumn, toRanges

5680             createEmptySelection={this.getCachedCreateEmptySelection(model)}
                 ~~~~~~~~~~~~~~~~~~~~

  packages/grid/dist/Grid.d.ts:135:9
    135         createEmptySelection: typeof RangedSelection.empty;
                ~~~~~~~~~~~~~~~~~~~~
    The expected type comes from property 'createEmptySelection' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Grid> & Pick<Readonly<GridProps>, "children" | "model" | "renderer" | "metricCalculator"> & InexactPartial<...> & InexactPartial<...>'
```
According to the AI the `defaultProps` narrows the `.d.ts` prop type.

I can remove the cast by using an arrow function:

```
    createEmptySelection: (getModel: GetModel): Selection =>
      RangedSelection.empty(getModel),
```
But I think that's what we had before. Do you have a preference or a different recommendation?

packages/grid/src/GridRenderer.ts
dgodinez-dh: Now that we need to render using the Selection interface, I had code to coalesce adjacent row selections to render as a single rect with no stroke in between (the isRowSelected case). But I forgot about input table single cell selection.

This code was added to coalesce partial rows with identical columns to render as a single rect. (the isCellSelected case)

image

But now that you bring it up, I realized that since we already have coalesce for the isRowSelected case, we can combine that with the isCellSelected case to simplify. So I have pushed that.

packages/iris-grid/src/IrisGrid.tsx
dgodinez-dh: Updated

packages/grid/src/Grid.tsx
mofojed: Well that's dumb that `defaultProps` narrows the type. That's probably an issue with how we set up the typing in the Grid component (e.g. the `type GridProps = typeof Grid.defaultProps & { ... }` pattern we have)... If we restructured how we created the types, then we wouldn't need to cast/duplicate the typing, e.g.:
```
// Props that have defaults — required in the type, optional at call sites
type GridDefaultProps = {
  ...
  createEmptySelection: (getModel: GetModel) => Selection;
  ...
};

export type GridProps = GridDefaultProps & {
  model: GridModel; // truly required props
  children?: ReactNode;
  ...
};

class Grid extends PureComponent<GridProps, GridState> {
  static defaultProps: GridDefaultProps = {
    // ...
    createEmptySelection: RangedSelection.empty, // no cast
  };
}
```

Not sure that's worth doing in this PR though.

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,131 @@
...
+ * - `withCommittedRanges` writes into the **committed** selection state
+ *   (programmatic entry points like `Grid.setSelectedRanges`).
+ * - `withMouseGestureRanges` writes into the **transient overlay** state
+ *   used for mid-gesture rendering (drag / shift-click on every mouse move).
+ *
+ * For `RangedSelection` these look the same because there's no separate
+ * overlay concept. For `KeyedSelection` they're distinct: overlay ranges
+ * drive gesture preview only; committed key sets change on `commitMouseGesture`.
+ */
+export interface Selection extends SelectionQueries, SelectionTransforms {}
```
mofojed: Wanting to be thorough with this `Selection` interface, since this is pretty important to Grid and we're going to be sticking with this interface for a while. With that being said, going through the `Selection`, want to just make sure our interface makes sense for other potential models to implement. Got a few comments for the interface here...

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/** Read-only inspection of a `Selection`. Every method is side-effect-free. */
+export interface SelectionQueries {
+  /** True when the selection contains no cells and no in-progress gesture. */
+  isEmpty: () => boolean;
+  /** True when `(row, column)` is part of the selection. */
+  isCellSelected: (row: VisibleIndex, column: VisibleIndex) => boolean;
+  /** True when the entire row is part of the selection. */
+  isRowSelected: (row: VisibleIndex) => boolean;
```
mofojed, line 36: `isEmpty`, `isCellSelected`, `isRowSelected` make sense. Though I feel like we should also have an `isColumnSelected` then. (We need to think about this from Pivot Table context as well (we'll definitely need to test with pivot tables in Enterprise)).

That being said, `isCellSelected` takes `(row, column)`. It should be `(column, row)`, as that's more like `x, y`, and how GridRange does it.

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,131 @@
...
+  /** False if any selected range exceeds `columnCount` or `rowCount`. */
+  isValid: (columnCount: number, rowCount: number) => boolean;
```
mofojed: `isValid` feels a bit strange, taking the `columnCount` and `rowCount`. At least make it clear that's the column/row count of the grid. Wondering if we should have it as an options object there; any thoughts?

packages/grid/src/Selection.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * The current `{row, column}` of the gesture anchor, or `null` if none is
+   * set or the anchor is no longer resolvable (e.g. a keyed anchor whose
+   * row has scrolled out of the viewport with no row hint fallback).
+   */
+  getGestureAnchor: () => {
+    row: GridRangeIndex;
+    column: GridRangeIndex;
+  } | null;
+}
```
mofojed: This part is where I get a little wary - does it make sense for _all_ of these to be in the `Selection` interface? Something seems a little off here to me... for example, we have the `toActiveRanges()` method, and then in Grid, in our `moveSelection` method we're immediately converting `selection.toActiveRanges()`, and getting the `selection.getGestureAnchor()`... should the `Selection` interface not handle how the move occurs? Why do we need to expose `toActiveRanges()/getLastSingleSelectedRow()getGestureAnchor()` if we have methods like `withCommittedRanges` etc., like shouldn't the selection just have a `moveSelection`... e.g.

```
type Selection = SelectionQueries & SelectionTransforms;

type SelectionQueries {
  isEmpty(): boolean;
  isCellSelected(...): boolean
  isRowSelected(...): boolean
  isColumnSelected(...): boolean

  // Return the selected ranges limited to the provided bounds.
  // Can be used by the Selection implementation to build the ranges and allow the GridRenderer to focus on drawing those ranges.
  selectedRanges(bounds?: GridRange): GridRange[];
}

type KeyboardSelection {
  withKeyboardSelect(...): Selection;
}

type MouseSelection {
  withMouseSelect(...): Selection;
}

// TickRange should be a separate optional interface, then we check `isTickRangeSelection` to know whether to render them, rather than having KeyedSelection implement the interface and return empty
type TickRangeSelection {
  readonly rowTickRanges: BoundedAxisRange[];
  readonly columnTickRanges: BoundedAxisRange[];
}

```

mofojed (changes requested):
Still want to iterate on this Selection interface a bit - the division of work is a little strange and it has some odd methods.

packages/grid/src/Grid.tsx
dgodinez-dh: Since defaultProps is deprecated, I think we should put this off until we change that. I think trying to fix it now will add a lot of noise to an already large change.

packages/grid/src/Selection.ts
dgodinez-dh: I added `isColumnSelected`. Trying to integrate it into the rendering logic did not improve anything, so it is not currently called in the code. It is unit tested, and I did hand test it with some throw away code.

packages/grid/src/Selection.ts
dgodinez-dh: You are right about this. KeyedSelection ignores the arguments and always returns true. It doesn't belong in the interface. I made the following changes:
- Removed isValid from the interface and KeyedSelection
- In RangedSelection, renamed it to `isInBounds` to better describe behavior
- Removed arguments as the only caller passed in bounds from the model, and the selection object already has access to the model

packages/grid/src/Selection.ts
dgodinez-dh: This was a huge refactor.

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -164,123 +158,85 @@ class SelectionKeyHandler extends KeyHandler {
...
-        grid.moveCursorToPosition(
-          moveToColumn,
-          moveToRow,
-          isShiftKey,
-          true,
-          maximizePreviousRange
+        grid.handleKeySelectAt(
+          { column: moveToColumn, row: moveToRow },
+          isShiftKey ? 'maximize' : 'replace'
         );
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -1206,6 +1219,15 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
     { max: 100 }
   );
 
+  getCachedCreateEmptySelection = memoize(
+    (model: IrisGridModel) =>
+      isKeyedGridModel(model)
+        ? (getModel: () => GridModel): Selection =>
+            KeyedSelection.empty(getModel as GetKeyedModel)
+        : undefined,
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/iris-grid/src/KeyedSelection.ts
```diff
@@ -0,0 +1,647 @@
...
+      // on a cell — toggle its selection. A multi-row overlay is a ctrl+shift+click
+      // grown range — add all rows without toggling off overlapping ones.
+      const shouldToggle = rowCount === 1;
+      const nextKeyValues = new Map(this.selectedKeyValues);
+      for (let i = 0; i < this.overlayRanges.length; i += 1) {
+        const { startRow, endRow } = this.overlayRanges[i];
+        if (startRow == null) continue; // eslint-disable-line no-continue
+        const rEnd = endRow ?? startRow;
+        for (let r = startRow; r <= rEnd; r += 1) {
+          const { key: k, values } = this.getRowKeyData(r);
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/key-handlers/SelectionKeyHandler.ts
```diff
@@ -164,123 +158,94 @@ class SelectionKeyHandler extends KeyHandler {
...
+    const { autoSelectRow = false, autoSelectColumn = false } = theme;
+    if (autoSelectRow && deltaColumn !== 0) {
+      const { lastLeft } = grid.metrics;
+      const left = clamp(grid.state.left + deltaColumn, 0, lastLeft);
+      grid.handleKeySelectAt({ column: left, row: cursorRow }, mode, {
+        keepCursorInView: false,
+      });
+      grid.setViewState({ left });
+      return true;
     }
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/GridRenderer.ts
```diff
@@ -2673,11 +2674,13 @@ export class GridRenderer {
...
-            GridUtils.compareRanges
-          )
-        );
+        const mergedRanges = isTickRangeSelection(state.selection)
+          ? GridUtils.mergeSortedRanges(
+              [...state.selection.getColumnTickRanges()].sort(
+                GridUtils.compareRanges
+              )
+            )
+          : [];
```
mofojed, line 2768: If we're going to get the ranges from the selection, why not do this compare/merge right in there? Then we don't need to do it every render cycle.

packages/grid/src/Selection.ts
```diff
@@ -1,79 +1,77 @@
...
-  /** Column `[start, end]` pairs for scrollbar tick rendering. */
-  getColumnTickRanges: () => readonly BoundedAxisRange[];
-  /** Row `[start, end]` pairs for scrollbar tick rendering. */
-  getRowTickRanges: () => readonly BoundedAxisRange[];
+  isSingleRowSelection: () => boolean;
   /**
    * The single selected visible row, or `null` when zero or multiple rows
    * are selected. Drives `gotoRow` sync.
    */
   getLastSingleSelectedRow: () => VisibleIndex | null;
```
mofojed: These two still seem a little odd for the abstraction. `getLastSingleSelectedRow` is only used by IrisGrid (not used by Grid proper itself):
```typescript
    const singleRow = selection.getLastSingleSelectedRow();
    if (singleRow != null) {
      this.setState({ gotoRow: `${singleRow + 1}` });
    }
```
Elsewhere in IrisGrid, when setting gotoRow, we do it based on the `cursorRow` of the grid itself:
```typescript
    const cursorRow = this.grid?.state.cursorRow;
    const cursorColumn = this.grid?.state.cursorColumn;

   ...

    this.setState({
      gotoRow: `${cursorRow + 1}`,
     ...
    });
  }
```
Would it make sense to instead of the `Selection` have a `cursorCell` property that returns the cursor column/row - I think that could be used here instead of `getLastSingleSelectedRow()` (and also then could be used instead of accessing `cursorRow`/`cursorColumn` directly from Grid, keeping the source of truth in the Selection object itself). (Should those even be removed from the Grid state and be part of  Selection? Though maybe a bigger change than we want now).

Then `isSingleRowSelection` is used by the SelectionKeyHandler for the same thing twice:
```
    // Avoid deselection when the target cell is already selected in a single-row selection.
    if (
      !isShiftKey &&
      grid.state.selection.isSingleRowSelection() &&
      grid.state.selection.isCellSelected(targetColumn, targetRow)
    ) {
      grid.handleKeyMoveCursor(targetColumn, targetRow);
      return true;
    }
```
I actually don't think this logic is correct, or something is wrong here anyways. If I'm holding Ctrl, I should be able to click, then click and drag another selection, and keep adding to it, e.g. like our old behaviour:

https://github.com/user-attachments/assets/a8924fc6-49a0-4253-9384-965d3b8e1fbf

However, it seems if I click and drag now, it's reverting the selection:

https://github.com/user-attachments/assets/ad78d486-89b9-428a-8d25-b2997e63c596

So I think first need to clean up that behaviour, we may not even need this last single row selection at all. We should also add a test for this case.

packages/grid/src/Grid.tsx
mofojed: So thinking about my other comment about cursorRow/cursorColumn - should these be deprecated as well, and instead codified into the Selection? Seems like it would make more sense to have the Selection be the source of truth about the selection.
Perhaps could argue that cursor is not necessarily selection; but I think it is?

packages/grid/src/GridSelectionUtils.ts
mofojed, line 1: Should have some unit tests around this file/behaviour

mofojed (changes requested):
Liking the look of this a lot better.

packages/grid/src/GridRenderer.ts
dgodinez-dh: Updated

packages/grid/src/mouse-handlers/GridSelectionMouseHandler.ts
```diff
@@ -250,9 +214,8 @@ class GridSelectionMouseHandler extends GridMouseHandler {
       return false;
     }
 
-    grid.clearSelectedRanges();
-    grid.moveCursorToPosition(column, row);
-    grid.commitSelection();
+    // Double-click behaves as a plain click at the target: replace selection.
+    grid.handleMouseSelectStart({ row, column }, 'replace');
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/grid/src/mouse-handlers/GridSelectionMouseHandler.ts
```diff
@@ -262,20 +225,17 @@ class GridSelectionMouseHandler extends GridMouseHandler {
...
 
-    // only change the selected range if the selected cell is not in the selected range
-    if (!isInRange && gridPoint.row !== null && gridPoint.column !== null) {
+    // only change the selected range if the clicked cell is not already selected
+    if (!isInRange && row != null && column != null) {
       this.startPoint = undefined;
       this.stopTimer();
-      grid.clearSelectedRanges();
-      grid.moveCursorToPosition(gridPoint.column, gridPoint.row);
+      grid.handleMouseSelectStart({ row, column }, 'replace');
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -242,6 +244,30 @@ class IrisGridTableModelTemplate<
...
+        return idx;
+      });
+    }
+  );
+
+  get selectionKeyColumnIndices(): readonly ModelIndex[] {
+    return this.getMemoizedSelectionKeyColumnIndices(
+      this.columns,
+      (this.table as DhType.Table).getAttribute?.('keyColumns')
+    );
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/Selection.ts
dgodinez-dh: Fixed

packages/grid/src/Grid.tsx
dgodinez-dh: Added cursor to selection

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
```diff
@@ -938,18 +938,24 @@ class IrisGridContextMenuHandler extends GridMouseHandler {
...
-        const keyValues = new Map<string, readonly unknown[]>([[key, values]]);
-        effectiveSelection = new KeyedSelection({
-          getModel: getModel as GetKeyedModel,
-          selectedKeys: new Set([key]),
-          lastSingleRow: rowIndex,
-          selectedKeyValues: keyValues,
-        });
+        if (model.isKeyableRow(rowIndex)) {
+          // Construct a committed single-row KeyedSelection from the model's key columns.
+          const getModel = () => model as IrisGridModel & KeyedGridModel;
```
mofojed: We should have a `isKeyedGridModel` type predicate instead of casting. Could put as a function exported from KeyedGridModel.ts (see `isExpandableGridModel` as another example)

packages/iris-grid/src/IrisGridTableModelTemplate.ts
```diff
@@ -268,6 +268,12 @@ class IrisGridTableModelTemplate<
     return (this.table as DhType.Table).getAttribute?.('uniqueKeys') === 'true';
   }
 
+  // Totals rows have aggregate column values (not real row keys), so exclude
+  // them from keyed selection to keep copy / download filters meaningful.
+  isKeyableRow(row: VisibleIndex): boolean {
```
mofojed, line 274: Did we note somewhere/create a ticket for Copying totals rows not working?
Side note, in `IrisGridTableModelTemplate.fetchKeyValuesForRowRanges`, we call `serializeKeyValues` to get the key, which means each key for a regular row is basically a JSON string array (e.g. `["foo","bar"]`). We could totally prefix those keys, or even a different kind of JSON object, e.g. `{"type":"totals","key":"sum"}`.
Don't need to get into it now, as we should be able to change the key format without it being a breaking change, just want to make sure we record the issue as a ticket (and that it's currently broken).

mofojed (changes requested):
Looks great overall. Couple minor cleanup changes.

packages/iris-grid/src/mousehandlers/IrisGridContextMenuHandler.tsx
dgodinez-dh: There is an isKeyedGridModel, which is calleed above, and this is an unnecessary cast. Removed the cast.

packages/iris-grid/src/IrisGridTableModelTemplate.ts
dgodinez-dh: The ticket to follup on this is https://deephaven.atlassian.net/browse/DH-23589
Added a comment referencing the ticket.

packages/grid/src/GridRenderer.ts
```diff
@@ -828,18 +827,9 @@ export class GridRenderer {
...
-        startRow != null &&
-        endRow != null &&
-        startRow <= row &&
-        endRow >= row
-      ) {
-        if (theme.selectedRowHoverBackgroundColor != null) {
-          context.fillStyle = theme.selectedRowHoverBackgroundColor;
-        }
-        break;
+    if (selection.isRowSelected(row)) {
```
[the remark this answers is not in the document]
dgodinez-dh: fixed

packages/grid/src/RangedSelection.ts
```diff
@@ -0,0 +1,471 @@
...
+      const range = this.ranges[i];
+      const rowSelected =
+        range.startRow === null ||
+        (range.startRow <= row && row <= (range.endRow ?? 0));
+      const columnSelected =
+        range.startColumn === null ||
+        (range.startColumn <= column && column <= (range.endColumn ?? 0));
+      if (rowSelected && columnSelected) {
+        return true;
+      }
```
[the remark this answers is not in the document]
dgodinez-dh: Fixed

mofojed (dismissed):

mofojed (approved):

PR #2737, opened 2026-08-18
fix: DH-22976: control IrisGrid sorts and filters

SimonVutov:
Adds controlled sorts and quick filters to IrisGrid. User changes are reported through callbacks without updating internal state, while uncontrolled behavior remains unchanged.

This PR should be merged and released before [deephaven-plugins#1377](https://github.com/deephaven/deephaven-plugins/pull/1377).

### DIFF
--- a/packages/iris-grid/src/IrisGrid.tsx
+++ b/packages/iris-grid/src/IrisGrid.tsx
@@ -978,6 +978,8 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       customFilters,
       quickFilters,
       sorts,
+      isQuickFiltersControlled,
+      isSortsControlled,
       getMetricCalculator,
     } = this.props;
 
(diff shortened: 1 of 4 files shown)


dgodinez-dh (changes requested):
The AI pointed this out, but it looks like applyInputFilters, rebuildFilters, and setFilters call setState on quickFilters. Also handleUpdateCustomColumns calls setState on quickFilters and sorts. I think this need to be requests.

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2081,22 +2113,16 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
         filter: this.makeQuickFilter(column, text, formatter.timeZone),
       });
     });
-    this.startLoading('Rebuilding filters...', { resetRanges: true });
```
dgodinez-dh, line 2084: You removed startLoading from rebuildFilters. This relies on requestQuickFilterChange to trigger a start loading. But it is possible that there are only advanced filters, so they rebuild without a start loading.

dgodinez-dh (changes requested):

packages/iris-grid/src/IrisGrid.tsx
```diff
@@ -2082,21 +2114,16 @@ class IrisGrid extends Component<IrisGridProps, IrisGridState> {
       });
     });
     this.startLoading('Rebuilding filters...', { resetRanges: true });
-
-    this.setState({
-      quickFilters: newQuickFilters,
-      advancedFilters: newAdvancedFilters,
-    });
+    this.requestQuickFiltersChange(newQuickFilters);
```
dgodinez-dh: Don't call this.requestQuickFiltersChange if the quickFilters size is 0. For that case, the filters do not change because there are no quick filters, and we don't want an unnecessary onChange event.
Also, I think the AI is wrong on this one. If the quick filters change due to a timezone change, I would expect that onChange to fire.

dgodinez-dh (changes requested):
Left one more comment.
Additionally, there are a lot of AI comments on this review. You should either address them or close them as not relevant.

dgodinez-dh (approved):

PR #2742, opened 2026-08-24
feat: grid accessibility content on demand, playwright-grid testing package

mofojed:
- The grid draws to a canvas, so there were no per-cell elements for screen readers or browser automation to inspect, and the fallback content was a single line telling the user to update their browser.
- Grid now renders a summary of its size and selection there, plus a button that describes the viewport as a table of the visible column headers and cells. Walking the viewport is too expensive to do every frame, so the description is generated on request and discarded as soon as the grid scrolls. Each cell carries its position on screen, which lets external tooling locate it without knowing the grid's layout maths.
  - Did it this way rather than exposing an object on the Grid element - for a testing library like Playwright, which is run in a separate process, it was not able to interact with that object directly; instead it had to do `evaluate` calls in the context of the UI for that to work. By doing it this way, it is also actually useful as an accessibility guide, as screen readers can be instructed to show all the details
- Also adds @deephaven/playwright-grid, which wraps that description in helpers so e2e tests can address cells by column name rather than by pixel offset, and covers both with e2e tests that verify grid contents instead of comparing screenshots.
- Added `table-content.spec.ts` for testing with the playwright-grid package and copying many of the tests from `table.spec.ts`; also added `table-a11y.spec.ts` to ensure the accessibility layer makes sense


dsmmcken (changes requested):
We should ensure the default state displays no ticking data updates, featuring only a button that activates the desired behaviour without the description triggering dom changes. The only automatic dom change I think we should do is a single one for the initial render complete, so that playwright can easily watch for that for easy screenshots.

packages/grid/src/GridA11yFallback.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const { model, metrics, selectedRanges } = renderState;
+    return createGridA11ySnapshot(model, metrics, selectedRanges);
+  }, [renderState, showSnapshot]);
+
+  return (
+    <div>
+      {renderState != null && (
+        <p>
+          {getGridA11ySummary(renderState.model, renderState.selectedRanges)}
+        </p>
```
dsmmcken: extract wrapper div uneccessary. This live grid summary is the only thing I am against.

dsmmcken (changes requested):

packages/playwright-grid/src/gridHelpers.ts
```diff
@@ -0,0 +1,432 @@
...
+ */
+export async function waitForGrid(grid: Locator): Promise<void> {
+  const canvas = await resolveGridCanvas(grid);
+  await expect
+    .poll(async () => {
+      const snapshot = await showSnapshot(canvas);
+      return snapshot.locator(`[${ATTRIBUTES.row}]`).count();
+    })
+    .toBeGreaterThan(0);
+}
```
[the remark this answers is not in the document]
mofojed: We can have an explicit function for an empty grid. This is a reliable indicator for any non-empty grid.

dsmmcken (dismissed):

packages/grid/src/GridA11yFallback.tsx
```diff
@@ -0,0 +1,121 @@
...
+  useEffect(() => {
+    if (!showSnapshot) {
+      setRenderState(null);
+      return undefined;
+    }
+    return registerDrawListener(setRenderState);
+  }, [registerDrawListener, showSnapshot]);
+
+  const snapshot = useMemo(() => {
+    if (renderState == null) {
```
jnumainville, line 50: `renderState` is initialized as `{}`. Is it ever possible to hit that? Not sure, this was flagged by my local AI review as a possible edge case.
A bit more detail:
If the describe button is toggled on before the grid has ever drawn (e.g. a zero-size container where drawCanvas never completes past assertNotNull(metrics)), renderState is {} (not null), so createGridA11ySnapshot(undefined, undefined, undefined) runs and throws on const { rowCount, columnCount } = model

jnumainville (changes requested):
Just one thing, maybe.

packages/grid/src/GridA11yFallback.tsx
mofojed: Fixed to make sure the renderState of {} never comes through.

jnumainville (approved):

PR #2751, opened 2026-09-09
fix: DH-22388: Fix web settings daylight savings not reflected in timezone labels

cryolins:
- TimeUtils previously hardcoded timezone UTC+/-X labels, causing discrepancies on what the label showed and true UTC offset during daylight savings time.
- Now extracts UTC offset from DateTimeFormat directly to construct the label so it would align with actual UTC offset.


cryolins:

cryolins:
I put a helper object and function outside of the TimeUtils class in the TimeUtils.ts file. Not sure if this is a preferred organization, so it can be reorganized as needed.

packages/utils/src/TimeUtils.ts
```diff
@@ -1,6 +1,44 @@
...
+  { name: 'Los Angeles', value: 'America/Los_Angeles', noDst: false },
+  { name: 'Anchorage', value: 'America/Anchorage', noDst: false },
+  { name: 'Honolulu', value: 'Pacific/Honolulu', noDst: true },
+  { name: 'Zurich', value: 'Europe/Zurich', noDst: false },
+  { name: 'Amsterdam', value: 'Europe/Amsterdam', noDst: false },
+  { name: 'Taipei', value: 'Asia/Taipei', noDst: true },
+  { name: 'Sydney', value: 'Australia/Sydney', noDst: false },
+] as const;
+
+function getTimeZoneOffset(timeZone: string, date = new Date()): string {
```
mofojed: I think this makes sense to add as a static function on TimeUtils. When adding new public functions, we should also document appropriately with JSDoc, and add unit tests.

mofojed (changes requested):
Looks good. Let's make this function public and add unit tests.
Also take a look at the comments Copilot made on the review.

packages/utils/src/TimeUtils.ts
```diff
@@ -16,29 +40,40 @@ class TimeUtils {
...
+  static get TIME_ZONES(): readonly { label: string; value: string }[] {
+    return Object.freeze(
+      TIME_ZONE_DEFINITIONS.map(({ name, value, noDst }) => ({
+        label: `${name} ${TimeUtils.getTimeZoneOffset(value)}${
+          noDst ? ' No DST' : ''
+        }`,
+        value,
+      }))
+    );
+  }
```
mofojed, line 63: Hmm. Previously this was a static array that was frozen, and it was only created once on initialization. Now, we're returning a _new_ array every time a consumer calls `TimeUtils.TIME_ZONES`.

While it isn't _too_ expensive to generate this array, it's unexpected to be doing any work when calling a property like this, someone could be using `TIME_ZONES` repeatedly, e.g.
```
for (let i = 0; i < TimeUtils.TIME_ZONES.length; i += 1) {
  let timeZone = TimeUtils.TIME_ZONES[i];
  ...
}
```
From a consumer perspective, that seems fine to loop over what is seemingly a static array; however, this means that every iteration of the loop is going to call this function to regenerate the time zones _twice_ (once for the condition of the for loop, then once again inside the loop to get the value). That could get expensive. So by doing it this way, we're creating a foot gun that someone could get themselves stuck with.

Since the array is going to be the same every time anyways, we should just build the value once and cache it (memoization). Though it _could_ be different if you keep the browser open long enough... and we _could_ have a check for the current date, and have it re-generate if the date is different than the last time we generated. Then it's always up to date and correct, and we just have to make sure it's still the same day every time it's called (which is pretty cheap).

So - cache the array, only regenerate if the date has changed since the last call.

mofojed (changes requested):

packages/utils/src/TimeUtils.ts
cryolins: The previous intention was to allow for dynamic updates if the user leaves it on long enough without reloading, would that be necessary or is preserving the build-once static nature more preferable? Caching would add a bit more complexity and it would mainly just help in that scenario.

mofojed: We might be off for a day if the user doesn't refresh their browser over a DST transition; I think that's a fine compromise for not having to refresh every time it's called. If we wanted to be more accurate, we could key to the minute instead of the date. Not really that big a deal.

mofojed (changes requested):
As Copilot suggested, we may as well key it on the minute instead of the day. It's technically more correct.

mofojed (approved):
Looks great!

PR #2756, opened 2026-09-15
fix: Cancel pending layout save when the layout is replaced

dsmmcken:
This PR attempts to reduce failing and flakey golden layout tests.

Bug: Replacing a layout left the throttled save describing the outgoing one still pending, so it could land afterwards and put the old layout back. Cancels that pending save in `loadNewConfig` before hydrating the new config

Covers the cancel with unit tests plus a reload assertion in the e2e import test; both screenshot assertions now share one named baseline

- Moves the golden-layout e2e suite off a shared `beforeAll` page onto a per-test fixture with a seeded workspace, making the tests isolated and parallel-safe, parallel tests had a chance of stepping on each other depending on timing
- Changes how tests layout loads to skip loading manually via import UI, to seeding the layout in "local storage". Faster and more reliable.
- Adds `stableBoundingBox` and `waitForModalSettled` to settle the measurement and click-timing for the 1d and 2d move test that was flakey.


mofojed (approved):
Great idea to use the storageState for initializing the layout.

PR #2760, opened 2026-09-18
fix: DH-17594: useWindowedListData prunes selectedKeys

cryolins:
`useWindowedListData` now cleans up dangling keys in selectedKeys in the `remove` method to prevent errors from them pointing to no longer existing data, reflecting similar behaviour in `react-stately`'s `useListData`.

Since we also expose `setItems`, it now has a wrapper to maybe clean up selectedKeys when items are being set.


packages/react-hooks/src/useWindowedListData.ts
```diff
@@ -73,29 +96,56 @@ export function useWindowedListData<T>({
     [findItem]
   );
 
+  /** Sets items and prunes selected keys based on new items */
+  const setItemsAndPruneKeys = useCallback(
+    (upd: React.SetStateAction<T[]>) => {
```
dgodinez-dh: I recommend renaming `upd` to `itemsOrUpdater` for clarity.

packages/react-hooks/src/useWindowedListData.ts
```diff
@@ -73,29 +96,56 @@ export function useWindowedListData<T>({
...
+      setDataState(prev => {
+        const nextItems = upd instanceof Function ? upd(prev.items) : upd;
+        let prevSelectedKeys = prev.selectedKeys;
+        if (prevSelectedKeys !== 'all') {
+          const newItemKeys = new Set(nextItems.map(item => getKey(item)));
+          prevSelectedKeys = new Set(
```
dgodinez-dh: Add a check here. If nothing was pruned, keep the previous set rather than creating a new set.

packages/react-hooks/src/useWindowedListData.ts
```diff
@@ -73,29 +96,56 @@ export function useWindowedListData<T>({
...
+            [...prevSelectedKeys].filter(key => newItemKeys.has(key))
+          );
+        }
+        return {
```
dgodinez-dh, line 144: Add a check here. If neither items nor selection changed (equality checks), then return prev. This one is important to prevent unnecessary re-renders.

packages/react-hooks/src/useWindowedListData.ts
```diff
@@ -38,10 +43,28 @@ export interface UseWindowedListDataOptions<T> {
...
 }: UseWindowedListDataOptions<T> = {}): WindowedListData<T> {
-  const [items, setItems] = useState<T[]>([]);
+  const [{ items, selectedKeys }, setDataState] = useState<
+    WindowedListDataState<T>
+  >({
+    items: [],
+    selectedKeys: new Set(),
+  });
+
+  const setItems = useCallback((upd: React.SetStateAction<T[]>) => {
```
dgodinez-dh: I would recommend a different name for this callback. Maybe setItemsWitoutPrune? It becomes confusing later when we assign setItemsAndPruneKeys to different setItems.

dgodinez-dh (changes requested):

dgodinez-dh (approved):

